Issue #97, opened 2018-06-26
Add GCC 5,6,8 compiler support


ejfitzgerald:
Add compiler support for all of the following compiler configurations:

- [ ] GCC 5
- [ ] GCC 6
- [x] GCC 7
- [x] GCC 8

These compilers should be supported with the following error flags:

`-Wall -Wexra -Wpedantic -Wconversion -Werror` -> essentially the same as our clang configuration

PR #98, opened 2018-06-26
GCC 5+ Compiler Support

ejfitzgerald:
The warning configuration between Clang and GCC vary quite considerably. For the moment and so that Nathan can develop using GCC if he wishes I have introduced a GCC based warning flag configuration. However, this is not ideal. I have created the issue #97 to track this issue.

Rather worryingly there are a number of conversion errors that are generated by `g++`. After we have the initial test net up and running we should revisit this issue. Certainly this need to be done for public code release.


troels-im (dismissed):
There is one trivial comment. It'd be great if you could take care of it and otherwise, just drop me a line and I'll do it.

PR #106, opened 2018-07-05
CMake: Add build support for gtest and gmock

ejfitzgerald:
Adding google test and google mock support for component tests.  This can be used by adding the following line to the `CMakeLists.txt` file.

`fetch_add_gtest(<test_target_name>, <component library>, <directory name>)`

This means that all the source files that are in the target directory are compiled together to make the test suite.


n-hutton (dismissed):

pbukva (approved):
LGTM

PR #113, opened 2018-07-11
Limit individual test executions to 2 minutes

ejfitzgerald:
Place a hard limit of 2minutes for each individual test case. Currently I think the longest test we have is the tcp_stress_test which is around 80 seconds when running correctly. This figure will give us a reasonable about of headroom.


troels-im (approved):

pbukva (commented):
Perhaps we can set this limitation on jenkins job level (per whole job or per stage) in Jenkinsfile?

PR #331, opened 2018-10-31
Bugfix/speed tcp stresstest

engineerkatie:
The random number generator is slow in debug mode and means the tcp client stress test sometimes timesout, making it difficult to run the whole test suite in debug mode to verify a change.

Here, we add a test for the speed of the random number generation (so we can detect if it slows down too much) and remove the tcp stress test reliance on random numbers.


ejfitzgerald (commented):
What @n-hutton said + my 2 pennies worth

ejfitzgerald (approved):
LGTM

PR #430, opened 2018-12-07
Feature/server disconnect

bipll:
This is probably an implementation of server-side disconnect, plus small fixes every here and there, like more correct invocation in generics/callbacks.hpp etc.

Changed FETCH_LOCK definition in core/mutex.hpp to accept locks of any type, could probably be useful later.


ejfitzgerald (changes requested):
This looks quite good to me. However, I was expecting that you implement the disconnect at the Muddle level instead of the service level.

We will actually in the process of migrating away some of the connections (like the lane connections) away from the raw RPC level.

I was thinking that the disconnect can be integrated as a "direct message" at the Muddle level. This would then trigger the update to the peer connection list. I apologise for the confusion.

In addition @engineerkatie and @qati have recently merged a mammoth set of changes back which includes a simple version of the blacklist inside the . I would probably remove that one in preference for your version, but you might have to make a few interface changes.

Hit me up on Slack if you want to discuss etc.

PR #466, opened 2019-01-07
Fix flow

pierre-wilmot-fetch-ai:
Separate Variable into Data and Ops.
 Done so far :
- Matrix Multiply
- Relu


baykaner (changes requested):
mostly great! A few naming convention issues I'd like you to fix though

baykaner (approved):
great thanks

PR #529, opened 2019-01-14
Feature/trust update

qati:


engineerkatie (commented):
I'm probably too close to this code to review it well... I think it's terrific and we should merge it....

Actually as a general note, we're passing a LOT of parameters around, mostly in order to get at config/argument information. It might be worth setting up a single datastore for these values to merge cmdline, config files & defaults and passing that around at some point -- possibly when we move to a more config-file oriented view?

PR #541, opened 2019-01-16
Feature/poll tx from lanes

### LINKED ISSUE #424: Transaction Summary RPC Poll
Add the infrastructure required to implement a constellation (hub) originating poll to the lane shards in order to get the newly transactions that have arrived.

- [ ] RPC Poll for new transactions
- [ ] Considation of transaction caches (if appropriate) on the shard side of things
- [ ] Add response handler for the poll to add transactions to the mining queue

**Design must be compatible with the non-mining use case**

n-hutton:
tx poll implementation. Done by having the transient object store keep a fixed size queue that fills with new (network) transactions. Transactions that arrive too fast will be dropped. The TX verifier has a thread that will periodically poll the storage unit interface, which will poll each of the lanes' transient object stores for fresh TXs.

I rate limited the TXs coming in this way as the presumption is that if you send a TX to a node it has a higher chance of getting mined by that node than by another node.

In future, the verifier should balance the volume of TXs coming in, and
there should be some sort of fee sorting so that high fee TXs are more likely to be polled.

Closes #424


ejfitzgerald (dismissed):
Very minor comments, you might want to fix up before merge back to develop. Nice one dude, as always you knock it out of the park

ejfitzgerald (approved):

PR #589, opened 2019-01-30
Make sure manifests are still set even when P2P churn is disabled.

engineerkatie:
We make the P2P workcycle run on a loop. It checks to see if peer updates are due & does them. Checks to see if manifest updates are due and does them.

This could probably do some sort of "sleep until the minimum time at which the next thing is due" and we could add that in the future when the code stabalises, but AB has an inbound change to this area.


ejfitzgerald (changes requested):
Mostly questions due to me not really understanding. 😁

ejfitzgerald (approved):

PR #598, opened 2019-01-31
Feature/mock smart ledger

baykaner:
initial mock smart ledger example
Various small interface improvements through libs/auctions


n-hutton (approved):
Lot to look through. Since I am going to integrate the vm next I will have to come back to this and we can work together to get it integrated in whatever form happens

PR #603, opened 2019-02-01
Bug/584

bipll:
`Identity` is now checked for emptiness when constructing, or verifying with, `ECDSAVerifier`, and the latter can be converted to bool.


ejfitzgerald (dismissed):

n-hutton (approved):
lgtm

PR #624, opened 2019-02-07
Feature/unit test for interleaving

n-hutton:
step 1: replicate the bad magic error seen by Troels.

Two tests try to interleave packets, one from client -> server and the other server -> client.

The server client one fails with bad magic. The client has strands which is presumably why it doesn't fail.


troels-im (approved):
LGTM

PR #651, opened 2019-02-18
State template skeleton

robertdickson:


n-hutton (approved):
Great, although ideally there would be an example for this state functionality

PR #656, opened 2019-02-18
Feature/math sfinae cleanup

baykaner:
Note - this commit will break all applications that explicitly depended upon NDArray, and possibly some that depended upon ShapelessArray and RectangularArray. Specifically, this means some of the python script examples may require updating from NDArray to Tensor, and there may possibly be other repositories that depend upon fetch-ledger that would also require updating

- fully removed ndarray
- some math SFINAE rationalisation
- removed some old redundant math operations which had multiple implementations (e.g. log & exp)


CMakeFiles/cmake.check_cache
```diff
@@ -0,0 +1 @@
+# This file is generated by cmake for dependency checking of the CMakeCache.txt file
```
ejfitzgerald: This should not be commited

libs/math/include/math/free_functions/fundamental_operators.hpp
```diff
@@ -67,82 +64,132 @@ meta::IfIsMathArray<ArrayType, ArrayType> Add(ArrayType const &array1, ArrayType
...
+  }
+}
+
+template <typename ArrayType, typename T>
+meta::IfIsMathArray<ArrayType, void> Subtract(ArrayType const &array1, ArrayType const &array2,
+                                              ArrayType &ret)
+{
+  for (std::size_t i = 0; i < ret.size(); ++i)
+  {
+    ret.At(i) = array1 - array2.At(i);
```
ejfitzgerald: Not `array1.At(i)`?

libs/math/include/math/free_functions/fundamental_operators.hpp
```diff
@@ -447,82 +486,51 @@ meta::IfIsMathShapelessArray<ArrayType, void> Subtract(ArrayType const &obj1, Ar
...
-  assert(obj1.shape() == obj2.shape());
-  NDArray<T, C> ret{obj1.shape()};
-  Subtract(obj1, obj2, ret);
-  return ret;
-}
+/////////////////////////////////////////
+///// ARRAY BROADCASTING SUBTRACTION  ///
+/////////////////////////////////////////
+//
+///**
```
ejfitzgerald: Not really blown away with all this commented out code

CMakeFiles/cmake.check_cache
pierre-wilmot-fetch-ai: Should go in gitignore

libs/math/include/math/free_functions/matrix_operations/matrix_operations.hpp
```diff
@@ -261,6 +261,19 @@ T Max(ShapelessArray<T, C> const &array)
   return ret;
 }
 
+template <typename ArrayType, typename T>
+meta::IfIsMathArray<ArrayType, void> Max(ArrayType const &array, T &ret)
+{
+  ret = -std::numeric_limits<T>::max();
```
pierre-wilmot-fetch-ai: I still thoink that should be std::numeric_limits<T>::min();
If we ever pass an unsigned type in there, this will break.

libs/math/include/math/free_functions/matrix_operations/matrix_operations.hpp
```diff
@@ -261,6 +261,19 @@ T Max(ShapelessArray<T, C> const &array)
...
+  for (std::size_t j = 0; j < array.size(); ++j)
```
pierre-wilmot-fetch-ai: WE should also implement the begin and end in tensor to allow range baser for

ejfitzgerald (dismissed):
There are a lot of changes here. Not sure if we have that great test coverage for this stuff. Have you generated any coverage reports.

Also for 10,000ft glance there is at least one case where a function looks incorrect (might just be my understanding)

CMakeFiles/cmake.check_cache
ejfitzgerald: This whole folder should not be even generated. If you name your build folder something with the name "build" in it then it will be ignored 😁

libs/math/include/math/free_functions/standard_functions/exp.hpp
```diff
@@ -28,23 +29,49 @@
...
-void Exp(fetch::fixed_point::FixedPoint<I, F> &x)
+template <typename T>
+fetch::math::meta::IfIsMathArray<T, void> Exp(T const &array, T &ret)
 {
-  x = fetch::fixed_point::FixedPoint<I, F>(std::exp(double(x)));
+  for (std::size_t i = 0; i < array.size(); ++i)
+  {
+    ret.Set(i, std::exp(array.At(i)));
+  }
 }
```
pierre-wilmot-fetch-ai: template <typename ArrayType>
fetch::math::meta::IfIsNonBlasArray<ArrayType, void> Exp(ArrayType &x)

template <typename ArrayType>
fetch::math::meta::IfIsMathFixedPointArray<ArrayType, void> Exp(ArrayType &x)

template <typename ArrayType>
fetch::math::meta::IfIsMathArray<T, void> Exp(T const &array, T &ret)

All three method have the same implementation.
All we need is to template at the last level -- In this case the function that does exp on one element.
There's no need to template the loop as well

libs/math/include/math/free_functions/standard_functions/log.hpp
```diff
@@ -43,35 +43,53 @@ fetch::math::meta::IfIsMathShapeArray<ArrayType, ArrayType> Log(ArrayType const
...
+template <typename T>
+meta::IfIsMathArray<T, void> Log(T const &array, T &ret)
+{
+  for (std::size_t i = 0; i < array.size(); ++i)
+  {
+    ret.Set(i, std::log(array.At(i)));
+  }
+}
 
 }  // namespace math
```
pierre-wilmot-fetch-ai: Same here, no need for 3 copies of the same loop

libs/math/tests/math/free_functions/free_functions.cpp
```diff
@@ -23,109 +23,88 @@
...
+#include "math/tensor.hpp"
+
+template <typename T>
+class FreeFunctionsTest : public ::testing::Test
+{
+};
+
+// using MyTypes = ::testing::Types<fetch::math::Tensor<float>, fetch::math::Tensor<double>,
+//                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<32, 32>>>;
+using MyTypes = ::testing::Types<fetch::math::Tensor<float>>;
```
pierre-wilmot-fetch-ai: Any reason for disabling fixed point tests ?

libs/math/include/math/free_functions/standard_functions/exp.hpp
baykaner: delayed until further cleanup in the next math cleaning PR removes shapeless array and rectangular array

libs/math/tests/math/free_functions/free_functions.cpp
baykaner: accident. fixed

libs/math/include/math/free_functions/fundamental_operators.hpp
baykaner: yup, I knew i missed some!

libs/math/include/math/free_functions/fundamental_operators.hpp
baykaner: good catch!

libs/math/include/math/free_functions/matrix_operations/matrix_operations.hpp
baykaner: yep, agreed.

libs/math/include/math/free_functions/matrix_operations/matrix_operations.hpp
baykaner: we have learned that std::numeric_limits::lowest() is apparently the correct way to go. Fixed

CMakeFiles/cmake.check_cache
baykaner: whoops - sorry about the garbage!

baykaner:
So I've addressed reviewer comments now, but is the plan to delay squash and merge of this PR until some time after 25th? @ejfitzgerald

baykaner:
Most of the original changes here were merged into a different PR. This PR now contains only a few fixes that were not merged in previously.

pierre-wilmot-fetch-ai (approved):

libs/math/include/math/free_functions/fundamental_operators.hpp
pierre-wilmot-fetch-ai: Good thing it's fixed, but the fact it didn't break Jenkins indicate this is actually dead (&untested) code. We should probably run a pass of coverage test and test or remove what is unused.

PR #659, opened 2019-02-18
Bugfix/tensor wrapper fix

baykaner:
replaced size_t with uint64_t for tensor and associated math & ml libraries.


pbukva (approved):
LGTM

PR #660, opened 2019-02-19
Fix for Tx syncing (make it working in both directions)

### LINKED ISSUE #563: TX sync appears to be unidirectional
when using swarm and sending TXs to a node, the TX sync only goes forward, so 0->1->2->3 etc

pbukva:
* This is just band-aid.
   - First: there is one outstanding issue - connection to own lane muddle
   (between lane service and muddle on own node side) is counted in to the
   connections for tx syncing, what is not correct.
   - Second: Whole concept of how connections are handled - tripple mapping
   uri vs handle vs address and design around it makes things very hard
   to maintain or change. Thus it would need some redesigning to make
   the code more readable & maintainable.

This closes #563.

### DIFF
--- a/apps/constellation/constellation.cpp
+++ b/apps/constellation/constellation.cpp
@@ -184,8 +184,8 @@ Constellation::Constellation(CertificatePtr &&certificate, Config config)
   reactor_.Attach(block_coordinator_.GetWeakRunnable());
 
   // configure all the lane services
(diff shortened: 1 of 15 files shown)


ejfitzgerald (changes requested):

pbukva:
closed since it was implemented in another PR.

PR #661, opened 2019-02-19
Feature/simplification

bipll:
`fetch::type_util::Disjunction`, basically a backport of its namesake from C++17's `<type_traits>`, plus two more templates, rather basic, applied to `fetch::vm::IsPrimitive` to make the latter more readable perhaps.


pbukva (dismissed):
This is just suggestion - see my inline comments. I'm out of office this week, so please discuss @ejfitzgerald, he can explain.

ejfitzgerald (changes requested):
Looks great to me, only suggestion is that the header file should live with the other meta stuff in the meta "library"

ejfitzgerald (approved):

PR #662, opened 2019-02-19
Bug/vm fix

n-hutton:


ejfitzgerald (approved):
We should investigate if the autogenerated code can be ignored from the commit but whatever.

PR #668, opened 2019-02-19
Feature/vm draft2

n-hutton:


libs/vm/tests/vm_tests.cpp
```diff
@@ -133,11 +301,62 @@ TEST_F(VMTests, CheckCustomBinding)
...
+
+  EXPECT_EQ(out, 8);
+}
+
+TEST_F(VMTests, CheckCustomBindingWithStateDefault)
+{
+  const std::string source =
+      "function main()                                      \n "
+      " var s = State<Int32>('hello', 9);                   \n "
+      " if(s.existed())                                     \n "
```
n-hutton: able to check if default initialization happened or not

libs/ledger/include/ledger/chain/mutable_transaction.hpp
```diff
@@ -73,6 +74,9 @@ struct TransactionSummary
   // TODO(issue 33): Needs to be replaced with some kind of ID
   ContractName contract_name;
 
+  // Hashes of the body of the smart contract(s) accessed
+  ContractHashes contract_hashes;
```
n-hutton: TXs will have to specify their raw contracts IDs

libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -94,14 +95,20 @@ class Contract
...
   void SetStateRecord(T const &record, byte_array::ByteArray const &address);
 
+  bool CheckRawState(byte_array::ByteArray const &address);
+  void SetRawState(byte_array::ByteArray const &payload, byte_array::ByteArray const &address);
+  bool GetRawState(byte_array::ByteArray &payload, byte_array::ByteArray const &address);
+
 private:
-  bool LockResources(ResourceSet const &resources);
-  bool UnlockResources(ResourceSet const &resources);
+  bool LockResources(ResourceSet const &resources, ContractHashes const &hashes);
```
n-hutton: Resources are now locked by appending state etc. to resources. also a raw lock of the contract hash.

libs/ledger/include/ledger/chaincode/database_interface.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include "core/byte_array/const_byte_array.hpp"
+#include "ledger/chaincode/contract.hpp"
+#include "ledger/chaincode/smart_contract.hpp"
+#include "vm/state_sentinel.hpp"
+
+namespace fetch {
+namespace ledger {
+
+class DatabaseInterface : public vm::ReadWriteInterface
```
n-hutton, line 28: Acts as a buffer between the state db and the SC. this way if there is some error during 'runtime', no writeback need occur. Also, permissions can be used to ensure there are no lock violations.

libs/ledger/include/ledger/chaincode/database_interface.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+  bool exists(uint8_t const *const key, uint64_t key_size, bool &exists) override;
+
+  bool write(uint8_t const *const source, uint64_t dest_size, uint8_t const *const keyy,
+             uint64_t key_size) override;
+
+  bool read(uint8_t *dest, uint64_t dest_size, uint8_t const *const keyy,
+            uint64_t key_size) override;
+
+  bool AccessResource(ByteArray const &key);
```
n-hutton, line 50: for checking permissions. I am totally open to renaming suggestions as I wrote this hurriedly

libs/ledger/include/ledger/chaincode/smart_contract.hpp
```diff
@@ -18,31 +18,30 @@
...
-  Status Invoke(Transaction const &tx);
-  Status DeleteContract(Transaction const &tx);
+  Status InvokeContract(Transaction const &tx);
+
+  bool RunSmartContract(std::string &source, std::string const &target_fn,
+                        byte_array::ConstByteArray const &hash, Transaction const &tx);
 
-  // queries
-  /* Status Balance(Query const &query, Query &response); */
+  std::string source_;
```
n-hutton, line 46: the only state a smart contract keeps is the original source. Note, I am in keeping in mind the requirement that the SC attaches handlers for as many functions as there are. It's just a bit inconvenient/unclean because in order to do this you will need to access the state db which creates a bit of a circular dependency. Do-able though.

libs/ledger/include/ledger/chaincode/smart_contract_manager.hpp
```diff
@@ -0,0 +1,39 @@
...
+{
+public:
+  SmartContractManager();
+  ~SmartContractManager() = default;
+
+  static constexpr char const *LOGGING_NAME = "SmartContractManager";
+
+private:
+  // transaction handlers
+  Status CreateInitialContract(Transaction const &tx);
```
n-hutton, line 35: SC manager is going to be the chaincode for managing SCs - pretty straightforward atm.

libs/ledger/src/chaincode/contract.cpp
```diff
@@ -123,6 +123,7 @@ bool Contract::ParseAsJson(Transaction const &tx, variant::Variant &output)
   }
   catch (json::JSONParseException &ex)
   {
+    // TODO(HUT): this can't be good for performance
```
n-hutton, line 123: IRL discussion?

libs/ledger/src/chaincode/contract.cpp
```diff
@@ -152,7 +153,7 @@ Contract::TransactionHandlerMap const &Contract::transaction_handlers() const
 storage::ResourceAddress Contract::CreateStateIndex(byte_array::ByteArray const &suffix) const
 {
   byte_array::ByteArray index;
-  index.Append(contract_identifier_.name_space(), ".state.", suffix);
+  index.Append(contract_identifier_[0], contract_identifier_[1], ".state.", suffix);
```
n-hutton: The namespace concept needs a little refactoring. This forces SC hash + pubkey to be hashed.

libs/ledger/src/chaincode/database_interface.cpp
```diff
@@ -0,0 +1,141 @@
...
+void DatabaseInterface::WriteBackToState()
+{
+  for (auto const &cached_item : cached_resources_)
+  {
+    auto address = cached_item.first;
+    auto data    = cached_item.second;
+
+    FETCH_LOG_WARN(LOGGING_NAME, "Writing back to state: ", ToHex(address), " ", ToHex(data));
+
+    context_->Set(data, address);
```
n-hutton, line 136: Note: for performance there could be a 'dirty bit' here

libs/ledger/src/chaincode/factory.cpp
```diff
@@ -66,8 +71,10 @@ ChainCodeFactory::ContractPtr ChainCodeFactory::Create(ContractName const &name)
   auto it = global_registry.find(name);
   if (it == global_registry.end())
   {
-    FETCH_LOG_ERROR(LOGGING_NAME, "Unable to lookup requested chain code: ", name);
-    throw std::runtime_error("Invalid chain code name");
+    FETCH_LOG_INFO(LOGGING_NAME, "Unable to lookup requested chain code: ", name,
+                   ". Creating new SC");
+
+    return std::make_shared<SmartContract>(name);
```
n-hutton, line 77: Default to creating a SC here if chain code can't be found.

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -17,40 +17,168 @@
...
+    if (!GetRawState(contract_source, byte_array::FromHex(contract_hash)))
+    {
+      FETCH_LOG_WARN(LOGGING_NAME, "Failed to lookup hash!", contract_hash);
+      return Status::FAILED;
+    }
+
+    source_ = std::string{contract_source};
+
+    size_t index = 0;
+    while (true)
```
n-hutton: Ignore this, as discussed the contract source will be kept as a base64 which voids this.

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -17,40 +17,168 @@
...
+    return Status::FAILED;
+  }
+
+  byte_array::ConstByteArray contract_hash;    // Lookup
+  byte_array::ByteArray      contract_source;  // Fill this
+
+  if (!Extract(data, CONTRACT_HASH, contract_hash))
+  {
+    FETCH_LOG_WARN(LOGGING_NAME, "Failed to parse contract hash from SC");
+    return Status::FAILED;
```
n-hutton: Before this gets merged in, I will change it such that FAILED can basically almost never be returned by the SC. That is, if it is a badly formatted SC or makes a bad access, that's just a 'no state change' scenario. In this circumstance you would lose your funds.

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -17,40 +17,168 @@
...
+  vm->SetIOInterface(&interface);
+
+  // Execute our fn
+  if (!vm->Execute(script, target_fn, error, output))
+  {
+    FETCH_LOG_INFO(LOGGING_NAME, "Runtime error: ", error);
+  }
+
+  // Successfully ran the VM. Write back the state
+  interface.WriteBackToState();
```
n-hutton: I am going to write an Array<N> error stack trace back to the SC class, that way you can do a query to see what went wrong.

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -17,40 +17,168 @@
...
+
+  return true;
 }
 
-Contract::Status SmartContract::DeleteContract(Transaction const & /*tx*/)
+bool SmartContract::Exists(byte_array::ByteArray const &address)
```
n-hutton, line 149: these fns are called by the database interface (context) to get state

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+  OnTransaction("create_initial_contract", this, &SmartContractManager::CreateInitialContract);
+}
+
+Contract::Status SmartContractManager::CreateInitialContract(Transaction const &tx)
+{
+  variant::Variant data;
+  if (!ParseAsJson(tx, data))
+  {
+    return Status::FAILED;
```
n-hutton: todo: don't fail this - just log it.

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  byte_array::ConstByteArray contract_source;
+  byte_array::ByteArray      contract_build_up;
+  byte_array::ByteArray      contract_build_up_final;
+  byte_array::ByteArray      contract_with_quotes{'"', 1};
```
n-hutton: ignore this annoyance.

libs/ledger/src/execution_manager.cpp
```diff
@@ -142,42 +142,47 @@ bool ExecutionManager::PlanExecution(Block::Body const &block)
...
       if (contract)
       {
-        auto item = std::make_unique<ExecutionItem>(tx.transaction_hash, slice_index);
-
         // transform the resources into lane allocation
         for (auto const &resource : tx.resources)
         {
-          storage::ResourceID const id{contract->CreateStateIndex(resource)};
-          item->AddLane(id.lane(block.log2_num_lanes));
+          storage::ResourceID const resource_id{contract->CreateStateIndex(resource)};
```
n-hutton: note: the condition of there being no contract should basically never happen

libs/vm/include/vm/state_sentinel.hpp
```diff
@@ -0,0 +1,118 @@
...
+public:
+  virtual ~ReadWriteInterface()                                                  = default;
+  virtual bool read(uint8_t *dest, uint64_t dest_size, uint8_t const *const key,
+                    uint64_t key_size)                                           = 0;
+  virtual bool write(uint8_t const *const source, uint64_t dest_size, uint8_t const *const key,
+                     uint64_t key_size)                                          = 0;
+  virtual bool exists(uint8_t const *const key, uint64_t key_size, bool &exists) = 0;
+};
+
+class StateSentinel
```
n-hutton, line 43: state sentinel decides what the VMs behaviuor is for a bad access

libs/vm/include/vm/state_sentinel.hpp
```diff
@@ -0,0 +1,118 @@
...
+//------------------------------------------------------------------------------
+
+#include <cstdint>
+#include <stdexcept>
+#include <string>
+
+namespace fetch {
+namespace vm {
+
+class ReadWriteInterface
```
n-hutton, line 32: other people use this to allow the VM access to state/file systems

libs/vm/src/module.cpp
```diff
@@ -72,6 +72,7 @@ Module::Module()
   istate.CreateTypeConstuctor<Ptr<String>, TemplateParameter>();
   istate.CreateInstanceFunction("get", &IState::Get);
   istate.CreateInstanceFunction("set", &IState::Set);
+  istate.CreateInstanceFunction("existed", &IState::Existed);
```
n-hutton, line 75: probably needs a rename

libs/vm/tests/vm_tests.cpp
```diff
@@ -22,6 +22,168 @@
...
+  {
+    data_ = data;
+  }
+
+private:
+  uint8_t *data_   = nullptr;
+  uint64_t length_ = 0;
+};
+
+class DummyReadWriteInterface : public ReadWriteInterface
```
n-hutton: this just acts as storage for the test, avoiding the bytearray library dependency etc.

libs/ledger/include/ledger/chain/mutable_transaction.hpp
n-hutton: this is currently encoded as base64

libs/ledger/include/ledger/chain/mutable_transaction.hpp
```diff
@@ -408,6 +427,11 @@ class MutableTransaction
     summary_.resources.insert(res);
   }
 
+  void PushContractHash(byte_array::ConstByteArray const &res)
```
n-hutton, line 442: side note: this TX stuff needs cleaned - there are like 4 fields in a tx and 100 getters and setters

libs/http/include/http/key_value_set.hpp
```diff
@@ -88,11 +88,20 @@ class KeyValueSet : private std::map<byte_array::ConstByteArray, byte_array::Con
...
 
-  byte_array::ConstByteArray const &operator[](byte_array::ConstByteArray const &name) const
+  byte_array::ConstByteArray operator[](byte_array::ConstByteArray const &name) const
   {
     LOG_STACK_TRACE_POINT;
 
-    return this->find(name)->second;
+    auto element = this->find(name);
+
+    if(element == this->end())
```
n-hutton: bug: could return end() as byte array crashing the system. Might be worth looking over this class once more.

libs/ledger/include/ledger/chain/mutable_transaction.hpp
```diff
@@ -92,6 +97,17 @@ struct TransactionSummary
   {
     if (resources.size() > 0 && transaction_hash.size() > 0 && contract_name.size() > 0)
     {
+      for (auto const &hash : contract_hashes)
+      {
+        // TODO(HUT): this should possibly check for base64 length if that's what we're going with here
```
n-hutton: todo here

libs/ledger/include/ledger/chain/mutable_transaction.hpp
```diff
@@ -495,9 +535,37 @@ void Serialize(T &stream, TxSigningAdapter<MUTABLE_TX> const &tx)
 template <typename T, typename MUTABLE_TX>
 void Deserialize(T &stream, TxSigningAdapter<MUTABLE_TX> &tx)
 {
+  const char * log_name = "DESER_MUTABLE_TX";
```
n-hutton: this fn needs cleaned - it's also hard to debug when your tx is malformatted. we need versioning for TXs pronto so future changes don't break ledger-api

libs/ledger/include/ledger/chaincode/chain_code_cache.hpp
```diff
@@ -32,8 +32,9 @@ class ChainCodeCache
 {
 public:
   using ContractPtr = ChainCodeFactory::ContractPtr;
+  using StoragePtr = ledger::StorageInterface;
 
-  ContractPtr Lookup(byte_array::ConstByteArray const &contract_name);
+  ContractPtr Lookup(byte_array::ConstByteArray const &contract_name, StoragePtr *state);
```
n-hutton, line 37: chaincode lookup now responsible for populating new Contracts with relevant info. I suspect that they should lock appropriately also.

libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -35,6 +35,13 @@ class Variant;
 }
 namespace ledger {
 
+inline storage::ResourceAddress CreateStateIndexWrapped(byte_array::ByteArray const &global_id, byte_array::ByteArray const &local_id, byte_array::ByteArray const &suffix)
```
n-hutton: new free fn lets you create resource addresses independently

libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -94,14 +106,23 @@ class Contract
...
   template <typename T>
   bool GetStateRecord(T &record, byte_array::ByteArray const &address);
   template <typename T>
   void SetStateRecord(T const &record, byte_array::ByteArray const &address);
 
+  bool CheckRawState(byte_array::ByteArray const &address);
+  void SetRawState(byte_array::ByteArray const &payload, byte_array::ByteArray const &address);
+  bool GetRawState(byte_array::ByteArray &payload, byte_array::ByteArray const &address);
+
+  std::vector<std::string> print_strings_;
```
n-hutton, line 121: contracts own 'print strings' - the idea being that you can look through the whole trace if you turn it on.

This is actually safe as contracts are unique by whole name SC_B64.PUBKEY and should not be scheduled in the same slice.

libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -148,9 +169,8 @@ void Contract::OnQuery(std::string const &name, C *instance,
 template <typename T>
 bool Contract::GetOrCreateStateRecord(T &record, byte_array::ByteArray const &address)
 {
-
   // create the index that is required
-  auto index = CreateStateIndex(address);
+  auto index = CreateStateIndexWrapped(contract_identifier_[0], contract_identifier_[1], address);
```
n-hutton, line 171: I explicitly break the contract identifier here to avoid a bug where the name only has two fields

libs/ledger/include/ledger/chaincode/database_interface.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+  void WriteBackToState();
+
+private:
+  SmartContract *                context_;
```
n-hutton, line 55: this should be a ref & - and others in this PR

libs/ledger/include/ledger/chaincode/smart_contract.hpp
n-hutton: ^ this is resolved now

libs/ledger/src/chaincode/contract_http_interface.cpp
```diff
@@ -112,16 +113,25 @@ ContractHttpInterface::ContractHttpInterface(StorageInterface &    storage,
     }
   }
 
-  Post("/api/contract/submit",
+  // Hex is transmitted in urls to avoid ugly regex parse scenarios
+  // TODO(HUT): pubkey hex should be parsed as hex (and sent as hex from ledger api)
```
n-hutton, line 117: todo here

libs/ledger/src/chaincode/contract_http_interface.cpp
```diff
@@ -163,9 +173,37 @@ http::HTTPResponse ContractHttpInterface::OnQuery(byte_array::ConstByteArray con
...
 http::HTTPResponse ContractHttpInterface::OnTransaction(
-    http::ViewParameters const &, http::HTTPRequest const &request,
+    http::ViewParameters const &params, http::HTTPRequest const &request,
     byte_array::ConstByteArray const *const expected_contract_name)
 {
+  /* ERROR_BACKTRACE; */
+
+  byte_array::ByteArray contract_name;
+
+  if(!expected_contract_name)
```
n-hutton: this shouldn't go in as - is. ugly logic. Basically it will try to parse from the url the keys etc. It's duplicated in this file.

libs/ledger/src/chaincode/contract_http_interface.cpp
```diff
@@ -304,5 +366,84 @@ ContractHttpInterface::SubmitTxStatus ContractHttpInterface::SubmitNativeTx(
   return SubmitTxStatus{submitted, transactions.size()};
 }
 
+http::HTTPResponse ContractHttpInterface::OnTransactionSpeculation(
```
n-hutton, line 360: this is quite useful to do - basically exactly the same code path as normal tx submission with more verbose responses, plus when it 'executes' the tx (without allowing writeback) it returns the print strings.

libs/ledger/src/chaincode/database_interface.cpp
```diff
@@ -0,0 +1,141 @@
...
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "ledger/chaincode/database_interface.hpp"
+
+namespace fetch {
+namespace ledger {
+
+DatabaseInterface::DatabaseInterface(SmartContract *context)
```
n-hutton, line 24: acts as a passthrough from the VM to the state DB, letting the VM know that bad accesses have occured. This way, if the VM does attempt to make a bad access, the write (for example) will return false and it can decide what to do with that rather than crashing the whole vm or invalidating the TX.

libs/ledger/src/chaincode/database_interface.cpp
```diff
@@ -0,0 +1,141 @@
...
+{}
+
+DatabaseInterface::~DatabaseInterface() = default;
+
+void DatabaseInterface::Allow(ByteArray const &resource)
+{
+  allowed_resources_.insert(resource);
+}
+
+bool DatabaseInterface::exists(uint8_t const *const keyy, uint64_t key_size, bool &exists)
```
n-hutton, line 35: bad name: keyy rename.

Also, naming for these interfaces in general has been poorly chosen

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -17,41 +17,186 @@
...
+  if (!ParseAsJson(tx, data))
+  {
+    FETCH_LOG_WARN(LOGGING_NAME, "Failed to parse data from SC. TX data: ", tx.data());
+    /* malformatted TXs should cause block execution failure (although there is no guard for this) */
+    return Status::FAILED;
+  }
+
+  byte_array::ConstByteArray contract_hash;    // Lookup
+  byte_array::ByteArray      contract_source;  // Fill this
+  byte_array::ConstByteArray fn_name = Identifier{tx.contract_name()}.name();
```
n-hutton, line 61: TX name is used to determine the function to call. This replicates information since the fn is called by this same handler and should be reworked. In its current form it's a bit cumbersome to add handers that take more argumentts than a single tx, it would be nice if this was reworked.

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -17,41 +17,186 @@
...
+  if (!Extract(data, CONTRACT_HASH, contract_hash))
+  {
+    FETCH_LOG_WARN(LOGGING_NAME, "Failed to parse contract hash from SC");
+    /* this is an acceptable use case - user loses fees, tx does nothing. */
+    return Status::OK;
+  }
+
+  if (!CheckRawState(contract_hash))
+  {
+    FETCH_LOG_WARN(LOGGING_NAME, "Failed to find SC ", contract_hash," in main DB");
```
n-hutton: not sure the code should be able to get here, now since this handler won't be populated.

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -0,0 +1,109 @@
...
+Contract::Status SmartContractManager::CreateInitialContract(Transaction const &tx)
+{
+
+  std::cerr << "inside SC 1" << std::endl;
+
+  variant::Variant data;
+  if (!ParseAsJson(tx, data))
+  {
+    FETCH_LOG_WARN(LOGGING_NAME, "FAILED TO PARSE TRANSACTION");
+    return Status::FAILED; // TODO(HUT):  // discuss with ed
```
n-hutton: I think this logic should be moved to some earlier verifier stage and not in the hotpath as it were.

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -0,0 +1,109 @@
...
+    FETCH_LOG_ERROR(LOGGING_NAME, "Failed to find SC after trying to set it! : ", source_hashed," in main DB");
+    std::cerr << "bad stuff." << std::endl;
+    std::cerr << "bad stuff." << std::endl;
+    std::cerr << "bad stuff." << std::endl;
+    std::cerr << "bad stuff." << std::endl;
+    std::cerr << "bad stuff." << std::endl;
+    return Status::FAILED;
+  }
+
+  FETCH_LOG_WARN(LOGGING_NAME, "<---------------- Added smart contract, ID: ", source_hashed);
```
n-hutton, line 92: delete noisy log messages.

libs/ledger/src/execution_manager.cpp
```diff
@@ -142,42 +143,41 @@ bool ExecutionManager::PlanExecution(Block::Body const &block)
...
-        auto item = std::make_unique<ExecutionItem>(tx.transaction_hash, slice_index);
+      auto item = std::make_unique<ExecutionItem>(tx.transaction_hash, slice_index);
 
-        // transform the resources into lane allocation
-        for (auto const &resource : tx.resources)
-        {
-          storage::ResourceID const id{contract->CreateStateIndex(resource)};
-          item->AddLane(id.lane(block.log2_num_lanes));
-        }
+      /* auto contract = contracts_.Lookup(id.name_space()); */ // no longer done like this
```
n-hutton: left comment for reviewer - plz delete.

libs/ledger/src/executor.cpp
```diff
@@ -91,23 +91,68 @@ Executor::Status Executor::Execute(TxDigest const &hash, std::size_t slice, Lane
...
+    {
+      return Status::RESOURCE_FAILURE;
+    }
+
+    successfully_locked_resources.push_back(std::move(address));
+  }
 
   // Lookup the chain code associated with the transaction
-  auto chain_code = chain_code_cache_.Lookup(identifier.name_space());
+  auto chain_code = chain_code_cache_.Lookup(id.name_space(), resources_.get());
```
n-hutton, line 135: I think this should be, but isn't, making locked calls to populate the SC

libs/ledger/src/executor.cpp
```diff
@@ -91,23 +91,68 @@ Executor::Status Executor::Execute(TxDigest const &hash, std::size_t slice, Lane
...
+  auto deleter=[&](uint16_t *dummy)
+  {
+    for(auto const &resource : successfully_locked_resources)
+    {
+      resources_->Unlock(resource);
+    }
+  };
+
+  uint16_t  dummyy = 9;
+  std::unique_ptr<uint16_t, decltype(deleter)> on_function_exit(&dummyy, deleter);
```
n-hutton: sadly, this smart pointer with custom deleter should probably be reworked. It's so that all locked resources are unlocked even if there is a throw or a return at any point. Obv this isn't safe if the closure isn't noexcept.

libs/ledger/src/executor.cpp
```diff
@@ -91,23 +91,68 @@ Executor::Status Executor::Execute(TxDigest const &hash, std::size_t slice, Lane
...
+    successfully_locked_resources.push_back(std::move(address));
+  }
 
   // Lookup the chain code associated with the transaction
-  auto chain_code = chain_code_cache_.Lookup(identifier.name_space());
+  auto chain_code = chain_code_cache_.Lookup(id.name_space(), resources_.get());
+
   if (!chain_code)
   {
+    FETCH_LOG_WARN(LOGGING_NAME, "Chain code lookup failure!");
```
n-hutton, line 139: this should probably just penalize the TX for fees and do nothing to the state.

libs/ledger/tests/chaincode/dummy_contract_tests.cpp
```diff
@@ -89,6 +89,6 @@ TEST_F(DummyContractTests, CheckDispatch)
   Identifier identifier;
   identifier.Parse(tx.contract_name());
 
-  contract_->DispatchTransaction(identifier.name(), VerifiedTransaction::Create(tx));
+  contract_->DispatchTransaction(identifier.name(), VerifiedTransaction::Create(tx), nullptr);
```
n-hutton, line 92: I did nullptr in a hurry here, I am not sure I'd like this pattern for retreiving debug info

libs/vm/examples/01_basic_vm/01_example.cpp
```diff
@@ -117,11 +117,12 @@ int main(int argc, char **argv)
...
 
-  std::string        error;
-  fetch::vm::Variant output;
+  std::string              error;
+  std::vector<std::string> print_strings;
+  fetch::vm::Variant       output;
 
   // Setting VM up and running
-  if (!vm->Execute(script, "main", error, output))
+  if (!vm->Execute(script, "main", error, print_strings, output))
```
n-hutton, line 125: this isn't great as the vm will still expend work populating its internal print strings even if its not used externally

libs/vm/include/vm/defs.hpp
```diff
@@ -1032,6 +1032,18 @@ struct Script
     }
     return nullptr;
   }
+
+  std::vector<std::string> GetFunctions() const
```
n-hutton, line 1077: needs to add a way to extract annotations so as to connect both query and regular endpoints.

libs/vm/include/vm/free_functions.hpp
```diff
@@ -28,14 +29,15 @@ namespace vm {
 // VM free functions. These will be available for all smart contracts and MUST NOT RETAIN STATE (be
 // thread safe)
 
-fetch::vm::Ptr<fetch::vm::String> toString(fetch::vm::VM *vm, int32_t const &a)
+static fetch::vm::Ptr<fetch::vm::String> toString(fetch::vm::VM *vm, int32_t const &a)
```
n-hutton: @baykaner will connect this to the math stuff and delete this.

libs/vm/include/vm/state.hpp
```diff
@@ -45,26 +46,44 @@ template <typename T>
...
         TemplateParameter const &value)
     : IState(vm, type_id)
   {
     value_         = value.Get<Value>();
     value_type_id_ = value_type_id;
     name_          = name->str;
+
+    if (this->vm_->state_sentinel_.exists(name_))
+    {
+      this->vm_->state_sentinel_.get(name_, value_);
```
n-hutton, line 75: only get if not exist, because I made the behaviour of the DB interface such that a get to non existing location just creates that much zeroed memory for you.

libs/vm/include/vm/state_sentinel.hpp
```diff
@@ -0,0 +1,135 @@
...
+
+    if (!success)
+    {
+      throw std::runtime_error("Failed to access state in VM! Bad access.");
+    }
+
+    return ret;
+  }
+
+  bool get(std::string const &str, vm::Ptr<vm::Object> &value)
```
n-hutton, line 68: Khan is going to arrange it such that gets and sets are properly templated to use our serializer framework (probably for primitives too for ease/readability).

Ed pointed out is probably a good idea to put all the serializers in a single file that is included here, that way easy to delete from the math library should we desire it.

libs/vm/tests/vm_tests.cpp
```diff
@@ -133,11 +307,62 @@ TEST_F(VMTests, CheckCustomBinding)
...
     EXPECT_EQ(res, true);
   }
 
   EXPECT_EQ(binding_called_count, 3);
 }
+
+TEST_F(VMTests, CheckCustomBindingWithState)
+{
+  const std::string source =
+      "function main()                                      \n "
```
n-hutton: I just have the newlines here for 'error compiling line 4' to have meaning. In retrospect the \n should be at the beginning.

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -17,41 +17,186 @@
...
+    source_ = std::string{FromBase64(contract_source)};
+  }
+
+  auto        module       = vm::VMFactory::GetModule();
+  fetch::vm::Script        script;
+  std::vector<std::string> errors = vm::VMFactory::Compile(module, source_, script);
+
+  if(errors.size() == 0)
+  {
+    for(auto const &fn : script.GetFunctions())
```
n-hutton: we parse our SC for functions and attach them here. Todo: parse annotations so that query handlers can be added.

PR #669, opened 2019-02-19
Main Chain Internals Update

### LINKED ISSUE #538: Default main chain to use disk writes
For historic purposes, the main chain defaults to not using an object store (disk writes) if no 'miner number' is supplied. Change this to default to writing to disk, take a miner signature instead, and create tests for a) file recovery. b) network syncronisation.

ejfitzgerald:
A reworking of the main chain internals and reinstating disk based persistence

Fixes #538


n-hutton (approved):
chat IRL

libs/ledger/include/ledger/chain/block.hpp
```diff
@@ -115,7 +114,7 @@ void Deserialize(T &serializer, Block::Body &body)
 template <typename T>
 inline void Serialize(T &serializer, Block const &block)
 {
-  serializer << block.body << block.nonce << block.proof;
+  serializer << block.body << block.nonce << block.proof << block.weight << block.total_weight;
```
n-hutton, line 117: if we are going to do this, it might make sense to think about doing away with the in-memory map in the mainchain since the object store can now hold complete information

libs/ledger/include/ledger/protocols/main_chain_rpc_protocol.hpp
```diff
@@ -48,13 +50,26 @@ class MainChainProtocol : public service::Protocol
...
   }
 
   Blocks GetChainPreceding(BlockHash const &at, uint32_t maxsize)
   {
     LOG_STACK_TRACE_POINT;
-    return chain_.ChainPreceding(at, maxsize);
+    return Copy(chain_.GetChainPreceding(at, maxsize));
+  }
+
+  static Blocks Copy(MainChain::Blocks const &blocks)
```
n-hutton, line 62: if I understand this correctly it's a bit confusing because mainchain::blocks and protocol::blocks are not the same
in that main chain blocks are smart pointers

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -256,24 +256,25 @@ BlockCoordinator::State BlockCoordinator::OnPreExecBlockValidation()
...
   if (!is_genesis)
   {
-    if (!chain_.Get(current_block_->body.previous_hash, previous))
+    BlockPtr previous = chain_.GetBlock(current_block_->body.previous_hash);
+    if (!previous)
     {
       FETCH_LOG_WARN(LOGGING_NAME, "Block validation failed: No previous block in chain (",
                      ToBase64(current_block_->body.hash), ")");
-      chain_.InvalidateBlock(current_block_->body.hash);
+      chain_.RemoveBlock(current_block_->body.hash);
```
n-hutton, line 267: as I understand this now, mainchiain should never hold loose blocks because they will always be removed?

n-hutton: i see now my assumption was not correct

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -291,7 +292,7 @@ BlockCoordinator::State BlockCoordinator::OnPreExecBlockValidation()
     {
       FETCH_LOG_WARN(LOGGING_NAME, "Block validation failed: Lane count mismatch (",
                      ToBase64(current_block_->body.hash), ")");
-      chain_.InvalidateBlock(current_block_->body.hash);
+      chain_.RemoveBlock(current_block_->body.hash);
```
n-hutton, line 295: I would suggest that some of this verifying logic is wrapped up in a block function similar to 'iswellformed' for TX.

Then, when we get a TX in or a block in we call 

```
isWellFormed();
Verify();
```

ensuring that bad blocks etc are caught immediately. Related: this logic would allow a MITM by changing the miner id to yours

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -393,29 +394,29 @@ BlockCoordinator::State BlockCoordinator::OnPostExecBlockValidation()
   // After the checks have been completed, if the validation has failed, the system needs to recover
   if (invalid_block)
   {
-    bool  revert_successful{false};
-    Block previous_block;
+    bool revert_successful{false};
 
     // we need to restore back to the previous block
-    if (chain_.Get(current_block_->body.previous_hash, previous_block))
+    BlockPtr previous_block = chain_.GetBlock(current_block_->body.previous_hash);
```
n-hutton, line 400: might be worth writing a convenience fn chain_.GetPrevBlock(current_block) since the use case of passing a hash is common

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -428,21 +429,20 @@ BlockCoordinator::State BlockCoordinator::OnPackNewBlock()
...
   {
     // call the block packer
-    block_packer_.GenerateBlock(*current_block_, num_lanes_, num_slices_, chain_);
+    block_packer_.GenerateBlock(*next_block_, num_lanes_, num_slices_, chain_);
 
 #if 1
-    FETCH_LOG_INFO(LOGGING_NAME, "New Block: Hash.........: ", ToBase64(current_block_->body.hash));
-    FETCH_LOG_INFO(LOGGING_NAME,
-                   "New Block: Previous.....: ", ToBase64(current_block_->body.previous_hash));
+    FETCH_LOG_INFO(LOGGING_NAME, "New Block: Hash.........: ", ToBase64(next_block_->body.hash));
```
n-hutton, line 435: do we want this if 1

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -428,21 +429,20 @@ BlockCoordinator::State BlockCoordinator::OnPackNewBlock()
...
-                   "New Block: Miner........: ", ToBase64(current_block_->body.miner));
-    FETCH_LOG_INFO(LOGGING_NAME, "New Block: Log2 Lanes...: ", current_block_->body.log2_num_lanes);
+                   "New Block: Merkle.......: ", ToBase64(next_block_->body.merkle_hash));
+    FETCH_LOG_INFO(LOGGING_NAME, "New Block: Block Number.: ", next_block_->body.block_number);
+    FETCH_LOG_INFO(LOGGING_NAME, "New Block: Miner........: ", ToBase64(next_block_->body.miner));
+    FETCH_LOG_INFO(LOGGING_NAME, "New Block: Log2 Lanes...: ", next_block_->body.log2_num_lanes);
 
     std::size_t slice_index{1};
-    for (auto const &slice : current_block_->body.slices)
+    for (auto const &slice : next_block_->body.slices)
```
n-hutton, line 445: slice index type here hardcoded type which might be out of line with the way we usually ```using = ```

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -578,15 +579,13 @@ BlockCoordinator::State BlockCoordinator::OnTransmitBlock()
 
 BlockCoordinator::State BlockCoordinator::OnReset()
 {
-  // mostly for debug
   current_block_.reset();
+  next_block_.reset();
```
n-hutton, line 583: on reset - state doesn't go to reset?

libs/ledger/src/chain/block_coordinator.cpp
n-hutton: I noticed that when you fail to verify a block, AND you can't revert to the prev block, you set the state to genesis. I think there should be a check there to say 'if this fails, (genesis state not reverted). You basically need to rm -rf * your system'

pbukva (commented):
LGTM

pbukva (approved):

libs/ledger/include/ledger/protocols/main_chain_rpc_protocol.hpp
ejfitzgerald: That is correct

libs/ledger/src/chain/block_coordinator.cpp
ejfitzgerald: It gets removed in the next PR

PR #670, opened 2019-02-20
feature/RangeBasedLoopForTensor

pierre-wilmot-fetch-ai:


libs/math/include/math/tensor.hpp
```diff
@@ -177,6 +178,23 @@ class Tensor  // Using name Tensor to not clash with current NDArray
     }
   }
 
+  /////////////////
+  /// Iterators ///
+  /////////////////
+
+  TensorIterator<T> begin()  // Need to stay lowercase for range basedloops
```
baykaner: hmm.. that's annoying. But we can maintain a bit more interface consistency if we also implement an overloaded interface as well:
i.e. 
TensorIterator<T> Begin(){return this->begin();}

libs/math/include/math/tensor.hpp
```diff
@@ -384,7 +402,14 @@ class Tensor  // Using name Tensor to not clash with current NDArray
   {
     std::stringstream ss;
     ss << std::setprecision(5) << std::fixed << std::showpos;
-    if (shape_.size() == 2)
+    if (shape_.size() == 1)
```
baykaner, line 406: depending on which PR goes in first one of us will have to resolve a merge conflict here. shouldn't be tricky though

libs/math/include/math/tensor_iterator.hpp
```diff
@@ -0,0 +1,88 @@
...
+//
+//------------------------------------------------------------------------------
+
+namespace fetch {
+namespace math {
+template <typename T>
+class TensorIterator
+{
+  using Type     = T;
+  using SizeType = std::uint64_t;
```
baykaner: hmm.... if we decide to change the SizeType in Tensor but forget to change it here there could be problems.
How about we make a common header called tensor_type_defs.hpp containing SizeType, DefaultAlignment, and anything else we later decide needs to be added.

libs/math/include/math/tensor_iterator.hpp
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  bool operator!=(const TensorIterator<T> &other) const
+  {
+    return !(*this == other);
+  }
+
+  bool operator==(const TensorIterator<T> &other) const
+  {
+    return pointer_ == pointer_ && coordinate_ == other.coordinate_;
```
baykaner: this looks like a bug. isn't it meant to be:
return (pointer_ == other.pointer_) && (coordinate_ == other.coordinate_)
also parentheses for clarity please

libs/math/include/math/tensor_iterator.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+  TensorIterator &operator++()
+  {
+    pointer_ += strides_.back();
+    coordinate_.back()++;
+    if (coordinate_.back() < shape_.back())
+    {
+      return *this;
+    }
+    std::uint64_t i(shape_.size() - 1);
```
baykaner: style police; we usually prefer:
std::uint64_t i{shape_.size() - 1};
since it's clear that it's not a function invokation

libs/math/include/math/tensor_iterator.hpp
```diff
@@ -0,0 +1,88 @@
...
+    while (i && coordinate_[i] >= shape_[i])
```
baykaner: parentheses around coordinate_[i] >= shape_[i]
then we don't have to try to figure out whether this is checking
i && (coordinate_[i] >= shape_[i])
or
(i && coordinate_[i]) >= shape_[i]

libs/math/include/math/tensor_iterator.hpp
```diff
@@ -0,0 +1,88 @@
...
+namespace fetch {
+namespace math {
+template <typename T>
+class TensorIterator
+{
+  using Type     = T;
+  using SizeType = std::uint64_t;
+
+public:
+  TensorIterator(std::vector<SizeType> const &shape, std::vector<SizeType> const &strides,
```
baykaner, line 30: is it worth also having a constructor that takes a tensor - so that you can just get the shape, strides, etc. from the tensor; minimises input parameters required and therefore opportunity for error.
Or alternatively if you don't like the idea of passing a tensor to a tensor iterator (for ownership reasons), you could write a const method in Tensor that returns shape, stride, etc. pre-packaged, then when you want to invoke the iterator from inside Tensor you just call something like:
TensorIterator ti{this->MetaData()};
Feel free to disagree if your way is better! just thinking out loud.

libs/math/tests/math/tensor/tensor_indexing.cpp
```diff
@@ -24,7 +24,7 @@ class TensorIndexingTest : public ::testing::Test
 {
 };
 
-using MyTypes = ::testing::Types<int, unsigned int, long, unsigned long, float, double>;
+using MyTypes = ::testing::Types<int>;  // , unsigned int, long, unsigned long, float, double>;
```
baykaner: not testing other types? was this leftover from debugging?

baykaner (changes requested):
nice work!
The main thing that needs changing is the bug (i think) in iterator operator== that doesn't check if pointer_ == other.pointer_.
The rest are mainly interface design choices and a few style points.

libs/math/include/math/tensor_iterator.hpp
pierre-wilmot-fetch-ai: The thing is you never explicitly instanciate an iterator, you always get it throught Tensor::begin().

libs/math/include/math/tensor_iterator.hpp
pierre-wilmot-fetch-ai: Not a big fan of the idea, I'd rather pass it as a template parameter so that we're sure the iterator usees the same type as the tensor it's refering too.

libs/math/include/math/tensor_iterator.hpp
pierre-wilmot-fetch-ai: I'd actually put the iterator constructor private 
And add TensorIterator as a friend class of tensor so that the only way to get an iterator is through Tensor::begin() / Tensor::end()

libs/math/include/math/tensor.hpp
pierre-wilmot-fetch-ai: I doubt we'll manually call that one often.
This is mostly for C++11 range based loop

baykaner: Agreed, it's definitely a nice to have, rather than a requirement

libs/math/include/math/tensor_iterator.hpp
baykaner: Works for me

libs/math/include/math/tensor_iterator.hpp
baykaner: Yup, I like this idea!

baykaner (approved):

PR #674, opened 2019-02-21
feature/ML Embedings module

pierre-wilmot-fetch-ai:
Ready for review


libs/ml/include/ml/ops/embedings.hpp
```diff
@@ -0,0 +1,87 @@
+#pragma once
```
baykaner: please mv/rename to "embeddings" (english spelling/typo) and find-replace embedings->embeddings

libs/ml/include/ml/ops/embedings.hpp
```diff
@@ -0,0 +1,87 @@
...
+namespace fetch {
+namespace ml {
+namespace ops {
+
+template <class T>
+class Embedings : public fetch::ml::ops::Weights<T>
+{
+public:
+  using ArrayType    = T;
+  using DataType     = typename ArrayType::Type;
```
baykaner: in fetch::Math::Tensor we have Type but no DataType.
We should probably be consistent about this and replace one or other. Fine to do this in a future PR, I will raise an Issues

baykaner (changes requested):
Please rename embedings for embeddings everywhere. Otherwise I don't see any problems.

libs/math/include/math/tensor.hpp
```diff
@@ -136,24 +136,12 @@ class Tensor
    * @param other
    * @return
    */
-  SelfType Copy(SelfType const &other)
+  void Copy(SelfType const &other)
```
baykaner, line 139: I'm a bit confused about the removal of most of this function. It's supposed to be a deep copy; unless I misunderstand we do need to copy the shape, padding etc. values, right?

baykaner (changes requested):
Not sure about the changes to Copy - seems unsafe if you remove copying the shape, padding etc.?

baykaner (approved):

PR #677, opened 2019-02-26
AssociateHandleWithAddress doesn't update indirect connections if the…

### LINKED ISSUE #569: AssociateHandleWithAddress doesn't update indirect connections if the update is also an indirect link

pierre-wilmot-fetch-ai:
… update is also an indirect link [closes #569]

### DIFF
--- a/libs/network/src/muddle/router.cpp
+++ b/libs/network/src/muddle/router.cpp
@@ -535,9 +535,8 @@ bool Router::AssociateHandleWithAddress(Handle handle, Packet::RawAddress const 
 
     bool const is_empty = (routing_data.handle == 0);
 
-    // an update is only valid when the connection is direct.
     bool const is_different = (routing_data.handle != handle) || (routing_data.direct != direct);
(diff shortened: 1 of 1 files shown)


bipll:
A simple scenario currently should break this IMO:

node A sends a packet aimed at towards a non-direct unknown (not in handles table) peer Z via a random handle to node B;
B forwards it to C — C records that Hb is a handle that eventually leads to A;
C forwards it to D — D notes that Hc is a handle that eventually leads to A;
D forwards it back to C;
now C's handle that leads to A is actually Hd, and, even more important, they form a closed loop;
C handles it to E that luckily has a direct connection with Z — Z records He as pointing to A;
Z sends a response towards A that back-traces the handles path and follows: Z→E→C→D→C→D→C... (repeat indefinitely*).

----------

*Or maybe, until packet's TTL dies out, which is still not a desired outcome.

----------

In other words, changing indirect handles each time seems unreasonable while the only unknown route lookup algorithm we have is a plain random walk.

ejfitzgerald:
Will sort this out this week

PR #681, opened 2019-02-27
Provide a way to export model weights structure

pierre-wilmot-fetch-ai:
Method name copied from pytorch.
Will be needed for loading / saving of models and for colaborative learning.

### DIFF
--- a/libs/ml/include/ml/ops/weights.hpp
+++ b/libs/ml/include/ml/ops/weights.hpp
@@ -23,6 +23,8 @@ namespace fetch {
 namespace ml {
 namespace ops {
 
+// StateDict is an utility class to extract the network trainable parameter and serialize them for
+// saving / sharing
 template <class T>
 struct StateDict
 {
@@ -32,6 +34,8 @@ struct StateDict
   std::map<std::string, StateDict<T>> dict_;
 };
 
+// Trainaible provide an interface that needs to be matched by any ops that has trainable
+// parameters, that's how the Graph class know with tensors to update during a gradient step
 template <class T>
 class Trainable
 {
@@ -39,9 +43,9 @@ public:
   using ArrayType    = T;
   using ArrayPtrType = std::shared_ptr<ArrayType>;
 
-  virtual void         Step(typename T::Type learningRate)     = 0;
-  virtual StateDict<T> GetStateDict() const                    = 0;
-  virtual void         LoadStateDict(StateDict<T> const &dict) = 0;
+  virtual void         Step(typename T::Type learningRate)                     = 0;
+  virtual StateDict<T> StateDict() const                                       = 0;
+  virtual void         LoadStateDict(fetch::ml::ops::StateDict<T> const &dict) = 0;
 };
 
 template <class T>
@@ -84,14 +88,14 @@ public:
     gradientAccumulation_->Fill(typename T::Type(0));
   }
 
-  virtual StateDict<T> GetStateDict() const
+  virtual StateDict<T> StateDict() const
   {
-    StateDict<T> d;
+    fetch::ml::ops::StateDict<T> d;
     d.weights_ = this->output_;
     return d;
   }
 
-  virtual void LoadStateDict(StateDict<T> const &dict)
+  virtual void LoadStateDict(fetch::ml::ops::StateDict<T> const &dict)
   {
     assert(dict.dict_.empty());
     SetData(dict.weights_);
--- a/libs/ml/include/ml/ops/weights.hpp
+++ b/libs/ml/include/ml/ops/weights.hpp
@@ -24,14 +24,28 @@ namespace ml {
 namespace ops {
 
 template <class T>
+struct StateDict
+{
+  using ArrayType    = T;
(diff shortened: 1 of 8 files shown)


libs/ml/tests/ml/ops/fully_connected.cpp
```diff
@@ -134,3 +134,20 @@ TYPED_TEST(FullyConnectedTest, graph_forward_test)  // Use the class as a Node
   ASSERT_EQ(prediction->shape()[0], 1);
   ASSERT_EQ(prediction->shape()[1], 42);
 }
+
+TYPED_TEST(FullyConnectedTest, getStateDict)
```
baykaner, line 138: seems like it would be worth adding a test that actually checks the weights values as well (rather than just the shape)

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -23,15 +23,29 @@ namespace fetch {
 namespace ml {
 namespace ops {
 
+template <class T>
+struct StateDict
```
baykaner, line 29: might be nice to add a comment above this line to indicate what a StateDict is for

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -23,15 +23,29 @@ namespace fetch {
...
+{
+  using ArrayType    = T;
+  using ArrayPtrType = std::shared_ptr<ArrayType>;
+  ArrayPtrType                        weights_;
+  std::map<std::string, StateDict<T>> dict_;
+};
+
 template <class T>
 class Trainable
 {
```
baykaner, line 41: same for Trainable. I know what it's for, but it's still good to add one line explaining it

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -70,6 +84,19 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<typename
     gradientAccumulation_->Fill(typename T::Type(0));
   }
 
+  virtual StateDict<T> GetStateDict() const
```
baykaner: Function comments would be helpful.
Also, not sure about the name - this isn't really a 'getter' its more of an assignment operation

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -70,6 +84,19 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<typename
...
+  {
+    StateDict<T> d;
+    d.weights_ = this->output_;
+    return d;
+  }
+
+  virtual void LoadStateDict(StateDict<T> const &dict)
```
baykaner: function comment please

libs/ml/tests/ml/ops/graph.cpp
```diff
@@ -75,3 +75,12 @@ TEST(graph_test, node_relu)
   // test correct values
   ASSERT_TRUE(prediction->AllClose(*gt));
 }
+
+TEST(graph_test, getStateDict)
```
baykaner, line 79: does this work for Tensor<FixedPoint<32, 32>>? Worth adding a test/typed_test?

libs/ml/tests/ml/ops/weigths.cpp
```diff
@@ -61,3 +61,30 @@ TYPED_TEST(WeightsTest, gradient_step_test)
   ASSERT_TRUE(w.Forward({}) == data);         // Test pointed address, should still be the same
   ASSERT_TRUE(w.Forward({})->AllClose(*gt));  // whit new values
 }
+
```
baykaner, line 64: this filename actually has a typo - might be nice to rename

libs/ml/tests/ml/ops/fully_connected.cpp
baykaner: actually - this is probably covered by your test in weights.cpp i guess?

baykaner (changes requested):
a few small changes only.
nice work

libs/ml/tests/ml/ops/fully_connected.cpp
pierre-wilmot-fetch-ai: This is randomly initialised weights, I don't know the value they hold.
And since FullyConnected doesn't has getter for them I can't do any comparaison

libs/ml/tests/ml/ops/graph.cpp
pierre-wilmot-fetch-ai: Does not work in this PR, I have another one open with FixedPoint serialisation

libs/ml/tests/ml/ops/weigths.cpp
pierre-wilmot-fetch-ai: 😅

libs/ml/include/ml/ops/weights.hpp
pierre-wilmot-fetch-ai: Yeah, pytorch actually names it just sate_dict(), I'll copy that

baykaner (approved):

PR #682, opened 2019-02-28
feature/Add serializer support for tensor / ML

pierre-wilmot-fetch-ai:


libs/math/include/math/tensor.hpp
```diff
@@ -156,12 +156,27 @@ class Tensor
...
 
-  std::vector<SizeType> const &shape()
-      const  // TODO(private, 520) fix capitalisation (kepping it consistent with NDArray for now)
+  // TODO(private, 520) fix capitalisation (kepping it consistent with NDArray for now)
+  std::vector<SizeType> const &shape() const
   {
     return shape_;
   }
 
+  std::vector<SizeType> const &Strides() const
```
baykaner, line 165: Depending on how you instantiated Tensor, this method is going to return an uninitialised vector (i.e. if you called Tensor(size)).
I'm not 100% clear on why you need separate input_strides_ and strides_ vectors.

libs/math/tests/math/serializers/math_types.cpp
```diff
@@ -0,0 +1,58 @@
...
+};
+
+using MyTypes = ::testing::Types<int, long, float, double>;
+TYPED_TEST_CASE(SerializersTest, MyTypes);
+
+TYPED_TEST(SerializersTest, serialize_empty_tensor)
+{
+  fetch::math::Tensor<TypeParam>      t1;
+  fetch::serializers::ByteArrayBuffer b;
+  b << t1;
```
baykaner, line 36: I think it would be nice to add trivial serialise and deserialise helper functions in tensor so that this line reads:
t1.Serialise(b);

libs/math/tests/math/serializers/math_types.cpp
```diff
@@ -0,0 +1,58 @@
...
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "math/serializers/math_types.hpp"
+#include "core/serializers/byte_array_buffer.hpp"
```
baykaner, line 20: I think you should rework your dependencies such that you don't have to include core/serializers here. Include it in math/serializers.

Also if you forward declare tensor in math/serializers (instead of including it), then you can include math/serializers in tensor.

The includes in this test should just read tensor.hpp and gtest.

libs/ml/include/ml/serializers/ml_types.hpp
```diff
@@ -0,0 +1,71 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "ml/ops/weights.hpp"
```
baykaner, line 20: forward declare weights, and include the serializer in weights rather than including weights in the serializer.

libs/math/include/math/serializers/math_types.hpp
```diff
@@ -0,0 +1,67 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "math/tensor.hpp"
```
baykaner, line 20: Okay this was very confusing at first. Because you haven't included anything from core/serializers here it looked like the calls to Serialize are recursive (which makes no sense). Presumably it's actually a call to something in core/serializers that works because you happen to make your includes in the right order when you later invoke serialization in a test.cpp file later.

This dependency structure is problematic. It leaves us prone to making errors that seem to appear out of nowhere (i.e. if I re-order the includes in your test file it should break).

What needs to be done is to include what you need from core/Serializers here. Don't include tensor here, forward declare it. Include this serialiser in tensor.hpp, and add a convenience method for serialising.

If you do all of this it should be robust to whatever include order you use, and it will be nicer for the developer to just call my_tensor.Serialise(byte_array)

libs/ml/tests/ml/serializers/ml_types.cpp
```diff
@@ -0,0 +1,60 @@
...
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "ml/serializers/ml_types.hpp"
+#include "core/serializers/byte_array_buffer.hpp"
```
baykaner, line 20: remove the relevant includes after reorganising dependencies

baykaner (changes requested):
It's great that you've worked on this, it really needed doing!
The dependency structure needs a bit of reworking.

libs/math/include/math/tensor.hpp
pierre-wilmot-fetch-ai: No, shouldn't return uninitialised. Every constructor is calling Init().
What is happening is that I need I need to be able to recover the parameter I used to instanciate a tensor in order to serialize and recreate during deserialisation.
At the moment the stride vector that you pass as constructor paramater is not kept, it's overwriten in Init().

libs/math/tests/math/serializers/math_types.cpp
pierre-wilmot-fetch-ai: Thought about it too, since it would solve the problem with input_strides_, but I'm trying to match the interface provided for every other type

libs/math/tests/math/serializers/math_types.cpp
pierre-wilmot-fetch-ai: Not sure about that one, all the code in serializeer is templated, so I need to specify what actual type of serializer I want to use at callsite.

@ejfitzgerald : Any insight on why the serializer is templated ?

baykaner (dismissed):

baykaner (approved):

PR #683, opened 2019-02-28
Add serialiser for fixed_point

pierre-wilmot-fetch-ai:


libs/core/include/core/serializers/fixed_point.hpp
```diff
@@ -0,0 +1,40 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "core/fixed_point/fixed_point.hpp"
```
baykaner, line 20: dependencies need reworking as in PR 682.

I see that you've probably copied the dependency ordering as done for byte_array - we'll check that there isn't some excellent reason that it has to be this way around, since i can't see any advantages to it and I see several disadvantages

libs/core/tests/serializers/fixed_point.cpp
```diff
@@ -0,0 +1,57 @@
...
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "core/serializers/fixed_point.hpp"
+#include "core/serializers/byte_array_buffer.hpp"
```
baykaner, line 20: after dependency reorg remove the two serializer includes and just include core/fixed_point

baykaner (changes requested):

baykaner (approved):

PR #684, opened 2019-03-01
Python Bindings for (de)serialisation of statedicts

pierre-wilmot-fetch-ai:


baykaner (dismissed):

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -38,6 +38,21 @@ struct StateDict
     return !((bool(weights_) ^ bool(o.weights_)) || (weights_ && (*weights_ != *(o.weights_))) ||
              (dict_ != o.dict_));
   }
+
+  StateDict &Merge(StateDict const &o, float a = .5f)
```
baykaner: okay "a" is not a self-explanatory variable name and there is also no comment here explaining what it does.
Please add doxygen style function comment (e.g.  /** @param a  this variable dictates the mixing ratio of the weights **/

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -38,6 +38,21 @@ struct StateDict
...
+  {
+    assert(a >= 0.0f && a <= 1.0f);
```
baykaner: you probably should assert that a > 0.0 and a < 1.0 since if a == 0 or a == 1 then why are we merging?

baykaner (changes requested):

baykaner (approved):

PR #685, opened 2019-03-01
Feature/ml reorg

baykaner:
some ml library reorg and slight unification with math lib


libs/math/include/math/tensor.hpp
```diff
@@ -334,6 +334,24 @@ class Tensor
...
+    Tensor<T> ret(
+        std::vector<SizeType>(std::next(tmp.shape_.begin()), tmp.shape_.end()),     /* shape */
+        std::vector<SizeType>(std::next(tmp.strides_.begin()), tmp.strides_.end()), /* stride */
+        std::vector<SizeType>(std::next(tmp.padding_.begin()), tmp.padding_.end()), /* padding */
+        tmp.storage_, tmp.offset_ + i * DimensionSize(0));
+    ret.strides_ = std::vector<SizeType>(std::next(tmp.strides_.begin()), tmp.strides_.end());
+    ret.padding_ = std::vector<SizeType>(std::next(tmp.padding_.begin()), tmp.padding_.end());
+    return ret;
+  }
+
```
pierre-wilmot-fetch-ai: Let's just make the other Slice method const, it has no reason not to be.
Then if you need to copy the underlying memory you can do t->Slice(i)->Copy()
Makes that you actually copy just what you need, rather than copying everything, extracting a slice, and having the rest of the memory sitting there unused.

libs/math/include/math/tensor_iterator.hpp
```diff
@@ -17,6 +17,8 @@
 //
 //------------------------------------------------------------------------------
 
+#include "tensor.hpp"
```
pierre-wilmot-fetch-ai: Probably not needed, tensor_iterator.hpp is always included through tensor, it's should even be available to the user for include

libs/ml/include/ml/graph.hpp
```diff
@@ -138,6 +138,50 @@ class Graph : public ops::Trainable<T>
     }
   }
 
+private:
+  template <typename OperationType, typename... Params>
+  void AddNodeImpl(std::string const &node_name, std::vector<std::string> const &inputs,
+                   std::shared_ptr<Node<ArrayType, OperationType>> op, bool trainable,
+                   Params... params)
```
pierre-wilmot-fetch-ai: Params seems to be unused here

libs/ml/include/ml/graph.hpp
```diff
@@ -138,6 +138,50 @@ class Graph : public ops::Trainable<T>
...
+   * generates a new variable name if necessary to ensure uniqueness within graph
+   * @param pre_string
+   * @return
+   */
+  template <typename OperationType>
+  std::string UpdateVariableName(std::string const &name)
+  {
+    std::string ret = name;
+    // search graph for existing variable names
+    if (ret == "")
```
pierre-wilmot-fetch-ai: if (ret.empty()) seem cleaner IMO

libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -53,21 +58,20 @@ class FullyConnected : public SubGraph<T>
...
-    }
+    ArrayPtrType weights = std::make_shared<ArrayType>(std::vector<std::uint64_t>({in, out}));
+    this->InitialiseWeights(weights);
     this->SetInput(name + "_Weights", weights);
+
     ArrayPtrType bias = std::make_shared<ArrayType>(std::vector<std::uint64_t>({1, out}));
     this->SetInput(name + "_Bias", bias);
   }
+
+  static std::string Descriptor()
```
pierre-wilmot-fetch-ai: I think all those descriptor method should be virtual rather than static, with a pure virtual defined in Ops.
That way we can still get a description of the actual op when manipulating through an Ops pointer

libs/ml/include/ml/layers/layer.hpp
```diff
@@ -0,0 +1,66 @@
...
+  Layer(std::uint64_t in, std::uint64_t out)
+    : in_size(in)
+    , out_size(out)
+  {}
+
+  /**
+   * xavier weights initialisation
+   * @param weights
+   */
+  void InitialiseWeights(ArrayPtrType weights)
```
pierre-wilmot-fetch-ai: I'd moce that bit into the weight class rather than having an intermediate layer in the hierachy.

libs/ml/include/ml/ops/activation.hpp
```diff
@@ -0,0 +1,21 @@
...
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "ml/ops/activations/relu.hpp"
+#include "ml/ops/activations/softmax.hpp"
```
pierre-wilmot-fetch-ai, line 21: This somehow breaks the C++ philosophy of pay only for what you use.
You're likely to end up including headers you don't need by doing that

libs/ml/include/ml/ops/activations/softmax.hpp
```diff
@@ -90,6 +69,11 @@ class SoftmaxLayer : public fetch::ml::Ops<T>
     }
     return {errorSignal};
   }
+
+  static std::string Descriptor()
+  {
+    return "Add";
```
pierre-wilmot-fetch-ai: Should be Softmax
+ Same Comment regarding the vistual vs static

libs/ml/include/ml/ops/multiply.hpp
```diff
@@ -0,0 +1,84 @@
...
+   * elementwise multiplication is not trainable - just pass the error signal back
+   */
+  virtual std::vector<ArrayPtrType> Backward(std::vector<ArrayPtrType> const &inputs, ArrayPtrType errorSignal)
+  {
+      std::vector<ArrayPtrType> err_sig{};
+      for (std::size_t i = 0; i < inputs.size(); ++i)
+      {
+        err_sig.emplace_back(errorSignal);
+      }
+      return err_sig;
```
pierre-wilmot-fetch-ai: This one could be a one liner 😉 

return std::vector<ArrayPtrType>(input.size(), errorSignal);

libs/ml/tests/ml/ops/softmax.cpp
```diff
@@ -16,9 +16,9 @@
 //
 //------------------------------------------------------------------------------
 
-#include "ml/ops/softmax.hpp"
 #include "core/fixed_point/fixed_point.hpp"
 #include "math/tensor.hpp"
+#include "ml/ops/activation.hpp"
```
pierre-wilmot-fetch-ai, line 21: As said before, you're now including the definition of Relu through that file even though you don't need it

libs/python/include/python/math/py_tensor.hpp
```diff
@@ -37,7 +37,8 @@ void BuildTensor(std::string const &custom_name, pybind11::module &module)
       .def("ToString", &ArrayType::ToString)
       .def("Size", &ArrayType::size)
       .def("Fill", &ArrayType::Fill)
-      .def("Slice", &ArrayType::Slice)
+      .def("Slice", [](ArrayType &a, SizeType i) { return a.Slice(i); })
+      .def("Slice", [](ArrayType const &a, SizeType i) { return a.Slice(i); })
```
pierre-wilmot-fetch-ai, line 41: Goes with previous comment regarding const Slice

pierre-wilmot-fetch-ai (changes requested):

libs/ml/include/ml/layers/layer.hpp
baykaner: I'm happy to move the initialisation to the weights class, however I'm inclined to leave the layer class as an abstract class that all layer types should inherit from to enforce a descriptor is present and hold any other common resources.

libs/ml/include/ml/ops/activation.hpp
baykaner: correct. It's here as a convenience, but generally it's preferred to us the activation you actually want. So in library header files, only include what you need, but if it's an example it's handy to just batch include all the activations to demonstrate a few.

libs/ml/tests/ml/ops/softmax.cpp
baykaner: this gets fixed in my other PR tidying up ml stuff

pierre-wilmot-fetch-ai (dismissed):

pierre-wilmot-fetch-ai (dismissed):

pierre-wilmot-fetch-ai (approved):

PR #688, opened 2019-03-02
Stability Fixes

### LINKED ISSUE #242: Block Scheduling does not happen consistently
Issue is about processing in block coordinator

### LINKED ISSUE #484: Startup Block Catchup optimisation
Current node block catch up is triggered from receiving a new block. Start up of the node can be optimised if the node on start up, triggers the synchronisation process.

### LINKED ISSUE #536: Investigate interplay between main chain manager and versioned RAS / state DB
Does forking actually work? What about switching to a different fork then back again? That old fork currently would be blasted and an attempt to check out its state hash would fail. At mininum create a unit test that tests fork switching.

### LINKED ISSUE #540: Block Coordination: Missing transaction and blocks
Add mechanism for limited the amount of time waiting for missing transactions and blocks.

### LINKED ISSUE #563: TX sync appears to be unidirectional

ejfitzgerald:
Pretest net stability fixes + HTTP request logging

* Closes #242 
* Closes #540
* Closes #563
* Closes #484 
* Closes #536


apps/constellation/constellation.cpp
```diff
@@ -251,34 +252,66 @@ void Constellation::Run(UriList const &initial_peers)
...
-  auto lane_connections_map = BuildLaneConnectionMap(cfg_.manifest, cfg_.num_lanes(), true);
+    // build the complete list of Uris to all the lane services across the internal network
+    Muddle::UriList uris;
+    uris.reserve(shard_cfgs_.size());
 
-  std::size_t const count = storage_->AddLaneConnectionsWaiting(
-      BuildLaneConnectionMap(cfg_.manifest, cfg_.num_lanes(), true), LANE_CONNECTION_TIME);
+    for (auto const &shard : shard_cfgs_)
+    {
+      uris.emplace_back(Uri{Peer{"127.0.0.1", shard.internal_port}});
```
pbukva, line 276: This feels like we are expecting that all shards are running on localhost.

libs/network/src/muddle/muddle.cpp
```diff
@@ -149,19 +155,27 @@ Muddle::ConnectionMap Muddle::GetConnections(bool direct_only)
...
+    if (!connection)
+    {
+      // do not care about connections that we are not connected too
+      continue;
+    }
 
-    if (direct_only && !entry.second.direct)
+    if (!connection->is_alive())
     {
+      // do not care about connections to whom we have yet fully connect
```
pbukva: Nit-picking (even if I would not be wrong): there is something about that sentence what does not feel right to my poorly trained English parser (perhaps "connect" -> "connect to", or "connections to whom we are not fully connected yet").

apps/constellation/constellation.cpp
```diff
@@ -251,34 +252,66 @@ void Constellation::Run(UriList const &initial_peers)
...
     FETCH_LOG_ERROR(LOGGING_NAME, "Unable to start lane server instances");
     return;
   }
+  FETCH_LOG_INFO(LOGGING_NAME, "Starting shard services...complete");
 
   /// LANE / SHARD CLIENTS
 
-  // add the lane connections
-  storage_->SetNumberOfLanes(cfg_.num_lanes());
+  {
```
pbukva, line 266: I think nesting in to new block scope is not necessary.

apps/constellation/constellation.hpp
```diff
@@ -87,7 +89,7 @@ class Constellation : public ledger::BlockSinkInterface
     uint32_t    max_peers{0};
     uint32_t    transient_peers{0};
     uint32_t    block_interval_ms{0};
-    uint32_t    block_difficulty{17};
+    uint32_t    block_difficulty{DEFAULT_BLOCK_DIFFICULTY};
```
pbukva, line 92: Note: This line sees as most appropriate to put my remotely associated comment about code in `proof_of_work.cpp` which is not part of this review since code has not been changed there:

I think there is slight miss in the `ProofOfWork::SetTarget(std::size_t zeros)` method - I think that the `-1` should be dropped to get exact number of zero bits defined by `zeros` input parameter:

```cpp
  target_ = 1;
  
  //* Instead of following line:
  //target_ <<= 8 * sizeof(uint8_t) * super_type::size() - 1 - zeros;
  
  //* This should be used (`-1` should be dropped):
  target_ <<= 8 * sizeof(uint8_t) * super_type::size() - zeros;
```

Specific example, lets say we are working with this setup to make things trivial:
 * 1 Byte long big number => `super_type::size() == 1`
 * number of zeros is 1 => `zeros=1`
 * thus MAX number `X` conforming to above defined conditions is `X = 0111 1111[bin] = 127[dec]`
 * and thus to satisfy the `X < target_`  condition in the `bool ProofOfWork::operator()()` then `target_` must be  1000 0000 [bin] = 128[dec]`

And so going back to code:
 `target_ = 1 <<  (8 * sizeof(uint8_t) * super_type::size() - zeros) = 1 << (8 * 1 * 1 - 1) = 1 << 7 = 128`
and so then `0111 1111[bin] < target_`

libs/core/src/json/document.cpp
```diff
@@ -60,7 +60,7 @@ void JSONDocument::ExtractPrimitive(Variant &variant, JSONToken const &token,
   case NUMBER_INT:
     // TODO(private issue #566): the `std::strtoll(...)` should be used here instead (converting to
     // `long long`)
-    variant = std::atoi(document.char_pointer() + token.first);
+    variant = std::atoll(document.char_pointer() + token.first);
```
pbukva, line 63: I know that this is pain in back, but `std::strtoXX(const char *str, char **str_end, ...)` has advantage when compared to `std::atoXX(...)` - it is ability to give feedback/hint whether conversion failed - via the `str_end` param:

`The functions sets the pointer pointed to by `str_end` to point to the character past the last character interpreted. If str_end is NULL, it is ignored. `

libs/ledger/include/ledger/storage_unit/lane_identity_protocol.hpp
```diff
@@ -31,11 +31,11 @@ class LaneIdentityProtocol : public service::Protocol
     GET_TOTAL_LANES,
   };
 
-  LaneIdentityProtocol(LaneIdentity *ctrl)
+  LaneIdentityProtocol(LaneIdentity &ctrl)
   {
-    this->Expose(GET_IDENTITY, ctrl, &LaneIdentity::Identity);
-    this->Expose(GET_LANE_NUMBER, ctrl, &LaneIdentity::GetLaneNumber);
-    this->Expose(GET_TOTAL_LANES, ctrl, &LaneIdentity::GetTotalLanes);
+    this->Expose(GET_IDENTITY, &ctrl, &LaneIdentity::Identity);
```
pbukva, line 36: Nit-picking: `this->` should not be necessary

libs/core/include/core/state_machine.hpp
```diff
@@ -78,14 +78,16 @@ class StateMachine : public Runnable
...
+  using Clock       = std::chrono::steady_clock;
   using Timepoint   = Clock::time_point;
+  using Duration    = Clock::duration;
   using CallbackMap = std::unordered_map<State, Callback>;
-  using Mutex       = mutex::Mutex;
+  using Mutex       = std::mutex;
 
   std::string const   logging_name_;
-  Mutex               callbacks_mutex_{__LINE__, __FILE__};
+  Mutex               callbacks_mutex_;
```
pbukva: this should be probably marked as `mutable` to signify that it does not contribute to the state of the `SateMachine<...>` instance.

libs/core/include/core/state_machine.hpp
```diff
@@ -78,14 +78,16 @@ class StateMachine : public Runnable
   StateMachine &operator=(StateMachine &&) = delete;
 
 private:
-  using Clock       = std::chrono::high_resolution_clock;
+  using Clock       = std::chrono::steady_clock;
```
pbukva, line 81: I'm miss-using this line to make comment about this class (since review does not allow me to make comment on class directly):
It would be really nice if the state machine would allow to define full transition as *triplet* {`from_state, to_state, callback`} rather than just pair {`from_state, callback`} (that is related to `void RegisterHandler(...)` methods).
Reason being that *triplet* this would completely decouple state handling/switching from external `callback` - meaning that external callbacks would not need to and would not be able to switch state - it would be completely controlled by `StateMachine` class.

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -95,6 +95,8 @@ BlockStatus MainChain::AddBlock(Block const &blk)
   auto block = std::make_shared<Block>(blk);
 
   // update the weight based on the proof and the number of transactions
+  block->weight       = 1;
+  block->total_weight = 1;
   for (auto const &slice : block->body.slices)
   {
     block->weight += slice.size();
```
pbukva, line 102: This is quite a simplification what it should be. This is the most trivial way how to estimate compression factor.
However, in normally we should use real compression factor which we can get from the list of slices in the block + we should factor in the second POW algorithm we use - leading zeoros brute-force alg.

pbukva (changes requested):
LGTM
I made some comments, please have a look and let me know what you think.
The code looks good - I can approve to get things moving and then we can discuss and possibly create issues from point raised.

libs/network/src/muddle/muddle.cpp
ejfitzgerald: Fixed in next

apps/constellation/constellation.hpp
ejfitzgerald: Raised #695

libs/core/include/core/state_machine.hpp
ejfitzgerald: Raised #696

libs/ledger/src/chain/main_chain.cpp
ejfitzgerald: Raised #697

pbukva (approved):
LGTM

PR #698, opened 2019-03-04
Execution Status

ejfitzgerald:
Add ability to query recent transaction execution status. This also includes a reworked the HTTP interface which is required for including the transaction hashes in the response.

Also add missing `-standalone` flag needed to run a single node network

I have made a series of corresponding updates to the fetch ledger api as well which I will submit in due course.

### DIFF
--- a/libs/core/include/core/state_machine.hpp
+++ b/libs/core/include/core/state_machine.hpp
@@ -74,7 +74,6 @@ public:
   }
 
   // Operators
-  // Operators
   StateMachine &operator=(StateMachine const &) = delete;
   StateMachine &operator=(StateMachine &&) = delete;
 
--- a/libs/core/include/core/state_machine.hpp
+++ b/libs/core/include/core/state_machine.hpp
@@ -74,6 +74,7 @@ public:
   }
 
   // Operators
+  // Operators
   StateMachine &operator=(StateMachine const &) = delete;
   StateMachine &operator=(StateMachine &&) = delete;
 
--- a/libs/ledger/src/chaincode/contract_http_interface.cpp
+++ b/libs/ledger/src/chaincode/contract_http_interface.cpp
@@ -21,15 +21,21 @@
 #include "core/logger.hpp"
 #include "core/serializers/stl_types.hpp"
 #include "core/string/replace.hpp"
+#include "variant/variant.hpp"
 #include "http/json_response.hpp"
 #include "ledger/chain/mutable_transaction.hpp"
 #include "ledger/chain/transaction.hpp"
 #include "ledger/chain/wire_transaction.hpp"
 #include "ledger/transaction_processor.hpp"
 
+#include <string>
+
 namespace fetch {
 namespace ledger {
 namespace {
+
+using fetch::variant::Variant;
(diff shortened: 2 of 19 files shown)


libs/core/include/core/state_machine.hpp
```diff
@@ -73,6 +73,7 @@ class StateMachine : public Runnable
     return current_state_;
   }
 
+  // Operators
```
pbukva: nit picking: delete

libs/ledger/src/chaincode/contract_http_interface.cpp
```diff
@@ -308,12 +321,15 @@ ContractHttpInterface::SubmitTxStatus ContractHttpInterface::SubmitJsonTx(
 
       MutableTransaction tx{FromWireTransaction(tx_obj)};
 
-      if (expected_contract_name && tx.contract_name() != *expected_contract_name)
+      if (!expected_contract.empty() && (tx.contract_name() != expected_contract))
```
pbukva, line 324: Nit picking: `!empty()` check should not be necessary, `tx.contract_name() != expected_contract` should do.

libs/ledger/src/chaincode/contract_http_interface.cpp
```diff
@@ -216,58 +235,53 @@ http::HTTPResponse ContractHttpInterface::OnTransaction(
...
-      oss << R"({ "submitted": false, "count": )" << submitted.processed
-          << R"(, "expected_count": )" << submitted.received
-          << R"(, "error": "Some transactions have NOT been submitted due to miss-matching contract name."})";
-    }
-    else
-    {
-      // success report the statistics
-      oss << R"({ "submitted": true, "count": )" << submitted.processed << " }";
-      error_response = false;
+      json["error"] =
```
pbukva, line 271: Caller won't get back info how many transaction passed/failed.

libs/ledger/src/transaction_status_cache.cpp
```diff
@@ -0,0 +1,107 @@
...
+    last_clean_ = now;
+  }
+}
+
+void TransactionStatusCache::PruneCache(Timepoint const &now)
+{
+  MilliTimer timer{"TxStatusCache::Prune"};
+
+  auto it = cache_.begin();
+  while (it != cache_.end())
```
pbukva, line 91: nit picking: `for(auto const& item : cache_)` feels like more concise & readable here.

pbukva (changes requested):
LGTM.
There are very minor comments - please let me know what you think.

libs/ledger/src/transaction_status_cache.cpp
pbukva: eh, never mind, disregard my comment since it it wrong.

libs/ledger/src/chaincode/contract_http_interface.cpp
pbukva: never mind, I was wrong, it actually is set in the code above.

pbukva (approved):
LGTM

PR #701, opened 2019-03-05
Feature/ml attention

baykaner:
initial implementation of a self-attention layer as described in https://papers.nips.cc/paper/7181-attention-is-all-you-need.pdf
the vector normalisation part is not implemented however awaiting upcoming PR on normalisation ops


libs/math/include/math/free_functions/fundamental_operators.hpp
```diff
@@ -463,11 +463,65 @@ template <typename ArrayType>
...
+      for (std::size_t j = 0; j < array2.shape()[1]; ++j)
+      {
+        ret.Set({i, j}, array.At({0, j}) - array2.At({i, 0}));
+      }
+    }
+  }
+  else
   {
-    ret.Set(i, array.At(i) - array2.At(i));
+    throw;
```
pierre-wilmot-fetch-ai: Throw used without specifying a object ot throw:

"Rethrows the currently handled exception. Abandons the execution of the current catch block and passes control to the next matching exception handler (but not to another catch clause after the same try block: its compound-statement is considered to have been 'exited'), reusing the existing exception object: no new objects are made. This form is only allowed when an exception is presently being handled (it calls std::terminate if used otherwise). The catch clause associated with a function-try-block must exit via rethrowing if used on a constructor."

Since there shouldn't be any curently handled exception here, I'm not sure what the behaviour is.

libs/math/include/math/tensor.hpp
```diff
@@ -322,6 +322,18 @@ class Tensor
...
+   */
+  void SetSlice(SizeType i, SelfType const &other)
+  {
+    assert(shape_.size() > 1 && i < shape_[0]);
+    for (std::size_t j = 0; j < shape_[1]; ++j)
+    {
+      this->Set({i, j}, other[j]);
+    }
+  }
+
```
pierre-wilmot-fetch-ai: This can be done at callsite with A.Slice[i].Copy(B);

libs/ml/include/ml/graph.hpp
```diff
@@ -44,9 +44,17 @@ class Graph : public ops::Trainable<T>
...
   {
-    return nodes_[nodeName]->Evaluate();
+    if (nodes_[node_name])
+    {
+      return nodes_[node_name]->Evaluate();
+    }
+    else
+    {
+      throw std::runtime_error(
+          std::string("Cannot evaluate: node [" + node_name + "] not in graph"));
```
pierre-wilmot-fetch-ai: Do we need to explicitly call std::string() here ?

libs/ml/include/ml/graph.hpp
```diff
@@ -145,7 +156,7 @@ class Graph : public ops::Trainable<T>
   {
     if (!(nodes_.find(node_name) == nodes_.end()))
```
pierre-wilmot-fetch-ai, line 202: Not really important since graph setup neer runs in a loop, but for optimisation purposes you could use std::map::insert

libs/ml/include/ml/ops/activations/softmax.hpp
```diff
@@ -44,6 +44,10 @@ class Softmax : public fetch::ml::Ops<T>
     }
 
     fetch::math::Softmax(*inputs[0], *this->output_);
+
+    std::cout << "this->output_->shape()[0]: " << this->output_->shape()[0] << std::endl;
```
pierre-wilmot-fetch-ai: Leftover debug code

libs/ml/include/ml/ops/transpose.hpp
```diff
@@ -0,0 +1,65 @@
...
+                                             ArrayPtrType                     errorSignal)
+  {
+    ASSERT(inputs.size() == 1);
+    std::shared_ptr<ArrayType> ret = std::make_shared<ArrayType>(input_shape_);
+    for (std::uint64_t i(0); i < ret->size(); ++i)
+    {
+      ret->At(i) = errorSignal->At(i);
+    }
+    return {ret};
+  }
```
pierre-wilmot-fetch-ai, line 48: That feels wrongs,
I think it should look, like 

return errorSignal->Clone()->Transpose();

libs/ml/include/ml/ops/transpose.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+  Transpose()          = default;
+  virtual ~Transpose() = default;
+
+  virtual ArrayPtrType Forward(std::vector<ArrayPtrType> const &inputs)
+  {
+    ASSERT(inputs.size() == 1);
+    input_shape_ = inputs[0]->shape();
+
+    this->output_ = std::make_shared<ArrayType>(inputs[0]->Transpose());
```
pierre-wilmot-fetch-ai: I would do it in two lines
```
this->output_ = std::make_shared<ArrayType>(inputs[0]);
this->output_->Transpose()
```
That way inputs[0] is let untouched.
We should probably also change the Ops interface from
```
virtual ArrayPtrType Forward(std::vector<ArrayPtrType> const &inputs)
// to
virtual ArrayPtrType Forward(std::vector<const ArrayPtrType> const &inputs)
```

pierre-wilmot-fetch-ai (changes requested):

libs/math/include/math/tensor.hpp
baykaner: I don't see how...
Slice() is a const function and returns a new Tensor with copied data.
so surely A.Slice(i).Copy(B) just gives you some new tensor returned from A.Slice(i) into which you copy B - but that isn't actually A?
Or did I misunderstand something?

libs/ml/include/ml/graph.hpp
baykaner: i suppose not :-p

libs/ml/include/ml/graph.hpp
baykaner: yep, you're right! we'll pick it up in a future PR

libs/ml/include/ml/ops/transpose.hpp
baykaner: hmm... you're probably right. I guess most of our code is only testing 1D cases that iterate through size so maybe it doesn't get picked up currently.

libs/ml/include/ml/ops/transpose.hpp
baykaner: okay so again similar to Slice() maybe I've misunderstood something?
Tranpose() returns a new tensor; so inputs[0]->Tranpose() shouldn't affect inputs[0]?

libs/math/include/math/tensor.hpp
pierre-wilmot-fetch-ai: Yeah, that's a bit counterIntuitive, 

Slice() is a const -- True
just gives you some new tensor returned from A.Slice(i) -- True
but that isn't actually A --False

A tensor is just a view into some memory.

A and A.Slice(i) are 2 different Tensor objects, but they point to the same underlying memory.
(Their storage pointer point to the same address)

libs/ml/include/ml/ops/transpose.hpp
baykaner: okay storage_ is shared; makes sense

pierre-wilmot-fetch-ai (approved):

PR #702, opened 2019-03-05
Packet signing and verification

bipll:
This implementation adds basic packet verification facilities:
`Muddle`s can now work in packet-signing mode;
`Router`s are `Prover`-aware, and in packet-signing mode automatically sign formatted packets and verify those incoming;
`Packet`s can now be `Sign`ed and `Verifi`ed.


libs/ledger/src/execution_manager.cpp
```diff
@@ -72,7 +73,7 @@ ExecutionManager::ExecutionManager(std::size_t num_executors, StorageUnitPtr sto
 {
   // setup the executor pool
   {
-    std::lock_guard<Mutex> lock(idle_executors_lock_);
+    FETCH_LOCK(idle_executors_lock_);
```
ejfitzgerald, line 78: 👍

libs/ledger/src/execution_manager.cpp
```diff
@@ -410,10 +415,10 @@ void ExecutionManager::MonitorThreadEntrypoint()
         // done before the thread pool dispatch)
         counters_.Set(Counters{0, slice_plan.size()});
 
+        auto self = shared_from_this();
```
ejfitzgerald, line 409: 👍

libs/network/include/network/management/network_manager.hpp
```diff
@@ -41,9 +41,8 @@ class NetworkManager
   static constexpr char const *LOGGING_NAME = "NetworkManager";
 
   NetworkManager(std::string name, std::size_t threads)
-  {
-    pointer_ = std::make_shared<implementation_type>(std::move(name), threads);
-  }
+    : pointer_{std::make_shared<implementation_type>(std::move(name), threads)}
```
ejfitzgerald, line 44: 👍

libs/network/include/network/muddle/packet.hpp
```diff
@@ -76,7 +85,7 @@ class Packet
     uint32_t direct    : 1;   ///< Flag to signal that a direct message is being sent (no routing)
     uint32_t broadcast : 1;   ///< Flag to signal that the packet is a broadcast packet
     uint32_t exchange  : 1;   ///< Flag to signal that this is an exchange packet
-    uint32_t reserved  : 1;   ///< Reserved for padding
+    uint32_t stamped   : 1;   ///< Flag to signal that the packet is signed by sender
```
ejfitzgerald, line 88: Nit picking but I think I would have renamed this to `signed`

libs/network/include/network/muddle/packet.hpp
```diff
@@ -246,77 +273,142 @@ inline Packet::Address const &Packet::GetSender() const
...
+  DropStamped();
 }
 
 inline void Packet::SetPayload(Payload const &payload)
 {
   payload_ = payload;
+  DropStamped();
+}
+
+inline void Packet::SetStamped() noexcept
```
ejfitzgerald: Maybe `SetSigned(bool set = true)` ?

libs/network/include/network/muddle/packet.hpp
```diff
@@ -246,77 +273,142 @@ inline Packet::Address const &Packet::GetSender() const
...
+{
+  RoutingHeader retVal{header_};
+  retVal.ttl = 0;
+  return *reinterpret_cast<BinaryHeader const *>(&retVal);
+}
+
+inline void Packet::Sign(crypto::Prover &prover)
+{
+  SetStamped();
+  if (prover.Sign((serializers::ByteArrayBuffer() << StaticHeader() << payload_).data()))
```
ejfitzgerald, line 367: Think I would have just left this on another line:

`serializers::ByteArrayBuffer() << StaticHeader() << payload_`

But whatever

libs/network/src/muddle/router.cpp
```diff
@@ -76,18 +77,7 @@ std::size_t GenerateEchoId(Packet const &packet)
...
-  {
-    if (a[i] != b[i])
-    {
-      equal = false;
-      break;
-    }
-  }
-
-  return equal;
+  return std::memcmp(a, b, Packet::ADDRESS_SIZE * sizeof(uint8_t)) == 0;
```
ejfitzgerald, line 80: 👍

libs/vm/src/tokeniser.cpp
```diff
@@ -1934,6 +1934,7 @@ static void yy_load_buffer_state  (yyscan_t yyscanner)
     if (b != YY_CURRENT_BUFFER){
         b->yy_bs_lineno = 1;
         b->yy_bs_column = 0;
+    }
```
ejfitzgerald: How come these changed?

apps/constellation/main.cpp
```diff
@@ -164,17 +164,18 @@ struct CommandLineArguments
...
-    p.add(args.cfg.max_peers,             "max-peers",             "The number of maximal peers to send to peer requests.",                         DEFAULT_MAX_PEERS);
-    p.add(args.cfg.transient_peers,       "transient-peers",       "The number of the peers which will be random in answer sent to peer requests.", DEFAULT_TRANSIENT_PEERS);
-    p.add(args.cfg.peers_update_cycle_ms, "peers-update-cycle-ms", "How fast to do peering changes.",                                               uint32_t{0});
+    p.add(args.cfg.max_peers,             "max-peers",             "The number of maximal peers to send to peer requests",                          DEFAULT_MAX_PEERS);
+    p.add(args.cfg.transient_peers,       "transient-peers",       "The number of the peers which will be random in answer sent to peer requests",  DEFAULT_TRANSIENT_PEERS);
+    p.add(args.cfg.peers_update_cycle_ms, "peers-update-cycle-ms", "How fast to do peering changes",                                                uint32_t{0});
+    p.add(args.cfg.sign_packets,          "sign-packets",          "Sign outbound packets (and verify those inbound)", bool{});
```
ejfitzgerald: I would change this CLI to be `disable-signing` so that the default is on.

apps/constellation/constellation.cpp
```diff
@@ -144,9 +144,11 @@ Constellation::Constellation(CertificatePtr &&certificate, Config config)
   , reactor_{"Reactor"}
   , network_manager_{"NetMgr", CalcNetworkManagerThreads(cfg_.num_lanes())}
   , http_network_manager_{"Http", HTTP_THREADS}
-  , muddle_{muddle::NetworkId{"IHUB"}, certificate, network_manager_}
+  , muddle_{muddle::NetworkId{"IHUB"}, std::move(certificate), network_manager_,
```
ejfitzgerald, line 147: If you see my comment below. I think that when signing is disabled then this should only affect the external facing muddle instances.

For all internal muddle instances we should disable signing by default.

Since these two are just the constellation (hub) muddle instances the lane shards will also need to be updated also.

libs/core/include/core/serializers/byte_array_buffer.hpp
```diff
@@ -171,7 +171,7 @@ class ByteArrayBufferEx
   }
 
   // FIXME: Incorrect naming
-  void seek(std::size_t const &p)
+  void seek(std::size_t p)
```
ejfitzgerald, line 174: 👍

ejfitzgerald (changes requested):
Really nice job, there are just a couple of CLI args changes that I would make and some possible renaming of stuff in the packet level.

Also I would not 100% rely on our unit tests. Have you tried firing up a small network of nodes and seeing if this works? Not sure that we have ever discussed how to do that. 😁

libs/network/include/network/muddle/packet.hpp
bipll: Yes, that would be more in-line with the whole sign/verify framework, it's just I've named it 'a stamp' for a time being, picturing a huge stamp jammed on a packet. %)

libs/network/include/network/muddle/packet.hpp
bipll: And drop the `DropStamped()` consequently?

libs/vm/src/tokeniser.cpp
bipll: Really strange. github shows this brace in this commit's diff but my local `git blame` attributes it to Nathan on Feb'15. Perhaps merging somehow got tangled. I'll remove it.

libs/network/include/network/muddle/packet.hpp
bipll: Oh, you wouldn't rename it to signed because `signed` is a keyword in C++ (and we don't use underscores in public struct fields.)

libs/vm/src/tokeniser.cpp
bipll: My guess is, an opening brace without a space confuses clang-format and, consequently, my bracing procedure in apply_style.py.

ejfitzgerald: Yeah a bit of a weird one

apps/constellation/constellation.cpp
ejfitzgerald: I think you probably need to apply this message signing flag to the `LaneService`'s

ejfitzgerald (changes requested):
Nice one just think we need to enable the signing by default on the `LaneService`'s and we will be in the game. 👍

libs/network/include/network/muddle/packet.hpp
bipll: You put it on a separate line, then clang-format glues it back. %-)

ejfitzgerald (approved):
Looks good to me!

PR #716, opened 2019-03-07
Very initial (v0.5) smart contract integration

ejfitzgerald:


libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -56,151 +66,224 @@ class Contract
...
+/**
+ * Register class member function transaction handler
+ *
+ * @tparam C The class type
+ * @param name The action name
+ * @param instance The pointer to the class instance
+ * @param func The member function pointer
+ */
 template <typename C>
 void Contract::OnTransaction(std::string const &name, C *instance,
```
pbukva, line 190: Just thinking: wouldn't it be safer to pass here rather `shared_ptr<C> instance` ?
Of course, it would impose till new requirement for `instance` that it would need to be created as `shared_ptr<>`.

libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -56,151 +66,224 @@ class Contract
...
-}
+  uint64_t buffer_length = buffer.size();
+  auto     status        = state().Read(std::string{key}, buffer.pointer(), buffer_length);
 
-template <typename T>
-bool Contract::GetStateRecord(T &record, byte_array::ByteArray const &address)
-{
+  // in rare cases the initial buffer might be too small in this case we need to reallocate and then
+  // re-query
+  if (vm::IoObserverInterface::Status::BUFFER_TOO_SMALL == status)
```
pbukva, line 237: Shouldn't this `if` become rather `while` ?
But what I actually wanted point out is that this test shall not be necessary at all - see my comment bellow for `Read(...)` method.

libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -56,151 +66,224 @@ class Contract
...
+  {
+    // reallocate the buffer
+    buffer.Resize(buffer_length);
 
-  // create the index that is required
-  auto index = CreateStateIndex(address);
+    // retry the read
+    status = state().Read(std::string{key}, buffer.pointer(), buffer_length);
```
pbukva, line 243: This `Read(...)` interface does not make much sense, I'm really struggling with why it is done like this? Why not pass rather `ByteArray&`, or lets if one go crazy also `std::vector<uint_8>&` or `unique_ptr<uint8_t>&`, or etc. ... , basically anything but raw pointer & length to per-allocated memory.

libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -56,151 +66,224 @@ class Contract
...
-  serializers::ByteArrayBuffer buffer(document.document);
-  buffer >> record;
+    success = true;
+    break;
+  }
+  case vm::IoObserverInterface::Status::ERROR:
+    break;
+  case vm::IoObserverInterface::Status::PERMISSION_DENIED:
+    break;
+  case vm::IoObserverInterface::Status::BUFFER_TOO_SMALL:
```
pbukva, line 261: see my 2 comments above.

libs/ledger/include/ledger/chaincode/smart_contract_exception.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+  const char *what() const noexcept override
+  {
+    if (errors_.empty())
+    {
+      return "Unknown Smart Contract Error";
+    }
+    else
+    {
+      return errors_.front().c_str();
```
pbukva, line 63: This is quite simplification, but fair enough given the `std` interface.

libs/ledger/include/ledger/chaincode/smart_contract_manager.hpp
```diff
@@ -0,0 +1,46 @@
...
+//------------------------------------------------------------------------------
+
+#include "ledger/chaincode/contract.hpp"
+
+namespace fetch {
+namespace ledger {
+
+class Identifier;
+
+class SmartContractManager : public Contract
```
pbukva, line 27: Judging solely based on `SmartContractManager` class name, it does not really feel that this class should be derived from `Contract`.
Renaming of the class would help.

libs/ledger/include/ledger/executor_interface.hpp
```diff
@@ -39,7 +39,8 @@ class ExecutorInterface
     RESOURCE_FAILURE,
     INEXPLICABLE_FAILURE,
```
pbukva, line 40: good one! :)
C++ compiler uses `Fatal` failure :)

libs/ledger/src/chaincode/contract_http_interface.cpp
```diff
@@ -153,6 +159,16 @@ ContractHttpInterface::ContractHttpInterface(StorageInterface &    storage,
     }
   }
 
+  Post("/api/contract/(digest=[a-fA-F0-9]{64})/(identifier=[a-fA-F0-9]{128})/(query=.+)",
+       [this](http::ViewParameters const &params, http::HTTPRequest const &request) {
+         // build the contract name
+         auto const contract_name =
+             ToBase64(FromHex(params["digest"])) + "." + ToBase64(FromHex(params["identifier"]));
```
pbukva, line 166: Nit picking:
```cpp
ByteArray contract_name;
contract_name.Append(ToBase64(FromHex(params["digest"])),  ".", ToBase64(FromHex(params["identifier"])));
```

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -18,38 +18,254 @@
...
+ * @return The type of the function
+ */
+Kind DetermineKind(vm::Script::Function const &fn)
 {
-  OnTransaction("create_initial_contract", this, &SmartContract::CreateInitialContract);
-  OnTransaction("invoke_contract", this, &SmartContract::Invoke);
+  Kind kind{Kind::NORMAL};
+
+  // loop through all the function annotations
+  if (1u == fn.annotations.size())
```
pbukva, line 61: Should here be rather `0u <` instead of `1u ==` ?

libs/ledger/src/state_sentinel.cpp
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * @param scope The contract context for the state variable
+ * @param key The input key to be converted
+ * @return The generated resource address for this key
+ */
+ResourceAddress StateAdapter::CreateAddress(Identifier const &scope, ConstByteArray const &key)
+{
+  FETCH_LOG_DEBUG("StateAdapter", "Creating address for key: ", key.ToBase64(),
+                  " scope: ", scope.full_name());
+  return ResourceAddress{scope.full_name() + ".state." + key};
```
pbukva, line 130: We could use here `ByteArray::Append(...)`.

libs/ledger/src/storage_unit/cached_storage_adapter.cpp
```diff
@@ -0,0 +1,229 @@
...
+  // ensure the key exists
+  auto it = cache_.find(address);
+  if (it != cache_.end())
+  {
+    value = it->second.value;
+  }
+  else
+  {
+    // sanity check
+    assert(false);
```
pbukva, line 209: this feels like overkill.  It feels like either exception or `bool CachedStorageAdapter::GetCacheEntry(CachedStorageAdapter::StateValue& state,
    ResourceAddress const &address) const` would be better?

libs/vm/include/vm/state.hpp
```diff
@@ -45,48 +46,78 @@ template <typename T>
...
     : IState(vm, type_id)
+    , name_{name}
+    , value_{value.Get<Value>()}
+    , value_type_id_{value_type_id}
+  {
+    // if we have a IO observer then
+    if (vm_->HasIoObserver())
+    {
+      // attempt to read the value from the storage engine
+      uint64_t   buffer_size = sizeof(value_);  // less important now with primitive types
```
pbukva, line 65: We had discussion about this topic (`sizeof(...)` usage is not future resistant) - perhaps //TODO would be better than just comment.

libs/vm/include/vm/state.hpp
```diff
@@ -45,48 +46,78 @@ template <typename T>
...
-  std::string name_;
+  using Value  = typename GetStorageType<T>::type;
+  using Status = IoObserverInterface::Status;
+
+  void FlushIO()
+  {
+    // if we have an IO observer then inform it of the changes
+    if (vm_->HasIoObserver())
+    {
+      vm_->GetIOObserver().Write(name_->str, &value_, sizeof(value_));
```
pbukva, line 105: as above regarding `sizeof`

pbukva (changes requested):

libs/vm/include/vm/vm.hpp
```diff
@@ -148,6 +166,27 @@ class VM
...
+  {
+    io_observer_ = &observer;
+  }
+
+  bool HasIoObserver() const
+  {
+    return io_observer_ != nullptr;
+  }
+
+  IoObserverInterface &GetIOObserver()
```
pbukva, line 179: This method could be declared as const:
`IoObserverInterface &GetIOObserver() const`

However based on my comment for `SetIOObserver(...)` method, it might not be even necessary here in VM class.

libs/vm/include/vm/vm.hpp
```diff
@@ -148,6 +166,27 @@ class VM
     return new T(this, GetTypeId<T>(), std::forward<Args>(args)...);
   }
 
+  void SetIOObserver(IoObserverInterface &observer)
```
pbukva, line 169: This feels like observer shall not be stored in the state of the `VM` class (as its data member) - it is not used anywhere in the `VM` class (besides setter and getter) - the only user/caller of the observer is `State<...>` class.

libs/ledger/include/ledger/chaincode/contract.hpp
ejfitzgerald: Yes this is mostly about convenience for the user. Given the design it is unlikely that this will be an issue (remember this is just being saved to its base class). The problem with the shared pointer approach is that is enforces you to always create on the heap.

libs/ledger/include/ledger/chaincode/contract.hpp
ejfitzgerald: The reason is to decouple. @troelsfr was keen that the VM interfaces should not really be tied to the other parts of the ledger (where possible).

I agree with this since we will want to make a number of external tools from this VM source code.

libs/ledger/include/ledger/chaincode/contract.hpp
ejfitzgerald: In the case that `BUFFER_TOO_SMALL` is returned, the function will also return a the required size. This means that this is a `if` rather than a `while`

libs/ledger/include/ledger/chaincode/smart_contract_manager.hpp
ejfitzgerald: I think the name could do with some update, but this (all all other contracts) must inherit from `Contract` otherwise the whole design falls on its head.

libs/ledger/src/chaincode/smart_contract.cpp
ejfitzgerald: Haha - this was YOUR suggested change. No I don't think so. More than one annotation not currently supported

libs/ledger/src/storage_unit/cached_storage_adapter.cpp
ejfitzgerald: This should not happen in normal cases because it is checked in the calling function. Mostly a nod, to a changing developer to be careful.

libs/vm/include/vm/state.hpp
ejfitzgerald: No need for a todo, this is what I was hoping you and I could tackle today

ejfitzgerald: In any case it is the next think on the list to support

libs/vm/include/vm/vm.hpp
ejfitzgerald: I disagree, logically the observer is at the VM level

libs/ledger/include/ledger/chaincode/contract.hpp
pbukva: That is fine, but for example `std::vector<uint8_t>&` is general enough (no dependency on anything but STL.

libs/ledger/include/ledger/chaincode/smart_contract_manager.hpp
pbukva: Inheriting from the `Contract` was not essence of my point, it was that name of inherited class (and it so it's meaning) didn't somehow match to what one would expect (contract behavior).

libs/vm/include/vm/vm.hpp
pbukva: We have had discussion about this.

libs/ledger/src/chaincode/smart_contract.cpp
pbukva: Not quite - that `More than one annotation not currently supported` is actually reflected in rather soft way by further code bellow, which takes on board just **first** annotation from provided annotations, so I would say that that suggested `0u <` is better.

pbukva (approved):
LGTM

PR #723, opened 2019-03-08
Fix Main chain not correctly flushing to disk

ejfitzgerald:


libs/ledger/src/chain/main_chain.cpp
```diff
@@ -516,10 +516,15 @@ void MainChain::WriteToFile()
...
 
       // Walk down the file to check we have an unbroken chain
-      while (LookupBlock(block->body.previous_hash, block) &&
-             !block_store_->Has(storage::ResourceID(block->body.hash)))
+      while (LookupBlockFromCache(block->body.previous_hash, block))
       {
-        block_store_->Set(storage::ResourceID(block->body.hash), *block);
+        storage::ResourceID block_storage_key{block->body.hash};
+
+        // add the block to the file storage if it
```
pbukva, line 523: Nit picking: comment is a bit unfinished, could be something like `add the block to the file storage if it is not there already`

pbukva (approved):

PR #728, opened 2019-03-11
Smart Contract Primitive Parameters

ejfitzgerald:


libs/vm/include/vm/vm.hpp
```diff
@@ -99,37 +99,153 @@ struct AssignParameters<POSITION>
...
+
+  template <typename T>
+  IfIsPtr<T, bool> Add(T const &obj)
+  {
+    bool success{false};
+
+    if (obj)
+    {
+      TypeId type_id = Getter<T>::GetTypeId(registered_types_, obj);
+
```
pbukva: This method shall have the same check as it is in method above for primitive type.

However, in that case both methods (for pointer and primitive) will have identical implementation, thus in name of avoiding redundant / copy-paste code,  it is enough to have single one and rather extend sfinae condition instead (suggested code has been successfully compiled):

```cpp
  template <typename T>
   meta::EnableIf<type_util::ConjunctionV<IfIsPrimitive<T>, IfIsPtr<T>>, bool>
  Add(T const &parameter)
  {
    bool success{false};

    // determine the type if the parameter based on the input
    TypeId type_id = Getter<T>::GetTypeId(registered_types_, parameter);

    if (type_id != TypeIds::Unknown)
    {
      // add the parameter to the list
      params_.emplace_back(parameter, type_id);

      success = true;
    }

    return success;
  }
```

pbukva (changes requested):
just for visibility

libs/core/src/byte_array/const_byte_array.cpp
```diff
@@ -0,0 +1,36 @@
...
+//
+//------------------------------------------------------------------------------
+
+#include "core/byte_array/const_byte_array.hpp"
+#include "core/byte_array/encoders.hpp"
+
+namespace fetch {
+namespace byte_array {
+
+ConstByteArray ConstByteArray::ToBase64() const
```
n-hutton, line 25: Couldn't you move most of constbytearray to a cpp while you are doing this?

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -90,6 +93,165 @@ ConstByteArray GenerateDigest(std::string const &source)
   return hash.Final();
```
n-hutton, line 93: Note: we seem to mix hash and digest as variable names etc. can we pick one?

n-hutton: smart contracts have an enum 'Kind' - this implies you get different kinds of SC, in reality it's different kinds of function. Suggest a rename (FunctionAnnotation? FunctionKind?)

n-hutton: Suggest functions have a fourth 'kind' - invalid. For example, atm more than one annotation should cause this.

libs/vm/include/vm/vm.hpp
```diff
@@ -99,37 +99,153 @@ struct AssignParameters<POSITION>
 // Forward declarations
 class Module;
 
+class ParameterPack
```
n-hutton, line 102: Needs doxygen comment - not clear what this class is for at first glance

libs/vm/include/vm/vm.hpp
```diff
@@ -99,37 +99,153 @@ struct AssignParameters<POSITION>
...
+{
+public:
+  // Construction / Destruction
+  explicit ParameterPack(RegisteredTypes const &registered_types)
+    : registered_types_{registered_types}
+  {}
+
+  ParameterPack(ParameterPack const &) = delete;
+  ParameterPack(ParameterPack &&)      = delete;
+  ~ParameterPack()                     = default;
```
n-hutton, line 112: minor: I'd move deleted operators here

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -90,6 +93,165 @@ ConstByteArray GenerateDigest(std::string const &source)
...
+
+/**
+ * Convert a byte array into a std::vector
+ *
+ * @param buffer The reference to the buffer
+ * @return The converted buffer
+ */
+std::vector<uint8_t> Convert(ConstByteArray const &buffer)
+{
+  return {buffer.pointer(), buffer.pointer() + buffer.size()};
```
n-hutton, line 136: Not a fan of this - it looks like you are using an init list but it's an iterator. Couldn't you do an assign?

pbukva (dismissed):
LGTM

pbukva (approved):
LGTM

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -90,6 +93,165 @@ ConstByteArray GenerateDigest(std::string const &source)
...
+ * Extract an address from a JSON object
+ *
+ * @param vm The pointer to the VM
+ * @param pack The parameter pack to be populated
+ * @param obj The variant to extract the address from
+ */
+void AddAddressToParameterPack(vm::VM *vm, vm::ParameterPack &pack, variant::Variant const &obj)
+{
+  // create the instance of the address
+  auto address = vm::Address::Constructor(vm, vm::TypeIds::Address);
```
n-hutton, line 190: This doesn't seem right long term. Highly coupled to the VM. Opinion: We need to come up with a serialization methodology between the VM and the outside world and the address class should be one of those.

libs/ledger/include/ledger/fetch_msgpack.hpp
```diff
@@ -0,0 +1,40 @@
...
+#pragma GCC diagnostic ignored "-Wpedantic"
+#endif
+
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wconversion"
+#pragma clang diagnostic ignored "-Wpedantic"
+#endif
+
+#include "msgpack.hpp"
```
n-hutton, line 32: this fails to find in my docker for some reason

libs/vm/include/vm/vm.hpp
```diff
@@ -99,37 +99,153 @@ struct AssignParameters<POSITION>
...
+  // Operators
+  ParameterPack &operator=(ParameterPack const &) = delete;
+  ParameterPack &operator=(ParameterPack &&) = delete;
+
+private:
+  using VariantArray = std::vector<Variant>;
+
+  RegisteredTypes const &registered_types_;
+  VariantArray           params_{};
+};
```
n-hutton, line 191: You said that this parameter pack class is only used for testing, is that right? If so, I would not put it in vm.hpp

libs/vm/include/vm/address.hpp
```diff
@@ -0,0 +1,77 @@
...
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "vm/vm.hpp"
+
+namespace fetch {
+namespace vm {
+
+class Address : public Object
```
n-hutton, line 25: needs doxy comment

libs/vm/include/vm/address.hpp
```diff
@@ -0,0 +1,77 @@
...
+{
+public:
+  static constexpr std::size_t SIZE = 64;
+  using Buffer                      = std::vector<uint8_t>;
+
+  static Ptr<Address> Constructor(VM *vm, TypeId id)
```
n-hutton, line 31: Could default id = TypeIds::Address

libs/vm/include/vm/address.hpp
```diff
@@ -0,0 +1,77 @@
...
+    , signed_tx_{signed_tx}
+  {}
+  ~Address() override = default;
+
+  bool HasSignedTx() const
+  {
+    return signed_tx_;
+  }
+
+  void SetSignedTx(bool set = true)
```
n-hutton, line 48: set does nothing here

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -90,6 +93,165 @@ ConstByteArray GenerateDigest(std::string const &source)
...
+
+/**
+ * Extract the a given type from the container type and insert it into the parameter pack
+ *
+ * @tparam T The type to be extracted
+ * @param pack The parameter pack to populate
+ * @param obj The source object to extract the value from
+ */
+template <typename T>
+void AddToParameterPack(vm::ParameterPack &pack, variant::Variant const &obj)
```
n-hutton, line 122: this code is odd. Surely the vm should handle parsing of parameters?

libs/vm/include/vm/vm.hpp
n-hutton: note: state.hpp - this does a flushIO in its destructor. I do not think we want to do this.

libs/vm/include/vm/map.hpp
```diff
@@ -79,14 +79,12 @@ struct Map : public IMap
...
-  Map()          = delete;
-  virtual ~Map() = default;
-
   Map(VM *vm, TypeId type_id)
     : IMap(vm, type_id)
   {}
+  ~Map() override = default;
 
-  virtual int32_t Count() const override
+  int32_t Count() const override
```
n-hutton, line 87: uint ?

libs/ledger/tests/chaincode/smart_contract_tests.cpp
```diff
@@ -318,4 +316,94 @@ TEST_F(SmartContractTests, CheckQueryReturnTypes)
...
+    Variant response;
+    EXPECT_EQ(SmartContract::Status::OK, SendQuery("offset", request, response));
+
+    // check the response is as we expect
+    ASSERT_TRUE(response.Has("result"));
+    EXPECT_EQ(response["result"].As<int32_t>(), 130);
+
+    ASSERT_TRUE(response.Has("status"));
+    EXPECT_EQ(response["status"].As<ConstByteArray>(), "success");
+  }
```
n-hutton, line 406: Might be useful to create a 'bad access' test

PR #734, opened 2019-03-12
Feature/on init decorator

n-hutton:


libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -76,7 +74,7 @@ class Contract
 
   /// @name Contract Lifecycle Handlers
   /// @{
-  void Attach(vm::IoObserverInterface &state);
+  void Attach(ledger::StateAdapter &state);
```
n-hutton, line 77: switch to contract having the state adapter - it will almost cert. need to access extended functionality like setting the scope as I do here.

libs/ledger/include/ledger/chaincode/smart_contract_manager.hpp
```diff
@@ -40,6 +46,13 @@ class SmartContractManager : public Contract
   /// @{
   Status OnCreate(Transaction const &tx);
   /// @}
+
+  using Script    = vm::Script;
```
n-hutton, line 50: this is now duplicated by both SC and SC_manager - needs thinking about

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -41,56 +42,87 @@ using fetch::byte_array::FromBase64;
...
 /**
- * Determine the type of the VM function
+ * Construct a smart contract from the specified source
  *
- * @param fn The reference to function entry
- * @return The type of the function
+ * @param source Reference to the script text
  */
-Kind DetermineKind(vm::Script::Function const &fn)
+SmartContract::SmartContract(std::string const &source)
```
n-hutton, line 64: put this at top - easier when opening the file to find the constructor

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -87,7 +92,92 @@ Contract::Status SmartContractManager::OnCreate(Transaction const &tx)
...
+
+  auto tx_signatures = tx.signatures();
+
+  if (tx_signatures.size() != 1)
+  {
+    FETCH_LOG_WARN(LOGGING_NAME, "Only one signature allowed when setting up smart contract");
+    return Status::FAILED;
+  }
+
+  auto pub_key_b64 = tx_signatures.begin()->first.identifier().ToBase64();
```
n-hutton, line 143: non-great here

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -87,7 +92,92 @@ Contract::Status SmartContractManager::OnCreate(Transaction const &tx)
...
+
+  Identifier scope;
+  if (!scope.Parse(calculated_hash + "." + pub_key_b64))
+  {
+    FETCH_LOG_WARN(LOGGING_NAME, "Failed to parse scope for smart contract");
+    return Status::FAILED;
+  }
+
+  state().PushContext(scope);
```
n-hutton, line 152: Push new scope/context when changing

libs/vm/include/vm/state.hpp
```diff
@@ -71,9 +71,7 @@ class State : public IState
   }
 
   ~State() override
-  {
-    FlushIO();
-  }
+  {}
```
n-hutton, line 74: Don't want to auto-flush state (let the user control this). Plus, this is pointless and dangerous if there is a bad access during program termination.

libs/ledger/src/storage_unit/cached_storage_adapter.cpp
```diff
@@ -190,8 +186,6 @@ void CachedStorageAdapter::AddCacheEntry(ResourceAddress const &address, StateVa
 CachedStorageAdapter::StateValue CachedStorageAdapter::GetCacheEntry(
     ResourceAddress const &address) const
 {
```
n-hutton, line 188: I can't see that locking is neccesary - the VM is single threaded

PR #737, opened 2019-03-12
Forward batch

pierre-wilmot-fetch-ai:


libs/math/include/math/tensor_operations.hpp
```diff
@@ -0,0 +1,43 @@
+#pragma once
```
baykaner, line 1: I think this should live together with all the other matrix operations (currently in free_functions/matrix_operations). Perhaps we should gradually move those here rather than the other way around though.

libs/math/include/math/tensor_operations.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include "core/assert.hpp"
+#include "math/tensor.hpp"
+#include <memory>
+#include <vector>
+
+namespace fetch {
+namespace math {
+
+template <typename T>
```
baykaner, line 33: please add a function comment with parameter descriptors

libs/math/include/math/tensor_operations.hpp
```diff
@@ -0,0 +1,43 @@
...
+  std::vector<typename fetch::math::Tensor<T>::SizeType> retSize;
+  retSize.push_back(tensors.size());
+  retSize.insert(retSize.end(), tensors.front().shape().begin(), tensors.front().shape().end());
+  fetch::math::Tensor<T> ret(retSize);
+  for (typename fetch::math::Tensor<T>::SizeType i(0); i < tensors.size(); ++i)
+  {
+    ret.Slice(i).Copy(tensors[i]);
+  }
+  return ret;
+}
```
baykaner, line 45: Either this function should live inside the namespace fetch::math::details (if you don't intend for other developers to call it), or you should set up two interfaces to the function like this:

void ConcatenateTensors(Tensor &x, Tensor& y, Tensor &ret)
{
// do the work;
}

Tensor ConcatenateTensors(Tensor &x, Tensor &y)
{
Tensor ret{x.shape()};
ConcatenateTensors(x, y, ret);
return ret;
}

libs/ml/examples/basic/main.cpp
```diff
@@ -57,11 +57,10 @@ int main(int ac, char **av)
 
   CrossEntropy<ArrayType> criterion;
 
-  std::pair<size_t, std::shared_ptr<ArrayType>> input;
-  std::shared_ptr<ArrayType>                    gt =
-      std::make_shared<ArrayType>(std::vector<typename ArrayType::SizeType>({1, 10}));
+  std::pair<size_t, ArrayType> input;
```
baykaner: std::size_t

libs/ml/include/ml/ops/activations/softmax.hpp
```diff
@@ -35,40 +35,37 @@ class Softmax : public fetch::ml::Ops<T>
   Softmax()          = default;
   virtual ~Softmax() = default;
 
-  virtual ArrayPtrType Forward(std::vector<ArrayPtrType> const &inputs)
+  virtual ArrayType ForwardBatch(std::vector<std::reference_wrapper<ArrayType const>> const &inputs)
+  {
+    // Batch detection is handled in fetch::math::Softmax
```
baykaner: hmm... perhaps it shouldn't be though? Any opinion on this?
Personally I don't like the fact that in the math library we assume that a particular dimension refers to the batch; it introduces more opportunities for error just by having a tensor transposed the wrong way. If we rework the math operations to require dimensionality {1, N} then we pass slices from within the ML library we can't make that error.

baykaner (dismissed):

libs/ml/include/ml/ops/activations/softmax.hpp
pierre-wilmot-fetch-ai: Agreed, I've just been too lazy to fix it

libs/math/include/math/tensor_operations.hpp
pierre-wilmot-fetch-ai: For now I don't need methods with these signatures, I'll add them if/when I need them.
Else it's gonna be unused / badly tested code laying around

baykaner (approved):

PR #740, opened 2019-03-12
Fix Chain Sync Issues

ejfitzgerald:


libs/ledger/include/ledger/chain/constants.hpp
```diff
@@ -22,6 +22,8 @@
 namespace fetch {
 namespace ledger {
 
+static const uint64_t FINALITY_PERIOD = 50;
```
pbukva: You can replace `static const` with `constexpr`:

```cpp
constexpr uint64_t FINALITY_PERIOD = 50;
```

apps/constellation/constellation.cpp
```diff
@@ -336,9 +336,12 @@ void Constellation::Run(UriList const &initial_peers)
   // monitor loop
   while (active_)
   {
+    // determine the status of the main chain server
+    bool const is_in_sync =
+        ledger::MainChainRpcService::State::SYNCHRONISED == main_chain_service_->state();
```
n-hutton, line 341: Minor: this is a bit verbose

n-hutton (dismissed):

pbukva (changes requested):
LGTM
Just look for inline comments and let me know (nothing critical only tweaks, PR can go as is).

pbukva (dismissed):
LGTM

pbukva (approved):

PR #743, opened 2019-03-14
Improved Deployment Updates

ejfitzgerald:
- Added `-v` or `--version` flag support to the main application
- Added new API `/api/status/states` to query the various state machine states


pbukva (approved):
LGTM

PR #744, opened 2019-03-14
Common Backward Batch implementation

pierre-wilmot-fetch-ai:


libs/ml/include/ml/ops/ops.hpp
```diff
@@ -74,6 +94,32 @@ class BatchOps : public Ops<T>
...
+        results[i].push_back(ret[i]);
+      }
+    }
+    std::vector<ArrayType> concatenatedResults;
+    for (auto const &tensorList : results)
+    {
+      concatenatedResults.push_back(ConcatenateTensors(tensorList));
+    }
+    return concatenatedResults;
+  }
```
baykaner, line 122: great - but does it work? do we have any unit tests for batched vs non-batched training?

baykaner (approved):

PR #746, opened 2019-03-14
Feature/on init decorator draft

n-hutton:
add on init functionality, the corresponding python script to drive it at n-hutton:feature/lock_resources_for_init

the main thing this is missing is on_init being able to take parameters, as pointed out in one of my comments.

This should now support a token contract fully, with minor workarounds.


libs/core/include/core/mutex.hpp
```diff
@@ -195,7 +195,7 @@ class DebugMutex : public AbstractMutex
 #ifdef NDEBUG
 using Mutex = ProductionMutex;
 #else
-using Mutex = DebugMutex;
+using Mutex = ProductionMutex;
```
n-hutton: not intentional - I'll remove this.

libs/ledger/include/ledger/chain/mutable_transaction.hpp
```diff
@@ -60,15 +60,20 @@ void Deserialize(T &serializer, Signature &b)
 
 struct TransactionSummary
 {
-  using Resource     = byte_array::ConstByteArray;
-  using TxDigest     = byte_array::ConstByteArray;
-  using ContractName = byte_array::ConstByteArray;
-  using ResourceSet  = std::set<Resource>;
-  using Fee          = uint64_t;
+  using Resource       = byte_array::ConstByteArray;
```
n-hutton, line 63: Added 'contract hashes' to mutable tx - these are raw hashes provided by the client

libs/ledger/include/ledger/state_adapter.hpp
```diff
@@ -50,36 +57,18 @@ class StateAdapter : public vm::IoObserverInterface
   Status Exists(std::string const &key) override;
   /// @}
 
-protected:
-  StorageInterface &storage_;
-  Identifier        scope_;
-};
+  void PushContext(Identifier const &scope);
```
n-hutton: In the case that there is another context you want to go to, you can push and pop it on. Reads and writes to "value", get prepended with the current state and checked against allowed accesses.

libs/ledger/include/ledger/storage_unit/cached_storage_adapter.hpp
```diff
@@ -75,9 +74,8 @@ class CachedStorageAdapter : public StorageInterface
 
   /// @name Cache Data
   /// @{
-  mutable Mutex lock_{__LINE__, __FILE__};
-  Cache         cache_{};                ///< The local cache
-  bool          flush_required_{false};  ///< Top level cache flush flag
+  Cache cache_{};                ///< The local cache
```
n-hutton: VM is single threaded - I don't think there needs to be locking here

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -329,7 +298,7 @@ Contract::Status SmartContract::InvokeAction(std::string const &name, Transactio
   std::vector<msgpack::object> input_params;
 
   // if the tx has a payload parse it
-  if (!tx.data().empty())
+  if (!tx.data().empty() && tx.data() != "{}")
```
n-hutton: I can now send message pack args to the SC using ledger-api. However, this does not yet work for contract_manager/on_init because the data field in the TX is used to hold the source. This needs a little sorting out (probably message pack the source + args in a meaningful way).

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -87,11 +91,103 @@ Contract::Status SmartContractManager::OnCreate(Transaction const &tx)
...
+      if (on_init_function.size() > 0)
+      {
+        FETCH_LOG_WARN(LOGGING_NAME, "More than one init function found in SC. Terminating.");
+        return Status::FAILED;
+      }
+      FETCH_LOG_WARN(LOGGING_NAME, "Found init function for SC");
+      on_init_function = fn.name;
+      break;
+
+    case vm::Kind::INVALID:
```
n-hutton, line 136: Add functionality to notice invalid function decorators.

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -87,11 +91,103 @@ Contract::Status SmartContractManager::OnCreate(Transaction const &tx)
...
+  }
+
+  FETCH_LOG_WARN(LOGGING_NAME, "Adding smart contract, ID: ", calculated_hash,
+                 " on_init called: ", on_init_function.size() != 0);
+
+  state().PopContext();
 
-  // store the contract
+  // store the contract TODO(HUT): this needs to check for success too.
+  auto success = true;
```
n-hutton, line 166: this is a bug! It is not checked that writing the SC itself to the storage works or was locked correctly.

libs/ledger/src/execution_manager.cpp
```diff
@@ -170,6 +171,11 @@ bool ExecutionManager::PlanExecution(Block::Body const &block)
             StateAdapter::CreateAddress(contract_id, resource).lane(block.log2_num_lanes));
       }
 
+      for (auto const &contract_hash : tx.contract_hashes)
```
n-hutton: I think it might make more sense to construct a state adapter, pass it the TX and ask for the lanes that will be locked.

libs/ledger/src/executor.cpp
```diff
@@ -109,8 +111,11 @@ Executor::Status Executor::Execute(TxDigest const &hash, std::size_t slice, Lane
       // detach the chain code from the current context
       chain_code->Detach();
 
-      // force the flushing of the cache
-      storage_cache.Flush();
+      // force the flushing of the cache *** only if the contract was successful ***
+      if (result == Contract::Status::OK)
```
n-hutton, line 115: Need to watch out for this! The cache must not write back until everything has come back 100% successful. My original intention for the cache was not efficiency, but safety. This makes it mandatory to provide to the state adapter.

libs/ledger/src/state_adapter.cpp
```diff
@@ -48,8 +76,17 @@ StateAdapter::Status StateAdapter::Read(std::string const &key, void *data, uint
 {
   Status status{Status::ERROR};
 
+  std::string new_key{scope_.back().full_name() + ".state." + std::string{key}};
```
n-hutton: The state adapt. now wraps state accesses with whatever the appropriate scope is at the time.

libs/ledger/tests/chaincode/smart_contract_tests.cpp
```diff
@@ -130,7 +130,7 @@ TEST_F(SmartContractTests, CheckSimpleContract)
     // from the action
     EXPECT_CALL(*storage_, Lock(expected_resource));
     EXPECT_CALL(*storage_, Get(expected_resource));
-    EXPECT_CALL(*storage_, Set(expected_resource, expected_value)).Times(2);
+    EXPECT_CALL(*storage_, Set(expected_resource, expected_value)).Times(1);
```
n-hutton, line 133: We need to decide whether the state will auto-write back on destruct. I do not think it should.

libs/vm_modules/include/vm_modules/core/print.hpp
```diff
@@ -26,9 +26,13 @@ namespace vm_modules {
 static void Print(fetch::vm::VM *vm, fetch::vm::Ptr<fetch::vm::String> const &s)
 {
   vm->AddOutputLine(s->str);
+
+#ifndef NDEBUG
```
n-hutton, line 30: lets keep this in, it's quite helpful when doing this stuff

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -38,59 +39,150 @@
...
+  fetch::crypto::SHA256 hash;
+  hash.Update(source);
+  return hash.Final();
+}
 
-  // loop through all the function annotations
-  if (1u == fn.annotations.size())
+void ValidateAddressesInParams(Transaction const &tx, vm::ParameterPack const &params)
+{
+  // This doesn't work with a set (???)
```
n-hutton, line 68: Look into this! inserting into a set does nothing, it seems

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -145,7 +237,7 @@ std::vector<uint8_t> Convert(ConstByteArray const &buffer)
  */
 void AddAddressToParameterPack(vm::VM *vm, vm::ParameterPack &pack, msgpack::object const &obj)
 {
-  static int8_t const   ADDRESS_ID   = static_cast<int8_t>(0xAD);
+  static uint8_t const  ADDRESS_ID   = static_cast<uint8_t>(0x4d);  // 77
```
n-hutton, line 220: AD exceeds the allowed bounds for message pack types

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -403,6 +471,63 @@ Contract::Status SmartContract::InvokeAction(std::string const &name, Transactio
...
+  auto vm = vm_modules::VMFactory::GetVM(module_);
+  vm->SetIOObserver(state());
+
+  FETCH_LOG_WARN(LOGGING_NAME, "Running SC init function: ", name);
+
+  vm::ParameterPack params{vm->registered_types()};
+
+  // lookup the function / entry point which will be executed
+  Script::Function const *target_function = script_->FindFunction(name);
+  if (target_function->num_parameters == 1)
```
n-hutton, line 446: on init will auto populate yourself as the address if that is valid (to remain compatible with waqqas' stuff)

libs/ledger/src/state_adapter.cpp
```diff
@@ -86,12 +91,18 @@ StateAdapter::Status StateAdapter::Read(std::string const &key, void *data, uint
...
-  FETCH_UNUSED(data);
-  FETCH_UNUSED(size);
+  if (!enable_writes_)
+  {
+    return Status::OK;
+  }
 
-  // this operation is not supported in this version of the adapter
-  return Status::PERMISSION_DENIED;
+  auto new_key = WrapKeyWithScope(key);
```
n-hutton, line 96: needs a little bit of work because the child wraps the key and then calls the parent which also wraps the key

libs/vm/include/vm/defs.hpp
```diff
@@ -76,6 +77,15 @@ struct IsVariant<T, typename std::enable_if_t<std::is_base_of<Variant, T>::value
 {
 };
 
+template <typename T, typename = void>
```
n-hutton, line 80: I was forced to put this in so Address would be a serializable exception

libs/vm/include/vm/module/base.hpp
```diff
@@ -71,6 +71,16 @@ struct TypeGetter<T, typename std::enable_if_t<IsPtr<T>::value>>
   }
 };
 
+template <typename T>
```
n-hutton, line 74: Forced to make exception for Address - needs review

libs/vm/include/vm/state.hpp
```diff
@@ -42,12 +42,61 @@ class IState : public Object
   static Ptr<IState> Construct(VM *vm, TypeId type_id, Args &&... args);
 };
 
+template <typename T, typename = std::enable_if_t<std::is_arithmetic<T>::value>>
```
n-hutton, line 45: read/write helpers since need an exceptional case for Address. Also, needs to check if Exists for Address etc.

libs/vm/src/analyser.cpp
```diff
@@ -1552,14 +1552,15 @@ bool Analyser::AnnotateInvokeOp(ExpressionNodePtr const &node)
...
+
     FunctionGroupPtr fg = ConvertToFunctionGroupPtr(symbol);
     TypePtrArray     actual_parameter_types;
     FunctionPtr f = FindFunction(lhs->type, fg, supplied_parameter_types, actual_parameter_types);
     if (f == nullptr)
     {
       // No matching constructor, or ambiguous
       AddError(lhs->token,
-               "unable to find matching constructor for type '" + lhs->type->name + "'");
+               "unable to find matching constructor for type/function '" + lhs->type->name + "'");
```
n-hutton, line 1563: Need to beef up this debug - I thought State<>(one_arg) still worked

libs/vm_modules/include/vm_modules/core/type_convert.hpp
```diff
@@ -27,18 +27,28 @@ template <typename T>
...
+static void CreateToString(std::shared_ptr<fetch::vm::Module> &module)
 {
   module->CreateFreeFunction("toString", &toString<int32_t>);
   module->CreateFreeFunction("toString", &toString<uint32_t>);
   module->CreateFreeFunction("toString", &toString<int64_t>);
   module->CreateFreeFunction("toString", &toString<uint64_t>);
   module->CreateFreeFunction("toString", &toString<float_t>);
   module->CreateFreeFunction("toString", &toString<double_t>);
+  module->CreateFreeFunction("toString", &toString<bool>);
+  /* module->CreateFreeFunction("toString", &toStringClass<fetch::vm::Address>); */
```
n-hutton: It would be really great if string concat called toString

cmake/BuildTargets.cmake
```diff
@@ -70,6 +70,9 @@ macro(setup_compiler)
   # warnings
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wconversion -Wpedantic")
 
+  # Suppress visibility link warnings
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden")
```
n-hutton, line 74: ld warnings are effecting my workflow

ejfitzgerald: I am sure that this caused issues with a previous piece of code. I think that has been deleted now, lets see what happens 😬

libs/ledger/include/ledger/chain/mutable_transaction.hpp
```diff
@@ -104,13 +117,13 @@ struct TransactionSummary
 template <typename T>
 void Serialize(T &serializer, TransactionSummary const &b)
 {
-  serializer << b.resources << b.fee << b.transaction_hash << b.contract_name;
+  serializer << b.resources << b.raw_resources << b.fee << b.transaction_hash << b.contract_name;
```
ejfitzgerald, line 120: When this commit lands can you make sure that you add something like BREAKING into the commit description (probably not the title). It should remind me to update the version number for this.

libs/ledger/include/ledger/chain/mutable_transaction.hpp
```diff
@@ -351,6 +369,15 @@ class MutableTransaction
       hash.Update(e);
     }
 
+    std::vector<byte_array::ConstByteArray> raw_resources;
```
ejfitzgerald, line 374: Since we know the size call `reserve`?

libs/ledger/include/ledger/chain/mutable_transaction.hpp
```diff
@@ -495,9 +533,42 @@ void Serialize(T &stream, TxSigningAdapter<MUTABLE_TX> const &tx)
...
+
+  uint16_t success_counter = 1;
+
+  auto deleter = [&](uint16_t *dummy) {
+    if (success_counter != 0)
+    {
+      FETCH_LOG_INFO(log_name, "Failed to deser. tx! Counter: ", success_counter);
+    };
+  };
+  std::unique_ptr<uint16_t, decltype(deleter)> on_exit(&success_counter, deleter);
```
ejfitzgerald: You have used this pattern a few times, I would suggest make a template class which can take the lambda directly for this scope exit stuff

libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -76,7 +73,7 @@ class Contract
 
   /// @name Contract Lifecycle Handlers
   /// @{
-  void Attach(vm::IoObserverInterface &state);
+  void Attach(ledger::StateAdapter &state);
```
ejfitzgerald, line 76: I would still probably have derived a new interface for this just to aid testing

libs/ledger/include/ledger/storage_unit/cached_storage_adapter.hpp
ejfitzgerald: That is true, however, we might want to use it somewhere else in the future. Don't change it could you just add a note in the class description. (for the autogenerated documentation we are produce at some point 😉 )

libs/ledger/src/chaincode/smart_contract.cpp
ejfitzgerald: Yep think it is probably because `ConstByteArray` doesn't have `<` operator. Either way I probably would have used an `unordered_set` anyway

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -38,59 +39,130 @@
...
     {
-      kind = Kind::ACTION;
+      auto &var = *(params[i].Get<vm::Ptr<vm::Address>>());
+
+      ConstByteArray address_data{var.GetBytes().data(), var.GetBytes().size()};
+
+      if (std::find(valid_addresses.begin(), valid_addresses.end(), address_data) !=
+          valid_addresses.end())
+      {
+        var.SetSignedTx(true);
```
ejfitzgerald, line 87: Nice one dude! 👍

libs/ledger/src/chaincode/smart_contract.cpp
ejfitzgerald: 👍

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -326,18 +340,20 @@ SmartContract::SmartContract(std::string const &source)
  */
 Contract::Status SmartContract::InvokeAction(std::string const &name, Transaction const &tx)
 {
+  // Important to keep the handle alive as long as the msgpack::object is needed to avoid segfault!
+  msgpack::object_handle       h;
```
ejfitzgerald, line 344: 👍

libs/ledger/src/chaincode/smart_contract.cpp
ejfitzgerald: Maybe it is already done, but I would add a check to the compile stage, so in the constructor to not allow anything other than an `Address` as the first parameter (if one is specified)

libs/ledger/src/chaincode/smart_contract_manager.cpp
ejfitzgerald: Lets raise an issue but don't stop the mergin' 😁

libs/ledger/src/state_adapter.cpp
```diff
@@ -35,6 +35,9 @@ StateAdapter::StateAdapter(StorageInterface &storage, Identifier scope)
   , scope_{std::move(scope)}
 {}
 
+StateAdapter::~StateAdapter()
```
ejfitzgerald: Can be made default

libs/ledger/tests/chaincode/smart_contract_tests.cpp
ejfitzgerald: I agree - nice change

libs/vm/include/vm/defs.hpp
ejfitzgerald: I guess I don't really understand can we discuss tomorrow?

libs/vm/include/vm/state.hpp
```diff
@@ -62,18 +108,15 @@ class State : public IState
...
-      uint64_t   buffer_size = sizeof(value_);  // less important now with primitive types
-      auto const status      = vm_->GetIOObserver().Read(name_->str, &value_, buffer_size);
+      auto const status = ReadHelper(name_->str, &value_, vm_->GetIOObserver());
 
       // mark the variable as existed if we get a positive result back
       existed_ = (Status::OK == status);
     }
   }
 
   ~State() override
```
ejfitzgerald: `= default`

libs/vm/include/vm/state.hpp
ejfitzgerald: I am not 💯 sure that this is necessary - lets talk tomorrow

libs/vm/src/analyser.cpp
ejfitzgerald: Sorry I removed that in the first PR, mostly for my own debugging

libs/vm_modules/include/vm_modules/core/print.hpp
ejfitzgerald: Hmmm, this feels like it should be a cmake option or something really. Debug mode is fine I guess. Can we change the logging name to just `"VM"` and remove the `***` that sort of just triggers my OCD (sorry) 😬

libs/vm_modules/include/vm_modules/core/type_convert.hpp
```diff
@@ -27,18 +27,27 @@ template <typename T>
...
-void CreateToString(std::shared_ptr<fetch::vm::Module> &module)
+static void CreateToString(std::shared_ptr<fetch::vm::Module> &module)
 {
   module->CreateFreeFunction("toString", &toString<int32_t>);
   module->CreateFreeFunction("toString", &toString<uint32_t>);
   module->CreateFreeFunction("toString", &toString<int64_t>);
   module->CreateFreeFunction("toString", &toString<uint64_t>);
   module->CreateFreeFunction("toString", &toString<float_t>);
   module->CreateFreeFunction("toString", &toString<double_t>);
+  module->CreateFreeFunction("toString", &toString<bool>);
```
ejfitzgerald, line 50: 👍 although I dislike the inconsistency of `toString` and `Print` (don't change that I am just winging, maybe a candidate for a FPR' 😁)

ejfitzgerald (changes requested):
Really nice job dude, just a couple of small requests from my side. I will pull my finger out and merge in the state mapper tomorrow and we will be in the game.

ejfitzgerald:
I wasn't sure if you wanted me to review this yet. Either way, surprise 😁

libs/ledger/include/ledger/storage_unit/cached_storage_adapter.hpp
n-hutton: change reverted with no comment

libs/ledger/src/chaincode/smart_contract_manager.cpp
n-hutton: ticket #767

libs/vm_modules/include/vm_modules/core/print.hpp
n-hutton: Gah. fine.

libs/ledger/src/chaincode/smart_contract.cpp
n-hutton: I would have thought that the set template would scream... Also, it does have the < operator I believe

libs/ledger/src/chaincode/smart_contract.cpp
n-hutton: I suggest to raise this as a ticket if you want that feature - in my mind this is transient since init should accept args normally like ```@action```. If the args are not an Address it will fail in the usual way with mismatched number of parameters.

libs/vm/include/vm/defs.hpp
n-hutton: Hit me up on slack

libs/vm/include/vm/state.hpp
n-hutton: hit me up on slack

cmake/BuildTargets.cmake
n-hutton: great

libs/ledger/include/ledger/chain/mutable_transaction.hpp
n-hutton: ok

libs/ledger/include/ledger/chain/mutable_transaction.hpp
n-hutton: added, also reserving for regular resources.

libs/ledger/include/ledger/chain/mutable_transaction.hpp
n-hutton: Maybe in future. I am going to remove it in this PR since it's not a safe pattern to use

libs/ledger/include/ledger/chaincode/contract.hpp
n-hutton: Chat on slack?

libs/ledger/src/state_adapter.cpp
n-hutton: done

libs/vm/include/vm/state.hpp
n-hutton: do you refer to the destructor here?

libs/vm_modules/include/vm_modules/core/type_convert.hpp
n-hutton: Agreed. I think that free functions like this should be ALL_CAPS

libs/ledger/src/chaincode/smart_contract.cpp
ejfitzgerald: Yep please raise a ticket and reference here

ejfitzgerald (approved):

ejfitzgerald:
LGTM

PR #766, opened 2019-03-19
Static linking for OpenSSL

pbukva:
* Set the  -dFETCH_STATIC_LINK=ON on cmake command-line to enable
static linking


cmake/BuildTargets.cmake
```diff
@@ -151,8 +151,15 @@ macro(setup_compiler)
 endmacro(setup_compiler)
 
 function(configure_vendor_targets)
+  #set(OPENSSL_USE_STATIC_LIBS TRUE)
+  if(OPENSSL_USE_STATIC_LIBS)
```
ejfitzgerald: I think this needs to be added as an explicit cmake option. Something like `FETCH_STATIC_BUILD` or similar

cmake/BuildTargets.cmake
```diff
@@ -151,8 +151,15 @@ macro(setup_compiler)
...
+    message(STATUS "OpenSSL linked STATICALLY")
+  else()
+    message(STATUS "OpenSSL linked DYNAMICALLY. Static linking can be achieved by setting `-DOPENSSL_USE_STATIC_LIBS=TRUE`.")
```
ejfitzgerald: If you add the option lets remove this helper text to we have something more general like:

`"OpenSSL Link: Static"`
`"OpenSSL Link: Shared"`

cmake/BuildTargets.cmake
```diff
@@ -151,8 +151,15 @@ macro(setup_compiler)
...
+  endif()
 
   find_package(OpenSSL REQUIRED)
+  find_package(Threads)
```
ejfitzgerald, line 161: For my curiosity why?

cmake/BuildTargets.cmake
```diff
@@ -161,7 +168,13 @@ function(configure_vendor_targets)
 
   # OpenSSL
   add_library(vendor-openssl INTERFACE)
-  target_link_libraries(vendor-openssl INTERFACE ${OPENSSL_LIBRARIES})
+
+  if(CMAKE_THREAD_LIBS_INIT)
+    target_link_libraries(vendor-openssl INTERFACE ${OPENSSL_LIBRARIES} ${CMAKE_DL_LIBS} ${CMAKE_THREAD_LIBS_INIT})
```
ejfitzgerald: This is a little more confusing to read this way. How about:

```cmake
target_link_libraries(vendor-openssl INTERFACE ${OPENSSL_LIBRARIES} ${CMAKE_DL_LIBS})

# comment about this inclusion funky, presumably static pthread support?
if (CMAKE_THREAD_LIBS_INIT)
    target_link_libraries(vendor-openssl INTERFACE ${CMAKE_THREAD_LIBS_INIT})
endif (CMAKE_THREAD_LIBS_INIT)

ejfitzgerald (changes requested):

cmake/BuildTargets.cmake
pbukva: OpenSSL depends on it - it is visible from `target_link_libraries(..., ${CMAKE_THREAD_LIBS_INIT})` bellow.

cmake/BuildTargets.cmake
pbukva: I was thinking about doing it this way, but then I realised that this would force us to make changes for full static build (no dynamic libraries at all).
Thus we would need to do a bit more then handle just OpenSSL linking.

ejfitzgerald (approved):

PR #771, opened 2019-03-19
Feature/text loader updates

baykaner:


libs/math/include/math/free_functions/ml/activation_functions/sigmoid.hpp
```diff
@@ -35,17 +35,33 @@ template <typename ArrayType>
...
+    {
+      Exp(val, ret.At(idx));
+      Divide(ret.At(idx), ret.At(idx) + typename ArrayType::Type(1.0), ret.At(idx));
+    }
+    ++idx;
+  }
+  //  Multiply(typename ArrayType::Type(-1.0), t, ret);
+  //  Exp(ret);
+  //  Add(ret, typename ArrayType::Type(1.0), ret);
+  //  Divide(typename ArrayType::Type(1.0), ret, ret);
```
pierre-wilmot-fetch-ai: Leftover comments

libs/ml/examples/word2vec_CBOW/main.cpp
```diff
@@ -60,26 +61,15 @@ std::string findWordByIndex(std::map<std::string, uint64_t> const &vocab, uint64
...
-  for (uint64_t i(0); i < embeddings.shape()[0]; ++i)
-  {
-    DataType d = fetch::math::distance::Cosine(wordVector, embeddings.Slice(i).Unsqueeze());
-    distances.emplace_back(i, d);
-  }
-  std::nth_element(distances.begin(), distances.begin() + k, distances.end(),
-                   [](std::pair<uint64_t, float> const &a, std::pair<uint64_t, float> const &b) {
-                     return a.second > b.second;
-                   });
+  std::vector<std::pair<std::string, double>> results = loader.GetKNN(embeddings, word, k);
```
pierre-wilmot-fetch-ai: GetKNN shouldn't be part of the loader

libs/ml/examples/word2vec_SGNS/main.cpp
```diff
@@ -212,7 +171,8 @@ int main(int argc, char **argv)
       context.At(j) = data.first.At(1);
 
       // assign label
-      gt.At(j) = DataType(data.second);
+      //      gt.At(j) = DataType(data.second);
```
pierre-wilmot-fetch-ai: Leftover comment

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -68,19 +67,44 @@ struct TextParams
...
+
+  std::vector<std::pair<std::string, SizeType>> BottomKVocab(SizeType k);
+  std::vector<std::pair<std::string, SizeType>> TopKVocab(SizeType k);
+
+  SizeType              GetSentenceIdxFromWordIdx(SizeType word_idx);
+  std::vector<SizeType> GetSentenceFromWordIdx(SizeType word_idx);
+  SizeType              GetWordOffsetFromWordIdx(SizeType word_idx);
+
+  virtual void                                AddData(std::string const &training_data);
+  std::vector<std::pair<std::string, double>> GetKNN(ArrayType const &  embeddings,
```
pierre-wilmot-fetch-ai: This doesn't belong here

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -68,19 +67,44 @@ struct TextParams
...
+                                                     std::string const &word, unsigned int k) const;
+
 protected:
   // training data parsing containers
   SizeType                                               size_ = 0;  // # training pairs
   std::unordered_map<std::string, std::vector<SizeType>> vocab_;  // full unique vocab + frequencies
```
pierre-wilmot-fetch-ai: Since the vector is always going to contain only 2 elements, let's switch it for a pair std::unordered_map<std::string, std::pair<SizeType, SizeType>>

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
+}
 
-  /**
-   * helper function for getting the Top K most frequent words
-   */
-  std::vector<std::pair<std::string, SizeType>> TopKVocab(SizeType k)
-  {
-    return FindK(k, true);
-  }
+/**
```
pierre-wilmot-fetch-ai: Comment is out of sync

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -110,168 +134,147 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
+      Reset();
+      is_done_ = true;
     }
 
-    SizeType label = buffer_idxs.at(p_.n_data_buffers);
-    cursor_++;
-
-    return std::make_pair(full_buffer, label);
-  }
+    bool valid_idx = CheckValidIndex(cursor_);
```
pierre-wilmot-fetch-ai: Checking if the index is valid at every single call seems like wasted compute power.

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -110,168 +134,147 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
+/**
+ * get the next data point in order
+ * @return
+ */
+template <typename T>
+std::pair<T, typename TextLoader<T>::SizeType> TextLoader<T>::GetNext()
+{
+  // loop until we find a non-discarded data point
+  bool not_found = true;
+  while (not_found)
```
pierre-wilmot-fetch-ai: Looks like this could potentially turn into an infinite loop in some edge cases

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -110,168 +134,147 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
+  return GetAtIndex(cursor_);
+}
 
-    return GetAtIndex(cursor_);
+/**
+ * gets the next data point at random
+ * @return
+ */
+template <typename T>
+std::pair<T, typename TextLoader<T>::SizeType> TextLoader<T>::GetRandom()
```
pierre-wilmot-fetch-ai: Get random looks more complex than it should be.
Either we just return a random point, without caring about uneven sampling.
Or we create a random sequence in the reset and GetRandom implementation should look like

```GetAtCursor(randomSequence[cursor++]);```

File: libs/ml/include/ml/dataloaders/text_loader.hpp [hunk 10]
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
+/**
+ * lookup a vocab index for a word
+ * @param idx
+ * @return
+ */
+template <typename T>
+typename TextLoader<T>::SizeType TextLoader<T>::VocabLookup(std::string const &word) const
+{
+  assert(vocab_.at(word).at(0) < vocab_.size());
+  assert(vocab_.at(word).at(0) != 0);  // dont currently handle unknowns elegantly
```
pierre-wilmot-fetch-ai: This will throw anyway

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
-
-    if (sentence_idx == 0)
-    {
-      word_offset = word_idx;
-    }
-    else
+    SizeType cur_word_idx = word_idx - 1;
+    SizeType word_idx_for_offset_zero;
+    bool     not_found = true;
+    while (not_found)
```
pierre-wilmot-fetch-ai: May lead to problem too.
Loop should loop over the container, not loop until we've found what we're looking for. If what we're looking for is not here, we're stuck in the loop forever (or in this situation we try to access array[-1])

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
-  virtual std::vector<SizeType> GetData(SizeType idx)
+/**
+ * print k nearest neighbours
+ * @param vocab
+ * @param embeddings
+ * @param word
+ * @param k
+ */
+template <typename T>
+std::vector<std::pair<std::string, double>> TextLoader<T>::GetKNN(ArrayType const &  embeddings,
```
pierre-wilmot-fetch-ai: Does not belong here

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
-                        std::back_inserter(result),  // Store output
-                        std::ptr_fun<int, int>(&std::ispunct));
-    word = result;
+    ret.emplace_back(std::make_pair(VocabLookup(distances.at(i).first), distances.at(i).second));
   }
+  return ret;
+}
+
+/**
+ * Override this method with a more interesting implementation
```
pierre-wilmot-fetch-ai: Should be virtual then

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
+  assert(p_.n_data_buffers == 1);
+  SizeType sentence_idx = this->word_idx_sentence_idx.at(idx);
+  SizeType word_idx     = this->GetWordOffsetFromWordIdx(idx);
+  ret.At(0)             = DataType(this->data_.at(sentence_idx).at(word_idx));
+}
+
+template <typename T>
+typename TextLoader<T>::SizeType TextLoader<T>::GetLabel(typename TextLoader<T>::SizeType idx)
+{
+  return 1;
```
pierre-wilmot-fetch-ai: Make it a pure virtual rather than returning hardcoded 1

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
+}
+
+/**
+ * helper for stripping punctuation from words
+ * @param word
+ */
+template <typename T>
+std::vector<std::string> TextLoader<T>::StripPunctuationAndLower(std::string &word) const
```
pierre-wilmot-fetch-ai:
```
std::vector<std::string> TextLoader<T>::StripPunctuationAndLower(std::string &word) const
{
  std::vector<std::string> ret(1);

  // replace punct with space and lower case
  SizeType word_idx = 0;
  for (auto &c : word)
  {
    if (std::isalpha(c))
    {
        ret.back() += (char)to_lower(c);
    }
    else if (c == '-')
    {
        ret.push_back(std::string());
    }
    // non-hyphens are assumed to be punctuation that we should ignore
    else
    {

    }
  }

  return ret;
}
```

Also, that doesn't handle apostrophes

Can't -> Cant
Won't -> Wont
Khan's -> Khans

May become a problem in French where we sometime squash together the article and the name.

l'{ANY_NOUN_THAT+STARTS_WITH_A_VOWEL}

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
+  return std::string(".!?").find(word.back()) != std::string::npos;
+}
+
+/**
+ * returns a vector of filenames of txt files
+ * @param dir_name  the directory to scan
+ * @return
+ */
+template <typename T>
+std::vector<std::string> TextLoader<T>::GetAllTextFiles(std::string dir_name)
```
pierre-wilmot-fetch-ai: Let's not put that here -- Will cause issues when compiling cross platform + the C style looks very out of place in the codebase.

reading file should be user responsability. My guess is that people will mostly use python for that

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
-        std::ifstream t(cur_file);
+      std::string   cur_file = training_data + "/" + file_names[j] + ".txt";
+      std::ifstream t(cur_file);
 
-        std::string cur_text((std::istreambuf_iterator<char>(t)), std::istreambuf_iterator<char>());
+      std::string cur_text((std::istreambuf_iterator<char>(t)), std::istreambuf_iterator<char>());
 
-        full_training_text += cur_text;
-      }
+      full_training_text += cur_text;
```
pierre-wilmot-fetch-ai: Probably should add a whitespace in between chunks of text

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
-
-        assert(cur_idx < p_.unigram_table_size);
-
-        double adjusted_word_probability = cur_val / sum_adj_vocab;
-
-        // word frequency
-        n_rows = SizeType(adjusted_word_probability * double(p_.unigram_table_size));
-
-        for (SizeType k = 0; k < n_rows; ++k)
+          vocab_[cur_word] = std::vector<SizeType>({word_idx, word_freq});
```
pierre-wilmot-fetch-ai: ``` vocab_[cur_word][1]++; ``` This should work

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
+        }
+        else
         {
-          unigram_table_[cur_idx] = e.second.at(0);
-          ++cur_idx;
+          vocab_.emplace(std::make_pair(cur_word, std::vector<SizeType>({unique_word_count_, 1})));
+          word_idx = unique_word_count_;
```
pierre-wilmot-fetch-ai: unique_word_count_ should be replaced by vocab_.size() since it's duplicated information

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
+          unique_word_count_++;
         }
+        data_.at(sentence_count_).push_back(word_idx);
```
pierre-wilmot-fetch-ai: ```data_.at(sentence_count_)``` => ```data.back()```

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -280,468 +283,560 @@ class TextLoader : public DataLoader<T, typename T::SizeType>
...
+  std::vector<std::pair<std::string, std::vector<SizeType>>> top_k(k);
 
-    std::vector<std::pair<std::string, SizeType>> ret(k);
-    SizeType                                      tmp = 0;
-    for (auto &e : ret)
-    {
-      e = std::make_pair(top_k.at(tmp).first, top_k.at(tmp).second.at(1));
-      ++tmp;
-    }
+  if (mode)
```
pierre-wilmot-fetch-ai: This can be implemented with a XOR operation

```
    std::partial_sort_copy(vocab_.begin(), vocab_.end(), top_k.begin(), top_k.end(),
                           [mode](std::pair<const std::string, std::vector<SizeType>> const &l,
                              std::pair<const std::string, std::vector<SizeType>> const &r) {
                             return (l.second.at(1) > r.second.at(1)) ^ mode;
                           });
```

libs/ml/include/ml/layers/skip_gram.hpp
```diff
@@ -20,13 +20,14 @@
...
 #include "ml/layers/layer.hpp"
-#include "ml/ops/activations/sigmoid.hpp"
+#include "ml/ops/activations/softmax.hpp"
 #include "ml/ops/embeddings.hpp"
 #include "ml/ops/matrix_multiply.hpp"
 #include "ml/ops/placeholder.hpp"
 #include "ml/ops/transpose.hpp"
 
 #include <cmath>
+#include <ml/ops/activation.hpp>
```
pierre-wilmot-fetch-ai: Likely unused

libs/ml/include/ml/ops/concat.hpp
```diff
@@ -0,0 +1,56 @@
...
+  using ArrayPtrType = std::shared_ptr<ArrayType>;
+
+  Concat()          = default;
+  virtual ~Concat() = default;
+
+  virtual ArrayType Forward(std::vector<std::reference_wrapper<ArrayType const>> const &inputs)
+  {
+    ASSERT(inputs.size() == 1);
+    this->output_ = std::make_shared<ArrayType>(inputs.front().get().Concat(inputs.at(1)));
+    return *this->output_;
```
pierre-wilmot-fetch-ai: ASSERT(inputs.size() == 1);
inputs.at(1) // This is out of range

libs/ml/include/ml/ops/concat.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+  virtual std::vector<ArrayType> Backward(
+      std::vector<std::reference_wrapper<ArrayType const>> const &inputs,
+      ArrayType const &                                           errorSignal)
+  {
+    ASSERT(inputs.size() == 1);
+    return {errorSignal.Clone().Concat()};
+  }
+
+  static constexpr char const *DESCRIPTOR = "Concat";
```
pierre-wilmot-fetch-ai: Tensor doesn't have a Concat method.

This is untested dead code 👎

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -155,6 +155,15 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<T>
     }
   }
 
+  /**
+   * exports the weights Array
+   * @return
+   */
+  ArrayType GetWeights()
```
pierre-wilmot-fetch-ai: Make it const and return const &

libs/ml/tests/ml/dataloader/cbow_loader.cpp
```diff
@@ -0,0 +1,91 @@
...
+    left_and_right = loader.GetNext().first;
+
+    std::string left  = loader.VocabLookup(SizeType(double(left_and_right.At(0))));
+    std::string right = loader.VocabLookup(SizeType(double(left_and_right.At(1))));
+
+    ASSERT_TRUE(std::find(gt_left_right_pairs.begin(), gt_left_right_pairs.end(),
+                          std::pair<std::string, std::string>(left, right)) !=
+                gt_left_right_pairs.end());
+  }
+}
```
pierre-wilmot-fetch-ai, line 99: Bring back the old cbow loader tests.

This is very loose test.
Your loader could return the same value at each call and the test would still pass.
Pluis it doesn't test any of the Size() / IsDone() method

libs/ml/tests/ml/dataloader/skipgram_loader.cpp
```diff
@@ -95,6 +96,7 @@ TYPED_TEST(SkipGramDataloaderTest, loader_test)
 
     std::string input   = loader.VocabLookup(SizeType(double(input_and_context.At(0))));
     std::string context = loader.VocabLookup(SizeType(double(input_and_context.At(1))));
+
     ASSERT_TRUE(std::find(gt_input_context_pairs.begin(), gt_input_context_pairs.end(),
```
pierre-wilmot-fetch-ai: Same here, this test pretty much doesn't test anything

libs/ml/tests/ml/dataloader/text_loader.cpp
```diff
@@ -110,6 +111,41 @@ TYPED_TEST(TextDataLoaderTest, adddata_loader_test)
...
+       "make",   "things",      "so",       "easy", "because",  "it",         "has",
+       "some",   "punctuation", "doesnt",   "it",   "indeed",   "it",         "does",
+       "and",    "this",        "sentence", "even", "has",      "hyphenation"});
+
+  std::string cur_word;
+  for (std::size_t j = 0; j < 20; ++j)
+  {
+    std::pair<TypeParam, SizeType> output = loader.GetNext();
+    cur_word = loader.VocabLookup(SizeType(double(output.first.At(0))));
+    ASSERT_TRUE(cur_word.compare(gt_input.at(j)) == 0);
```
pierre-wilmot-fetch-ai: ASSERT_EQ(cur_word, gt_input.at(j) // This will print the value of the strings if test fails

libs/ml/tests/ml/dataloader/w2v_cbow_dataloader.cpp
```diff
@@ -1,128 +0,0 @@
-//------------------------------------------------------------------------------
```
pierre-wilmot-fetch-ai, line 1: Bring back that guy

pierre-wilmot-fetch-ai (changes requested):
This pull request is way more complex than it needs to be.
Code is difficult to understand and almost not tested.

I'd recommand starting again from scratch with a clean definition of what each class responsabilities are.

The TextLoader base class should have:
- A map maping each distinc word to a unique index + it frequency in the dataset (stored as number of occurences)
- Some data structure that record the order in which words / sentence appear (vector<vector<unint>>) 
- A tokeniser / preprocessor (for punctuation / lowercase)
- Probably shouldn't inherit the DataLoader interface


CBOWLoader & SkipgramLoader:
- Inherit from TextLoader and implement the Dataloader interface
- Their job is to sample from the vector<vector<unint>> in parent class

CBOW:
- Can be as simple as moving a cursor and returning the 8 words that come in order

Skipgram:
- Sample word and build positive pair (by sampling nearby) or negative pair (By sampling randomly and making sure the word is not in the window)

if the samplng is computationally expensive, we can build a cache class (probably based on TensorLoader), that just gets filled with results of sampling from CBOW/Skipgram at the begining of the program.

libs/ml/include/ml/dataloaders/text_loader.hpp
baykaner: seems to me that there is little advantage to doing that (only advantage i can think of is that it makes it clearer to developers that there is only expected to be two elements - which is nice but not essential), but there is one disadvantage compared with using a vector - unordered map does not have a hash function for pair so we would need to implement it ourselves and that seems like a little code bloat for not much gain.

libs/ml/include/ml/dataloaders/text_loader.hpp
baykaner: Hmm.. something like this is mandatory though - to handle the window edge checking and discarded word checking. The only obvious change I can see to make is that I could replace the window edge checking with a list of stored indices; probably worth doing at some point

libs/ml/include/ml/dataloaders/text_loader.hpp
baykaner: The only edge case I can think of where that would happen is if the user sets the hyper-parameters very foolishly, but I don't think we should be putting bounds on permissible hyper-parameters since they're also data dependent. What do you propose? throwing after some large number of iterations? I guess that could be helpful.

libs/ml/include/ml/dataloaders/text_loader.hpp
baykaner: good point - random sequence makes sense to be in the reset. and the rest of the logic can be re-used from getnext

libs/ml/include/ml/dataloaders/text_loader.hpp
baykaner: I don't see how.   if (sentence_idx == 0) guards against accessing array[-1], and this is much faster than iterating through the entire container; this method will be take on average sentence length iterations to complete; much faster than iterating through the entire container.

libs/ml/include/ml/dataloaders/text_loader.hpp
baykaner: it is; see the function declaration in the class body.

libs/ml/include/ml/dataloaders/text_loader.hpp
baykaner: I'm not sure about that. Returning 1 is pointless, I agree, but using the text loader to just process and return data seems conceivably useful to me. I guess I could make another text loader class that inherits from this and does nothing except implement GetLabel, but that seems a bit pointless also to me

libs/ml/include/ml/dataloaders/text_loader.hpp
baykaner: this handles apostrophes in precisely the way I expected it to.
can't -> cant is the intended behaviour; but I can see what can't -> can t could also be a good result (have you read any literature on which results in better embeddings?). behaviour for plurals is not great and we're pretty far from multi-language support at the moment. I think this falls under work for a future PR of improving punctuation handling

libs/ml/include/ml/dataloaders/text_loader.hpp
baykaner: isnt that slower - since it's operated in the sort_copy loop? Or does the compiler just unroll into two separate code blocks anyway?
The xor is more concise though so maybe I'll just change it anyway

libs/ml/include/ml/ops/concat.hpp
baykaner: yep - I didn't implement any of this; it's from a demonstration I was giving to subcontractors about our workflow. Thought I deleted - glad you caught it.

libs/ml/examples/word2vec_CBOW/main.cpp
baykaner: yep, fair point

libs/ml/include/ml/dataloaders/text_loader.hpp
baykaner: It doesn't matter - I decided to rework not to use either, but keep a separate vocab_frequencies variable

libs/ml/tests/ml/dataloader/skipgram_loader.cpp
baykaner: for the skipgram case it's a little different because there is random sampling (hence why I tested in this looser way), but I can still narrow it down to sampling pairs which is better, so I'll do that.

libs/ml/examples/word2vec_CBOW/main.cpp
```diff
@@ -91,13 +90,22 @@ int main(int ac, char **av)
     return 1;
   }
 
-  fetch::ml::CBOWLoader<DataType> loader(CONTEXT_WINDOW_SIZE);
+  fetch::ml::dataloaders::CBoWTextParams<ArrayType> p;
+  p.window_size    = CONTEXT_WINDOW_SIZE;
+  p.n_data_buffers = CONTEXT_WINDOW_SIZE * 2;
+  p.min_sentence_length =
+      SizeType((CONTEXT_WINDOW_SIZE * 2) + 1);  // maximum number of sentences to use
```
pierre-wilmot-fetch-ai: Comment out of sync

libs/ml/examples/word2vec_CBOW/main.cpp
```diff
@@ -60,27 +62,24 @@ std::string findWordByIndex(std::map<std::string, uint64_t> const &vocab, uint64
...
+
+  for (std::size_t j = 0; j < output.size(); ++j)
   {
-    std::cout << findWordByIndex(vocab, distances[i].first) << " -- " << distances[i].second
-              << std::endl;
+    std::cout << "output.at(j).first: " << dl.VocabLookup(output.at(j).first) << std::endl;
+    std::cout << "output.at(j).second: " << output.at(j).second << "\n" << std::endl;
   }
+
+  std::cout << "hot-cold distance: "
```
pierre-wilmot-fetch-ai: Leftover debug

libs/ml/examples/word2vec_CBOW/main.cpp
```diff
@@ -154,8 +163,9 @@ int main(int ac, char **av)
       iteration++;
     }
     std::cout << "End of epoch " << epoch << std::endl;
+
     // Print KNN of word "one"
-    PrintKNN(loader.GetVocab(), *g.StateDict().dict_["Embeddings"].weights_, "mind", 6);
+    PrintKNN(dl, *g.StateDict().dict_["Embeddings"].weights_, "cold", 10);
```
pierre-wilmot-fetch-ai: Not sure how to deal with that one.
Hardcoding a specific word is bad, but since training data is user specified we have no garentee that any given word will be in the trained embeddings.

libs/ml/examples/word2vec_SGNS/main.cpp
```diff
@@ -0,0 +1,261 @@
...
+  std::vector<std::pair<typename ArrayType::SizeType, typename ArrayType::Type>> output =
+      fetch::math::clustering::KNN(arr, one_vector, k);
+
+  for (std::size_t j = 0; j < output.size(); ++j)
+  {
+    std::cout << "output.at(j).first: " << dl.VocabLookup(output.at(j).first) << std::endl;
+    std::cout << "output.at(j).second: " << output.at(j).second << "\n" << std::endl;
+  }
+
+  std::cout << "hot-cold distance: "
```
pierre-wilmot-fetch-ai: Leftover debug

libs/ml/examples/word2vec_SGNS/main.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+  // first get hold of the skipgram layer by searching the return name in the graph
+  std::shared_ptr<fetch::ml::layers::SkipGram<ArrayType>> sg_layer =
+      std::dynamic_pointer_cast<fetch::ml::layers::SkipGram<ArrayType>>(g.GetNode(skip_gram_name));
+
+  // next get hold of the embeddings
+  std::shared_ptr<fetch::ml::ops::Embeddings<ArrayType>> embeddings =
+      sg_layer->GetEmbeddings(sg_layer);
+
+  PrintKNN(dl, embeddings->GetWeights(), "cold", 10);
```
pierre-wilmot-fetch-ai: Same problem here

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -17,731 +17,322 @@
...
+  using SentencesType = std::vector<std::vector<std::string>>;
+
+  TextLoader() = default;
+
+  SizeType    Size() const;
+  SizeType    VocabSize() const;
+  SizeType    VocabLookup(std::string const &word) const;
+  std::string VocabLookup(SizeType idx) const;
+
+  std::unordered_map<std::string, SizeType> GetVocab() const;
```
pierre-wilmot-fetch-ai: Let's return VocabType here.
(And fix L42)
And return a const & (else we copy a potentially very large object at each call)

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -17,731 +17,322 @@
...
-  {
-    return word_count_;
-  }
+/**
+ *
+ * @tparam T Array Type
+ * @return unordered map from string to word index & vocab frequency
+ */
+template <typename T>
+std::unordered_map<std::string, typename TextLoader<T>::SizeType> TextLoader<T>::GetVocab() const
```
pierre-wilmot-fetch-ai: Let's return VocabType here.
(And fix L42)

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -17,731 +17,322 @@
...
-public:
-  TextLoader(std::string &data, TextParams<ArrayType> const &p, SizeType seed = 123456789)
-    : cursor_(0)
-    , p_(p)
-    , lfg_(seed)
-    , lcg_(seed)
-  {
+private:
+  std::vector<char> word_break_{'-', '\'', '.', '\t', '\n', '!', '?'};
+  std::vector<char> sentence_break_{'.', '\t', '\n', '!', '?'};
```
pierre-wilmot-fetch-ai: These can be const, or maybe even constexpr

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -17,731 +17,322 @@
...
-    SizeType buffer_count = 0;
-    for (SizeType j = 0; j < p_.n_data_buffers; ++j)
+    if (vocab_.find(parsed_word.at(0)) != vocab_.end())
     {
-      SizeType sentence_idx = word_idx_sentence_idx[buffer_idxs.at(buffer_count)];
-      SizeType word_idx     = GetWordOffsetFromWordIdx(buffer_idxs.at(buffer_count));
-      full_buffer.At(j)     = DataType(data_.at(sentence_idx).at(word_idx));
-      ++buffer_count;
+      // return found word
+      return vocab_.at(parsed_word.at(0));
```
pierre-wilmot-fetch-ai: Unimportant optimisation trick.
Rather that doing a find followed by a at (wich leads to serching the map twice), you can save the iteratoe returned by find and just return the pointed value.

```
if ((auto it = vocab_.find(parsed_word.at(0))) != vocab_.end())
{
 return *it;
}
```
Somathing like that should work

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -17,731 +17,322 @@
...
-   */
-  virtual std::pair<T, SizeType> GetNext()
+/**
+ * lookup the string corresponding to a vocab index
+ * if the index doesn't occur in the vocabulary, returns "UNKNOWN"
+ * @param idx
+ * @return
+ */
+template <typename T>
+std::string TextLoader<T>::VocabLookup(typename TextLoader<T>::SizeType const idx) const
```
pierre-wilmot-fetch-ai: Not a big deal either, but could return a const &

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -17,731 +17,322 @@
...
+    SizeType word_idx;
+    for (std::string const &cur_word : sentence)
     {
-      if (sentence_count_ + sentence_count > p_.max_sentences)
+      // if already seen this word
+      if (vocab_.find(cur_word) != vocab_.end())
       {
-        break;
+        vocab_frequencies.at(vocab_.at(cur_word))++;
+        word_idx = vocab_.at(cur_word);
```
pierre-wilmot-fetch-ai: Same remark, we're looking through the vocab for the word index 3 times

libs/ml/include/ml/dataloaders/word2vec_loaders/basic_textloader.hpp
```diff
@@ -0,0 +1,511 @@
...
+{
+  // loop until we find a non-discarded data point
+  bool     not_found     = true;
+  SizeType timeout_count = 0;
+  while (not_found)
+  {
+    if (IsDone())
+    {
+      Reset();
+    }
```
pierre-wilmot-fetch-ai: This kind of while(not_found) loop is problematic, especialy since you automatically reset inside the loop.
If you look at libs/ml/examples/word2vec_SGNS/main.cpp L185, you're looping until the dataloader is done.
If it turns out that your last word is discarded, your GetNextValidIndex will automatically reset the data loader and the calling loop is gonna run through a full new epoch

libs/ml/include/ml/dataloaders/word2vec_loaders/basic_textloader.hpp
```diff
@@ -0,0 +1,511 @@
...
+
+    if (random ? CheckValidIndex(ran_idx_.at(cursor)) : CheckValidIndex(cursor))
```
pierre-wilmot-fetch-ai: if (CheckValidIndex(random ? ran_idx_.at(cursor) : cursor))

libs/ml/include/ml/dataloaders/word2vec_loaders/basic_textloader.hpp
```diff
@@ -0,0 +1,511 @@
...
+  cursor = random ? ran_idx_.at(cursor) : cursor;
+  return true;
+}
+
+/**
+ * checks if a data point may be indexed to create a training pair
+ * @return
+ */
+template <typename T>
+bool BasicTextLoader<T>::CheckValidIndex(SizeType idx)
```
pierre-wilmot-fetch-ai, line 438: I think any index pointed by the cursor should be valid.
The logic of stopping iterating at the end of a sentence should live with the cursor incrementing logic.

libs/ml/include/ml/dataloaders/word2vec_loaders/basic_textloader.hpp
```diff
@@ -0,0 +1,511 @@
...
+{
+  cursor_ = 0;
+
+  // generate a new random sequence for random sampling
+  ran_idx_ = std::vector<SizeType>(this->Size());
+  std::iota(ran_idx_.begin(), ran_idx_.end(), 0);
+  std::random_shuffle(ran_idx_.begin(), ran_idx_.end());
+
+  // recompute which words should be ignored based on their frequency
+  DiscardFrequent();
```
pierre-wilmot-fetch-ai, line 250: I think the full discarding logic need to be discused and potentially reworked, but if we end up keeping it like it is, we're missing a 
```
discards_.clear();
discard_count_ = 0;
```
At the moment we add more and more words to the discarded list on each reset call.
If we run a large number of epoch, we may end up spending more time looking for a non-discarded word than actually training.

libs/ml/include/ml/dataloaders/word2vec_loaders/basic_textloader.hpp
```diff
@@ -0,0 +1,511 @@
...
+ * discards words in training data set based on word frequency
+ */
+template <typename T>
+void BasicTextLoader<T>::DiscardFrequent()
+{
+  if (p_.discard_frequent)
+  {
+    SizeType sentence_count = 0;
+
+    // iterate through all sentences
```
pierre-wilmot-fetch-ai, line 478: If I understood the code correctly, ```discards_.clear()``` should be here.

libs/ml/include/ml/dataloaders/word2vec_loaders/skipgram_dataloader.hpp
```diff
@@ -0,0 +1,376 @@
...
+    SizeType ran_val     = SizeType(this->lcg_());
+    SizeType max_val     = this->unigram_table_.size();
+    SizeType idx         = ran_val % max_val;
+    negative_context_idx = this->unigram_table_.at(idx);
+    assert(negative_context_idx > 0);
+    assert(negative_context_idx < this->vocab_.size());
+    ongoing = false;
+
+    // check if selected negative word actually within context window
+    for (SizeType j = 0; j < (2 * p_.window_size) + 1; j++)
```
pierre-wilmot-fetch-ai, line 225: Not sure how this is meant to work, but shoudn't this be ?
```for (SizeType j = -p_.window_size ; j <=p_.window_size ; j++)```

File: libs/ml/tests/ml/dataloader/cbow_loader.cpp [hunk 47]
```diff
@@ -0,0 +1,202 @@
...
+  EXPECT_FALSE(loader.IsDone());
+  data = loader.GetNext();  // fetchbot I am one year old and I eat
+  EXPECT_FALSE(loader.IsDone());
+  data = loader.GetNext();  // I am one year old and I eat tokens
+  EXPECT_FALSE(loader.IsDone());
+  data = loader.GetNext();  // am one year old and I eat tokens for
+  EXPECT_FALSE(loader.IsDone());
+  data = loader.GetNext();  // year old and I eat tokens for breakfast
+  EXPECT_TRUE(loader.IsDone());
+}
```
pierre-wilmot-fetch-ai, line 207: Thanks 😉

libs/ml/tests/ml/dataloader/text_loader.cpp
```diff
@@ -36,21 +36,137 @@ using MyTypes = ::testing::Types<fetch::math::Tensor<int>, fetch::math::Tensor<f
...
+  EXPECT_TRUE(loader.IsDone());
+  EXPECT_EQ(loader.GetVocab().size(), 0);
+
+  EXPECT_TRUE(loader.AddData("A longer, five-word sentence"));
+  EXPECT_EQ(loader.Size(), 1);       // indicates total number of valid words in training corpus
+  EXPECT_EQ(loader.VocabSize(), 5);  // indicates number of unique words
+  EXPECT_FALSE(loader.IsDone());
+  EXPECT_EQ(loader.GetVocab().size(), 5);
+
+  // since vocab stores words in an unordered map, we don't know where each word will
```
pierre-wilmot-fetch-ai: Incorrect comment, we have no gerentee on the order of the word when uing an iterator to traverse the map, but the mapping from string to indexes is guarented

libs/math/include/math/free_functions/clustering_algorithms/knn.hpp
```diff
@@ -0,0 +1,92 @@
...
+  assert(array.shape().size() == 2);
+  assert(array.shape().at(1) == one_vector.shape().at(1));
+  assert(one_vector.shape().at(0) == 1);
+
+  std::vector<std::pair<SizeType, DataType>> similarities;
+  similarities.reserve(array.shape().at(0));
+  for (SizeType i(0); i < array.shape().at(0); ++i)
+  {
+    typename ArrayType::Type d =
+        fetch::math::distance::Cosine(one_vector, array.Slice(i).Unsqueeze());
```
pierre-wilmot-fetch-ai, line 45: Not a big deal for now, but in the future we'll need a way which distance function to use for KNN

libs/math/include/math/free_functions/clustering_algorithms/knn.hpp
```diff
@@ -0,0 +1,92 @@
...
+  std::nth_element(similarities.begin(), similarities.begin() + k, similarities.end(),
+                   [](std::pair<SizeType, DataType> const &a,
+                      std::pair<SizeType, DataType> const &b) { return a.second > b.second; });
+
+  std::vector<std::pair<SizeType, DataType>> ret;
+  for (SizeType i(0); i < k; ++i)
+  {
+    ret.emplace_back(std::make_pair(similarities.at(i).first, similarities.at(i).second));
+  }
+  return ret;
```
pierre-wilmot-fetch-ai, line 65: May be nice to sort ret, since nth_element garantes the N max elements will be in the N first position, but doesn't enforce that they are sorted

pierre-wilmot-fetch-ai (changes requested):

libs/ml/include/ml/dataloaders/text_loader.hpp
pierre-wilmot-fetch-ai: On second look, \t && \n should be sentence break

libs/ml/examples/word2vec_CBOW/main.cpp
baykaner: that's right. I figure since it's an example it's okay - but I'll change it to a parameter so the test word selection is at the top of the file and clearly a parameter to set.

libs/ml/include/ml/dataloaders/text_loader.hpp
baykaner: why do you think that \t and \n shouldn't also be word break? if a word slips through with something like text\ntext surely we should also split that into two words? or did you mean something else?

libs/ml/include/ml/dataloaders/word2vec_loaders/basic_textloader.hpp
baykaner: good point.

libs/ml/tests/ml/dataloader/cbow_loader.cpp
baykaner: fetchbot may be gone from our PR, but he shall live on on our shirts and in our code :-p

libs/ml/tests/ml/dataloader/text_loader.cpp
baykaner: yup that's what i meant but it's poorly worded. will fix

libs/math/include/math/free_functions/clustering_algorithms/knn.hpp
baykaner: yup - leaving that for later improvements to math lib distance/similarity functionality. Opened up issue 784 so we don't forget

libs/math/include/math/free_functions/clustering_algorithms/knn.hpp
baykaner: yeah good point - this was annoying me too

libs/ml/include/ml/dataloaders/word2vec_loaders/skipgram_dataloader.hpp
baykaner: Don't think so, the calls inside the loop handle the fact that 0 actually refers to an offset windowsize prior to the target word; especially since we're dealing with SizeTypes (which are expected to be unsigned) writing the loop that way would involve casting and I don't think there's any need for it. Can be rewritten if you think it's too unclear this way though

libs/ml/include/ml/dataloaders/text_loader.hpp
pierre-wilmot-fetch-ai: On second look, \t && \n shouldN'T be sentence break
Typo in my comment

libs/ml/include/ml/dataloaders/text_loader.hpp
```diff
@@ -319,6 +319,12 @@ std::vector<std::string> TextLoader<T>::StripPunctuationAndLower(std::string con
 template <typename T>
 bool TextLoader<T>::CheckEndOfSentence(std::string &word)
 {
+  // if the word ends with two new-lines consecutively we assume it is a sentence break
+  if ((word.end()[-2] == '\n') && (word.end()[-1] == '\n'))
```
pierre-wilmot-fetch-ai: if (word.size() > 1 && (word.end()[-2] == '\n') && (word.end()[-1] == '\n'))

pierre-wilmot-fetch-ai (dismissed):

pierre-wilmot-fetch-ai (approved):

PR #777, opened 2019-03-20
#756 added coonditional probability distance

MissingNO57:


libs/math/include/math/distance/conditional_probabilities_distance.hpp
```diff
@@ -0,0 +1,74 @@
+#pragma once
```
baykaner: since this is already in the distance namespace we don't need distance in the filename, please rename to conditional_probabilities.hpp

libs/math/include/math/distance/conditional_probabilities_distance.hpp
```diff
@@ -0,0 +1,74 @@
...
+                                      typename ArrayType::Type const sigma,
+                                      typename ArrayType::Type &     r)
+{
+  assert(i < a.shape()[0] && j < a.shape()[0]);
+
+  // Calculate numerator
+  typename ArrayType::Type tmp_numerator = SquareDistance(a.Slice(i), a.Slice(j));
+
+  // Divide by 2 sigma squared
+  tmp_numerator = -Divide(tmp_numerator, 2.0 * sigma * sigma);
```
baykaner: negative sign supposed to be here? you didn't mention it in your comment so I wasn't sure

libs/math/include/math/distance/conditional_probabilities_distance.hpp
```diff
@@ -0,0 +1,74 @@
...
+namespace fetch {
+namespace math {
+namespace distance {
+
+template <typename ArrayType>
+void ConditionalProbabilitiesDistance(ArrayType const &a, size_t i, size_t j,
+                                      typename ArrayType::Type const sigma,
+                                      typename ArrayType::Type &     r)
+{
+  assert(i < a.shape()[0] && j < a.shape()[0]);
```
baykaner: please use .at(idx) instead of [idx] since .at(idx) checks the bounds of the vector/array

libs/math/include/math/distance/conditional_probabilities_distance.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+  // Divide by 2 sigma squared
+  tmp_numerator = -Divide(tmp_numerator, 2.0 * sigma * sigma);
+
+  Exp(tmp_numerator, tmp_numerator);
+
+  // Calculate denominator
+  typename ArrayType::Type tmp_denominator = 0.0;
+
+  for (size_t k = 0; k < a.shape()[0]; k++)
```
baykaner: a.shape().at(0)

libs/math/include/math/distance/conditional_probabilities_distance.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include "math/free_functions/matrix_operations/matrix_operations.hpp"
+
+#include <cmath>
+
+namespace fetch {
+namespace math {
+namespace distance {
+
+template <typename ArrayType>
+void ConditionalProbabilitiesDistance(ArrayType const &a, size_t i, size_t j,
```
baykaner: please use std::size_t instead of size_t

libs/math/include/math/distance/conditional_probabilities_distance.hpp
```diff
@@ -0,0 +1,74 @@
...
+    Exp(tmp_val, tmp_val);
+
+    tmp_denominator += tmp_val;
+  }
+
+  r = tmp_numerator / tmp_denominator;
+}
+
+template <typename ArrayType>
+typename ArrayType::Type ConditionalProbabilitiesDistance(ArrayType const &a, size_t i, size_t j,
```
baykaner: std::size_t instead of size_t (coding style guide - you should have access now by the way)

libs/math/tests/math/distance/gtest/conditional_probabilities_distance_tests.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <iomanip>
+#include <iostream>
+
+#include "math/distance/conditional_probabilities_distance.hpp"
+#include "math/tensor.hpp"
+
+using namespace fetch::math::distance;
+using namespace fetch::math;
+
+TEST(distance_tests, conditional_distance)
```
baykaner: please make this a TYPED_TEST which tests Tensor<float>, Tensor<double,  Tensor<FixedPoint<16, 16>>, Tensor<FixedPoint<32, 32>> (see other tests in the ML library for examples if you're not sure how to do this)

libs/math/tests/math/distance/gtest/conditional_probabilities_distance_tests.cpp
```diff
@@ -0,0 +1,54 @@
...
+  A.Set({2, 1}, 10);
+  A.Set({2, 2}, 11);
+  A.Set({2, 3}, 12);
+
+  A.Set({3, 0}, 13);
+  A.Set({3, 1}, 14);
+  A.Set({3, 2}, 15);
+  A.Set({3, 3}, 16);
+
+  ASSERT_TRUE(ConditionalProbabilitiesDistance(A, 1, 2, 1) == 0.5);
```
baykaner: can you set up the test in such a way that it tests more than just one value please? It's possible to have errors and still have this test pass.

libs/math/tests/math/distance/gtest/conditional_probabilities_distance_tests.cpp
```diff
@@ -0,0 +1,54 @@
...
+}
```
baykaner: please test including negative values in the array and make sure that you test the distances the following cases:
1. -x -> -y,
2. -x -> y,
3. and x -> y
because sign errors are very common

baykaner (changes requested):

libs/math/include/math/distance/conditional_probabilities_distance.hpp
MissingNO57: I will add that as comment, because it requires exp(-result), but I can't simply call  Exp(-tmp_numerator, tmp_numerator);, because it expects reference. I can do something like exp(--tmp_numerator,tmp_numerator), but it might be confusing.

baykaner: It was just a matter of clarity.
The other options could be something like:
-1 * Divide(....)
or
Multiply(ret, -1)
(functions exist in free_functions/fundamental_operators)

But I'm happy with the comment too

libs/math/include/math/free_functions/statistics/entropy.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include "vectorise/memory/range.hpp"
+
+#include <cmath>
+
+namespace fetch {
+namespace math {
+namespace statistics {
+
+template <typename ArrayType>
+typename ArrayType::Type Entropy(ArrayType const &a, size_t index, typename ArrayType::Type &ret)
```
baykaner: std::size_t

libs/math/include/math/free_functions/statistics/entropy.hpp
```diff
@@ -0,0 +1,55 @@
...
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "core/assert.hpp"
+#include "math/distance/conditional_probabilities.hpp"
+#include "math/shapeless_array.hpp"
```
baykaner: remove this include

libs/math/include/math/free_functions/statistics/entropy.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include "vectorise/memory/range.hpp"
```
baykaner: and this one

libs/math/include/math/free_functions/statistics/entropy.hpp
```diff
@@ -0,0 +1,55 @@
...
+    typename ArrayType::Type tmp_val = distance::ConditionalProbabilitiesDistance(a, j, index, 1);
+    tmp_val *= Log2(tmp_val);
+    ret += tmp_val;
+  }
+
+  return -ret;
+}
+
+template <typename ArrayType>
+typename ArrayType::Type Entropy(ArrayType const &a, size_t index)
```
baykaner: std::size_t

libs/math/include/math/distance/conditional_probabilities.hpp
```diff
@@ -29,24 +29,24 @@ namespace math {
 namespace distance {
 
 template <typename ArrayType>
-void ConditionalProbabilitiesDistance(ArrayType const &a, size_t i, size_t j,
+void ConditionalProbabilitiesDistance(ArrayType const &a, std::size_t i, std::size_t j,
```
baykaner, line 41: Okay great one more thing though - Please add a doxygen-style function comment. You can get clion to auto-generate the template for you, in this case it should be something like:

`
/**
 * Calculates the conditional probability of point j given point i in N-dimensional feature space
 * i.e.   e^(- x_i^2 / 2*sigma^2) / Sum(e^(- x_j^2 / 2*sigma^2))
 * @param a    input tensor of dimensions n_data x n_features
 * @param i     index of the data point on which distribution is centred
 * @param j     index of data point for which to estimate probability
 * @param sigma controls the distribution kurtosis/width
 * @param r   return value
 */
`

baykaner: okay i can't get the formatting right but those are mean to be stars not bullet points :-p

libs/math/tests/math/distance/gtest/conditional_probabilities_tests.cpp
```diff
@@ -0,0 +1,76 @@
...
+  A.Set({2, 1}, -1.2);
+  A.Set({2, 2}, -1.3);
+  A.Set({2, 3}, -1.4);
+
+  A.Set({3, 0}, 1.1);
+  A.Set({3, 1}, 1.2);
+  A.Set({3, 2}, 1.3);
+  A.Set({3, 3}, 1.4);
+
+  EXPECT_NEAR(ConditionalProbabilitiesDistance(A, 2, 1, 1), double(0.93083999), 1e-7);
```
baykaner, line 64: nice! 👍

baykaner (changes requested):

libs/math/tests/math/distance/gtest/conditional_probabilities_tests.cpp
MissingNO57: Thanks, I am just copying from other places.

baykaner (approved):
nice!

PR #780, opened 2019-03-22
Fix handling of Invalid Transaction format

ejfitzgerald:
- Add protection for empty HTTP requests
- Add protection for empty identity requests
- Remove display of chain status

### DIFF
--- a/libs/http/include/http/connection.hpp
+++ b/libs/http/include/http/connection.hpp
@@ -126,11 +126,15 @@ public:
       }
       else
       {
-        request->ParseHeader(*buffer_ptr, len);
-
-        if (is_open_)
+        // only parse the header if there is data to be parsed
+        if (len)
         {
-          ReadBody(buffer_ptr, request);
+          request->ParseHeader(*buffer_ptr, len);
+
+          if (is_open_)
+          {
+            ReadBody(buffer_ptr, request);
+          }
         }
       }
     };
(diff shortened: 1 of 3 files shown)


libs/http/include/http/connection.hpp
```diff
@@ -126,11 +126,15 @@ class HTTPConnection : public AbstractHTTPConnection,
       }
       else
       {
-        request->ParseHeader(*buffer_ptr, len);
-
-        if (is_open_)
+        // only parse the header if there is data to be parsed
```
n-hutton, line 129: 👍

libs/network/include/network/p2pservice/p2p_http_interface.hpp
```diff
@@ -248,21 +248,24 @@ class P2PHttpInterface : public http::HTTPModule
 
       if (include_transactions)
       {
-        auto const &slices = b->body.slices;
+        // create and allocate the variant array
```
n-hutton, line 251: this is for block explorer right?

n-hutton (approved):

libs/network/include/network/p2pservice/p2p_http_interface.hpp
ejfitzgerald: Yes and for @CobrasCobras mobile app

PR #786, opened 2019-03-25
VM-Lang JSON State Object

ejfitzgerald:


apps/vm-lang/main.cpp
```diff
@@ -193,5 +317,17 @@ int main(int argc, char **argv)
     std::cout << console << std::endl;
   }
```
troels-im, line 319: Would suggest to check for errors here and print them if there are any runtime errors.

troels-im (approved):
LGTM

PR #787, opened 2019-03-26
Feature/cross entropy tests

baykaner:
- tests/bugfixes/refactoring for cross entropy/softmax_cross_entropy/scaled_cross_entropy
- bugfix for Argmax which assumed tensors all 1-dim
- implementing first ml integration tests checking training/convergence in very basic scenarios
- minor tweaks/improvement to SGNS example


libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -350,22 +350,32 @@ class FixedPoint
...
+  {
+    return (static_cast<unsigned>(data_) / one);
+  }
+
+  explicit operator unsigned long() const
+  {
+    return (static_cast<unsigned long>(data_) / one);
+  }
+
+  explicit operator unsigned long long() const
```
pierre-wilmot-fetch-ai, line 378: Does that makes sense ? 
What's the expected result when the FixedPoint is representing a negative value ?
In what situation do we need that ?

libs/math/include/math/free_functions/matrix_operations/matrix_operations.hpp
```diff
@@ -544,26 +544,60 @@ void PeakToPeak(ArrayType arr)
  * @return
  */
 template <typename ArrayType>
-void ArgMax(ArrayType const &array, typename ArrayType::SizeType &ret)
+void ArgMax(ArrayType const &array, ArrayType &ret, typename ArrayType::SizeType axis = 0)
```
pierre-wilmot-fetch-ai, line 547: Why is ArgMax not returning a SizeType anymore ?

libs/math/include/math/free_functions/matrix_operations/matrix_operations.hpp
```diff
@@ -544,26 +544,60 @@ void PeakToPeak(ArrayType arr)
...
+  {
+    // 2D argmax result has size
+    if (axis == 0)
+    {
+      ret = ArrayType{{array.shape().at(axis), typename ArrayType::SizeType(1)}};
+    }
+    else
+    {
+      ret = ArrayType{{typename ArrayType::SizeType(1), array.shape().at(axis)}};
+    }
```
pierre-wilmot-fetch-ai, line 606: Processing for axis == 1 is most likely broken

libs/math/include/math/free_functions/standard_functions/log.hpp
```diff
@@ -70,25 +77,27 @@ meta::IfIsNonBlasArray<ArrayType, void> Log(ArrayType &x)
...
 template <typename T>
 meta::IfIsMathArray<T, void> Log(T const &array, T &ret)
 {
   ret = array;
-  for (typename T::Type &e : ret)
+  typename T::SizeType ret_count{0};
+  for (typename T::Type &e : array)
   {
-    Log(e);
+    Log(e, ret.At(ret_count));
```
pierre-wilmot-fetch-ai, line 102: Since ret = array (L95), you're effectivily overwriting your input.

libs/math/tests/math/ml_loss_functions/cross_entropy.cpp
```diff
@@ -105,6 +105,5 @@ TYPED_TEST(CrossEntropyTest, value_test)
   score = fetch::math::CrossEntropyLoss(test_array, gt_array);
 
   // test correct values
-  ASSERT_NEAR(score, typename TypeParam::Type(1.78777539730072021484375),
-              typename TypeParam::Type(1.0e-5f));
+  ASSERT_NEAR(double(score), double(0.893887639), double(1.0e-5f));
```
pierre-wilmot-fetch-ai, line 108: Where are these values coming from ? 
That's the second time I see ground thruth values for math function unit test being modified even though the input is the same.

libs/ml/examples/file_loader.hpp
```diff
@@ -42,7 +48,7 @@ std::vector<std::string> GetAllTextFiles(std::string const &dir_name)
       p1 = strtok(NULL, ".");
       if (p1 != NULL)
       {
-        txt_cmp = strcmp(p1, "txt");
+        txt_cmp = std::strcmp(p1, "txt");
```
pierre-wilmot-fetch-ai: I still think this C-code looks very out of place in out C++ codebase

libs/ml/examples/model_saver.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace fetch {
+namespace ml {
+namespace examples {
+
+/**
+ * Saves the state dict of a graph to a file location specified by user
+ * @param g the graph to save
+ * @param save_location a string specifying save location
+ * @return bool specifiy succes
```
pierre-wilmot-fetch-ai: function has no return value

libs/ml/examples/model_saver.hpp
```diff
@@ -0,0 +1,52 @@
...
+ */
+template <typename GraphType>
+void SaveModel(GraphType g, std::string const &save_location)
```
pierre-wilmot-fetch-ai: GraphType g => GraphType const &g.
No need to make a copy

libs/ml/examples/word2vec_SGNS/main.cpp
```diff
@@ -42,33 +48,34 @@ using SizeType     = typename ArrayType::SizeType;
...
-  SizeType    k               = 10;      // how many nearest neighbours to compare against
-  std::string test_word       = "cold";  // test word to consider
+  SizeType    output_size     = 1;
+  SizeType    batch_size      = 128;            // training data batch size
+  SizeType    embedding_size  = 16;             // dimension of embedding vec
+  SizeType    training_epochs = 100000;         // total number of training epochs
+  double      learning_rate   = 0.1;            // alpha - the learning rate
+  SizeType    k               = 10;             // how many nearest neighbours to compare against
+  std::string test_word       = "cold";         // test word to consider
+  std::string save_loc        = "./model.fba";  // small value for avoiding numerical instability
```
pierre-wilmot-fetch-ai: Out of sync comment

libs/ml/include/ml/ops/activations/sigmoid.hpp
```diff
@@ -45,6 +46,14 @@ class Sigmoid : public fetch::ml::ElementWiseOps<T>
     }
 
     fetch::math::Sigmoid(inputs.front().get(), *this->output_);
+
+    // ensures numerical stability
+    for (auto &val : *this->output_)
+    {
+      fetch::math::Max(val, epsilon_, val);
+      fetch::math::Min(val, fetch::math::Subtract(DataType(1), epsilon_), val);
```
pierre-wilmot-fetch-ai, line 54: Let maybe add a fetch::math::Clamp(T &val, Tconst &min, T const &max)

libs/ml/tests/ml/ops/activations/sigmoid.cpp
```diff
@@ -58,8 +58,7 @@ TYPED_TEST(SigmoidTest, backward_test)
   TypeParam           gt(8);
   std::vector<double> dataInput({1, -2, 3, -4, 5, -6, 7, -8});
   std::vector<double> errorInput({0, 0, 0, 0, 1, 0, 0, 0});
-  std::vector<double> gtInput({-0.196612, -0.104994, -0.0451767, -0.0176627, 0.993352, -0.00246651,
-                               -0.000910167, -0.000335238});
+  std::vector<double> gtInput({0, 0, 0, 0, 0.00664803, 0, 0, 0});
```
pierre-wilmot-fetch-ai, line 61: Where are these value coming from ?
Where were the previous values coming from ?

libs/ml/tests/ml/ops/loss_functions/cross_entropy.cpp
```diff
@@ -96,40 +96,136 @@ TYPED_TEST(CrossEntropyTest, one_dimensional_forward_test)
   }
 
   fetch::ml::ops::CrossEntropy<TypeParam> op;
-  ASSERT_FLOAT_EQ(float(op.Forward({data1, data2})), 1.78777539730072021484375);
+  ASSERT_FLOAT_EQ(float(op.Forward({data1, data2})), float(0.893887639));
```
pierre-wilmot-fetch-ai, line 99: Where are these value coming from ?
Where were the previous values coming from ?

scripts/python_ml_lib/cross_entropy_test.py
```diff
@@ -0,0 +1,55 @@
+import numpy as np
```
pierre-wilmot-fetch-ai, line 1: Leftover test code ?

libs/ml/tests/CMakeLists.txt
```diff
@@ -13,5 +13,5 @@ add_fetch_test(ml_ops_gtest fetch-ml ml/ops)
 add_fetch_test(ml_serializers_gtest fetch-ml ml/serializers)
 add_fetch_test(ml_dataloader_gtest fetch-ml ml/dataloader)
 add_fetch_test(ml_layers_gtest fetch-ml ml/layers)
-
+add_fetch_test(ml_training_gtest fetch-ml ml/training)
```
pierre-wilmot-fetch-ai: Let's mark these est as integration
Test #36: ml_training_gtest .................   Passed   16.59 sec

pierre-wilmot-fetch-ai (changes requested):

libs/core/include/core/fixed_point/fixed_point.hpp
baykaner: as discussed argmax or indexing in general requires DataType -> SizeType conversions

libs/math/include/math/free_functions/matrix_operations/matrix_operations.hpp
baykaner: so that ArgMax can handle a 2D matrix of for example (n_data x n_features) and return a vector of arg maxes. If you want argmax over the whole 2d matrix I guess it should be flatten first or call argmax on the resultant vector.

Do you think a different interface should be preferred for handling both argmax over entire 2d tensor AND batch argmax?

libs/math/include/math/free_functions/matrix_operations/matrix_operations.hpp
baykaner: yep nice catch

libs/math/include/math/free_functions/standard_functions/log.hpp
baykaner: yep, and the two log functions defined above have the wrong interface too. fixed

libs/math/tests/math/ml_loss_functions/cross_entropy.cpp
baykaner: python script uploaded showing how to calculate values in numpy for future reference if necessary

libs/ml/include/ml/ops/activations/sigmoid.hpp
baykaner: yes, this is going to be an extremely resuable piece of code. I'll open up an issue to create it and replace this implementation with it, but leave it out of this PR.

libs/ml/tests/ml/ops/activations/sigmoid.cpp
baykaner: python script should show how to generate test values now. In future we should add the python scripts demonstrating how the values are arrived at (whether its numpy, pytorch, tensorflow, or otherwise). Currently I just pushed to scripts/python_ml_lib/ but probably we can tidy up a bit where we keep these.

scripts/python_ml_lib/cross_entropy_test.py
baykaner: no; scripting code for validating the values in the unit tests

libs/math/include/math/free_functions/matrix_operations/matrix_operations.hpp
baykaner: fixed to return Tensor<SizeType> but calling Argmax(array, axis) to return a SizeType now requires the user to be explicit about the templates e.g. SizeArrayType ret = ArgMax<ArrayType, SizeArrayType>(array, axis)

baykaner: actually on second thought I'm not going to fix this now because it requires adding a new SizeArrayType template to Argmax, but that ends up propagating through most of the ML library annoyingly. The better solution is to rework math and ml library such that instead of just ArrayType template, we give <ArrayType, DataType, SizeType> (or possibly just <DataType, SizeType> in cases where tensor is already included). This is big job so I'm going to bundle this together with removing the shapeless array and do this all in one go. Will add to the shapelessarray issue

baykaner: issue #794 regarding this

libs/ml/examples/file_loader.hpp
baykaner: yup, updated

libs/ml/examples/file_loader.hpp
```diff
@@ -38,18 +38,15 @@ std::vector<std::string> GetAllTextFiles(std::string const &dir_name)
...
     {
-      strtok(ent->d_name, ".");
-      p1 = strtok(NULL, ".");
-      if (p1 != NULL)
+      p1 = ent->d_name;
+      if (p1.find('.'))
       {
-        txt_cmp = std::strcmp(p1, "txt");
-        if (txt_cmp == 0)
+        if (p1.compare("txt"))
```
pierre-wilmot-fetch-ai: That will never evaluate to true, unless the folder contains a file called just "txt"

pierre-wilmot-fetch-ai (dismissed):

libs/ml/examples/file_loader.hpp
baykaner: 👍

pierre-wilmot-fetch-ai (approved):

PR #789, opened 2019-03-26
Initial Token Contracts

ejfitzgerald:
- Rework of non-primitive serialisation format
- Removal of `vm::Address::AsString()`
- Rework of the contract initialiser code
- Added unit test for the basic token contract

### DIFF
--- a/libs/ledger/src/chaincode/smart_contract.cpp
+++ b/libs/ledger/src/chaincode/smart_contract.cpp
@@ -39,7 +39,6 @@
 
 using fetch::byte_array::ConstByteArray;
 using fetch::byte_array::FromBase64;
-using fetch::byte_array::ToBase64;
 
 namespace fetch {
 namespace ledger {
--- a/libs/ledger/src/chaincode/smart_contract.cpp
+++ b/libs/ledger/src/chaincode/smart_contract.cpp
@@ -107,7 +107,7 @@ SmartContract::SmartContract(std::string const &source)
                                  {"No source present in contract"});
   }
 
-  FETCH_LOG_WARN(LOGGING_NAME, "Constructing contract: ", contract_digest().ToBase64());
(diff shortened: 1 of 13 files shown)


ejfitzgerald:
FYI: @devjsc

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -425,43 +419,44 @@ Contract::Status SmartContract::InvokeAction(std::string const &name, Transactio
...
   // lookup the function / entry point which will be executed
-  Script::Function const *target_function = script_->FindFunction(name);
+  Script::Function const *target_function = script_->FindFunction(init_fn_name_);
   if (target_function->num_parameters == 1)
   {
-    FETCH_LOG_INFO(LOGGING_NAME, "One argument for init - defaulting to address population");
+    FETCH_LOG_DEBUG(LOGGING_NAME, "One argument for init - defaulting to address population");
+
+    // create the public key from the identity
+    auto const identity = owner.identifier();
```
pbukva: `auto const &identity`

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -149,8 +149,7 @@ Contract::Status SmartContractManager::OnCreate(Transaction const &tx)
     smart_contract.Attach(state());
 
     // Dispatch to the init. method
-    auto status = smart_contract.DispatchTransaction(on_init_function, tx);
-
+    auto const status = smart_contract.DispatchInitialise(tx.signatures().begin()->first);
```
pbukva, line 152: Here we assume some that either there will be just single singatory, or the first signatory is the one we use as identifier for the contract.

pbukva (approved):
LGTM

PR #790, opened 2019-03-28
Tanh implementation

juergendkeba:
Added Tanh functions to the standard_funcs in math, added tanh as op in computational graph


libs/math/include/math/free_functions/ml/activation_functions/tanh.hpp
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * The tanh function
+ * @tparam ArrayType
+ * @tparam T
+ * @param y
+ * @param y_hat
+ * @param ret
+ */
+template <typename ArrayType>
+void TanhLayer(ArrayType const &t, ArrayType &ret)
```
baykaner: As we discussed the interface here is good, but please move the implementation out from kernel to here and ditch the kernel file.
Afterwards it will be safe to rename this to Tanh (since it's not a layer).

libs/math/include/math/free_functions/standard_functions/trigonometric.hpp
```diff
@@ -0,0 +1,86 @@
...
+  for (typename ArrayType::Type &e : x)
+  {
+    Tanh(e, e);
+  }
+}
+
+template <typename T>
+fetch::math::meta::IfIsMathArray<T, void> Tanh(T const &array, T &ret)
+{
+  ret = array;
```
baykaner: I suspect this doesn't do what you intend.
Tensors are quite 'light' in that there will be no copy operation if you do this; you'll simply get a new tensor that points to the same storage data. In any case we would like to avoid unnecessary copies, I think you want the following:

assert(ret.shape() == array.shape())
T::SizeType idx{0};
for (auto &e : array)
{
  Tanh(e, ret.At(idx);
}

Strictly speaking this implementation is unsafe if since you could call it with a shapeless array and the check of .shape() is unsafe; however since this is a new feature and we're phasing out the shapeless array it shouldn't be a problem

libs/math/include/math/free_functions/standard_functions/trigonometric.hpp
```diff
@@ -0,0 +1,86 @@
...
+template <typename ArrayType>
+fetch::math::meta::IfIsNonBlasArray<ArrayType, void> Tanh(ArrayType &x)
+{
+  for (typename ArrayType::Type &e : x)
+  {
+    Tanh(e, e);
+  }
+}
+template <typename ArrayType>
+fetch::math::meta::IfIsMathFixedPointArray<ArrayType, void> Tanh(ArrayType &x)
```
baykaner: not the preferred interface; should take input parameter const &x, and return the outut.

libs/math/include/math/free_functions/standard_functions/trigonometric.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+template <typename ArrayType>
+fetch::math::meta::IfIsBlasArray<ArrayType, void> Tanh(ArrayType &x)
+{
+  free_functions::kernels::Tanh<typename ArrayType::Type> kernel;
+  x.data().in_parallel().Apply(kernel, x.data());
+}
+
+template <typename ArrayType>
+fetch::math::meta::IfIsNonBlasArray<ArrayType, void> Tanh(ArrayType &x)
```
baykaner: looking forward to deleting most of these duplicate implementations after we have only tensors :-p

libs/math/include/math/free_functions/standard_functions/trigonometric.hpp
```diff
@@ -0,0 +1,86 @@
...
+#include "math/meta/math_type_traits.hpp"
+
+/**
+ * tanh(x)
+ * @param x
+ */
+
+namespace fetch {
+namespace math {
+
```
baykaner: I'm wondering if we need a namespace for trig; have you noticed any collisions?

libs/math/include/math/free_functions/standard_functions/trigonometric.hpp
```diff
@@ -0,0 +1,86 @@
...
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "core/fixed_point/fixed_point.hpp"
+#include "math/kernels/standard_functions/trigonometric.hpp"
+#include "math/meta/math_type_traits.hpp"
+
+/**
+ * tanh(x)
```
baykaner: useless comments?

libs/math/include/math/free_functions/standard_functions/trigonometric.hpp
```diff
@@ -0,0 +1,86 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "core/fixed_point/fixed_point.hpp"
```
baykaner: do you actually need to include this?

libs/math/include/math/kernels/standard_functions/trigonometric.hpp
```diff
@@ -0,0 +1,39 @@
+#pragma once
+//------------------------------------------------------------------------------
```
baykaner: as discussed ditch this and move the implementation into standard_functions/trigonometric

libs/math/include/math/free_functions/ml/activation_functions/tanh.hpp
baykaner: alternatively it's fine to just move all of this into standard functions/trig
It's fairly simple reusable code and theres nothing unique about these interfaces to the ML library

libs/math/include/math/free_functions/ml/activation_functions/tanh.hpp
```diff
@@ -0,0 +1,54 @@
...
+    typename ArrayType::SizeType idx(0);
+    for(auto &val: t){        
+        Tanh(val, ret.At(idx));
+        ++idx;
+    };
+    
+}
+
+template <typename ArrayType>
+ArrayType TanhLayer(ArrayType const &t)
```
baykaner: not a layer

libs/ml/include/ml/ops/tanh.hpp
```diff
@@ -0,0 +1,71 @@
...
+namespace ml {
+namespace ops {
+
+template <class T>
+class Tanh : public fetch::ml::ElementWiseOps<T>
+{
+public:
+  using ArrayType    = T;
+  using DataType     = typename ArrayType::Type;
+  using ArrayPtrType = std::shared_ptr<ArrayType>;
```
baykaner: i suspect this is redundant

libs/ml/include/ml/ops/tanh.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+  virtual ArrayType Forward(std::vector<std::reference_wrapper<ArrayType const>> const &inputs)
+  {
+    assert(inputs.size() == 1);
+    if (!this->output_ || this->output_->shape() != inputs.front().get().shape())
+    {
+      this->output_ = std::make_shared<ArrayType>(inputs.front().get().shape());
+    }
+
+    fetch::math::TanhLayer(inputs.front().get(), *this->output_);
```
baykaner: rename - not a layer

libs/ml/include/ml/ops/tanh.hpp
```diff
@@ -0,0 +1,71 @@
...
+      std::vector<std::reference_wrapper<ArrayType const>> const &inputs,
+      ArrayType const &                                           errorSignal)
+  {
+    assert(inputs.size() == 1);
+
+    assert(inputs.front().get().shape() == errorSignal.shape());
+
+    ArrayType t            = this->Forward(inputs);
+    ArrayType returnSignal = errorSignal.Clone();
+    fetch::math::Add(errorSignal, fetch::math::Subtract(DataType(1), fetch::math::Multiply(t,t)),
```
baykaner: I think as you mentioned previously this Add should be a multiply due to chain rule.
Also if you can split these three operations over three lines without unecessary objection creation or copies that would be clearer to read. e.g.:

// gradient of tanh: 1 - tanh(x)^2
ArrayType t            = this->Forward(inputs);
fetch::math::Multiply(t,t,t);
fetch::math::Subtract(DataType(1), t, t)

// apply chain rule
fetch::math::Multiply(errorSignal, t, returnSignal);

libs/ml/include/ml/ops/tanh.hpp
```diff
@@ -0,0 +1,71 @@
...
+  virtual std::vector<ArrayType> Backward(
+      std::vector<std::reference_wrapper<ArrayType const>> const &inputs,
+      ArrayType const &                                           errorSignal)
+  {
+    assert(inputs.size() == 1);
+
+    assert(inputs.front().get().shape() == errorSignal.shape());
+
+    ArrayType t            = this->Forward(inputs);
+    ArrayType returnSignal = errorSignal.Clone();
```
baykaner: actually this is an example of bad style; i know you've copied existing bad style :-p

the general rule is classes and functions should be PascalCase and variable should be snake_case so this should be return_signal and error_signal.

libs/ml/tests/ml/ops/tanh.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <gtest/gtest.h>
+#include "ml/ops/tanh.hpp"
+
+
+template <typename T>
+class TanhTest : public ::testing::Test
+{
+};
+
+using MyTypes = ::testing::Types<fetch::math::Tensor<int>,
```
baykaner: is this really going to work for a Tensor<int>? seems a bit unlikely

libs/ml/tests/ml/ops/tanh.cpp
```diff
@@ -0,0 +1,80 @@
...
+                                 fetch::math::Tensor<float>,
+                                 fetch::math::Tensor<double>,
+                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<16, 16>>,
+                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<32, 32>>
```
baykaner: Just so you know, we sometimes run into overflow issues for FixedPoint<16, 16> and currently some tests don't cover it for this reason. If you have troubles with this feel free to drop it

libs/ml/tests/ml/ops/tanh.cpp
```diff
@@ -0,0 +1,80 @@
...
+                                 >;
+
+TYPED_TEST_CASE(TanhTest, MyTypes);
+
+TYPED_TEST(TanhTest, forward_all_positive_test)
+{
+  TypeParam     data(1);
```
baykaner: A very pedantic point but preferred style is that variable initialization is performed with {} rather than (), because () can look like function invokation.

libs/ml/tests/ml/ops/tanh.cpp
```diff
@@ -0,0 +1,80 @@
...
+  {
+    data.Set(i, typename TypeParam::Type(dataInput[i]));
+    gt.Set(i, typename TypeParam::Type(gtInput[i]));
+    i++;
+  }
+
+  fetch::ml::ops::Tanh<TypeParam> op;
+  TypeParam                       prediction = op.Forward({data});
+
+  ASSERT_TRUE(prediction.AllClose(gt));
```
baykaner: add some backwards tests

libs/ml/tests/ml/ops/tanh.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
```
baykaner, line 128: Probably also worth adding a numerical precision guard inside the Tanh op as well (please see my implementation in activations/sigmoid.hpp for open PR787  https://github.com/uvue-git/fetch-ledger/pull/787/files)
Then ideally we should add another test that checks passing extreme values results in safe operation.

baykaner (changes requested):
I've suggested some changes. Please fix them and also make sure you get the green light from jenkins then I'll re-review.
Nice work!

libs/math/include/math/free_functions/standard_functions/trigonometric.hpp
juergendkeba: In order to get rid of that inplace operations.. can we assume that all `MathFixedPointArray` and `NonBlasArrays` have a copy constructor? 

Cause then, I could use -- even though not perfect -- the following 
```
  ArrayType ret(x);
  T::SizeType idx{0};
  for (typename ArrayType::Type &e : x)
  {
    ret.At(idx) = some_func_on_(e)
    idx++;
  }

return ret
```

(( I'd more prefer to do somethling like `ArrayType ret(x.shape);` but I don't know, every ArrayType can be constructed by passing a shape?))

libs/ml/tests/ml/ops/tanh.cpp
juergendkeba: Do you mean, where you check against `DataType epsilon_ = DataType(1e-12);` 

If yes, I'll add it!

libs/ml/tests/ml/ops/tanh.cpp
juergendkeba: Done!

libs/ml/tests/ml/ops/tanh.cpp
juergendkeba: Good practice is never pedantic! :-) Thx for the info!

libs/ml/tests/ml/ops/tanh.cpp
juergendkeba: Jep, I ran also into numeric problems! Commented it out!

libs/ml/tests/ml/ops/tanh.cpp
juergendkeba: It worked, but not for extreme edge cases, cause
`int(f(a,b)) != f(int(a), int(b))` :-)
Dropped it now! (Just used it, cause I saw it, in the relu tests)

libs/ml/include/ml/ops/tanh.hpp
juergendkeba: K! If that's the preferred style, I'll stick to it!

libs/ml/include/ml/ops/tanh.hpp
juergendkeba: :-) Alright!

libs/ml/include/ml/ops/tanh.hpp
juergendkeba: Done!

libs/ml/include/ml/ops/tanh.hpp
juergendkeba: You mean the shared_ptr declaration? You mean because it's not used? (Took the relu implementation as a template for that and thought it's the preferred style)

libs/math/include/math/kernels/standard_functions/trigonometric.hpp
juergendkeba: Done

libs/math/include/math/free_functions/standard_functions/trigonometric.hpp
juergendkeba: Done!

libs/math/include/math/free_functions/standard_functions/trigonometric.hpp
juergendkeba: nope, haven't noticed any!

libs/math/include/math/free_functions/ml/activation_functions/tanh.hpp
juergendkeba: Done!

libs/math/include/math/free_functions/ml/activation_functions/tanh.hpp
juergendkeba: Put everything into "standard_funcs/trigonometry"!

libs/math/include/math/free_functions/standard_functions/trigonometric.hpp
juergendkeba: Removed!

libs/math/include/math/free_functions/standard_functions/trigonometric.hpp
juergendkeba: Removed!

libs/math/include/math/free_functions/ml/activation_functions/tanh.hpp
baykaner: yes; everything to do with trigonometry in the math library can go there I think, since in this case the math/free_functions/ml/activation_functions?tanh.hpp is really not doing a lot of useful work for us!

libs/math/include/math/free_functions/standard_functions/trigonometric.hpp
baykaner: No, you must be extremely careful with the copy constructor because it does NOT perform a deep copy of the data.
Please assume that ArrayType will always be a Tensor<DataType>, feel free to look at tensor.hpp for details on how it works, but in general you can assume the following:
- ArrayType ret(x); // Will instantiate a new Tensor but with shared data (so changes to ret will change x)
- ArrayType ret = x.Clone(); // This will perform a deep copy of the tensor
- ArrayType ret{x.shape()); // this will always work (i'm removing the shapeless array)
- ArrayType ret{x.size()); // this will also work

libs/ml/include/ml/ops/tanh.hpp
baykaner: Yup, right idea, but the existing code is not perfect - sometimes there is leftover junk that we haven't cleaned up yet!

libs/ml/tests/ml/ops/tanh.cpp
baykaner: yep, that's what i mean

juergendkeba: done

juergendkeba:
In the tests for the tanh-op (tests/ops/tanh.hpp), I had to comment out the tests for `fetch::fixed_point::FixedPoint<32ul, 32ul>`. The values diverge by quite some amount:
```
[ RUN      ] TanhTest/2.forward_all_negative_test
AllClose[1] - 0.00000000000000000000000000000000 != -0.19737499998882412910461425781250

[ RUN      ] TanhTest/2.backward_all_negative_test
AllClose[1] - -0.09999999986030161380767822265625 != -0.09610399999655783176422119140625
```
But, it works, when I don't use the "epsilon"-check in the op implementation! see comment https://github.com/uvue-git/fetch-ledger/pull/790#discussion_r270148856

libs/math/include/math/fundamental_operators.hpp
```diff
@@ -186,6 +138,54 @@ meta::IfIsMathArray<ArrayType, void> Add(ArrayType const &array1, ArrayType cons
...
+
+template <typename S>
+meta::IfIsArithmetic<S, S> Add(S const &scalar1, S const &scalar2)
+{
+  S ret;
+  Add(scalar1, scalar2, ret);
+  return ret;
+}
+// template <typename S>
+// meta::IfIsFixedPoint<S, S> Add(S const &scalar1, S const &scalar2)
```
baykaner: generally better not to leave in commented out code. Just remove this. I'll be pushing an update to this file soon anyway.

libs/ml/include/ml/ops/tanh.hpp
```diff
@@ -0,0 +1,92 @@
...
+      this->output_ = std::make_shared<ArrayType>(inputs.front().get().shape());
+    }
+
+    fetch::math::TanH(inputs.front().get(), *this->output_);
+
+    // ensures numerical stability
+    for (auto &val : *this->output_)
+    {
+      // Minimum value of tanh is restricted to -1+epsilon
+      fetch::math::Max(val, fetch::math::Add(DataType(1) * -1, epsilon_), val);
```
baykaner: presumably this line is the culprit for why it does not work with fixed_point. It might be that * is calling an operator that behaves unsafely with fixed_point at the moment. If in doubt use fetch::math::Multiply instead.
In this case however, what happens if you replace this with:
fetch::math::Max(val, fetch::math::Add(DataType(-1), epsilon_), val);

libs/ml/include/ml/ops/tanh.hpp
```diff
@@ -0,0 +1,92 @@
...
+  {
+    assert(inputs.size() == 1);
+
+    assert(inputs.front().get().shape() == error_signal.shape());
+
+    ArrayType return_signal = error_signal.Clone();
+    ArrayType t             = this->Forward(inputs);
+
+    // gradient of tanh: 1 - tanh(x)^2
+    fetch::math::Multiply(t, t, t);
```
baykaner, line 73: that's fine, no need to change it, but be aware that fetch::math::Square(t, t); would also work

baykaner (changes requested):

libs/ml/include/ml/ops/tanh.hpp
juergendkeba: Indeed, this works! (y)

libs/ml/include/ml/ops/tanh.hpp
juergendkeba: Thx!

libs/math/include/math/fundamental_operators.hpp
juergendkeba: done

baykaner (approved):

PR #802, opened 2019-04-01
math library reorg

baykaner:
-shapeless array removal
-rectangular array removal
-free_functions subdirectories moved
-some new unit tests
-re-implemented some functions for Tensor


libs/math/tests/math/distance/euclidean_distance.cpp
```diff
@@ -26,7 +26,18 @@
...
+
+using MyTypes = ::testing::Types<fetch::math::Tensor<float>, fetch::math::Tensor<double>,
+                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<16, 16>>,
+                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<32, 32>>>;
+
+TYPED_TEST_CASE(EuclideanTest, MyTypes);
+
+TYPED_TEST(EuclideanTest, simple_test)
 {
   Tensor<double> A = Tensor<double>(4);
```
pierre-wilmot-fetch-ai, line 42: Test typing is only halway done.
Should be TypeParam instead of Tensor<double>.
The fix is probably waiting for merge in Jiri PR

libs/math/tests/math/distance/hamming_tests.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+using MyTypes = ::testing::Types<fetch::math::Tensor<float>, fetch::math::Tensor<double>,
+                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<16, 16>>,
+                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<32, 32>>>;
+
+TYPED_TEST_CASE(HammingTest, MyTypes);
+
+TEST(HammingTest, simple_test)
+{
+  Tensor<double> A = Tensor<double>(4);
```
pierre-wilmot-fetch-ai, line 44:
```
TYPED_TEST(HammingTest, simple_test)
{
  TypeParam A = TypeParam(4);
...
```

libs/math/tests/math/distance/manhattan_tests.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+using MyTypes = ::testing::Types<fetch::math::Tensor<float>, fetch::math::Tensor<double>,
+                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<16, 16>>,
+                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<32, 32>>>;
+
+TYPED_TEST_CASE(ManhattenTest, MyTypes);
+
+TYPED_TEST(ManhattenTest, simple_test)
+{
+  Tensor<double> A = Tensor<double>(3);
```
pierre-wilmot-fetch-ai, line 43: Same

pierre-wilmot-fetch-ai (approved):
Thanks for cleaning up 👍 
Just some typed tests that are not actually typed properlly

libs/math/include/math/correlation/cosine.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+  dp_ret = fetch::math::DotTranspose(a, a);
+  assert(dp_ret.size() == 1);
+  a_r = dp_ret[0];
+
+  dp_ret = fetch::math::DotTranspose(b, b);
+  assert(dp_ret.size() == 1);
+  b_r = dp_ret[0];
+
+  denom_r = std::sqrt(fetch::math::Multiply(a_r, b_r));
```
MissingNO57, line 51: Shouldn't you need to cast return value from Multiply to double?  std::sqrt wouldn't work for non-primitive data types.

baykaner: yes you're right - I should have called our math libraries' Sqrt rather than std::sqrt

PR #803, opened 2019-04-02
further cleanup of math library

baykaner:
- typeparam bugfixes in tests
- invalid std::sqrt call in cosine
- reworked fundamental_operators
- added unit tests for fundamental_operators


libs/math/include/math/fundamental_operators.hpp
```diff
@@ -17,563 +17,644 @@
...
+ * Scalar-Array subtraction
+ * @tparam T
+ * @tparam ArrayType
+ * @param array
+ * @param scalar
+ * @param ret
+ * @return
+ */
+template <typename ArrayType, typename T, typename = std::enable_if_t<meta::IsArithmetic<T>>>
+meta::IfIsValidArrayScalarPair<ArrayType, T, void> Subtract(T const &scalar, ArrayType const &array,
```
bipll, line 143: Here  actually two SFINAE styles are mixed, template parameter default value and function ret type. Perhaps it would be more consistent to check both predicates at the same site.

libs/math/include/math/fundamental_operators.hpp
```diff
@@ -17,563 +17,644 @@
...
+ * scalar division implementation
+ * @tparam S
  * @param scalar1
  * @param scalar2
  * @param ret
+ * @return
  */
 template <typename S>
-meta::IfIsArithmetic<S, void> Add(S const &scalar1, S const &scalar2, S &ret)
+meta::IfIsArithmetic<S, void> Divide(S const &scalar1, S const &scalar2, S &ret)
```
bipll, line 252: `meta`'s type predicates all default their last parameter to `void`. Since it is `void` anywhere in this file, it probably makes sense to simply remove it (as is done with `std::enable_if`), slightly unscrew the boilerplate.

baykaner:
some of this became obsolete; and some has been merged into PR #897

PR #806, opened 2019-04-02
State Recovery

ejfitzgerald:
Implement and correctly handle the state recovery for the ledger nodes.

In addition this PR adds the SIGTERM handler to the `constellation` app this is needed to ensure that when a user or k8s kills the process it attempts to shutdown in a graceful way.


ejfitzgerald:
@pbukva when you get a chance can you have a look at this, I would be keen to try and get this one down for tomorrow morning.

libs/core/include/core/mutex.hpp
```diff
@@ -102,9 +103,11 @@ class DebugMutex : public AbstractMutex
     void Eval()
     {
       LOG_STACK_TRACE_POINT;
-      FETCH_LOG_ERROR(LOGGING_NAME, "Mutex timed out: ", filename_, " ", line_);
+      FETCH_LOG_ERROR(LOGGING_NAME, "The system will terminate, mutex timed out: ", filename_, " ",
+                      line_);
 
-      exit(-1);
+      // Send a sigint to ourselves since we have a handler for this
+      kill(0, SIGINT);
```
pbukva, line 110: | :thumbsup:
:+1:

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -815,6 +866,9 @@ char const *BlockCoordinator::ToString(State state)
   case State::RESET:
     text = "Reset";
     break;
+  default:
```
pbukva: Just detail, I would suggest drop `default:` from the switch, change all above `text=...; brake;` to `return ...;`, and `return "Unknown!"` a the end of the method.

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -537,31 +603,66 @@ void MainChain::WriteToFile()
       }
     }
 
-    // This block is now the head in our file
-    if (!failed)
+    if (failed)
     {
-      // store both the HEAD block as both its has and the "HEAD" entry
+      FETCH_LOG_WARN(LOGGING_NAME,
```
pbukva, line 608: just comment: this should be rather error.

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -483,6 +489,7 @@ void MainChain::RecoverFromFile(Mode mode)
   if (Mode::CREATE_PERSISTENT_DB == mode)
```
pbukva, line 489: this should be rather `switch(...)` (without `default:`) than if-else

ejfitzgerald: Yes I know, this used to be an if / else block. The switch was a little more difficult to read I think. I will have another look at this in the next chain update I am planning on doing today

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -537,31 +603,66 @@ void MainChain::WriteToFile()
...
         {
-          block_store_->Set(block_storage_key, *block);
+          block_store_->Get(storage::ResourceID(current_file_head->body.previous_hash),
+                            *current_file_head);
         }
 
-        // Clear the block from ram
-        FlushBlock(block);
+        // Successful case
+        if (current_file_head->body.hash == block->body.previous_hash)
```
pbukva, line 647: Perhaps it would be better to use rather block number to compare blocks -  it is more performant & more consistent with other code above.
```suggestion
        if (current_file_head->body.block_number == expected_block_num)
```

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -537,31 +603,66 @@ void MainChain::WriteToFile()
...
+      // touch it or it's root.
+      for (;;)
       {
-        storage::ResourceID block_storage_key{block->body.hash};
+        block_store_->Set(storage::ResourceID(block->body.hash), *block);
 
-        // add the block to the file storage if it
-        if (!block_store_->Has(block_storage_key))
+        // Keep the current_file_head one block behind
+        while (current_file_head->body.block_number != block->body.block_number - 1)
```
pbukva, line 640:
```suggestion
        auto const expected_block_num = block->body.block_number - 1;
        while (current_file_head->body.block_number != expected_block_num)
```

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -537,31 +603,66 @@ void MainChain::WriteToFile()
...
+
+      // Success - we kept a copy of the new head to write
+      block_store_->Set(storage::ResourceAddress("head"), *block_head);
     }
 
+    // Clear the block from ram
+    FlushBlock(block);
+
+    // Force flush of the file object!
+    block_store_->Flush(false);
```
pbukva, line 664: Perhaps it would be better to use lazy flush here, to take advantage of possible speculative lazy flush optimisation?
This would need to add the `block_store_->Flush(false);` at the end of this method to make sure that hard-core flush is executed once at the end.

libs/ledger/src/storage_unit/storage_unit_client.cpp
```diff
@@ -101,156 +106,163 @@ byte_array::ConstByteArray StorageUnitClient::LastCommitHash()
...
+    {
+      FETCH_LOG_INFO(LOGGING_NAME, "Reverting state to genesis.");
 
-  MerkleTreePtr tree;
-  if (genesis_state)
-  {
-    // create the new tree
-    tree = std::make_shared<MerkleTree>(num_lanes());
+      // create the new tree
+      MerkleTree tree{num_lanes()};
```
pbukva: this is not right, it should be deleted, and instance created above the `if(...)` shall be used instead

libs/ledger/src/storage_unit/storage_unit_client.cpp
```diff
@@ -101,156 +106,163 @@ byte_array::ConstByteArray StorageUnitClient::LastCommitHash()
...
+      FETCH_LOG_DEBUG(LOGGING_NAME, "Successfully found merkle at: ", index);
+    }  // End set merkle stack
+
+    // Note: we shouldn't be touching the lanes at this point from other threads
+    std::vector<service::Promise> promises;
+
+    // the check to see if the hash exists is only necessary for non genesis blocks since we know
+    // that we can always revert back to nothing! Note: this functionality doesn't exist and will
+    // crash the node
+    if (!genesis_state)
```
pbukva: also not right, should be executed for geneses state as well, or completely drop, since all what this `if(...){...}` block is doing is just checking if peers have block with the particular hash, what is not necessary, direct request to revert to that block is enough (in `for(...)` bellow).

libs/ledger/src/storage_unit/storage_unit_client.cpp
```diff
@@ -101,156 +106,163 @@ byte_array::ConstByteArray StorageUnitClient::LastCommitHash()
...
-    uint32_t              lane_counter = 0;
-
     for (auto &p : promises)
     {
       FETCH_LOG_PROMISE();
-
       if (!p->As<bool>())
       {
-        failed_lanes.push_back(lane_counter);
+        throw std::runtime_error(
```
pbukva: Ideally, this shouldn't throw, it should wait for all promises and accumulate result to single over-all Boolean result

libs/storage/include/storage/new_versioned_random_access_stack.hpp
```diff
@@ -409,12 +417,35 @@ class NewVersionedRandomAccessStack
...
     return internal_bookmark_index_ - 1;
   }
 
-  bool HashExists(DefaultKey const & /*key*/) const
+  bool HashExists(DefaultKey const &key) const
   {
-    throw std::runtime_error("HashExists functionality not yet implemented");
+    if (hash_history_.empty())
+    {
+      FETCH_LOG_WARN(LOGGING_NAME, "Attempted to find if hash exists, but history is empty!");
```
pbukva, line 430: this log msg should be either removed, or made debug rather than warn (objective of the method is to find out if has exists).

libs/storage/src/new_revertible_document_store.cpp
```diff
@@ -95,7 +95,9 @@ void NewRevertibleDocumentStore::Set(ResourceID const &rid, ByteArray const &val
 // State-based operations
 Hash NewRevertibleDocumentStore::Commit()
 {
-  return storage_.Commit();
+  Hash ret{std::move(storage_.Commit())};
```
pbukva, line 98: std::move should not be necessary here

pbukva (changes requested):
please see inline comments

pbukva (approved):
LGTM

PR #809, opened 2019-04-03
Feature/message signing

bipll:
Broadcasts signed/verified.


libs/network/src/muddle/router.cpp
```diff
@@ -230,12 +231,15 @@ void Router::Stop()
 
 inline bool Router::Genuine(PacketPtr const &p) const
 {
-  return p->Verify() || !(prover_ || p->IsStamped());
+  return p->IsBroadcast() ?
```
ejfitzgerald: I would suggest breaking this set of ternary statements up so that it is clearer

ejfitzgerald (changes requested):
Only one minor comment but looks good.

ejfitzgerald (approved):

PR #814, opened 2019-04-04
Style/not true and not false

bipll:
Mostly removed comparisons against boolean literals, also changed things like `EXPECT_TRUE(one < two)` into `EXPECT_LT(one, two)`; additionally fixed few member comparison predicates that lacked const-qualification.


ejfitzgerald (dismissed):
Looks good to me. Was there any clang tidy configurations here that might help enforce this in the future?

ejfitzgerald (approved):

bipll:
> Was there any clang tidy configurations here that might help enforce this in the future?

Not yet, I've done it in vim (with later apply-style.py).

PR #815, opened 2019-04-04
Clean Output From Ops

pierre-wilmot-fetch-ai:
This the step needed to get the batch processing multithreaded.
Now ops are just thread safe functions with no internal state.
Caching is handled at Node level, and threading can be added in BatchOps::ForwardBatch


libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -71,6 +72,13 @@ class FullyConnected : public Layer<T>
     this->SetInput(bias, bias_data);
   }
 
+  virtual std::vector<SizeType> ComputeOutputSize(
```
baykaner: Okay, so I think the design should change slightly regarding this ComputeOutputSize function. Currently a vector is being created on every call to ComputeOutputSize which is going to be inside a loop; A better implementation would be to create the vector once, and just access it.

So this should be a function more like this:
virtual std::vector<SizeType> const& output_size()
{
   return output_size_;
}

With output_size_ being set on construction. (we should use lower case for functions that will execute cheaply/quickly)

However, the problem with that is the user should be allowed to vary the batch size, which will change the output size. However, I think if we're smart about it we can get around this.

One solution could be as follows:
Let us start by working from the assumption that batch sizes will usually be constant. We should be able to handle any forward and backward call with any batch size, but generally there will be two batch sizes that will get used almost all of the time; batch_size = 1, and batch_size = {some_user_define_value}, and switching between these batch_sizes will be very infrequent also. If we make this assumption, then the efficient thing to do would be to store a "current_batch_size_" variable in the graph, and then implement an "UpdateOutputSizes" method in graph that gets triggered whenever batch_size != current_batch_size_.

That's going to be faster in almost all cases, and should be about the same speed in the worst case where a user continually changes batch size on every iteration.

libs/ml/include/ml/node.hpp
```diff
@@ -77,13 +77,17 @@ class Node : public NodeInterface<T>, public O
     {
       std::vector<std::reference_wrapper<const ArrayType>> inputs = GatherInputs();
       FETCH_LOG_INFO("ML_LIB", "Evaluating node [", name_, "]");
+      if (cachedOutput_.shape() != this->ComputeOutputSize(inputs))
```
baykaner: is this really how apply_style formatted this line? it looks super weird!

libs/ml/include/ml/ops/activations/sigmoid.hpp
```diff
@@ -32,29 +32,25 @@ class Sigmoid : public fetch::ml::ElementWiseOps<T>
...
   using ArrayType    = T;
   using DataType     = typename ArrayType::Type;
+  using SizeType     = typename ArrayType::SizeType;
   using ArrayPtrType = std::shared_ptr<ArrayType>;
 
   Sigmoid()          = default;
   virtual ~Sigmoid() = default;
 
-  virtual ArrayType Forward(std::vector<std::reference_wrapper<ArrayType const>> const &inputs)
+  virtual ArrayType Forward(std::vector<std::reference_wrapper<ArrayType const>> const &inputs,
```
baykaner, line 41: nice, this new way of defining Forward does seem cleaner 👍

libs/ml/include/ml/ops/activations/logsoftmax.hpp
```diff
@@ -37,8 +37,9 @@ class LogSoftmax : public fetch::ml::Ops<T>
   LogSoftmax()          = default;
   virtual ~LogSoftmax() = default;
 
-  virtual ArrayPtrType Forward(std::vector<ArrayPtrType> const &inputs)
+  virtual ArrayPtrType Forward(std::vector<ArrayPtrType> const &inputs, ArrayType &output)
   {
+    (void)output;
```
baykaner, line 42: looks like this has been changed only for compatibility but in a way that likely breaks it?
other methods no longer have a this->output_ right? So this method should be updated likewise?
If there is no test for logsoftmax that flagged this up please write one

libs/ml/include/ml/ops/add.hpp
```diff
@@ -35,19 +35,18 @@ class Add : public fetch::ml::ElementWiseOps<T>
...
-      this->output_ = std::make_shared<ArrayType>(inputs.at(0).get().shape());
-    }
+    ASSERT(output.shape() == this->ComputeOutputSize(inputs));
 
-    this->output_->Fill(DataType(0));
-    this->output_->InlineAdd(inputs[0]);
-    this->output_->InlineAdd(inputs[1]);
-    return *this->output_;
+    output.Fill(DataType(0));
+    output.InlineAdd(inputs[0]);
```
baykaner: This seems wasteful: Fill with zeros then Add inputs[0]; cant this be replaced with a single Assignment operation? (not sure if its output.Assign(inputs[0]) or output = inputs[0].Copy() due to ongoing rewrite)

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -45,9 +52,11 @@ class Ops
   virtual std::vector<ArrayType> BackwardBatch(
       std::vector<std::reference_wrapper<ArrayType const>> const &inputs,
       ArrayType const &                                           errorSignal) = 0;
+  virtual std::vector<SizeType> ComputeOutputSize(
+      std::vector<std::reference_wrapper<ArrayType const>> const &inputs) = 0;
 
-protected:
-  ArrayPtrType output_;  // TODO(private, 736) -- Remove
+  // protected:
+  //   ArrayPtrType output_;  // TODO(private, 736) -- Remove
```
baykaner, line 59: commented out code

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -57,11 +66,14 @@ template <class T>
...
 public:
-  using ArrayType = T;
+  using ArrayType    = T;
+  using SizeType     = typename ArrayType::SizeType;
+  using ArrayPtrType = std::shared_ptr<ArrayType>;
 
   virtual ArrayType ForwardBatch(std::vector<std::reference_wrapper<ArrayType const>> const &inputs)
   {
-    return this->Forward(inputs);
+    ArrayType output;  // Temporary Dummy
```
baykaner, line 75: is this actually temporary? If so what's the plan regarding replacing it? Can you open up an issue and put a TODO here?

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -90,7 +110,8 @@ class BatchOps : public Ops<T>
     for (typename ArrayType::SizeType b(0); b < inputs.front().get().shape()[0]; ++b)
     {
       ArrayType slice = inputs.front().get().Slice(b);
-      results.push_back(this->Forward({slice}));
+      ArrayType output;  // Temporary Dummy
```
baykaner, line 113: same as previous comment

baykaner (changes requested):

libs/ml/include/ml/ops/ops.hpp
pierre-wilmot-fetch-ai: Cleaned up in https://github.com/uvue-git/fetch-ledger/pull/821

libs/ml/include/ml/ops/ops.hpp
pierre-wilmot-fetch-ai: Cleaned up in https://github.com/uvue-git/fetch-ledger/pull/821

libs/ml/include/ml/ops/ops.hpp
pierre-wilmot-fetch-ai: Cleaned up in https://github.com/uvue-git/fetch-ledger/pull/821

libs/ml/include/ml/ops/add.hpp
pierre-wilmot-fetch-ai: Indeed 😅 

Still wastefull as we loop over the tensor twice.
Right way to do would be something like 
```
auto output_it = output.begin();
auto output_end = output.end();
auto a_it = inputs[0].begin();
auto b_it = inputs[1].begin();

while (output_begin != output_end)
{
   *output_it = *a_it + *b_it;
    output_it++;
    a_it++;
    b_it++;
}
```

libs/ml/include/ml/ops/activations/logsoftmax.hpp
pierre-wilmot-fetch-ai: Yeah, logsoftmax.hpp  is dead code, never compilled / used anywhere.
I flagged an issue for it : https://github.com/uvue-git/fetch-ledger/issues/818

libs/ml/include/ml/layers/fully_connected.hpp
pierre-wilmot-fetch-ai: Yeah, adding some kind of caching sounds like a good idea.
I'll try to find a clean way to do it, may be a but tricky as theoutput shape depends on the shape of all inputs, so the cache would have to look like ```vector<vector<SizeType>> cache_;```

libs/ml/include/ml/graph.hpp
```diff
@@ -163,11 +168,12 @@ class Graph : public ops::Trainable<T>
   /**
    * Resets graph cache, clearing stored evaluation outputs
```
baykaner, line 169: ...and recursively updating the input size for all downstream nodes

libs/ml/include/ml/node.hpp
```diff
@@ -125,9 +127,27 @@ class Node : public NodeInterface<T>, public O
...
+    return outputs_;
+  }
+
+  virtual void ResetCache(bool input_size_changed)
   {
     cachedOutputPresent_ = false;
+    if (input_size_changed)
+    {
+      std::vector<std::reference_wrapper<const ArrayType>> inputs = GatherInputs();
+      if (cachedOutput_.shape() != this->ComputeOutputSize(inputs))
```
baykaner: not sure whether the compiler will optimise this out or not, but just in case:
auto output_size = ComputeOutputSize(inputs);
if (cachedOutput_.shape() != output_size)
{
  cachedOutput_ = ArrayType(output_size);
}

- also shouldn't ComputeOutputSize really be ComputeOutputShape?
- also please replace all cachedOutput_ with cached_output_

libs/ml/include/ml/ops/add.hpp
```diff
@@ -35,19 +35,28 @@ class Add : public fetch::ml::ElementWiseOps<T>
...
+  virtual ArrayType Forward(std::vector<std::reference_wrapper<ArrayType const>> const &inputs,
+                            ArrayType &                                                 output)
   {
+    (void)output;
     ASSERT(inputs.size() == 2);
     ASSERT(inputs.at(0).get().size() == inputs.at(1).get().size());
-    if (!this->output_ || this->output_->shape() != inputs.at(0).get().shape())
+    ASSERT(output.shape() == this->ComputeOutputSize(inputs));
+
+    auto output_it  = output.begin();
```
baykaner, line 46: Implementations like this should not exist anywhere in the ML library, they should be in the Math library. Here the implementation should be either:
output = inputs[0].get() + inputs[1].get();
or
Add(inputs[0].get(),inputs[1].get() output); // (if you want to be very explicit)
and both of these calls should invoke the same implementation in the math library.

We have a couple other places in the ML library where the implementation needs to be removed and replaced with a call to the math library as well.

libs/ml/include/ml/ops/convolution.hpp
```diff
@@ -93,6 +88,16 @@ class Convolution : public BatchOps<T>
     return {errorSignal};
   }
 
+  virtual std::vector<SizeType> ComputeOutputSize(
+      std::vector<std::reference_wrapper<ArrayType const>> const &inputs)
+  {
+    std::vector<typename ArrayType::SizeType> outputShape;
```
baykaner, line 93: is it worth adding a test for this? I know I always get confused and slip up when trying to calculate the output shape for convolutions

libs/ml/include/ml/ops/matrix_multiply.hpp
```diff
@@ -35,38 +35,34 @@ class MatrixMultiply : public fetch::ml::BatchOps<T>
...
 
-    std::vector<SizeType> outputShape(
-        {inputs.at(0).get().shape()[0], inputs.at(1).get().shape()[1]});
-    if (!this->output_ || this->output_->shape() != outputShape)
-    {
-      this->output_ = std::make_shared<ArrayType>(outputShape);
-    }
-
     for (SizeType i(0); i < inputs.at(0).get().shape()[0]; ++i)
     {
```
baykaner: this is another place where the implementation needs to be removed and a call made the the math library

baykaner (changes requested):

libs/ml/include/ml/ops/convolution.hpp
pierre-wilmot-fetch-ai: Tested here : https://github.com/uvue-git/fetch-ledger/blob/develop/libs/ml/tests/ml/ops/convolution.cpp

libs/ml/include/ml/ops/add.hpp
pierre-wilmot-fetch-ai: fetch::math::Add is currently still implemented as naive lookup in a loop.
Keeping the iterator implementation until we optimisize the math library.

libs/ml/include/ml/ops/activations/leaky_relu.hpp
```diff
@@ -63,7 +58,7 @@ class LeakyRelu : public fetch::ml::ElementWiseOps<T>
     ArrayType t{inputs.front().get().shape()};
 
     // gradient of parametric relu function is for x<0 = a, x>=0 = 1.0
-    t = this->Forward(inputs);
+    t = this->Forward(inputs, t);
```
baykaner, line 61: as discussed we'll change this behaviour of forward at some point, but not important to do in this PR

libs/ml/include/ml/ops/add.hpp
baykaner: yepyep fine. Raised issue #844

libs/ml/include/ml/ops/convolution.hpp
baykaner: 👍

baykaner (approved):

Issue #680, opened 2019-02-27
FixedPoint operators for type casting to more types


baykaner:
tensor shuffle test has some unnecessary double casting that can be replaced when further type casting operators are implemented in fixed point

markos:
Closed in #819.

PR #819, opened 2019-04-05
FixedPoint class changes

### LINKED ISSUE #800: fixed point native implementations must replace casting to double throughout math library

markos:
Modify FixedPoint class to avoid conversions to/from doubles, arithmetic is only done in integer.
Add several constants like smallest_fraction, largest_fraction, min, max, etc
Add several tests
Inline all methods by default
Multiplication and division are done in the NextType type which is double the width. In the case of 64-bit types it uses the __int128_t provided by gcc/clang.
It will be possible to avoid this, if we restrict the ranges of the values and lose some accuracy in the fractional parts.
Optimise HighestBitSet method using the __builtin_clzll() function.
Extend tests for FixedPoint<32, 32>
Add tests for FixedPoint<16, 16>

TODO: Add tests for FixedPoint<8, 8>
TODO: Add provision for FixedPoint<128, 128>, this will also need provision of 256-bit arithmetic if we need the full range, or limit the range of the values and lose some accuracy so we do it only using 128bit arithmetic.

Closes #800, #680.


libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -130,23 +144,16 @@ void Multiply(const FixedPoint<I, F> &lhs, const FixedPoint<I, F> &rhs, FixedPoi
  * @return
  */
 template <typename T>
-std::size_t HighestSetBit(T n_input)
+inline std::size_t HighestSetBit(T n_input)
 {
-  int n = static_cast<int>(n_input);
+  uint64_t n = n_input;
```
baykaner: It may be that I've misunderstood, but this is an implicit cast from a T to a uint64_t? why is this safe? don't we get compiler errors complaining about this?

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -130,23 +144,16 @@ void Multiply(const FixedPoint<I, F> &lhs, const FixedPoint<I, F> &rhs, FixedPoi
...
 
-  std::size_t msb = 0;
-  while (n != 0)
-  {
-    n = n / 2;
-    msb++;
-  }
-
-  return msb;
+  return sizeof(n)*8 - __builtin_clzll(n);
```
baykaner: awesome! didn't know this builtin existed!

libs/core/tests/fixed_point/fixed_point.cpp
```diff
@@ -151,15 +459,128 @@ TEST(FixedPointTest, Division)
...
+  fetch::fixed_point::FixedPoint<32, 32> infinitesimal(0, one.smallest_fraction);
+  fetch::fixed_point::FixedPoint<32, 32> huge(0x40000000, 0);
+  fetch::fixed_point::FixedPoint<32, 32> small(0, 0x40000000);
+
+  EXPECT_EQ(small / infinitesimal, huge);
+  EXPECT_EQ(infinitesimal / one, infinitesimal);
+  EXPECT_EQ(one / huge, infinitesimal * 4);
+  EXPECT_EQ(huge / infinitesimal, zero);
+
+  try {
```
baykaner: is this debugging code you forgot to remove?
I'm a little confused as to why you need a try catch in a test?

libs/core/tests/fixed_point/fixed_point.cpp
```diff
@@ -151,15 +459,128 @@ TEST(FixedPointTest, Division)
...
     EXPECT_EQ(int(two / zero), 0);
     EXPECT_EQ(float(two / zero), 0.0f);
     EXPECT_EQ(double(two / zero), 0.0);
-  */
+  } catch (std::exception &e) {
+    std::cerr << "exception caught: " << e.what() << std::endl;
+  }
+  // TODO: Return a NaN
```
baykaner: There are two valid options with regards to TODOs in the codebase:
1. fix it and remove it
2. raise an issue, assign it to the relevant person, and put " // TODO(issue number): brief description " in the code

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -219,48 +226,64 @@ class FixedPoint
...
-    : data_(0)
+  inline FixedPoint()
+    : data_{0}
   {}  // initialise to zero
 
   template <typename T>
-  explicit FixedPoint(T n, meta::IfIsInteger<T> * = nullptr)
-    : data_(static_cast<Type>(n) << static_cast<Type>(fractional_bits))
+  inline explicit FixedPoint(T n, meta::IfIsInteger<T> * = nullptr)
+    : data_{static_cast<Type>(n) << static_cast<Type>(fractional_bits)}
```
baykaner: nice!

baykaner (changes requested):

libs/core/tests/fixed_point/fixed_point.cpp
markos: changed to EXPECT_THROW, thanks :+1:

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -43,17 +43,31 @@ struct TypeFromSize
...
 };
 
-// 64 bit implementation
+#if (__SIZEOF_INT128__ == 16)
+// 128 bit implementation
 template <>
-struct TypeFromSize<64>
+struct TypeFromSize<128>
 {
   static const bool        is_valid = true;
```
pbukva: I would suggest do make all these static `const` members rather static `constexpr` (this also applies for all template specializations).

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -219,48 +226,64 @@ class FixedPoint
   using NextType     = typename BaseTypeInfo::NextSize::ValueType;
   using UnsignedType = typename BaseTypeInfo::UnsignedType;
 
-  const Type fractional_mask = Type((2 ^ fractional_bits) - 1);
-  const Type integer_mask    = ~fractional_mask;
+  static const Type fractional_mask = Type((Type(1) << fractional_bits) - 1);
```
pbukva: `const` -> `constexpr`

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -219,48 +226,64 @@ class FixedPoint
...
+  static const Type integer_mask    = ~fractional_mask;
```
pbukva: `const` -> `constexpr`

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -219,48 +226,64 @@ class FixedPoint
...
 
+  // Constants/Limits
   static const Type one = Type(1) << fractional_bits;
```
pbukva: `const` -> `constexpr`

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -219,48 +226,64 @@ class FixedPoint
...
+  const Type smallest_fraction = 1;
```
pbukva: `const` -> `static constexpr`

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -219,48 +226,64 @@ class FixedPoint
...
+  const Type largest_fraction = fractional_mask;
```
pbukva: `const` -> `static constexpr`

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -219,48 +226,64 @@ class FixedPoint
...
+  const Type largest_int = Type(fractional_mask >> 1) << fractional_bits;
```
pbukva: `const` -> `static constexpr`

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -219,48 +226,64 @@ class FixedPoint
...
+  const Type smallest_int = integer_mask & ((Type(1) << (total_bits -1)));
```
pbukva: `const` -> `static constexpr`

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -219,48 +226,64 @@ class FixedPoint
...
+  const Type max = largest_int | largest_fraction;
```
pbukva: `const` -> `static constexpr`

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -219,48 +226,64 @@ class FixedPoint
...
+  const Type min = smallest_int - largest_fraction;
```
pbukva: `const` -> `static constexpr`

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -219,48 +226,64 @@ class FixedPoint
...
-  Type data_;  // the value to be stored
-
-public:
   ////////////////////
   /// constructors ///
   ////////////////////
 
-  FixedPoint()
-    : data_(0)
+  inline FixedPoint()
```
pbukva: Just comment: Perhaps I'm mistaken, but `inline` should not be necessary for any methods which are templated or members of templated class or both.
I suppose it does not hurt though.

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -43,17 +43,31 @@ struct TypeFromSize
...
   static const bool        is_valid = true;
-  static const std::size_t size     = 64;
-  using ValueType                   = int64_t;
-  using UnsignedType                = uint64_t;
-  using SignedType                  = int64_t;
+  static const std::size_t size     = 128;
+  using ValueType                   = __int128_t;
+  using UnsignedType                = __uint128_t;
+  using SignedType                  = __int128_t;
   using NextSize                    = TypeFromSize<128>;
```
pbukva: Shouldn't here be rather 256 than 128?

And on that note, if `size` data-member will be made `constexpr` as suggested in one of my previous comments, then the next size can be (and shall be) derived in general matter as follows in **every** specialization of this class:

```cpp
static constexpr std::size_t size     = 128;
using NextSize                    = TypeFromSize<(size << 1)>;
// OR:
//using NextSize                    = TypeFromSize<size * 2>;
```

However, since we do not provide implementation for the 256 bit version, we should either do 2 of following things:

1) **EITHER** completely omit declaration/typedefing of the `NextSize` type here inside of  `TypeFromSize<128>` in order to make compiler to fail to compile if someone actually wants to use `TypeFromSize<128>::NextSize` type,

2) **OR** keep `NextSize` typedef in `TypeFromSize<128>`, and rather declare **empty** type for 256 template class specialisation:
```
template<>
struct TypeFromSize<256>;
```

pbukva (changes requested):
See my inline comments

libs/core/include/core/fixed_point/fixed_point.hpp
bipll: It's simply methods defined inside class definition itself are implicitly `inline`.

libs/core/tests/fixed_point/fixed_point.cpp
```diff
@@ -134,9 +393,63 @@ TEST(FixedPointTest, Multiplication)
...
+  EXPECT_EQ(infinitesimal / one, infinitesimal);
+  EXPECT_EQ(one / huge, infinitesimal * 4);
+  EXPECT_EQ(huge / infinitesimal, zero);
+
+  try {
+    EXPECT_EQ(int(two / zero), 0);
+    EXPECT_EQ(float(two / zero), 0.0f);
+    EXPECT_EQ(double(two / zero), 0.0);
+    EXPECT_EQ(zero / zero, zero);
+  } catch (std::exception &e) {
```
bipll: `e` is not mutated inside the `catch` block, so there's no point in catching it by mutable reference.

libs/math/include/math/standard_functions/exp.hpp
```diff
@@ -44,7 +44,7 @@ meta::IfIsNonFixedPointArithmetic<Type, void> Exp(Type const &x, Type &ret)
 template <typename T>
 meta::IfIsFixedPoint<T, void> Exp(T const &n, T &ret)
 {
-  ret = T(std::exp(double(n)));
+  return T::Exp(n);
```
bipll: Function's ret type is `void` (the second argument to `IfIsFixedPoint`) and `ret` is the return argument.

bipll (dismissed):

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -310,204 +353,254 @@ class FixedPoint
...
 
   ////////////
   /// swap ///
   ////////////
 
-  void Swap(FixedPoint &rhs)
+  inline void Swap(FixedPoint &rhs)
   {
-    std::swap(data_, rhs.data_);
+    std::swap(data_, rhs.Data());
```
troels-im: This seems like a mistake unless `Data()` gives a reference to the underlying data. I.e. Data() returns a temporary so in effect this swap function is an equality.

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -310,204 +353,254 @@ class FixedPoint
...
   {
     return data_;
   }
 
+  static inline FixedPoint FromBase(Type n)
+  {
+    return FixedPoint(n, NoScale());
+  }
+
+  static inline FixedPoint Exp(FixedPoint &x)
```
troels-im: Const ref rather than ref?

libs/core/include/core/fixed_point/fixed_point.hpp
markos: I opt for 1, whenever we decide to support FixedPoint<128,128> we will have to deal with a 256-bit type. That's what I did temporarily in the code.

libs/core/tests/fixed_point/fixed_point.cpp
markos: replaced with EXPECT_THROW

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Good catch, will fix it asap.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Fixed locally.

libs/core/tests/fixed_point/fixed_point.cpp
markos: Added 843 ticket.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Just did.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: And committed.

libs/math/include/math/standard_functions/exp.hpp
markos: Fixed.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: @bipll is correct, inline is only needed outside of the class definition. Removed.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Converted.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Converted.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Converted.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Converted.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Opted for first choice.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Converted.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Converted.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Converted.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Converted.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: min still poses a problem, as it gives an 'integer overflow' problem.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Converted.

libs/core/include/core/fixed_point/fixed_point.hpp
pbukva: Not quite, the **only** thing what `inline` keyword actually really guarantees is that there are **NO duplicates** of the symbol during linking process (what would happen **if** the symbol **implementation / definition** is located in **include** file).
This (no symbol duplication = implicit inline) is automatically/implicitly guaranteed for all templated elements (class/struct, type alias, method or function) as long as they are **NOT** fully specialised (they can be **partially** specialised) - this rule is simply implication of the ultimate limitation of C++ templates (whole **implementation / definition** (not just declaration) of templated code **MUST** be delivered exclusively by **including**). 

Just to give a hint, **full** specialisation of templated code should be placed rather in .cpp than .hpp. Depending on which element we fully specialising (struct/class vs. function/method/data-member), **declaration** might need to be present in .hpp (for class/struct/aliased-type), or it does **not** need to be (but can be) declared in .hpp (for  function/method/data-member).

Final physical inlining of method/function is **solely & exclusively** upon compiler to decide whether to do it actually, or not.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Fixed.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Fixed in last commit.

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -313,200 +383,351 @@ class FixedPoint : public BaseFixedpointType
...
+    }
+    if (x.Data() == SMALLEST_FRACTION)
+    {
+      return FixedPoint{-FRACTIONAL_BITS};
+    }
+    if (x < CONST_ZERO)
+    {
+      throw std::runtime_error("Log2(): mathematical operation not defined: x < 0!");
+    }
+    // TODO(843): Add proper support for NaN/Infinity/etc in FixedPoint class
```
baykaner: private issues should be labelled: TODO(private 843)

baykaner (commented):
This is looking pretty good to me, but there's a lot here and it would be good to get a second pair of eyes on it. @troelsfr can you also take a look?

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -183,7 +193,7 @@ bool CheckNoOverflow(T n, std::uint16_t fractional_bits, std::uint16_t total_bit
  * @return
  */
 template <typename T>
-bool CheckNoRounding(T n, std::uint16_t fractional_bits)
+inline bool CheckNoRounding(T n, std::uint16_t fractional_bits)
```
pierre-wilmot-fetch-ai, line 197: > The value of std::numeric_limits<T>::max_digits10 is the number of base-10 digits that are necessary to uniquely represent all distinct values of the type T

fractional_bits is number of bits (base 2) we use to store the fractional part.

I think we're comparing apple to oranges here.

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -130,23 +147,16 @@ void Multiply(const FixedPoint<I, F> &lhs, const FixedPoint<I, F> &rhs, FixedPoi
...
 
-  std::uint16_t msb = 0;
-  while (n != 0)
-  {
-    n = n / 2;
-    msb++;
-  }
-
-  return msb;
+  return (int32_t)(sizeof(uint64_t) * 8) - __builtin_clzl(n);
```
troels-im: Please use the wrapper function for pop count in the vectorise library `platform.hpp`. If this particular version does not exist, then it needs to be made. Also, preferably, stick with C++ style casting `static_cast< int32_t >( ... )` as it is clearer.

libs/core/include/core/fixed_point/fixed_point.hpp
pierre-wilmot-fetch-ai: \+ Every call to that function seems to be commented out

libs/core/include/core/fixed_point/fixed_point.hpp
markos: should we move the platform.hpp inside core? Or is it ok to include it from here?

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -321,200 +383,351 @@ class FixedPoint : public BaseFixedpointType
...
     FixedPoint t(*this);
     t.data_ = -t.data_;
     return t;
   }
 
-  FixedPoint &operator++()
+  constexpr FixedPoint &operator++()
   {
-    data_ += CONST_ONE;
+    data_ += ONE_MASK;
```
pierre-wilmot-fetch-ai, line 409: operator++ uses ONE_MASK
operator-- uses CONST_ONE

That seems like a mistake, if it's not, a comment explaining the inconsitency would be welcome :)

pierre-wilmot-fetch-ai (dismissed):
Looks good, unit test coverage looks pretty intense too.
Since most of the ML library test various arithmetic opeartion using FixedPoint too and Jenkins didn't complain, I assume everything runs OK ;)

libs/core/include/core/fixed_point/fixed_point.hpp
markos: that is deliberate, CONST_ONE is actually the FixedPoint object for value of 1.0. ONE_MASK is the integer equivalent.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: this function could not be used due to constexpr problems and cout. I didn't remove it but it's currently commented out. But you are right that it's wrong to compare max_digits10 and base2 fractional bits. For this reason I will be adding a separate "tolerance" quantity -feel free to suggest another name- per type that will be used for the same purpose. But that will happen in a subsequent PR.

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -321,200 +383,351 @@ class FixedPoint : public BaseFixedpointType
...
+    return *this * FixedPoint(n);
   }
 
-  FixedPoint operator/(const FixedPoint &n) const
+  constexpr FixedPoint operator/(const FixedPoint &n) const
   {
-    // TODO(private, 501)
-    return FixedPoint(double(*this) / double(n));
+    if (n == CONST_ZERO)
+      throw std::overflow_error("Division by zero!");
```
troels-im, line 484: Please put `{` `}` around

markos: ack.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: I have added the static_cast, but would prefer to do the platform fix in a different PR.

pierre-wilmot-fetch-ai (dismissed):

baykaner (dismissed):

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -321,200 +384,353 @@ class FixedPoint : public BaseFixedpointType
...
+      {
+        return CONST_ZERO;
+      }
+    }
+    if (x < CONST_ZERO && y.fraction() != 0)
+    {
+      throw std::runtime_error(
+          "Pow(x, y): x^y where x < 0 and y non-integer: mathematical operation not defined!");
+    }
+    FixedPoint s   = CONST_ONE * ((y.integer() + 1) % 2) + Sign(x) * (y.integer() % 2);
```
pbukva, line 719: Don't fix this now, just something to think about:
The ```% 2``` operation can be done simply as ```& 1``` what should be faster. Perhaps it can be made even faster with little bit of additional wiggling.

markos: fixed locally.

baykaner:
I think @bipll is not available to approve the changes he requested but it looks to me like they have been addressed so for the sake of expediency I will dismiss his review.

pbukva (approved):
LGTM (I made one inline comment, nothing important, just to try something later).

PR #822, opened 2019-04-06
Bugfix updates for MacOS

ejfitzgerald:
Latest `develop` branch doesn't build and complete all tests in MacOS. This is due to the following:

1. A compilation warning, being treated as an error
2. Once compiling, the T-SNE test does not execute successfully under MacOS.

On the compilation side, the error is just a classic Mac weird `std::size_t` issue. On the test side I have not really investigated but the actual numerical result differs wildly from the expectation.

I would suggest that this probably warrants further investigation. In that light for the meantime, I have disabled this test.

Hopefully, by the end of next week we should have the CI-Mac attached to the Jenkins server so it will highlight these issues at PR time.

### DIFF
--- a/libs/ml/tests/ml/clustering/tsne_tests.cpp
+++ b/libs/ml/tests/ml/clustering/tsne_tests.cpp
@@ -34,7 +34,7 @@ using MyTypes = ::testing::Types<fetch::math::Tensor<float>, fetch::math::Tensor
 
 TYPED_TEST_CASE(TsneTests, MyTypes);
 
-TYPED_TEST(TsneTests, tsne_test_2d)
+TYPED_TEST(TsneTests, DISABLED_tsne_test_2d)
 {
   using DataType  = typename TypeParam::Type;
   using ArrayType = TypeParam;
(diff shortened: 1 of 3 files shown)


troels-im:
We need to get that Mac mini up running.

ejfitzgerald:
> We need to get that Mac mini up running.

Agreed

chr15murray:
On my list for this coming week on the new Jenkins install

ejfitzgerald:
Wow - we all need to get lives if we are answering GitHub comments on the weekend 😀

libs/ml/tests/ml/clustering/tsne_tests.cpp
```diff
@@ -34,7 +34,7 @@ using MyTypes = ::testing::Types<fetch::math::Tensor<float>, fetch::math::Tensor
 
 TYPED_TEST_CASE(TsneTests, MyTypes);
 
-TYPED_TEST(TsneTests, tsne_test_2d)
+TYPED_TEST(TsneTests, DISABLED_tsne_test_2d)
```
troels-im, line 37: Probably should open a ticket for this if you haven't already.

troels-im (approved):
Open a ticket for the disabled test - rest looks good.

libs/ml/tests/ml/clustering/tsne_tests.cpp
ejfitzgerald: Done!

Issue #257, opened 2018-10-01
ConstByteArray::Append(...args): Concatenating argumens done from the back of the buffer


pbukva:
This issue has been created to investigate suspicion that current implementation of concatenation (read bellow) might affect performance since it does not write arguments content to the memory buffer in forward sequential order from the start of the mem buffer to its end but other way around.

Arguments are written to the memory buffer from its back rather then start, iterating from the last argument (the most right one) to the first on.
This approach takes advantage of one-go approach natively coming out of  stack based recursive variadic function calls (when call stack is naturally unwinding back when functions return), what needs to involve seeking back from end of the memory buffer to the right position for actual argument to write it to the right position in the buffer.

PR #825, opened 2019-04-07
Feature/server disconnect

bipll:
The older PR got so complicated after three months of being put on hold, I've accidentally removed it so all its code, merged with current develop, is now here. Plus a few things more.

Disconnects are performed via sending MAINT messages. A router maintains two blacklists, those that to be blacklisted by it and those that have blacklisted it, to not create an unnecessary net load by bothering the latter. A holdback is a temporary blacklisting, valid until a specified timestamp (apparently, relative intervals are not applicable here as message sending times can bias them quite perceivably, and it's hardly worth the effort to waste effort on computing all the correct delays and pings when we can simply resort to UTC timestamps).

I've been long talking about passing scalars by value, finally this commit gets rid of every notion of `std::size_t const &`, after I've discovered a build is broken by `ConstByteArray::size()` declared as having such ret type but actually returning `data_.size()`, a reference to temporary, yeah.

There was Peter's comment in const_byte_array.hpp mentioning issue #257, about the fact that several arrays were appended in reverse order, right-to-left, probably raising the chances of a cache miss. Here, there's a proposed implementation that does it left-to-right. To do that I've added a `FoldL()` procedure to type_util.h, a kind of a pack left fold backported to C++14.


ejfitzgerald:
@bipll I am going to take over this review (as discussed). I will break up and submit as separate PRs

PR #828, opened 2019-04-08
Bug/muddle segf

n-hutton:
uniform_int_distribution is closed interval


ejfitzgerald (approved):

libs/network/src/muddle/router.cpp
```diff
@@ -708,13 +708,15 @@ Router::Handle Router::LookupRandomHandle(Packet::RawAddress const & /*address*/
     {
       // decide the random index to access
       std::uniform_int_distribution<decltype(routing_table_)::size_type> distro(
-          0, routing_table_.size());
+          0, routing_table_.size() - 1);
```
bipll, line 711: :+1:

PR #829, opened 2019-04-08
Tensor concat implementation

juergendkeba:


libs/math/include/math/tensor_operations.hpp
```diff
@@ -44,5 +44,163 @@ fetch::math::Tensor<T> ConcatenateTensors(std::vector<fetch::math::Tensor<T>> co
   return ret;
 }
 
+template <typename SizeType>
+std::vector<SizeType> _get_cumsum(std::vector<SizeType> inp)
```
baykaner: Function names should be PascalCase() usually
(rare exceptions are those that are very fast to access in a loop such as size() which just returns size_);

libs/math/include/math/tensor_operations.hpp
```diff
@@ -44,5 +44,163 @@ fetch::math::Tensor<T> ConcatenateTensors(std::vector<fetch::math::Tensor<T>> co
...
+  result[0] = inp[0];
+  for (SizeType i = 1; i < result.size(); i++)
+  {
+    result[i] = result[i - 1] + inp[i];
+  }
+  return result;
+}
+
+template <typename SizeType>
+SizeType _get_array_number(SizeType pos, std::vector<SizeType> array_sizes_cumsum)
```
baykaner: GetArrayNumber

libs/math/include/math/tensor_operations.hpp
```diff
@@ -44,5 +44,163 @@ fetch::math::Tensor<T> ConcatenateTensors(std::vector<fetch::math::Tensor<T>> co
...
+    if (pos < array_sizes_cumsum[i])
+    {
+      return i;
+    };
+  }
+  return SizeType{0};
+}
+
+template <typename T>
+void _assert_concat_tensor_shapes(std::vector<T> tensors, typename T::SizeType concat_axis)
```
baykaner: AssertConcatTensorShapes
presumably you call this assert.... because it asserts if the tensors cannot be concatenated?
Seems better to give it a sensible name ConcatImplemenation, and have it return a bool indicating success.

libs/math/include/math/tensor_operations.hpp
```diff
@@ -44,5 +44,163 @@ fetch::math::Tensor<T> ConcatenateTensors(std::vector<fetch::math::Tensor<T>> co
...
+        final_shape[i] += tensors[j].shape()[i];
+      }
+    }
+  }
+
+  return final_shape;
+}
+
+template <typename T>
+std::vector<typename T::SizeType> _get_dims_along_ax_cumsummed(std::vector<T>       tensors,
```
baykaner: all of these underscore leading methods are presumably intended to be 'private' in some sense. The way to do that would be to wrap them up in a namespace called details_concat{}, so as not to pollute the fetch::math namespace.

However, I think it might be preferable just to move these implementations all to the private part of Tensor, and have the following two public methods:
bool Concat(std::vector<SelfType const&> others, SizeType axis)
{
others.emplace_back(*this);
this->Copy(ConcatImplementation(others, axis));
}
and also:

static SelfType Concat(std::vector<SelfType const&> tensors, SizeType axis)
{
    return ConcatImplementation(tensors, axis);
}

libs/math/include/math/tensor_operations.hpp
```diff
@@ -44,5 +44,163 @@ fetch::math::Tensor<T> ConcatenateTensors(std::vector<fetch::math::Tensor<T>> co
...
+    }
+  }
+  else
+  {
+    _concatenate_assign_values(res_tensor, tensors, concat_axis, counter);
+  }
+}
+
+template <typename T>
+T concatenate(std::vector<T> tensors, typename T::SizeType concat_axis)
```
baykaner: PascalCase please

libs/math/tests/math/tensor/tensor_operations.cpp
```diff
@@ -16,10 +16,57 @@
 //
 //------------------------------------------------------------------------------
 
+#include "math/tensor_operations.hpp"
```
baykaner: this can be removed after moving implementation into tensor

baykaner: or possibly not - depends on whether my PR tensor refactor goes in before or after this PR

libs/math/tests/math/tensor/tensor_operations.cpp
```diff
@@ -16,10 +16,57 @@
...
+  static const bool value = false;
+};
+template <typename T>
+struct IsVector<std::vector<T>>
+{
+  static const bool value = true;
+};
+
+template <typename T, typename R = void>
+using IfIsVector = fetch::meta::EnableIf<IsVector<T>::value, R>;
```
baykaner, line 35: SFINAE just for a test? is that really necessary?

baykaner (changes requested):

libs/math/include/math/tensor_operations.hpp
juergendkeba: This overloading wouldnt work :-( 

> Member function declarations with the same name and the same parameter types cannot be overloaded if any of them is a static member function declaration (9.4).

In the next PR, I'll include the static version

libs/math/tests/math/tensor/tensor_operations.cpp
juergendkeba: Maybe we should not use it for this test exclusively, cause initialising a tensor like this, saved me a lot of typing!

And... I don't think we can avoid SFINAE here in favour of partial template specialisation ((cause then I'd need so specialise for all the possible scalar types in order to make the recursion work :-D ))

libs/math/include/math/tensor_operations.hpp
juergendkeba: Done

libs/math/include/math/tensor_operations.hpp
juergendkeba: Done

libs/math/include/math/tensor_operations.hpp
juergendkeba: Done

libs/math/include/math/tensor_operations.hpp
juergendkeba: Done

libs/math/tests/math/tensor/tensor_operations.cpp
juergendkeba: Done

PR #833, opened 2019-04-09
HTTPS Client

ejfitzgerald:
This PR adds a new object `HttpsClient` to the HTTP library. This is the SSL version of the HttpClient. These classes are primarily used for communication with the bootstrap/nebula service.

As part of this change I also derived a new common interface between these two clients and updated the `JsonClient` to work with both.

As a bonus feature I also updated the CI script to preferentially choose to use the Ninja build system in preference of regular make

### DIFF
--- a/libs/http/examples/client/client.cpp
+++ b/libs/http/examples/client/client.cpp
@@ -16,10 +16,20 @@
 //
 //------------------------------------------------------------------------------
 
-#include "http/client.hpp"
+#include "http/http_client.hpp"
+#include "http/https_client.hpp"
+#include "http/request.hpp"
+#include "http/response.hpp"
 #include "core/commandline/params.hpp"
 
+#include <memory>
 #include <iostream>
+
+using fetch::http::HttpClientInterface;
+using fetch::http::HttpClient;
+using fetch::http::HttpsClient;
+
+using ClientPtr = std::unique_ptr<HttpClientInterface>;
 
 int main(int argc, char **argv)
 {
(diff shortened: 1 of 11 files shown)


libs/http/examples/client/client.cpp
```diff
@@ -31,29 +41,47 @@ int main(int argc, char **argv)
...
+  }
+
+  std::cout << "Host     : " << host << std::endl;
+  std::cout << "Port     : " << port << std::endl;
+  std::cout << "Endpoint : GET " << endpoint << std::endl;
+  std::cout << "SSL      : " << ssl << std::endl;
+
   // create the client
-  fetch::http::HTTPClient client(host, port);
+  ClientPtr client =
```
n-hutton, line 65: 👍

libs/http/src/http_client.cpp
```diff
@@ -130,5 +153,42 @@ bool HTTPClient::Connect()
...
 
+/**
+ * Write the contents of the buffer to the socket
+ *
+ * @param buffer The input buffer to send
+ * @param ec The output error code from the operation
+ */
+void HttpClient::Write(asio::streambuf const &buffer, std::error_code &ec)
+{
+  socket_.write_some(buffer.data(), ec);
```
n-hutton, line 164: So you're going with all blocking writes for the http client?

n-hutton (approved):
Looks safe to me due to avoidance of async behaviour (although it does add another component which is doing its own networking thing which I think isn't great design practice).

libs/http/src/http_client.cpp
ejfitzgerald: That is correct, it is just a very simple client that is run in its own background thread.

PR #842, opened 2019-04-11
Feature/tensor refactor

baykaner:
main purpose of this refactor is to rework tensor implementations such that they can be compatible with SIMD implementations which are due to come soon; as well as this there are various reimplementations purely concerned with optimising performance.

Primarily this means TensorSlice, TensorIterator, and TensorBroadcast are implemented to permit very quick iteration over multiple tensors for math operations.

A few areas are notably non-performant currently, but due to be fixed in the next PR: specifically, broadcasting currently invokes Copies of Tensors. Likewise the ML library can be optimised if instead of dealing with Tensors it deals with TensorSlices; this would prevent the need to make copies in the training loop.


libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -153,7 +153,7 @@ std::size_t HighestSetBit(T n_input)
  * helper function that checks no bit overflow when shifting
  * @tparam T the input original type
  * @param n the value of the datum
- * @param fractional_bits the number of fractional_bits to be set
+ * @param FRANCTIONAL_BITS the number of FRANCTIONAL_BITS to be set
```
n-hutton: spelling? fractional

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -153,7 +153,7 @@ std::size_t HighestSetBit(T n_input)
  * helper function that checks no bit overflow when shifting
  * @tparam T the input original type
  * @param n the value of the datum
```
n-hutton, line 155: does fixed_point really belong in core? Also, there is another fixed_point.hpp in serializers, might make more sense to have the seralizer with the definition which is the usual way we are doing things.

libs/core/include/core/fixed_point/fixed_point.hpp
n-hutton: this whole namespace details might be able to be an anonymous namespace ?

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -201,55 +201,57 @@ bool CheckNoRounding(T n, std::size_t fractional_bits)
 
 }  // namespace details
 
+struct BaseFixedpointType
```
n-hutton, line 204: The reason for this is to be able to do ```std::is_base_of<fixed_point::base``` ? But there is only one use case - are there to be others?

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -201,55 +201,57 @@ bool CheckNoRounding(T n, std::size_t fractional_bits)
...
+{
+};
+
 template <std::size_t I, std::size_t F>
```
n-hutton: Didn't we have some edict about not using size_t when possible?

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -201,55 +201,57 @@ bool CheckNoRounding(T n, std::size_t fractional_bits)
...
+  enum
+  {
+    FRACTIONAL_MASK = Type(((1ull << FRANCTIONAL_BITS) - 1)),
+    INTEGER_MASK    = Type(~FRACTIONAL_MASK),
+    CONST_ONE       = Type(1) << FRANCTIONAL_BITS
+  };
 
 private:
-  Type data_;  // the value to be stored
+  Type data_{0};  // the value to be stored
```
n-hutton: any reason for alternating public and private?

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -201,55 +201,57 @@ bool CheckNoRounding(T n, std::size_t fractional_bits)
...
-
-  FixedPoint()
-    : data_(0)
-  {}  // initialise to zero
+  FixedPoint() = default;
 
   template <typename T>
   explicit FixedPoint(T n, meta::IfIsInteger<T> * = nullptr)
-    : data_(static_cast<Type>(n) << static_cast<Type>(fractional_bits))
+    : data_(static_cast<Type>(n) << static_cast<Type>(FRANCTIONAL_BITS))
```
n-hutton: this is a bit heavy, the class needs some commenting on what's happening

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -352,37 +354,37 @@ class FixedPoint
...
   explicit operator unsigned long long() const
   {
-    return (static_cast<unsigned long long>(data_) / one);
+    return (static_cast<unsigned long long>(data_) / CONST_ONE);
   }
 
   //  // casting operators
   //  operator uint32_t () const {
-  //    return (data_ & integer_mask) >> fractional_bits;
+  //    return (data_ & INTEGER_MASK) >> FRANCTIONAL_BITS;
```
n-hutton: delete commented out code?

libs/math/include/math/clustering/k_means.hpp
```diff
@@ -437,8 +437,7 @@ class KMeansImplementation
     // assign first cluster centre
     for (SizeType j = 0; j < n_dimensions_; ++j)
     {
-      k_means_.Set(std::vector<SizeType>({0, j}),
-                   data.At(std::vector<SizeType>({data_idxs_[0], j})));
+      k_means_.Set(std::vector<SizeType>({0, j}), data.At(data_idxs_[0], j));
```
n-hutton, line 440: maybe we should discuss using vectors as indexes - I think this should be avoided where possible

libs/math/include/math/fundamental_operators.hpp
```diff
@@ -17,10 +17,199 @@
...
 
+#include "core/assert.hpp"
 #include "math/meta/math_type_traits.hpp"
 
 namespace fetch {
 namespace math {
+
+namespace details_vectorisation {
+
+// TODO: vectorisation implementations not yet called
```
n-hutton: make this an issue

libs/math/include/math/fundamental_operators.hpp
```diff
@@ -17,10 +17,199 @@
 //
 //------------------------------------------------------------------------------
 
+#include "core/assert.hpp"
 #include "math/meta/math_type_traits.hpp"
 
 namespace fetch {
 namespace math {
+
+namespace details_vectorisation {
```
n-hutton, line 26: can this be anonymous?

libs/math/include/math/fundamental_operators.hpp
```diff
@@ -17,10 +17,199 @@
...
+
+// TODO: vectorisation implementations not yet called
+
+/**
+ * subtract a scalar from every value in the array
+ * @tparam T
```
n-hutton: doxygen comments incomplete

n-hutton: needs pretty explicit comments about what is happening imo. Everytime you use typename, I would say it becomes harder to follow

libs/math/include/math/fundamental_operators.hpp
```diff
@@ -329,18 +426,39 @@ template <typename ArrayType, typename T,
...
+
+  while (it1 != eit1)
+  {
+    *rit = (*it1) - (*it2);
+    ++it1;
+    ++it2;
+    ++rit;
+  }
+
+  /*
```
n-hutton: fair amount of commented out code in this file - delete.

libs/math/include/math/fundamental_operators.hpp
```diff
@@ -17,10 +17,199 @@
...
+/**
+ * Multiply array by another array within a range
+ * @tparam T
+ * @tparam C
+ * @param array1
+ * @param scalar
+ * @param ret
+ */
+template <typename ArrayType>
+meta::IfIsMathArray<ArrayType, void> Multiply(ArrayType const &obj1, ArrayType const &obj2,
```
n-hutton, line 115: This evaluates to void Multiply right?

libs/math/include/math/fundamental_operators.hpp
```diff
@@ -17,10 +17,199 @@
...
+
+  if (range.is_undefined())
+  {
+    Multiply(obj1, obj2, ret);
+  }
+  else
+  {
+    auto r = range.ToTrivialRange(ret.data().size());
+
+    ret.data().in_parallel().Apply(r,
```
n-hutton, line 133: I wonder if it might be cleaner to refer to a free fn than use closures like this. I don't know if Apply is set up to do that.

libs/math/include/math/fundamental_operators.hpp
```diff
@@ -17,10 +17,199 @@
...
+
+  assert(range.is_undefined() || range.is_trivial());
+
+  if (range.is_undefined())
+  {
+    Subtract(obj1, obj2, ret);
+  }
+  else
+  {
+    auto r = range.ToTrivialRange(ret.data().size());
```
n-hutton, line 165: What is trivialrange?

libs/math/include/math/fundamental_operators.hpp
```diff
@@ -65,10 +254,17 @@ template <typename ArrayType>
 ::fetch::math::meta::IfIsMathArray<ArrayType, void> Multiply(ArrayType const &obj1,
                                                              ArrayType const &obj2, ArrayType &ret)
 {
-  assert(obj1.size() == obj2.size());
-  for (std::size_t i = 0; i < ret.size(); ++i)
+  ASSERT(obj1.size() == obj2.size());
```
n-hutton: Why refer to global namespace ``` ::fetch ``` ?

libs/math/include/math/matrix_operations.hpp
```diff
@@ -20,41 +20,105 @@
...
+#include "math/base_types.hpp"
 #include "math/comparison.hpp"
 #include "math/fundamental_operators.hpp"  // add, subtract etc.
 #include "math/meta/math_type_traits.hpp"
 
 namespace fetch {
 namespace math {
+//
+// template <typename T, typename C>
+// class Tensor;
```
n-hutton: delete

libs/math/include/math/matrix_operations.hpp
```diff
@@ -20,41 +20,105 @@
...
+//
+// template <typename T, typename C, typename TensorType>
+// class TensorIterator;
 
-template <typename T>
-class Tensor;
+// TODO - vectorisation implementations not yet called
```
n-hutton: private issue

libs/math/include/math/matrix_operations.hpp
```diff
@@ -20,41 +20,105 @@
...
+/**
+ * Min function for returning the smallest value in an array
+ * @tparam ArrayType
+ * @param array
+ * @return
+ */
+template <typename ArrayType>
+inline void Min(ArrayType const &array, typename ArrayType::Type &ret)
+{
+  using vector_register_type = typename ArrayType::vector_register_type;
```
n-hutton: Style guide is VectorRegisterType

libs/math/include/math/matrix_operations.hpp
```diff
@@ -20,41 +20,105 @@
...
+ * @param obj1
+ * @param ret
+ */
+template <typename ArrayType>
+void Product(ArrayType const &obj1, typename ArrayType::Type &ret)
+{
+  ret =
+      obj1.data().in_parallel().Reduce(memory::TrivialRange(0, obj1.size()),
+                                       [](typename ArrayType::vector_register_type const &a,
+                                          typename ArrayType::vector_register_type const &b) ->
```
n-hutton: if you are intending to use arrow notation shouln't you apply it everywhere

libs/math/include/math/matrix_operations.hpp
```diff
@@ -232,6 +302,56 @@ meta::IfIsMathArray<ArrayType, typename ArrayType::Type> Min(ArrayType const &ar
   return ret;
 }
 
+/**
+ * Implementation of Max that returns the n-1 dim array by finding the max of all 1-d vectors within
```
n-hutton: this is min. comments also incomplete.

libs/math/include/math/matrix_operations.hpp
```diff
@@ -287,6 +412,25 @@ meta::IfIsMathArray<ArrayType, T> Product(ArrayType const &array1)
...
+/**
+ * return the product of all elements in the vector
+ * @tparam T
+ * @param obj1
+ * @param ret
+ */
+template <typename T>
+void Product(std::vector<T> const &obj1, T &ret)
+{
+  ret = std::accumulate(std::begin(obj1), std::end(obj1), std::size_t(1), std::multiplies<>());
```
n-hutton, line 414: Max dereferenced std::max_element, but this doesn't deref accumulate?

libs/math/include/math/matrix_operations.hpp
```diff
@@ -399,48 +546,66 @@ void PeakToPeak(ArrayType arr)
  * @return
```
n-hutton, line 536: no need for return when void

libs/math/include/math/matrix_operations.hpp
```diff
@@ -450,21 +615,28 @@ ArrayType ArgMax(ArrayType const &array, typename ArrayType::SizeType axis = 0)
...
+    ret_shape.erase(ret_shape.begin() + int(axis));
+    ret = ArrayType{ret_shape};
   }
 
   ArgMax(array, ret, axis);
   return ret;
 }
 
+template <typename T>
+void ArgMax(std::vector<T> const &obj1, T &ret)
```
n-hutton, line 618: Is there no way to have a convention that dictates one of these styles?

libs/math/include/math/matrix_operations.hpp
```diff
@@ -473,18 +645,14 @@ fetch::math::meta::IfIsMathArray<ArrayType, void> Dot(ArrayType const &A, ArrayT
...
-          A.At(std::vector<typename ArrayType::SizeType>({i, 0})) *
-          B.At(std::vector<typename ArrayType::SizeType>({0, j}));
-      for (typename ArrayType::SizeType k(1); k < A.shape()[1]; ++k)
+      ret.At(i, j) = A.At(i, 0) * B.At(0, j);
+      for (SizeType k(1); k < A.shape()[1]; ++k)
       {
-        ret.At(std::vector<typename ArrayType::SizeType>({i, j})) +=
-            A.At(std::vector<typename ArrayType::SizeType>({i, k})) *
-            B.At(std::vector<typename ArrayType::SizeType>({k, j}));
+        ret.At(i, j) += A.At(i, k) * B.At(k, j);
```
n-hutton, line 645: This is troels' change I expect. Presumably it's using some sort of variadics for indexing now

n-hutton (changes requested):

libs/math/include/math/matrix_operations.hpp
baykaner: correct. it might be worth prioritising your review to begin with changes to tensor.hpp to see how this is all meant to work. But yes, At(... indices) is variadics, but Set() isn't yet

libs/core/include/core/fixed_point/fixed_point.hpp
troels-im: No, this is the only use case. Feel free to propose another construct, but please be mindful that any solution is not allowed to alter the size of `FixedPoint<N, M >`.

libs/core/include/core/fixed_point/fixed_point.hpp
troels-im: Yes, this should be `uint16_t`

libs/core/include/core/fixed_point/fixed_point.hpp
troels-im: I believe the reason is to expose `Type` alongside the enums publicly, but check with @baykaner

libs/math/include/math/clustering/k_means.hpp
troels-im: This is due to be refactored as well - entirely my fault that it has not been. The key reason is that we need to do variadic template arguments with a the last argument being the value.

libs/math/include/math/matrix_operations.hpp
troels-im: In principle yes. We could opt for functions which are always "returning" `void`. This would have the benefit that it forces people to write efficient code, but the drawback that the code becomes slightly more inreadible.

libs/math/include/math/ml/activation_functions/sigmoid.hpp
```diff
@@ -34,19 +34,20 @@ void Sigmoid(ArrayType const &t, ArrayType &ret)
 {
 
   typename ArrayType::SizeType idx(0);
-  for (auto &val : t)
+  for (auto const &val : t)
   {
-    if (val >= typename ArrayType::Type(0))
+    if (val >= typename ArrayType::Type(0))  // TODO: Reimplement using iterators - this is ugly
```
n-hutton: private issue

libs/math/include/math/ml/activation_functions/softmax.hpp
```diff
@@ -42,23 +42,33 @@ namespace details {
...
+  Type sum = Type(0);
+  while (it1.is_valid())
+  {
+    *it2 = Exp(*it1 - array_max);
+    sum += *it2;
+    ++it2;
+    ++it1;
+  }
+
+  auto it3 = ret.begin();  // TODO: Fix implictly deleted copy const. for iterator
```
n-hutton: private issue

libs/math/include/math/ml/activation_functions/softmax.hpp
```diff
@@ -73,10 +83,10 @@ void Softmax2DImplementation(ArrayType const &array, ArrayType &ret,
 
   for (std::size_t i = 0; i < array.shape()[axis]; ++i)
   {
-    ArrayType cur_slice = array.Slice(i);
-    ArrayType ret_slice = ret.Slice(i);
+    auto cur_slice = array.Slice(i).Copy();
```
n-hutton, line 86: this way is more efficient?

libs/math/include/math/ml/loss_functions/cross_entropy.hpp
```diff
@@ -58,17 +58,19 @@ typename ArrayType::Type CrossEntropyLoss(
   // if not a one-hot, must be binary logistic regression cost
   if (n_dims == 1)
   {
-    assert(n_classes == SizeType(2));
+    ASSERT(n_classes == SizeType(2));
+
+    // TODO: Use iterators instead to be correct
```
n-hutton: private issue

libs/math/include/math/serializers/math_types.hpp
```diff
@@ -22,46 +22,14 @@
 namespace fetch {
 namespace serializers {
 
+/*
```
n-hutton: delete file?

libs/math/include/math/standard_functions/abs.hpp
```diff
@@ -31,16 +31,17 @@ namespace math {
 /// IMPLEMENTATIONS ///
 ///////////////////////
 
-template <typename Type>
-meta::IfIsNonFixedPointArithmetic<Type, void> Abs(Type const &x, Type &ret)
-{
-  ret = std::abs(x);
-}
+// template <typename Type>
```
n-hutton: delete

libs/math/include/math/standard_functions/pow.hpp
```diff
@@ -57,54 +59,63 @@ meta::IfIsMathArray<ArrayType, void> Pow(ArrayType const &               array1,
...
 template <typename Type>
 meta::IfIsArithmetic<Type, void> Square(Type const &x, Type &ret)
 {
-  Pow(x, Type(2), ret);
+  ret = x * x;
 }
 
 template <typename ArrayType>
-meta::IfIsMathArray<ArrayType, ArrayType> Square(ArrayType const &x)
+meta::IfIsMathArray<ArrayType, void> Square(ArrayType const &x, ArrayType &ret)
```
n-hutton, line 99: when I look at the templating for this, it's confusing and not consistent:

```
template <typename A, typename R>
struct IsMathArrayImpl
{
};
template <typename R, typename T, typename C>
struct IsMathArrayImpl<Tensor<T, C>, R>
{
  using Type = R;
};
template <typename T, typename R>
using IfIsMathArray = typename IsMathArrayImpl<T, R>::Type;

template <typename T, typename R>
using IfIsMathFixedPointArray = IfIsFixedPoint<typename T::Type, IfIsMathArray<T, R>>;

template <typename T, typename R>
using IfIsMathNonFixedPointArray = IfIsNotFixedPoint<typename T::Type, IfIsMathArray<T, R>>;

```

Template paramers don't appear to follow an order and their meaning isn't clear from just a single character

libs/math/include/math/tensor.hpp
```diff
@@ -17,360 +17,648 @@
 //
 //------------------------------------------------------------------------------
 
+//#include "math/free_functions/free_functions.hpp"
+//#include "math/shapeless_array.hpp"
+
+// https://github.com/uvue-git/fetch-ledger/tree/cf2fc8441f6ae33d6248559c52473a7f15c5aef2/libs/math/include/math
```
n-hutton: should this be here? delete commented out includes.

libs/math/include/math/tensor.hpp
```diff
@@ -17,360 +17,648 @@
...
-#include <vector>
+#include <utility>
 
 namespace fetch {
 namespace math {
 
-template <typename T>
+namespace details {
+template <typename DataType, typename ArrayType>
+static void ArangeImplementation(DataType const &from, DataType const &to, DataType const &delta,
```
n-hutton, line 48: spelling: arrange (?)

libs/math/include/math/tensor.hpp
```diff
@@ -17,360 +17,648 @@
...
-    , storage_(std::move(storage))
-    , offset_(offset)
-  {
-    ASSERT(padding.empty() || padding.size() == shape.size());
-    ASSERT(strides.empty() || strides.size() == shape.size());
-    Init(input_strides_, padding_);
-  }
+  using vector_slice_type = typename ContainerType::vector_slice_type;
+  using vector_register_type =
+      typename ContainerType::vector_register_type;  // TODO: Legacy style, replace with new
```
n-hutton: private issue

libs/math/include/math/tensor.hpp
```diff
@@ -17,360 +17,648 @@
...
-    , padding_(std::move(padding))
-    , input_strides_(std::move(strides))
-    , storage_(std::move(storage))
-    , offset_(offset)
-  {
-    ASSERT(padding.empty() || padding.size() == shape.size());
-    ASSERT(strides.empty() || strides.size() == shape.size());
-    Init(input_strides_, padding_);
-  }
+  using vector_slice_type = typename ContainerType::vector_slice_type;
```
n-hutton: mismatched casing: should be ``` using VectorSliceType ```

libs/math/include/math/tensor.hpp
```diff
@@ -381,217 +669,1567 @@ class Tensor
...
+//  {
+//    vector_register_type val(scalar);
+//
+//    this->data().in_parallel().Apply(
+//        [val](vector_register_type const &y, vector_register_type &z) { z = val - y; },
+//        this->data());
+//
+//    return *this;
+//  }
+//
```
n-hutton: delete all this commented out code

libs/math/include/math/tensor.hpp
```diff
@@ -17,360 +17,648 @@
...
+  }
+
+  void Fill(Type const &value)
+  {
+    vector_register_type val(value);
+
+    this->data().in_parallel().Apply([val](vector_register_type &z) { z = val; });
+  }
+
+  // TODO - general Set implementation
```
n-hutton: delete/todo private issues

libs/math/include/math/tensor.hpp
```diff
@@ -381,217 +669,1567 @@ class Tensor
...
+      LazyResize( x.size() );
+
+      kernels::ApproxExp< vector_register_type > aexp;
+      data_.in_parallel().Apply(aexp, x.data_);
+    }
+  */
+
+  void ApproxSoftMax(SelfType const & /*x*/)
+  {
+    //    kernels::ApproxSoftMax< Type, vector_register_type > kernel;
```
n-hutton: functions with no body

libs/math/include/math/tensor.hpp
```diff
@@ -381,217 +669,1567 @@ class Tensor
...
+    //    kernel( this->data_, x.data());
+  }
+
+  /**
+   * calculates the l2loss of data in the array
+   *
+   * @return       returns single value as Type
+   *
+   **/
+  Type L2Loss() const
```
n-hutton, line 1362: tensor class is huge. try to split the class into signatures only, then definitins, and make free functions where possible?

libs/math/include/math/tensor.hpp
```diff
@@ -381,217 +669,1567 @@ class Tensor
...
+
+  template <typename S>
+  typename std::enable_if<std::is_integral<S>::value, void>::type Resize(S const &n)
+  {
+    SizeType oldsize = size_;
+    LazyResize(n);
+    data_.SetZeroAfter(oldsize);
+  }
+
+  //  iterator begin()
```
n-hutton: clean up lots of commented out code

libs/math/include/math/tensor_iterator.hpp
```diff
@@ -17,76 +17,359 @@
...
+   * @return boolean indicating validity
+   */
+  operator bool()
+  {
+    return counter_ < size_;
+  }
+
+  bool is_valid() const
+  {
+    return counter_ < size_;
```
n-hutton: duplicates bool

libs/math/include/math/tensor_iterator.hpp
```diff
@@ -17,76 +17,359 @@
...
+
+      next                   = false;
+      TensorIteratorRange &s = ranges_[i];
+      s.index += s.step;
+      position_ += s.step_volume;
+      ++s.current_n_dim_position;
+
+      if (s.index >= s.to)
+      {
+
```
n-hutton: odd space

libs/math/include/math/tensor_iterator.hpp
```diff
@@ -17,76 +17,359 @@
...
+          r.index = r.from;
+          position_ += r.volume * r.index;
+        }
+      }
     }
-    pointer_ = original_pointer_;
-    for (std::size_t i(0); i < coordinate_.size(); ++i)
+
+#ifndef NDEBUG
+    // Test
```
n-hutton: remove comment

libs/math/tests/math/kernels/sign/sign.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include "math/tensor.hpp"
+#include <gtest/gtest.h>
+
+using namespace fetch::math;
+using data_type      = double;
+using container_type = fetch::memory::SharedArray<data_type>;
+
+Tensor<data_type, container_type> RandomArray(std::size_t n, data_type adj)
+{
+  static fetch::random::LinearCongruentialGenerator gen;
```
n-hutton, line 32: I wonder if this is thread safe

n-hutton: oh, just a test, doesn't matter.

libs/math/tests/math/kernels/sign/sign.cpp
```diff
@@ -0,0 +1,122 @@
...
+  Tensor<data_type, container_type> test_array   = ConstantArray(n, 0);
+  Tensor<data_type, container_type> test_array_2 = RandomArray(n, -1.0);
+
+  // sanity check that all values equal 0
+  for (std::size_t i = 0; i < n; ++i)
+  {
+    ASSERT_TRUE(test_array[i] == 0);
+  }
+
+  // check that sign(0) = 0
```
n-hutton: delete

libs/math/tests/math/matrix_operations/matrix_operations.cpp
```diff
@@ -73,10 +75,85 @@ TYPED_TEST(FreeFunctionsTest, ArgMax_TwoDimension)
...
+  }
+
+  //  TypeParam output{{n_data, 1}};
+  fetch::math::Tensor<int> output{n_data};
+  //  fetch::math::Tensor2<int> output{n_data};
+
+  auto start = std::chrono::system_clock::now();
+  fetch::math::ArgMax(array, output, SizeType(1));
+  auto end = std::chrono::system_clock::now();
+  //
```
n-hutton: del

libs/math/tests/math/numpy/basic.cpp
```diff
@@ -17,17 +16,15 @@
...
+#include "math/tensor.hpp"
+#include <gtest/gtest.h>
+
+using namespace fetch::math;
 
-}  // namespace kernels
-}  // namespace fetch
+TEST(numpy, arange)
+{
+  auto tensor = Tensor<double>::FromString(R"(1 3 4)");
```
n-hutton: is right?

libs/math/tests/math/serializers/math_types.cpp
```diff
@@ -50,9 +50,12 @@ TYPED_TEST(SerializersTest, serialize_tensor)
     i++;
   }
   fetch::serializers::ByteArrayBuffer b;
+
```
n-hutton, line 52: whitespace only

libs/math/tests/math/tensor/basic_tests.cpp
```diff
@@ -0,0 +1,461 @@
...
+  T rn{0};
+  for (std::size_t i = 0; i < n; ++i)
+  {
+    rn       = T(gen.AsDouble() * scale);
+    a1.At(i) = rn + adj;
+  }
+  return a1;
+}
+
+/// template for producing a random array of float types
```
n-hutton: three /

libs/math/tests/math/tensor/basic_tests.cpp
```diff
@@ -0,0 +1,461 @@
...
+  std::size_t n            = 10000;
+  Tensor<T>   test_array   = RandomArray(n, T(0));
+  Tensor<T>   result_array = test_array;
+
+  ASSERT_TRUE(result_array.AllClose(test_array));
+}
+
+TEST(shapeless_array_tests, int_equals)
+{
+  equal_test<int>();
```
n-hutton: could use templated fixtures?

libs/math/tests/math/tensor/tensor_indexing.cpp
```diff
@@ -284,54 +127,38 @@ TYPED_TEST(TensorIndexingTest, three_dimentional_tensor_test)
...
- * You can now store an arbitrary large tensor in an underlying storage of size 1
- * Will be useful for broadcasting
- */
-TYPED_TEST(TensorIndexingTest, zero_stride_tensor_test)
-{
-  fetch::math::Tensor<TypeParam> t({2, 3, 5}, {0, 0, 0}, {0, 0, 0});
-
-  ASSERT_EQ(t.size(), 30);
-  ASSERT_EQ(t.Capacity(), 1);
+//  std::vector<TypeParam> gt({0,  1,  2,  3,  4,  0, 0, 0, 5,  6,  7,  8,  9,  0, 0, 0,
```
n-hutton: del

libs/math/tests/math/tensor/tensor_operations.cpp
```diff
@@ -31,25 +31,38 @@ TYPED_TEST_CASE(TensorOperationsTest, MyTypes);
...
+    for (std::uint64_t j(0); j < 4; ++j)
+    {
+      EXPECT_EQ(t1.At(i, j), TypeParam(gtInput[counter]));
+      EXPECT_EQ(t2.At(i, j), TypeParam(t2Input[counter]));
+      ++counter;
+    }
   }
 }
 
+/*
```
n-hutton: del

libs/ml/include/ml/ops/activations/relu.hpp
```diff
@@ -46,18 +47,18 @@ class Relu : public fetch::ml::ElementWiseOps<T>
...
     assert(inputs[0].get().shape() == errorSignal.shape());
 
-    ArrayType returnSignal = errorSignal.Clone();
+    ArrayType returnSignal = errorSignal.Copy();
     for (std::size_t i = 0; i < inputs.front().get().size(); ++i)
     {
       if (inputs.front().get()[i] <= DataType(0))
       {
-        returnSignal.Set(i, DataType(0));
+        returnSignal.data().Set(i, DataType(0));  // TODO:
```
n-hutton: priv issue

libs/ml/include/ml/ops/loss_functions/cross_entropy.hpp
```diff
@@ -50,7 +50,7 @@ class CrossEntropy
     SizeType n_classes{2};
     if (inputs.size() == 3)
     {
-      n_classes = SizeType(inputs.at(2).At(0));
+      n_classes = SizeType(inputs.at(2).data()[0]);  // TODO: This is wrong
```
n-hutton: ???

libs/ml/tests/ml/clustering/tsne_tests.cpp
```diff
@@ -84,16 +87,86 @@ TYPED_TEST(TsneTests, tsne_test_2d)
...
+{
+  SizeType N_DATA_SIZE{100};
+  SizeType N_OUTPUT_FEATURE_SIZE{2};
+
+  Tensor<float> output_matrix = RunTest<Tensor<float>>(N_DATA_SIZE, N_OUTPUT_FEATURE_SIZE);
+
+  ASSERT_EQ(output_matrix.shape().at(0), N_DATA_SIZE);
+  ASSERT_EQ(output_matrix.shape().at(1), N_OUTPUT_FEATURE_SIZE);
+
+  EXPECT_FLOAT_EQ(output_matrix.At(0, 0), 0.25323879718780517578);
```
n-hutton, line 103: magic numbers?

libs/python/include/python/math/py_tensor.hpp
```diff
@@ -29,19 +29,21 @@ template <typename T>
...
-      .def("Fill", &ArrayType::Fill)
-      .def("Slice", [](ArrayType &a, SizeType i) { return a.Slice(i); })
-      .def("Slice", [](ArrayType const &a, SizeType i) { return a.Slice(i); })
-      .def("At", [](ArrayType &a, SizeType i) { return a.At(i); })
-      .def("Set", (void (ArrayType::*)(std::vector<SizeType> const &, T)) & ArrayType::Set)
-      .def("Set", [](ArrayType &a, std::vector<SizeType> const &i, float v) { a.Set(i, T(v)); })
-      .def("Set", (void (ArrayType::*)(SizeType, T)) & ArrayType::Set);
+  py::class_<ArrayType, std::shared_ptr<ArrayType>>(module, custom_name.c_str());
+  /*
+        .def(py::init<std::vector<SizeType> const &>())
```
n-hutton: del?

libs/vectorise/include/vectorise/memory/shared_array.hpp
```diff
@@ -44,13 +44,13 @@ class SharedArray : public VectorSlice<T, type_size>
   // TODO(check IfIsPodOrFixedPoint memory safe)
```
n-hutton: priv issue?

libs/vectorise/include/vectorise/memory/vector_slice.hpp
```diff
@@ -89,8 +89,8 @@ class VectorSlice
     return reverse_iterator(pointer_ - 1, pointer_ - 1);
   }
 
-  template <typename R = T>
-  typename std::enable_if<std::is_pod<R>::value>::type SetAllZero()
+  // TODO: ensure trivial type
```
n-hutton: p. issue

libs/vm/examples/01_basic_vm/scripts/calculate.fetch
```diff
@@ -4,4 +4,5 @@ function main()
 
   b = a + b;
   Print("The result is: " + toString(b));
+
```
n-hutton: whitespace only

libs/vm/include/vm/matrix.hpp
```diff
@@ -352,7 +352,7 @@ struct Matrix : public IMatrix
       RuntimeError("index out of bounds");
       return nullptr;
     }
-    return &matrix.At(std::vector<typename fetch::math::Tensor<T>::SizeType>(column, row));
+    return &matrix.At(column, row);
```
n-hutton, line 355: is the  & def safe?

also, matrix is a public member?

libs/vm_modules/include/vm_modules/ml/tensor.hpp
```diff
@@ -41,24 +41,26 @@ class TensorWrapper : public fetch::vm::Object, public fetch::math::Tensor<float
   {
     return {new TensorWrapper(vm, type_id, shape->elements)};
   }
-
-  void SetAt(uint64_t index, float value)
-  {
-    (*this).At(index) = value;
-  }
-
+  /* TODO: this is broken by design
```
n-hutton: ???

libs/vm_modules/include/vm_modules/ml/tensor.hpp
```diff
@@ -41,24 +41,26 @@ class TensorWrapper : public fetch::vm::Object, public fetch::math::Tensor<float
...
 };
 
 inline void CreateTensor(std::shared_ptr<fetch::vm::Module> module)
 {
   module->CreateClassType<TensorWrapper>("Tensor")
-      .CreateTypeConstuctor<fetch::vm::Ptr<fetch::vm::Array<TensorWrapper::SizeType>>>()
-      .CreateInstanceFunction("SetAt", &TensorWrapper::SetAt)
-      .CreateInstanceFunction("ToString", &TensorWrapper::ToString);
+      .CreateTypeConstuctor<fetch::vm::Ptr<fetch::vm::Array<TensorWrapper::SizeType>>>();
+  //      .CreateInstanceFunction("SetAt", &TensorWrapper::SetAt)
```
n-hutton: del

libs/vm_modules/include/vm_modules/ml/tensor.hpp
```diff
@@ -41,24 +41,26 @@ class TensorWrapper : public fetch::vm::Object, public fetch::math::Tensor<float
...
+  //      .CreateInstanceFunction("ToString", &TensorWrapper::ToString);
```
n-hutton: below: any reason submodules are update?

n-hutton (changes requested):

libs/math/include/math/matrix_operations.hpp
baykaner: I don't see a problem with having two implementations for each function. Yes, that's more code but the second function is always trivially implemented as:
Type Fn(input)
{
  Type ret;
  Fn(input, ret);
  return ret;
}
So there's almost no additional error surface (since the implementation is the same); plus the user doesn't have to think about/care which version exists - If they have the return object already they pass it in, if they don't then they shouldn't have to think about the object size/shape.

libs/core/include/core/fixed_point/fixed_point.hpp
baykaner: 1. I'm not sure. On the one hand it's a basic type. On the other hand it's (currently) only used in the Math library. For now I'm leaving it here
2. "have the seraliazer with the definition" - does not seem to be the case for byte_array, which I probably copied. Is byte_array a special case?

libs/core/include/core/fixed_point/fixed_point.hpp
baykaner: maybe there was a reason once, but doesn't seem to be necessary; will fix

libs/math/include/math/clustering/k_means.hpp
baykaner: Will implement Setter as Troels' described in Slack

libs/math/include/math/fundamental_operators.hpp
baykaner: yes, but for now lets leave it named so it's clear what it's for; we can anonymise it as part of #854

libs/math/include/math/fundamental_operators.hpp
baykaner: only if ArrayType is a MathArray yep :-)

libs/math/include/math/fundamental_operators.hpp
baykaner: This vectorisation code was previously in use, so will presumably work again when we come to reimplement it. I guess there's nothing wrong with using lambdas, but you're right that free functions might be clearer. I think this would be part of issue #854

libs/math/include/math/fundamental_operators.hpp
baykaner: defined in vectorise/memory/range.hpp
This will get re-included when we reimplement these vectorised methods in #854

libs/math/include/math/matrix_operations.hpp
baykaner: that's how they work:
max element returns an iterator - https://en.cppreference.com/w/cpp/algorithm/max_element
accumulate returns a type - https://en.cppreference.com/w/cpp/algorithm/accumulate

baykaner: with that said matrix_operations is scheduled to get a lot of extra test code after this tensor refactor so errors are possible

libs/math/include/math/ml/activation_functions/softmax.hpp
baykaner: No, next task will be to go through ml library updating to use Slices everywhere; but that's a big job and this is already a big PR. Not sure if ML library will be more or less efficient after this PR goes in (more efficient tensor vs less efficient copies in ML library); but it will be much more efficient after the next one.

libs/math/include/math/tensor.hpp
baykaner: I think it's a reference to numpy.arange - which is probably short for ArrayRange?
Or it could be a typo - who knows with Troels :-p

libs/ml/tests/ml/clustering/tsne_tests.cpp
baykaner: tsne is a learning algorithm. This test seems to check a specific value is arrived at given fixed initialisation. Not sure why this number has been changed in this PR; maybe carried over from a merge somehow.

libs/vm/include/vm/matrix.hpp
baykaner: not sure I understand your comments.
This change just mirrors the change to tensor accessors (used to take a vector of n-dims size, now takes a parameter pack)

libs/vm_modules/include/vm_modules/ml/tensor.hpp
baykaner: @troelsfr did this so I'm not certain - best guess; because the Setter works differently now, so presumably he wanted to disable this until the tensor setter was working properly.

baykaner: I propose we leave these here with the issue #861 raised, so as not to hold up further tensor reworking PRs; we'll come back and fix it soon after

libs/core/include/core/fixed_point/fixed_point.hpp
n-hutton: byte array is a special case I believe yes.

libs/math/include/math/fundamental_operators.hpp
n-hutton: Still not sure what trivial range is

libs/math/include/math/tensor.hpp
n-hutton: might be good to find out

libs/math/include/math/tensor.hpp
```diff
@@ -381,217 +655,1227 @@ class Tensor
...
+   * @return       returns single value as Type
+   *
+   **/
+  Type L2Loss() const
+  {
+    Type sum = data_.in_parallel().SumReduce([](VectorRegisterType const &v) { return v * v; });
+    return sum * Type(0.5);
+  }
+
+  //
```
n-hutton: lots of dead code still

libs/ml/include/ml/clustering/tsne.hpp
```diff
@@ -251,7 +249,8 @@ class TSNE
     ArrayType d =
         fetch::math::Multiply(DataType(-2), fetch::math::DotTranspose(input_matrix, input_matrix));
 
-    d = fetch::math::Add(fetch::math::Add(d, sum_x).Transpose(), sum_x);
+    d = (d + sum_x).Transpose() + sum_x;
+    //    d = fetch::math::Add(fetch::math::Add(d, sum_x).Transpose(), sum_x);
```
n-hutton: deleteme

libs/ml/tests/ml/ops/embeddings.cpp
```diff
@@ -71,9 +71,19 @@ TYPED_TEST(EmbeddingsTest, forward)
...
+  std::cout << " ---- " << std::endl;
+  for (unsigned int i{0}; i < 2; ++i)
   {
-    EXPECT_EQ(output.At(i), typename TypeParam::Type(gt[i]));
+    for (unsigned int j{0}; j < 6; ++j)
+    {
+      EXPECT_EQ(output.At(i, j), typename TypeParam::Type(gt[(i * 6) + j]));
+      if (output.At(i, j) != (typename TypeParam::Type(gt[(i * 6) + j])))
+      {
+        std::cerr << "ERROR: " << output.At(i, j) << " "
```
n-hutton, line 82: Don't change but usually use fetch logger error

libs/vm/include/vm/matrix.hpp
n-hutton: taking a reference to the matrix will be unsafe if the matrix goes out of scope and the pointer is still used - need to look at use case for this

libs/vm_modules/include/vm_modules/ml/tensor.hpp
```diff
@@ -42,23 +42,24 @@ class TensorWrapper : public fetch::vm::Object, public fetch::math::Tensor<float
...
-  void SetAt(uint64_t index, float value)
-  {
-    (*this).At(index) = value;
-  }
-
-  fetch::vm::Ptr<fetch::vm::String> ToString()
-  {
-    return new fetch::vm::String(vm_, (*this).Tensor::ToString());
-  }
+  //  // TODO(private 123): change in tensor design breaks these
```
n-hutton: dead code

libs/vm_modules/include/vm_modules/ml/tensor.hpp
n-hutton: Almost certain this is not intended - I would suggest to revert

n-hutton (changes requested):

n-hutton (approved):

PR #847, opened 2019-04-11
Key<...> (storage): Wrong calculation from bits to last block & last bit indexes

### LINKED ISSUE #846: Key<...> (storage): Wrong calculation from bits to last block * last bit indexes
This issue has serious impact on comparison of keys, omitting certain number bits during comparison, what ultimately manifests in upper storage layers as issue when multiple keys points to the same object in the storage.

pbukva:
This solves the #846 issue.

Fixes #846


libs/meta/include/meta/log2.hpp
```diff
@@ -17,18 +17,25 @@
...
-#include "vectorise/meta/log2.hpp"
+#include "meta/type_traits.hpp"
 
 namespace fetch {
 namespace meta {
 
-template <uint64_t VALUE>
-struct IsLog2
+template <typename T>
+constexpr EnableIf<std::is_integral<typename std::decay<T>::type>::value, bool> isLog2(
```
n-hutton: isLog2 should be IsLog2

libs/storage/include/storage/key.hpp
```diff
@@ -40,35 +43,34 @@ namespace storage {
...
  */
-template <std::size_t BITS = 256>
+template <std::size_t V_BITS = 256, typename BLOCK_TYPE = uint64_t>
 struct Key
 {
-  enum
+  static_assert(std::is_integral<BLOCK_TYPE>::value, "The BLOCK_TYPE must be integer type.");
+  static_assert(meta::isLog2(V_BITS) && (V_BITS >= 128),
+                "Keys expected to be a cryptographic hash function output");
+  using block_type = BLOCK_TYPE;
```
n-hutton: style guide should be using BlockType

libs/storage/include/storage/key_value_index.hpp
```diff
@@ -88,29 +88,26 @@ struct KeyValuePair
...
-  KeyValuePair()
-  {
-    memset(this, 0, sizeof(decltype(*this)));
-    parent = uint64_t(-1);
-  }
+  using key_type   = Key<S>;
+  using index_type = uint64_t;
 
-  using key_type = Key<S>;
+  static constexpr index_type TREE_ROOT_VALUE{~index_type{0}};
```
n-hutton: Just use std::numeric_limit?

libs/storage/include/storage/key_value_index.hpp
```diff
@@ -88,29 +88,26 @@ struct KeyValuePair
...
 
-  key_type key{};
-  uint8_t  hash[N]{};
+  key_type key;
+  uint8_t  hash[N]{{}};
```
n-hutton: Does calling the constructor like this initialize the memory to 0?

libs/storage/include/storage/key_value_index.hpp
```diff
@@ -88,29 +88,26 @@ struct KeyValuePair
...
 
   // Ref to parent, left and right branches
-  uint64_t parent = uint64_t(-1);
+  index_type parent = TREE_ROOT_VALUE;
 
   union
   {
-    uint64_t value = 0;
-    uint64_t left;
+    index_type value = 0;
```
n-hutton: Style guide IndexType

libs/storage/include/storage/key_value_index.hpp
```diff
@@ -152,15 +149,7 @@ struct KeyValuePair
...
-    byte_array::ByteArray ret;
-    ret.Resize(N);
-
-    for (std::size_t i = 0; i < N; ++i)
-    {
-      ret[i] = hash[i];
-    }
-
-    return ret;
+    return {hash, N};
```
n-hutton, line 152: 👍

libs/storage/include/storage/key_value_index.hpp
```diff
@@ -747,7 +735,7 @@ class KeyValueIndex
 
       stack_.Get(next, kv);
 
-      left_right = key.Compare(kv.key, pos, kv.split >> 8, kv.split & 63);
+      left_right = key.Compare(kv.key, pos, kv.split);
```
n-hutton, line 740: 👍

libs/storage/tests/gtest/object_store_tests.cpp
```diff
@@ -280,39 +273,39 @@ TEST(storage_object_store_with_STL_gtest, iterator_over_basic_struct)
 
     std::sort(objects.begin(), objects.end());
 
-    auto mmm = testStore.begin();
-    while (mmm != testStore.end())
-    {
-      ++mmm;
-    }
+    // auto mmm = testStore.begin();
```
n-hutton: delete

libs/storage/tests/gtest/object_store_tests.cpp
```diff
@@ -480,15 +475,39 @@ TEST(storage_object_store, correlated_strings_work_correctly_working)
...
+    EXPECT_EQ(testStore.size(), ++expected_size);
+  }
+
+  ASSERT_EQ(testStore.size(), unique_ids.size()) << "ERROR: Failed to verify final size!";
+}
+
+TEST(storage_object_store, correlated_strings_work_correctly_failing_dedicated)
+{
+  ObjectStore<std::string> testStore;
+  testStore.New("testFile_01.db", "testIndex_01.db");
```
n-hutton: should have unique *.db names

libs/storage/tests/gtest/object_store_tests.cpp
```diff
@@ -480,15 +475,39 @@ TEST(storage_object_store, correlated_strings_work_correctly_working)
...
+
+  std::unordered_set<ResourceID> unique_ids;
+  unique_ids.emplace(FromBase64("X+zrZv/IbzjbUnhsbWlsecLbwjndTpG0ZynXOif7V+k="));
```
n-hutton: No comments for this test to indicate whether these two keys are different or why they were chosen. There are now three correlated strings tests in this file - I think there should only be one

libs/storage/include/storage/key.hpp
```diff
@@ -157,14 +153,8 @@ struct Key
...
 
-    uint64_t *ret_reinterpret = reinterpret_cast<uint64_t *>(ret.pointer());
-
-    // Force the byte array to fill the 64 bit key from 'left to right'
-    for (std::size_t i = 0; i < BLOCKS; ++i)
-    {
-      ret_reinterpret[i] = platform::ConvertToBigEndian(key_[i]);
-    }
-
+    block_type(&ret_reinterpret)[BLOCKS] = *reinterpret_cast<block_type(*)[BLOCKS]>(ret.pointer());
```
ejfitzgerald: Don't change, but this seems more complicated than is necessary. It also appears that we do not actually use the `[BLOCKS]` array size infomation

libs/meta/include/meta/log2.hpp
ejfitzgerald: Style 🚓 : `IsLog2`

ejfitzgerald: Also lets use the standard `meta::IfIsInteger` constructs

libs/storage/tests/selective_test/key_tests.cpp
```diff
@@ -19,34 +19,293 @@
...
+    auto const &key = keys.back();
+
+    if (i == 0)
+    {
+      continue;
+    }
+
+    auto const &last_key{keys[i - 1]};
+
+    // auto const &last_bs_key{bs_keys[i-1]};
```
n-hutton: delete

libs/storage/tests/selective_test/key_tests.cpp
```diff
@@ -19,34 +19,293 @@
...
-using ConstByteArray = fetch::byte_array::ConstByteArray;
-using DefaultKey     = fetch::storage::Key<256>;
+using ByteArray          = fetch::byte_array::ByteArray;
+using ConstByteArray     = fetch::byte_array::ConstByteArray;
+using DefaultKey         = fetch::storage::Key<256>;
+using DefaultBitset      = std::bitset<DefaultKey::BITS>;
+using DefaultArray       = ArrayB<uint64_t, DefaultKey::BITS>;
+using UnorderedSetBitset = std::unordered_set<DefaultBitset>;
+
+TEST(new_key_test, test_compare_keys_shifted_by_single_bit__triangular_formation)
```
n-hutton: double underscore

libs/meta/include/meta/log2.hpp
```diff
@@ -17,18 +17,25 @@
...
-  static constexpr uint64_t calculated_value = 1u << log2_value;
-  static constexpr bool     value            = (calculated_value == VALUE);
-};
+  return static_cast<bool>(val && !(val & (val - 1)));
+}
+
+template <typename T>
+constexpr EnableIf<std::is_integral<typename std::decay<T>::type>::value,
+                   typename std::decay<T>::type>
+log2(T &&val) noexcept
```
ejfitzgerald: Style 🚓 : `Log2`

Also again lets use `meta::IfIsInteger`

libs/storage/tests/selective_test/key_tests.cpp
```diff
@@ -19,34 +19,293 @@
...
+    // Comparing the the key with itself, expected identity result
+    EXPECT_EQ(key.size_in_bits(), pos);
+    EXPECT_EQ(0, res);
+
+    // Comparing *current* key against *pevious* key which is BIGGER by value
+    res = key.Compare(last_key, pos, static_cast<uint16_t>(key.size_in_bits()));
+    EXPECT_EQ(key.size_in_bits() - i, pos);
+    EXPECT_EQ(-1, res);
+
+    // Reciprocally comparing *pevious* key to *current* key which is SMALLER by value
```
n-hutton: spelling: previous

libs/storage/include/storage/key.hpp
```diff
@@ -173,13 +163,13 @@ struct Key
...
    */
-  std::size_t size_in_bits() const
+  static constexpr std::size_t size_in_bits()
   {
     return BITS;
   }
 
 private:
-  uint64_t key_[BLOCKS];
+  KeyArray key_{{}};
```
ejfitzgerald: Not sure this double braces init. is needed? Am I wrong?

libs/storage/tests/gtest/object_store_tests.cpp
ejfitzgerald: Dead code?

libs/testing/include/testing/bitset_array_conversion.hpp
```diff
@@ -0,0 +1,115 @@
...
+template <typename T, std::size_t SIZE, std::size_t BITS = (SIZE * sizeof(T) * 8),
+          meta::EnableIf<std::is_integral<T>::value && meta::isLog2(BITS)> * = nullptr>
+using Array = std::array<T, SIZE>;
+
+template <typename T, std::size_t BITS,
+          meta::EnableIf<std::is_integral<T>::value && meta::isLog2(BITS)> * = nullptr>
+using ArrayB = std::array<T, (BITS >> meta::log2(sizeof(T) * 8))>;
+
+template <typename T, std::size_t BITS, std::size_t T_BITS_LOG2 = meta::log2(sizeof(T) * 8),
+          T T_BITS_MASK = ((1 << T_BITS_LOG2) - 1)>
```
ejfitzgerald: Style 🚓 : I know that you love this `T_` convention @pbukva but you are the only person in the code base who uses them. Please can we avoid them in the future.

ejfitzgerald: Also this SFINAE stuff here probably could in future do with a breaking out into its own type alias to make the reading easier.

ejfitzgerald (changes requested):

libs/storage/tests/selective_test/key_tests.cpp
```diff
@@ -19,34 +19,293 @@
...
+    res = last_key.Compare(key, pos, static_cast<uint16_t>(key.size_in_bits()));
+    EXPECT_EQ(key.size_in_bits() - i, pos);
+    EXPECT_EQ(1, res);
+  }
+}
+
+TEST(new_key_test, test_compare_for_keys_whis_shifted_single_zero_by_one_bit__moving_zero_formation)
```
n-hutton: I would suggest that the test name is short and you comment  the test appropriately

libs/storage/tests/selective_test/key_tests.cpp
```diff
@@ -19,34 +19,293 @@
...
+
+  std::vector<DefaultKey> seen_keys;
+  seen_keys.reserve(key_val.size());
+
+  for (std::size_t i = 0; i < key_val.size(); ++i)
+  {
+    auto       bs_key = (key_val >> i);
+    auto const k_arr{to_array<DefaultArray::value_type>(bs_key)};
+
+    byte_array::ConstByteArray const cba_key{
```
n-hutton: Isn't this testing the same thing as the test above? The way this is written is too complex to follow easily.

libs/testing/include/testing/bitset_array_conversion.hpp
```diff
@@ -0,0 +1,115 @@
...
+  {
+    ostream << std::setw(2) << std::setfill('0') << std::hex << static_cast<int>(bytes[i]);
+  }
+
+  ostream.copyfmt(orig_state);
+
+  return ostream;
+}
+
+template <typename T, std::size_t BITS, std::size_t T_BITS_LOG2 = meta::log2(sizeof(T) * 8)>
```
n-hutton: This is too complex

libs/testing/tests/bitset_array_conversion_tests.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+#include "testing/bitset_array_conversion.hpp"
+
+#include <gtest/gtest.h>
+
+namespace fetch {
+namespace testing {
+namespace {
+
+using namespace testing;
```
n-hutton, line 27: what's the reason for anonymous namespace and using namespace testing?

libs/testing/tests/bitset_array_conversion_tests.cpp
```diff
@@ -0,0 +1,113 @@
...
+using DefaultBitset = std::bitset<BITS>;
+
+TEST(array_bitset_test, test_conversion_bitset_to_array_and_back)
+{
+  std::vector<DefaultArray>  arr_keys;
+  std::vector<DefaultBitset> bs_keys;
+
+  auto const starting_bitset{~DefaultBitset{0}};  //= 111...1 (bin) = 0xff...f (hex)
+
+  //`bs_keys` container will contain (values *DISPLAYED* in *BIG* Endian encoding):
```
n-hutton, line 41: 👍 good comment

n-hutton (changes requested):
I don't understand what has happened here - I thought we agreed to have an initial pull request that fixed the bug in question, and then additional pull request(s) for more tests if neccessary. Lets chat IRL

libs/storage/include/storage/key.hpp
pbukva: This is to ensure initialization by 0.

libs/testing/include/testing/bitset_array_conversion.hpp
pbukva: That `T_` prefix in `T_BITES...` parameters is there not because my convention, but simply because of necessity to show association with `T` type which is actually the first template parameter (we already have `BITS` parameter which the second template parameter)

libs/storage/include/storage/key_value_index.hpp
pbukva: Yes it does, and it is intended to do so.

libs/testing/tests/bitset_array_conversion_tests.cpp
pbukva: This is the pattern we all should use for unit test files.
Enclosing whole unit test code in to anonymous namespace is very good defensive approach for writing unit tests.

The reason being that unit test is .cpp compilation unit, and all symbols defined(implemented) in that .cpp file can potentially leak out in to resulting **production** code binary **IF** it gets linked in.
Once that happens, and mangled symbol name from unit test matches mangled symbol name from production code (what surprisingly likely because people tend to use names for classes or functions based on production code which thy are writing test for),  linker has free hand to pick symbol from duplicates.
  
The reason(s) **why** unit test .o file (or whole unit test .a library) can get linked in to prod binary are not important, it simply can happen (e.g. due to copying unit test .cpp file by mistake to production code folder, or by messing with file names while reorganizing repo folder structure, or by error in build pipeline) - I personally experienced that on multiple occasions, and each occasion manifested with extremely weird/hard-to-explain behavior with extreme difficulty to locate the root cause.

libs/meta/include/meta/log2.hpp
pbukva: I'm aware of it, and decided not to use it since it excludes `bool` type from integer types, what I don't think we should do here - the `Log2(...)` and `IsLog2(...)` are general functions giving information about value of that particular type.

libs/storage/include/storage/key.hpp
ejfitzgerald: I believe that will happen provided that you simply initialise it. ie. `key_{}`

pbukva: Yeah, that feature came with C++14.

libs/meta/include/meta/log2.hpp
ejfitzgerald: Why would you take a log2 of a bool?

ejfitzgerald: We are basically only using lowercase function now as simple class accessors. So lets change this so that it more consistent with the rest of the code base

libs/storage/include/storage/key.hpp
ejfitzgerald: Okay so since we are using C++14 we can use it?

ejfitzgerald (changes requested):

libs/storage/include/storage/key_value_index.hpp
pbukva: The `~` operator is more reliable, since it works across all bits of the type in very defined & expected way.
The `std::numeric_limit<...>()` gives wrong numbers for signed integer types.

libs/meta/include/meta/log2.hpp
```diff
@@ -17,18 +17,22 @@
...
-#include "vectorise/meta/log2.hpp"
+#include "meta/type_traits.hpp"
 
 namespace fetch {
 namespace meta {
 
-template <uint64_t VALUE>
-struct IsLog2
+template <typename T>
+constexpr EnableIf<IsInteger<Decay<T>>, bool> IsLog2(T &&val) noexcept
```
ejfitzgerald: I am not sure what the benefit of the universal reference is here. I can only see this function being used for primitive types. Presumably if we removed the universal reference then we also loose the `Decay<T>` and therefore we can use the simple `IfIsInteger<>` alias

libs/storage/include/storage/key_value_index.hpp
```diff
@@ -661,8 +650,10 @@ class KeyValueIndex
 
     FindNearest(key, kv, split, pos, left_right, depth, max_bits);
 
-    pos = 0;
-    kv.key.Compare(key_str, pos, 0, 64);
+    // TODO(issue 874): Compare(...) call bellow seems to be unnecessary/redundant,
```
n-hutton, line 653: delete?

libs/meta/include/meta/log2.hpp
```diff
@@ -17,18 +17,22 @@
...
-#include "vectorise/meta/log2.hpp"
+#include "meta/type_traits.hpp"
 
 namespace fetch {
 namespace meta {
 
-template <uint64_t VALUE>
-struct IsLog2
+template <typename T>
+constexpr EnableIf<IsInteger<T>, bool> IsLog2(T val) noexcept
```
n-hutton: Does adding noexcept to a constexpr do anything?

libs/storage/include/storage/key.hpp
```diff
@@ -40,35 +43,35 @@ namespace storage {
  * all of 0xF
  *
  */
-template <std::size_t BITS = 256>
+template <std::size_t V_BITS = 256, typename BLOCK_TYPE = uint64_t>
```
n-hutton, line 46: This block type template is unneccessary (probably the V_BITS also). I think we need to agree to avoid templates if we can - templates are in my mind mostly to avoid code duplication, but if we only ever have one template instantiation of something in the codebase it's counter-productive.

In this case, the ```Key``` will immediately break if a BLOCK_TYPE other than 64 bits is used as it will fail on CountTrailingZeroes64.

libs/storage/include/storage/key.hpp
```diff
@@ -108,39 +102,42 @@ struct Key
...
+    uint16_t  bit  = (diff == 0) ? static_cast<uint16_t>(BLOCK_SIZE_BITS)
+                               : static_cast<uint16_t>(platform::CountTrailingZeroes64(diff));
     if (i == last_block)
     {
       bit = std::min(bit, last_bit);
     }
 
-    pos = bit + (i << 6);
-    if (pos >= int(this->size_in_bits()))
+    pos = bit + (i << BLOCK_SIZE_BITS_LOG2);
```
n-hutton, line 143: We discussed how this code was hard to follow - perhaps some inline comments would help

libs/storage/tests/selective_test/key_tests.cpp
```diff
@@ -19,34 +19,332 @@
...
+    EXPECT_EQ(-1, res);
+
+    // Reciprocally comparing *previous* key to *current* key which is SMALLER by value
+    res = last_key.Compare(key, pos, static_cast<uint16_t>(key.size_in_bits()));
+    EXPECT_EQ(key.size_in_bits() - i, pos);
+    EXPECT_EQ(1, res);
+  }
+}
+
+TEST_F(NewKeyTest, test_compare_for_keys_whis_moving_zero)
```
n-hutton: Spelling: ```whis```?

libs/storage/tests/selective_test/key_tests.cpp
```diff
@@ -19,34 +19,332 @@
...
+      auto const       itr = std::find(seen_keys.begin(), seen_keys.end(), key);
+
+      EXPECT_EQ(seen_keys.cend(), itr);  // Expected *NOT* to be found
+
+      seen_keys.emplace_back(std::move(key));
+    }
+  }
+};
+
+TEST_F(NewKeyTest, test_compare_keys_triang_bit_shift)
```
n-hutton, line 60: It's a bit odd that these tests derive from a fixture they don't use/doesn't do anything

libs/storage/tests/selective_test/key_tests.cpp
```diff
@@ -19,34 +19,332 @@
...
+    res = last_key.Compare(key, pos, static_cast<uint16_t>(key.size_in_bits()));
+    EXPECT_EQ(key.size_in_bits() - i, pos);
+    EXPECT_EQ(1, res);
+  }
+}
+
+TEST_F(NewKeyTest, test_compare_for_keys_whis_moving_zero)
+{
+  std::vector<DefaultKey>    keys;
+  std::vector<DefaultArray>  arr_keys;
```
n-hutton: arr_keys appears unused, bs_keys are also not used as a vector - is this so that it can be viewed that they populate correctly in an IDE?

libs/storage/tests/selective_test/key_tests.cpp
```diff
@@ -19,34 +19,332 @@
...
+  // unique_hashes[3]   = 000111...11
+  // unique_hashes[4]   = 000011...11
+  // unique_hashes[5]   = 000001...11
+  //             ...
+  // unique_hashes[255] = 000000...01
+
+  test_correlated_keys_are_unique(std::move(unique_hashes));
+}
+
+TEST_F(NewKeyTest, test_comparison_using_last_bit_value__moving_zero_formation)
```
n-hutton: double ```__```

libs/storage/tests/selective_test/key_tests.cpp
```diff
@@ -19,34 +19,332 @@
...
+      // Reciprocally comparing *ref* key to *current* key which are the same when
+      // given the *last position* limitation.
+      res = ref_key.Compare(key, pos, last_bit_position_BEFORE_actual_difference);
+      EXPECT_EQ(last_bit_position_BEFORE_actual_difference, pos);
+      EXPECT_EQ(1, res);
+    }
+  }
+}
+
+TEST_F(NewKeyTest, test_comparison_using_last_bit_value__triangular_formation)
```
n-hutton: ```__```

libs/testing/include/testing/bitset_array_conversion.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+    if (bit_value)
+    {
+      bs[i] = true;
+    }
+  }
+  return bs;
+}
+
+template <typename T, std::size_t SIZE>
```
n-hutton, line 78: Wouldn't it make sense to use ```ConstByteArray ToHex(...```?

libs/testing/include/testing/bitset_array_conversion.hpp
```diff
@@ -0,0 +1,116 @@
...
+template <typename T, std::size_t SIZE, std::size_t BITS = (SIZE * sizeof(T) * 8),
+          meta::EnableIf<meta::IsInteger<T> && meta::IsLog2(BITS)> * = nullptr>
+using Array = std::array<T, SIZE>;
+
+template <typename T, std::size_t BITS,
+          meta::EnableIf<meta::IsInteger<T> && meta::IsLog2(BITS)> * = nullptr>
+using ArrayB = std::array<T, (BITS >> meta::Log2(sizeof(T) * 8))>;
+
+template <typename T, std::size_t BITS>
+ArrayB<T, BITS> to_array(std::bitset<BITS> const &bs)
```
n-hutton: Code style: We do not use snake_case. Also, these free functions need comments - it looks logically wrong at first glance because you have an ArrayB<uint64_t, BITS> which might cause the reader to assume an array of 64 * bits

libs/testing/include/testing/bitset_array_conversion.hpp
```diff
@@ -0,0 +1,116 @@
...
+  {
+    ostream << std::setw(2) << std::setfill('0') << std::hex << static_cast<int>(bytes[i]);
+  }
+
+  ostream.copyfmt(orig_state);
+
+  return ostream;
+}
+
+template <typename T, std::size_t BITS>
```
n-hutton: This EnableIf looks redundant to me since ArrayB already checks for integer and log2

libs/vectorise/tests/meta/gtest/test_log2_tests.cpp
```diff
@@ -16,112 +16,269 @@
...
-  EXPECT_TRUE(TEST_LOG_EX__EQ__X(5));
-  EXPECT_TRUE(TEST_LOG_EX__EQ__X(6));
-  EXPECT_TRUE(TEST_LOG_EX__EQ__X(7));
-  EXPECT_TRUE(TEST_LOG_EX__EQ__X(8));
-  EXPECT_TRUE(TEST_LOG_EX__EQ__X(9));
-}
+using namespace testing;
 
-TEST(vectorise_exact_exponents_gtest, ten_to_nineteen)
+class LogTest : public Test
```
n-hutton, line 33: Test needs comment(s)

n-hutton (changes requested):

libs/meta/include/meta/log2.hpp
pbukva: Yes, because `constexpr` **function** can be called in both contexts - compile-time and run-time. 
The `noexcept` is for when this function is called in run-time.

libs/storage/include/storage/key.hpp
pbukva: I will first react to the 2nd topic (suggested Key break  for `BLOCK_TYPE` other than 64 bits):

It will **NOT** break, reason being that:
1) since it calculates **trailing** (not leading) zeroes (because of **Little**-Endian , our whole our code base assumes **Little**-Endianess on many places), 
1) `BLOCK_TYPE` can be only such that `sizeof(BLOCK_TYPE) =< sizeof(uint64_t)`, that is ensured by compile time checks put in place (see the `static_assert(...)` checks)
 2) value of any integral `BLOCK_SIZE` type (which conforms to compile time checks put in place), will be **implicitly** converted by C++ compiler to superseding type `uint64_t` of input parameter of the `CountTrailingZeroes64(uint64_t val)`. I removed messing with endianess conversion.

**Going back to 1st topic:**
The reason why this class is actually templated is to **avoid** dynamic allocation on heap and do rather everything statically on stack using native C++ static array type (originally of `TYPE[SIZE]`, now changed to be `std::array<TYPE<SIZE>` which is just thin wrapper around above native C++ static array).
Now, when class is already templated due to above mentioned reason, it does not make sense not to pull out another type parameter  representing underlying block type. Reason being that I would need to create (inside of class) type alias for block type (e.g. ```using BlockType = uint64_t;```) anyway, then why not to pull it out as template param?).

libs/meta/include/meta/log2.hpp
ejfitzgerald: This was my original suggestion (please check syntax)

```suggestion
constexpr IfIsInteger<T, bool> IsLog2(T val) noexcept
```

libs/meta/include/meta/log2.hpp
```diff
@@ -17,18 +17,22 @@
...
 {
-  static constexpr uint64_t log2_value       = Log2<VALUE>::value;
-  static constexpr uint64_t calculated_value = 1u << log2_value;
-  static constexpr bool     value            = (calculated_value == VALUE);
-};
+  return static_cast<bool>(val && !(val & (val - 1)));
+}
+
+template <typename T>
+constexpr EnableIf<IsInteger<T>, T> Log2(T val) noexcept
```
ejfitzgerald:
```suggestion
constexpr IfIsInteger<T, T> Log2(T val) noexcept
```

ejfitzgerald (changes requested):
I have some suggestions

libs/storage/include/storage/key_value_index.hpp
pbukva: I originally wanted to delete the code, but then decided rather to just comment it out and create issue for it just in case we observe unexpected behavior.

libs/storage/include/storage/key.hpp
pbukva: I put the comment on top of the method. Would that be sufficient?

libs/storage/include/storage/key_value_index.hpp
pbukva: I would like to keep this (due to reason I mentioned above), but if you think it should be deleted, let me know, I will do so.

libs/storage/tests/selective_test/key_tests.cpp
pbukva: Actually, these tests use shared method `test_correlated_keys_are_unique(...)` from the `NewKeyTest` class.
Yes it is possible to convert that method to standalone function.
But I like the presence of class since it allows easy run of all tests from the test class in environments like CLion, where you can select class and run all test belonging to it.

libs/testing/include/testing/bitset_array_conversion.hpp
pbukva: I relocated the `<<` operator to the bottom of this file, and used `ConstByteArray::ToHex()` there.

libs/vectorise/tests/meta/gtest/test_log2_tests.cpp
pbukva: I documented the whole file using doxygen. It should sufficient.

libs/storage/include/storage/key.hpp
n-hutton: That is true, it does break if your type is not signed though, for example int64_t, char etc. Apart from breaking encapsulation a bit it's also pretty inefficient (I assume) to promote 32 or lower bits to a 64 so as to use builtins on it.

on the first point: I view template params as something you want to minimize use of where possible. This case seems somewhat gratuitous to me but as you say, it would instead be an alias/array template, so sure

libs/storage/include/storage/key_value_index.hpp
n-hutton: not too concerned either way

n-hutton (approved):

ejfitzgerald (approved):

PR #864, opened 2019-04-15
Code2vec implementation

juergendkeba:
An implementation of the code2vec graph as proposed in https://arxiv.org/pdf/1803.09473.pdf


libs/ml/examples/code2vec/main_draft.cpp
```diff
@@ -0,0 +1,187 @@
...
+  g.AddNode<Concatenate<ArrayType>>("ContextVectors", {"EmbeddingSourcewords",
+  "EmbeddingPaths", "EmbeddingTargetwords"}, 1);
+
+  //In the original implementation its without bias
+  // Dimensions: (N_CONTEXTS, EMBEDDING_SIZE) = (EMBEDDING_SIZE, 3*EMBEDDING_SIZE) @ (N_CONTEXTS, 3*EMBEDDING_SIZE)
+  g.AddNode<fetch::ml::layers::FullyConnected<ArrayType>>("FC1", {"ContextVectors", "FC"},
+  BATCHSIZE, 3*EMBEDDING_SIZE);
+  
+  // elementwise tanh
+  g.AddNode<fetch::ml::ops::TanH<ArrayType>>("CombinedContextVector", {"FC1"});
```
baykaner: nothing wrong with doing this, but we do allow specifying the activation within the fully connected layer as well if you wanted to wrap these two into one line

libs/ml/examples/code2vec/main_draft.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+
+  // Dimensions: (1, vocab_size_functions) = Transpose (vocab_size_functions, 1)
+  std::string result = g.AddNode<fetch::mal::ops::Transpose<ArrayType>>("PredictionSoftMaxTransposed", {"PredictionSoftMax"});
+
+  //Initialising variables
+  //Here, the CrossEntropy eats two tensors of size (1, function_name_vocab_size); i.e. it has 1 example, and as many categories as vocab_size
+  fetch::ml::ops::CrossEntropy<ArrayType> criterion;
+  DataType     loss = 0;
+  fetch::math::Tensor<DataType> y_true_vec({1, cloader.GetCounterFunctionNames().size()});
```
baykaner: ArrayType

libs/ml/examples/code2vec/main_draft.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+  for(u_int32_t 0; i < y_true_vec.shape()[1]; i++){
```
baykaner: I think you missed an "i" here?
Anyway, please use SizeType

libs/ml/examples/code2vec/main_draft.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+  // Criterion: Cross Entropy Loss
+  // Here, the CrossEntropy eats two tensors of size (1, function_name_vocab_size); i.e. it has 1 example, and as many categories as vocab_size
+  fetch::ml::ops::CrossEntropy<ArrayType> criterion;
+  DataType     loss = 0;
+
+  // (One hot encoded) \y_{true} vector
+  fetch::math::Tensor<DataType> y_true_vec({1, cloader.GetCounterFunctionNames().size()});
+
+  for(u_int64_t i{0}; i < y_true_vec.shape()[1]; i++){
```
baykaner: better to use:
y_true_vec.Fill(0);

libs/ml/examples/code2vec/main_draft.cpp
```diff
@@ -0,0 +1,203 @@
...
+    // second: function_name_idx is the index of the function name in the vocabulary
+    ContextTensorsLabelPair input = cloader.GetNext();
+
+    // Feeding the tensors to the graph
+    g.SetInput("InputSourceWords", std::get<0>(input.first));
+    g.SetInput("InputPaths", std::get<1>(input.first));
+    g.SetInput("InputTargetWords", std::get<2>(input.first));
+
+    // Preparing the y_true vector (one-hot-encoded)
+    SizeType y_true_i{input.second};
```
baykaner: Not quite sure why you need to do this; isn't input.second already a SizeType? In which case why not:
y_true_vec.Set(0, input.second, 1);

libs/ml/examples/code2vec/main_draft.cpp
```diff
@@ -0,0 +1,203 @@
...
+    // dimension:  (1, vocab_size_functions)
+    ArrayType results = g.Evaluate(result);
+    // dimension:  (1, vocab_size_functions), (1, vocab_size_functions)
+    loss = criterion.Forward({results, y_true_vec});
+
+    // Making the backward pass
+    g.BackPropagate(result, criterion.Backward({results, y_true_vec}));
+
+    // Resetting the y_true vector for reusing it
+    y_true_vec.Set(0, y_true_i, 1);
```
baykaner: I think you mean to set it back to 0, right?

libs/ml/examples/code2vec/main_draft.cpp
```diff
@@ -0,0 +1,203 @@
...
+  }
```
baykaner: would be nice to print the loss periodically as well as something indicating steps/sec

baykaner (changes requested):
cool - nice I can see it's coming together

libs/ml/examples/code2vec/main_draft.cpp
juergendkeba: i will include it! 
I'll make an issue, cause the tanh needs to be incorporated in the activation.hpp first!

Edit: had to drop the FC anyways

libs/ml/examples/code2vec/main_draft.cpp
juergendkeba: Thx!!! :-)

libs/ml/examples/code2vec/main_draft.cpp
juergendkeba: Thx!
I'm using now `.fill(0)`!

libs/ml/examples/code2vec/main_draft.cpp
juergendkeba: Yes! Thx for catching that error!

libs/ml/examples/code2vec/main.cpp
```diff
@@ -0,0 +1,267 @@
...
+  int n_iter{0};
+
+  while (n_epochs < N_EPOCHS)
+  {
+    if (cloader.IsDone())
+    {
+      cloader.Reset();
+      n_epochs++;
+    }
+    g.Step(LEARNING_RATE);
```
baykaner: shouldn't this happen at the end of the loop?
In this case I think you're taking an initial step that probably does nothing at the start?

libs/ml/examples/code2vec/main.cpp
```diff
@@ -0,0 +1,267 @@
...
+    g.BackPropagate(result, criterion.Backward({results, y_true_vec}));
+
+    // Resetting the y_true vector for reusing it
+    y_true_vec.Set(0, input.second, 0);
+
+    n_iter++;
+    if (n_iter % 5 == 0)
+    {
+      std::cout << "MiniBatch: " << n_iter / 5 << " -- Loss : " << loss << std::endl;
+      loss = 0;
```
baykaner, line 274: would be nice to report training steps / second

baykaner (changes requested):

baykaner (approved):

Issue #857, opened 2019-04-15
Implement general Setter with variadic for Tensor


baykaner:

baykaner:
interface like this:
template< typename ... Indices >
void Set(Indices ... indices, Type &val)
{
}

bipll:
@baykaner Will it even compile, a value after a pack?

baykaner:
The interface, from the user's perspective, needs to ACT as though it is as specified above. 
The actual interface will look like this:
template <typename ... Args>
void Set(Args ... args)
{
}
but behave as described previously.
You can see the implementation when I close this issue later this week

Issue #856, opened 2019-04-15
change legacy style typenames


baykaner:
e.g. vector_register_type -> VectorRegisterType

PR #875, opened 2019-04-16
Tensor improvements

baykaner:
- updates from legacy style; closes issue #856
- general variadic Setter for Tensor; closes issue #857
- removed reshape force; closes issue #866
- refactoring Tensor class for readability
- implemented some trivial functions that were not implemented yet (e.g. with calls the math library)


libs/math/include/math/ml/loss_functions.hpp
```diff
@@ -18,5 +18,7 @@
 //------------------------------------------------------------------------------
 
 #include "math/ml/loss_functions/cross_entropy.hpp"
+#include "math/ml/loss_functions/kl_divergence.hpp"
+#include "math/ml/loss_functions/l2_loss.hpp"
```
n-hutton: This file has added includes but no change to code body - are the includes def used in this file? Of course, good practice is to try to include what you use

libs/math/include/math/tensor.hpp
```diff
@@ -59,1807 +63,2407 @@ template <typename T, typename C = memory::SharedArray<T>>
...
-  static SizeType SizeFromShape(SizeVector const &shape)
-  {
-    if (shape.size() == 0)
-    {
-      return SizeType{0};
-    }
-    return std::accumulate(std::begin(shape), std::end(shape), SizeType(1), std::multiplies<>());
-  }
+  template <typename Unsigned>
+  static fetch::meta::IfIsUnsignedInteger<Unsigned, SelfType> Arange(Unsigned const &from,
```
n-hutton, line 284: I am wondering - is it not possible to declare this as ``` SelfType Arange(T from, T to, T delta) ``` then specialize the definition for signed and unsigned integer?

libs/math/include/math/tensor.hpp
```diff
@@ -59,1807 +63,2407 @@ template <typename T, typename C = memory::SharedArray<T>>
...
+  Tensor            ret;
+  SizeType          n = 1;
+  std::vector<Type> elems;
+  elems.reserve(1024);
+  bool failed = false;
+
+  for (SizeType i = 0; i < c.size();)
+  {
+    SizeType last = i;
+    switch (c[i])
```
n-hutton, line 610: this seems like it could benefit with a comment or two (and most of tensor.hpp)

libs/math/include/math/tensor.hpp
```diff
@@ -59,1807 +63,2407 @@ template <typename T, typename C = memory::SharedArray<T>>
...
-      bool success                      = false;
-      (total == this->size()) ? success = true : success = false;
-      return success;
-    }
-    return false;
-  }
+/**
+ * Provides an Tensor that is a copy of the current Tensor
+ *
+ * @return       copy is a Tensor with the same data, size, and shape of this array.
```
n-hutton: Minor : return comment slightly wonky. Suggest : ``` @return A Tensor with the same... ```

libs/math/include/math/tensor.hpp
```diff
@@ -59,1807 +63,2407 @@ template <typename T, typename C = memory::SharedArray<T>>
...
-   *
-   **/
-  SizeVector const &shape() const
-  {
-    return shape_;
-  }
-  SizeType const &shape(SizeType const &n) const
+/**
+ * assign makes a deep copy of data from another tensor into this one
+ * @param other
```
n-hutton: incomplete comment

libs/math/include/math/tensor.hpp
```diff
@@ -59,1807 +63,2407 @@ template <typename T, typename C = memory::SharedArray<T>>
...
-      if (!(Broadcast([](T x, T y) { return x + y; }, self_copy, other_copy, *this)))
-      {
-        throw std::runtime_error("arrays not broadcastable!");
-      }
-    }
-    return *this;
-  }
+/**
+ *
+ * @tparam Indices
```
n-hutton: Incomplete comment

n-hutton: and so on

libs/math/include/math/tensor.hpp
```diff
@@ -59,1807 +63,2407 @@ template <typename T, typename C = memory::SharedArray<T>>
...
+
+/**
+ * One-dimensional reference index operator.
+ *
+ * This operator acts as a one-dimensional array accessor that is
+ * meant for non-constant object instances.
+ *
+ * @tparam S an integral type
+ * @param i the index to access
+ * @return
```
n-hutton: return comment incomplete - also I find it helpful when doing enable_if etc if you comment what the templates are restricting the usage to or what have you

libs/math/include/math/tensor.hpp
```diff
@@ -59,1807 +63,2407 @@ template <typename T, typename C = memory::SharedArray<T>>
...
+  else
   {
-    Type ret{0};
-    for (auto const &v : *this)
+    SelfType self_copy  = this->Copy();
+    SelfType other_copy = other.Copy();
+    if (!(Broadcast([](T x, T y) { return x + y; }, self_copy, other_copy, *this)))
     {
-      ret += v;
+      throw std::runtime_error("arrays not broadcastable!");
```
n-hutton: suggest : ``` arrays not broadcastable when trying to do inline add! ``` or something to that effect, otherwise when you see the message you won't know where it comes from as you have duplicate messages

libs/math/include/math/tensor.hpp
```diff
@@ -59,1807 +63,2407 @@ template <typename T, typename C = memory::SharedArray<T>>
...
-   * This method will initialise all memory with zero.
-   */
-  void SetAllZero()
+/**
+ * Divide Tensor by another Tensor from another and support broadcasting
+ * @param other
+ * @return
+ */
+template <typename T, typename C>
+typename Tensor<
```
n-hutton, line 1797: broken comment

n-hutton: dead code

n-hutton, line 2424: is this due to template loop unrolling?

n-hutton: minor : three ```///```

libs/network/include/network/management/network_manager.hpp
```diff
@@ -35,7 +35,7 @@ class NetworkManager
   using event_function_type = std::function<void()>;
 
   using implementation_type = details::NetworkManagerImplementation;
-  using pointer_type        = std::shared_ptr<implementation_type>;
+  using PointerType         = std::shared_ptr<implementation_type>;
```
n-hutton, line 38: Note: other types adjaent to here don't match style

libs/vectorise/benchmarks/parallel_dispatcher/kernel_bench.cpp
```diff
@@ -28,8 +28,8 @@ template <typename type, unsigned long N = 100000>
 class ParallelDispatcherKernelBench : public ::benchmark::Fixture
 {
 public:
-  using ndarray_type         = SharedArray<type>;
-  using vector_register_type = typename ndarray_type::vector_register_type;
+  using ndarray_type       = SharedArray<type>;
```
n-hutton, line 31: and here etc.

n-hutton (changes requested):

libs/math/include/math/ml/loss_functions.hpp
baykaner: this file (and a couple others like it) exists only as a convenience (so for example in testing/benchmarking/examples you can just include everything in one go); but you're right they should probably be removed.

libs/math/include/math/tensor.hpp
baykaner: is this comment in the wrong place?

baykaner: and this comment?

libs/network/include/network/management/network_manager.hpp
baykaner: I noticed; but I didn't want to break this code that I don't understand, so I fixed the style only minimally due to whatever was required.

n-hutton (approved):

Issue #852, opened 2019-04-12
Update fully connected layer to include activation function


MissingNO57:

pierre-wilmot-fetch-ai:
Do we really want to do that ?
What would be the reason ?

PR #878, opened 2019-04-16
#852 Made activation function as FullyConnected layer parameter

MissingNO57:


pierre-wilmot-fetch-ai:
Code is good.
But I think activation function shouldn't be intergrated into FullyConnected module.

baykaner:
@pierre-wilmot-fetch-ai why?
We all know you prefer PyTorch. In PyTorch you might do something like this:
model = torch.nn.Sequential(
          torch.nn.Linear(D_in, H),
          torch.nn.ReLU(),
          torch.nn.Linear(H, D_out),
        ).to(device)

In Tensorflow you can do something like this:
xx = layers.fully_connected(input, num_outputs=100, activation_fn=tf.Relu)
xx = layers.fully_connected(xx, num_outputs=100, activation_fn=None)

I don't see any reason not to have the convenience method available to users since stacking fully connected layers is going to be something that gets used over and over again

libs/math/include/math/standard_functions/log.hpp
```diff
@@ -35,7 +35,7 @@ namespace math {
 template <typename Type>
 meta::IfIsNonFixedPointArithmetic<Type, void> Log(Type const &x, Type &ret)
 {
-  ret = std::log(x);
+  ret = Type(std::log(double(x)));
```
baykaner: I've probably been guilty of doing this myself, but this isn't a good solution.
This will cast whatever kind of arithmetic comes in to a double, so that std::log(double) is always used, then cast back to the relevant type.
The correct solution, however, would be to do the SFINAE properly. If IfIsNonFixedPointArithmetic does not correctly protect against a type, then you should change to the correct SFINAE (which may need to be implemented if it doesn't exist).
Let me know if you're not sure exactly how to do that, but the examples in math/meta/math_type_traits.hpp should be illustrative.

libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -42,8 +41,9 @@ class FullyConnected : public Layer<T>
   using SizeType     = typename ArrayType::SizeType;
   using WeightsInit  = fetch::ml::ops::WeightsInitialisation;
 
-  FullyConnected(std::uint64_t in, std::uint64_t out, std::string const &name = "FC",
-                 WeightsInit init_mode = WeightsInit::XAVIER_GLOROT)
+  FullyConnected(std::uint64_t in, std::uint64_t out,
```
baykaner: should be a SizeType not a std::uint64_t

libs/ml/include/ml/ops/activation.hpp
```diff
@@ -17,7 +17,56 @@
...
+  RELU,
+  LEAKY_RELU,
+  SIGMOID,
+  LOG_SIGMOID,
+  SOFTMAX,
+  LOG_SOFTMAX
+};
+
+template <class T>
+std::string CreateActivationLayer(ActivationType type, Graph<T> *g, std::string name,
```
baykaner: I would suggest that since you're adding an op, not a layer, and since you're adding a node, you should rename this to something like AddActivationNode

libs/ml/include/ml/ops/activations/logsoftmax.hpp
```diff
@@ -27,49 +28,46 @@ namespace ml {
...
   LogSoftmax()          = default;
   virtual ~LogSoftmax() = default;
 
-  virtual ArrayPtrType Forward(std::vector<ArrayPtrType> const &inputs, ArrayType &output)
+  virtual ArrayType Forward(std::vector<std::reference_wrapper<ArrayType const>> const &inputs,
+                            ArrayType &                                                 output)
   {
-    (void)output;
+    ASSERT(output.shape() == ComputeOutputShape(inputs));
     assert(inputs.size() == 1);
```
baykaner: change to ASSERT please

libs/ml/include/ml/ops/activations/logsoftmax.hpp
```diff
@@ -27,49 +28,46 @@ namespace ml {
...
-    if (!this->output_ || this->output_->shape() != inputs[0]->shape())
-    {
-      this->output_ = std::make_shared<ArrayType>(inputs[0]->shape());
-    }
-
-    fetch::math::Softmax(*inputs[0], *this->output_);
-    fetch::math::Log(*this->output_, *this->output_);
-
-    return this->output_;
+    fetch::math::Softmax(inputs.front().get(), output);
```
baykaner, line 46: You seem to have changed the implementation here. If you've fixed it and it was broken before that means it wasn't covered properly by a test: please either implement the test or raise an issue to implement the test, and assign it to you, me & pierre, and one of us will get to it.

libs/ml/include/ml/ops/activations/logsoftmax.hpp
```diff
@@ -27,49 +28,46 @@ namespace ml {
...
-                                             ArrayPtrType                     errorSignal)
+  virtual std::vector<ArrayType> Backward(
+      std::vector<std::reference_wrapper<const ArrayType>> const &inputs,
+      ArrayType const &                                           errorSignal)
   {
     FETCH_UNUSED(inputs);
-
     assert(inputs.size() == 1);
-    assert(inputs[0]->shape() == errorSignal->shape());
+    assert(inputs.front().get().shape() == errorSignal.shape());
```
baykaner: ASSERT macro please

libs/ml/include/ml/ops/activations/logsoftmax.hpp
```diff
@@ -27,49 +28,46 @@ namespace ml {
...
+    fetch::math::Softmax(inputs.front().get(), output);
+    fetch::math::Log(output, output);
+    return output;
   }
 
-  virtual std::vector<ArrayPtrType> Backward(std::vector<ArrayPtrType> const &inputs,
-                                             ArrayPtrType                     errorSignal)
+  virtual std::vector<ArrayType> Backward(
+      std::vector<std::reference_wrapper<const ArrayType>> const &inputs,
+      ArrayType const &                                           errorSignal)
```
baykaner: errorSignal should be error_signal

libs/ml/include/ml/ops/activations/logsoftmax.hpp
```diff
@@ -27,49 +28,46 @@ namespace ml {
...
   {
     FETCH_UNUSED(inputs);
-
     assert(inputs.size() == 1);
-    assert(inputs[0]->shape() == errorSignal->shape());
+    assert(inputs.front().get().shape() == errorSignal.shape());
 
-    assert(errorSignal->size() == 1);
+    ArrayType returnSignal = fetch::math::Exp(errorSignal);
+    fetch::math::Add(DataType(1), returnSignal, returnSignal);
```
baykaner: returnSignal should be return_signal, or better yet just call it "ret"

libs/ml/include/ml/ops/activation.hpp
baykaner: the input name should be std::string const & name

libs/ml/include/ml/ops/activations/logsoftmax.hpp
```diff
@@ -27,49 +28,46 @@ namespace ml {
...
-    for (std::size_t j = 0; j < errorSignal->size(); ++j)
-    {
-      fetch::math::Add(DataType(1), errorSignal->At(j), errorSignal->At(j));
-    }
-    fetch::math::Divide(DataType(1), *errorSignal, *errorSignal);
-    return {errorSignal};
+    return {returnSignal};
+  }
+
+  virtual std::vector<SizeType> ComputeOutputShape(
```
baykaner: do we actually need these functions to be virtual? Presumably we only need it when we define an activation that inherits from another activation?

baykaner: this method should be const also right?

libs/ml/tests/ml/layers/fully_connected.cpp
```diff
@@ -82,7 +82,7 @@ TYPED_TEST(FullyConnectedTest, node_forward_test)  // Use the class as a Node
   placeholder->SetData(data);
 
   fetch::ml::Node<TypeParam, fetch::ml::layers::FullyConnected<TypeParam>> fc(
-      "FullyConnected", 50u, 42u, "FullyConnected");
+      "FullyConnected", 50u, 42u, fetch::ml::details::ActivationType::NOTHING, "FullyConnected");
```
baykaner: no need to change it, but we don't actually need to specify NOTHING & FullyConnected here

baykaner (changes requested):

libs/ml/include/ml/ops/activations/logsoftmax.hpp
MissingNO57: Implementation of logsoftmax was ancient and incompatible with current ops, so it didn't compile, because it was not used anywhere, so I added tests.

libs/math/include/math/standard_functions/log.hpp
MissingNO57: I found that problem is that std::log doesn't work for type int. So I if it's not floating point primitive we need to cast it to something else or call different function.

libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -18,14 +18,13 @@
 //------------------------------------------------------------------------------
 
 #include "ml/layers/layer.hpp"
+#include "ml/meta/ml_type_traits.hpp"
+#include "ml/ops/activation.hpp"
```
baykaner, line 22: I'm reconsidering whether this file should even exist - in general it was a convenience file for tests and examples so all activations could be included in one go. Internally in the library it's better not to use it; and since the convenience is tempting, I think we should ditch it at some point. For now please replace with a list of includes for the individual activation functions that are covered.

libs/ml/include/ml/layers/PRelu.hpp
```diff
@@ -60,7 +60,8 @@ class PRelu : public SubGraph<T>
   }
 
   virtual std::vector<SizeType> ComputeOutputShape(
-      std::vector<std::reference_wrapper<ArrayType const>> const &inputs)
+
```
baykaner: whitespace?

libs/ml/include/ml/layers/fully_connected.hpp
baykaner: actually - given the new things you've added to activation.hpp we can't remove it anymore. okay :-p in that case I guess it's fine!

libs/ml/include/ml/ops/activations/logsoftmax.hpp
```diff
@@ -27,49 +28,46 @@ namespace ml {
...
 {
 public:
-  using ArrayType      = T;
-  using DataType       = typename ArrayType::Type;
-  using ArrayPtrType   = std::shared_ptr<ArrayType>;
-  using ConstSliceType = typename ArrayType::ConstSliceType;
+  using ArrayType    = T;
+  using DataType     = typename ArrayType::Type;
+  using SizeType     = typename ArrayType::SizeType;
+  using ArrayPtrType = std::shared_ptr<ArrayType>;
```
baykaner: is ArrayPtrType used anywhere?

libs/ml/tests/ml/layers/fully_connected.cpp
```diff
@@ -64,13 +64,13 @@ TYPED_TEST(FullyConnectedTest, ops_backward_test)  // Use the class as an Ops
...
       std::vector<std::reference_wrapper<TypeParam const>>({inputData}));
-  TypeParam errorSignal(std::vector<typename TypeParam::SizeType>({1, 10}));
+  TypeParam error_signal(std::vector<typename TypeParam::SizeType>({1, 10}));
 
-  std::vector<TypeParam> backpropagatedErrorSignals = fc.Backward({inputData}, errorSignal);
-  ASSERT_EQ(backpropagatedErrorSignals.size(), 1);
-  ASSERT_EQ(backpropagatedErrorSignals[0].shape().size(), 2);
-  ASSERT_EQ(backpropagatedErrorSignals[0].shape()[0], 5);
-  ASSERT_EQ(backpropagatedErrorSignals[0].shape()[1], 10);
+  std::vector<TypeParam> backpropagatederror_signals = fc.Backward({inputData}, error_signal);
```
baykaner: back_propagated_error_signals would be the correct snake_case here. but it's a really long name, how about backprop_error?

libs/ml/tests/ml/ops/activations/logsoftmax.cpp
```diff
@@ -0,0 +1,75 @@
...
+                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<32, 32>>>;
+
+TYPED_TEST_CASE(LogSoftmaxTest, MyTypes);
+
+TYPED_TEST(LogSoftmaxTest, forward_test)
+{
+  TypeParam           data(8);
+  TypeParam           gt(8);
+  std::vector<double> dataInput({1, -2, 3, -4, 5, -6, 7, -8});
+  std::vector<double> gtInput({-6.14520134, -9.14520134, -4.14520134, -11.14520134, -2.14520134,
```
baykaner, line 39: did you generate these number using numpy? we should store them/the script somewhere.

baykaner (changes requested):

libs/ml/tests/ml/ops/activations/logsoftmax.cpp
MissingNO57: I did just np.log(np.exp(x)/sum(np.exp(x))) in console, but I will create script for that.

baykaner (approved):
LGTM

PR #879, opened 2019-04-16
TX-NG: New Transaction Classes

ejfitzgerald:
## Summary

This is the first in what will probably become 3 pull requests for the new transaction format changes. This first PR contains the new classes that have been added for the transaction format.

You will notice that these have been placed into a `v2` namespace. This is only temporary and is just a nice way when makes when doing the integration.

The first big change will be that I have remove for the moment the whole flow of `MutableTransaction` -> `UnverifiedTransaction` -> `VerifiedTransaction`. Some degree of this can be introduced if this proves too controversial. However the  new flow is incredible simple as compared.

You simply have an immutable `Transaction` object which effectively takes the place of the `VerifiedTransaction` in the old world.

In order to build a valid transaction a user would typically construct either a `TransactionBuilder` class that would allow the user to programatically construct the `Transaction` (as shown in the example below).

```c++
  auto tx = TransactionBuilder()
                .From(address1)
                .Transfer(address2, 2000)
                .Signer(identity1.identity())
                .Seal()
                .Sign(identity1)
                .Build();
```

Alternatively transactions can be deserialised from a byte stream by a `TransactionSerializer` instance.

The new system retains the existing benefits of the old system (tx payload caching) but should provide a cleaner interface.

The stream formatting is defined covered by this document:

https://docs.google.com/document/d/1q7gzBBJWFLHj27jKe_9O0DfZn_v-6BumCJYeFG3VDC4/edit

## Future improvements

At some point I will refactor the serializer to break it up across multiple files. I will make an issue about this shortly


libs/core/include/core/byte_array/const_byte_array.hpp
```diff
@@ -465,7 +466,15 @@ class ConstByteArray
   void AppendInternal(std::size_t const acc_size, self_type const &other, Arg const &... others)
   {
     AppendInternal(acc_size + other.size(), others...);
-    std::memcpy(pointer() + acc_size, other.pointer(), other.size());
+    memcpy(pointer() + acc_size, other.pointer(),
+           static_cast<size_t>(other.size()) & 0x7FFFFFFFFFFFFFFFull);
```
n-hutton, line 470: Should close issue 257 which is nearby?

Also, doesn't this mask rely on size_t being 64 bits? What is this fixing, it looks like a signed issue?

libs/crypto/include/crypto/ecdsa.hpp
```diff
@@ -37,7 +37,7 @@ class ECDSAVerifier : public Verifier
     , public_key_{identity_ ? PublicKey(identity_.identifier()) : PublicKey()}
   {}
 
-  bool Verify(byte_array_type const &data, byte_array_type const &signature) override
+  bool Verify(ConstByteArray const &data, ConstByteArray const &signature) override
```
n-hutton, line 40: 👍

libs/crypto/src/verify.cpp
```diff
@@ -22,14 +22,39 @@
 namespace fetch {
 namespace crypto {
 
+/**
+ * Build the corresponding Verifier based from the provided identity
+ *
+ * @param identity
```
n-hutton: incomplete comment

libs/ledger/include/ledger/chain/v2/address.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+#include <array>
+
+namespace fetch {
+namespace crypto {
+class Identity;
+}
+
+namespace ledger {
+namespace v2 {
```
n-hutton, line 30: side note: have you heard of inline namespaces?

libs/ledger/include/ledger/chain/v2/address.hpp
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * The Address is a unifying mechanism between multiple different public keys types into a single
+ * format.
+ *
+ * It is generated by creating a hash of the raw bytes of the public key. In addition, when
+ * displaying the address on interfaces an additional 4 bytes of checksum is appended to the
+ * address. This checksum is calculated by hashing the address and selecting the first 4 bytes from
+ * the digest. This mechanism is common place and allows interfaces to integrity check the address.
+ *
+ * ┌──────────────────────────────────────────────────┐
```
n-hutton, line 41: 👍

n-hutton: I wonder if 'Address' is too vague, much like 'Node'

n-hutton: Minor: '1st 4 bytes of the'

libs/ledger/include/ledger/chain/v2/address.hpp
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * @return The raw address
+ */
+inline Address::ConstByteArray Address::address() const
+{
+  return address_;
+}
+
+/**
+ * Get the raw bytes fo the display variant of the address (with checksum)
```
n-hutton: fo -> of

libs/ledger/include/ledger/chain/v2/transaction.hpp
```diff
@@ -0,0 +1,338 @@
...
+class Transaction
+{
+public:
+  using Identity       = crypto::Identity;
+  using ConstByteArray = byte_array::ConstByteArray;
+  using BitVector      = bitmanip::BitVector;
+  using TokenAmount    = uint64_t;
+  using BlockIndex     = uint64_t;
+
+  struct Transfer
```
n-hutton, line 48: It's possibly unneccessary, but I'd comment this liberally - after all people mainly want to know what our block format and TX format is I would expect

libs/ledger/include/ledger/chain/v2/transaction.hpp
```diff
@@ -0,0 +1,338 @@
...
+  /// @}
+
+  /// @name Metadata
+  /// @{
+  ConstByteArray digest_{};                       ///< The digest of the transaction
+  bool           verification_completed_{false};  ///< Signal that the verification has been done
+  bool           verified_{false};                ///< The cached result of the verification
+  /// @}
+
+  // These are the only two ways to generate a transaction from one of the two companion classes
```
n-hutton: 👍 👍 👍

libs/ledger/include/ledger/chain/v2/transaction_builder.hpp
```diff
@@ -0,0 +1,111 @@
...
+  using TransactionPtr = std::shared_ptr<Transaction>;
+  using ConstByteArray = byte_array::ConstByteArray;
+  using BitVector      = Transaction::BitVector;
+
+  /**
+   * Special version of builder only allowing signing. This pattern ensures that the user can not
+   * sign contents and then try and modify the transaction contents. It is also a nice placeholder
+   * generating a serial representation for the transaction.
+   */
+  struct Sealer
```
n-hutton, line 55: 👍 Really like the seal concept

libs/ledger/include/ledger/chain/v2/transaction_layout.hpp
```diff
@@ -0,0 +1,128 @@
...
+namespace v2 {
+
+class Transaction;
+
+/**
+ * A Transaction Layout is a summary class that extracts certain subset of information
+ * from a transaction. This minimal set of information should be useful only for the mining/packing
+ * of transactions into blocks.
+ */
+class TransactionLayout
```
n-hutton, line 34: Great! Much better paradigm than before, althogh don't get why it's a 'layout'

libs/ledger/include/ledger/chain/v2/transaction_layout.hpp
```diff
@@ -0,0 +1,128 @@
...
+private:
+  ConstByteArray digest_{};
+  ShardMask      mask_{};
+  TokenAmount    charge_{0};
+  BlockIndex     valid_from_{0};
+  BlockIndex     valid_until_{0};
+
+  // Native serializers
+  template <typename T>
+  friend void Serialize(T &s, TransactionLayout const &tx);
```
n-hutton, line 71: Do we not usually define the serializers in the same .hpp?

libs/ledger/include/ledger/chain/v2/transaction_serializer.hpp
```diff
@@ -0,0 +1,76 @@
...
+namespace fetch {
+namespace ledger {
+namespace v2 {
+
+class Transaction;
+
+/**
+ * The transaction serializer is one of the two methods for constructing a transaction object. This
+ * is intended to the main way that transaction are built in the system. I.e. they are received
+ * over the wire on a HTTP or similar interface.
```
n-hutton, line 31: 👍

libs/ledger/src/chain/v2/transaction_builder.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+/**
+ * Construct the sealed builder from a given transaction pointer
+ *
+ * @param tx The current partial transaction being built
+ */
+TransactionBuilder::Sealer::Sealer(TransactionPtr tx)
+  : partial_transaction_{std::move(tx)}
+{
+  // perform some sanity checks on the payload
```
n-hutton, line 45: Why not have a method on the TX like the old IsWellFormed() ?

libs/ledger/src/chain/v2/transaction_builder.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+      FETCH_LOG_DEBUG(LOGGING_NAME, "Signed: ", it->signature.ToHex(),
+                      " len: ", it->signature.size());
+      FETCH_LOG_DEBUG(LOGGING_NAME, "- Payload: ", serialized_payload_.ToHex());
+    }
+    else
+    {
+      FETCH_LOG_WARN(LOGGING_NAME, "Failed to sign transaction payload");
+    }
+  }
```
n-hutton, line 92: In the case the signatory isn't found, this still returns otherwise normally?

libs/ledger/src/chain/v2/transaction_builder.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+  // if valid, extract the transaction
+  TransactionPtr tx{};
+  if (valid)
+  {
+    // generate the final transaction
+    partial_transaction_->digest_ = hash_function.Final();
+
+    tx = std::move(partial_transaction_);
```
n-hutton, line 135: 👍

libs/ledger/src/chain/v2/transaction_serializer.cpp
```diff
@@ -0,0 +1,731 @@
...
+
+  return value;
+}
+
+ConstByteArray Encode(Address const &address)
+{
+  return address.address();
+}
+
+template <typename T>
```
n-hutton, line 77: Templates in a cpp are a bit unusual no?

n-hutton: Seems this math type stuff could go elsewhere

libs/ledger/src/chain/v2/transaction_serializer.cpp
```diff
@@ -0,0 +1,731 @@
...
+  }
+
+  if (has_valid_from)
+  {
+    buffer.Append(Encode(tx.valid_from()));
+  }
+
+  buffer.Append(Encode(tx.valid_until()));
+
+  // TODO(EJF): Increase efficiency by signaling with the charge_unit_flag
```
n-hutton: private issue

libs/ledger/src/chain/v2/transaction_serializer.cpp
```diff
@@ -0,0 +1,731 @@
...
+
+    switch (charge_unit)
+    {
+    case UNIT_MEGA:
+      tx.charge_ *= 10000000000000000ull;
+      break;
+    case UNIT_KILO:
+      tx.charge_ *= 10000000000000ull;
+      break;
+    case 0:  // unit
```
n-hutton: UNIT_DEFAULT?

libs/ledger/src/chain/v2/transaction_serializer.cpp
```diff
@@ -0,0 +1,731 @@
...
+        std::size_t const shard_mask_length_bits =
+            1u << (static_cast<std::size_t>(contract_header & 0x3fu) + 3u);
+
+        // create the mask of the correct size and decode the value
+        tx.shard_mask_.Resize(shard_mask_length_bits);
+        Decode(buffer, tx.shard_mask_);
+      }
+    }
+
+    if (contract_type == 1)
```
n-hutton: enums here?

libs/ledger/tests/chain/address_tests.cpp
```diff
@@ -0,0 +1,112 @@
...
+{
+  // create the private key
+  ECDSASigner signer;
+
+  // create the address from the identity
+  Address address{signer.identity()};
+
+  // generate our expectations
+  auto const expected_address = CreateAddress(signer.identity());
+  auto const expected_display = expected_address + CreateChecksum(expected_address);
```
n-hutton, line 88: 👍

libs/ledger/tests/chain/json_transaction_tests.cpp
```diff
@@ -0,0 +1,79 @@
...
+  // create 2 private/public key pairs
+  ECDSASigner identity1{};
+  ECDSASigner identity2{};
+
+  // create the associated addresses
+  Address const address1{identity1.identity()};
+  Address const address2{identity2.identity()};
+
+  // build the single transfer transaction
+  auto tx = TransactionBuilder()
```
n-hutton, line 49: Niiice

libs/ledger/tests/chain/transaction_serializer_tests.cpp
```diff
@@ -0,0 +1,1034 @@
...
+};
+
+// static data
+TransactionSerializerTests::Signers   TransactionSerializerTests::signers_{};
+TransactionSerializerTests::Addresses TransactionSerializerTests::addresses_{};
+
+TEST_F(TransactionSerializerTests, SimpleTransfer)
+{
+  static const std::string EXPECTED_DATA =
+      "a12400532398dd883d1990f7dad3fde6a53a53347afc2680a04748f7f15ad03cadc4d44235130ac5aab442e39f9a"
```
n-hutton, line 567: Hopefully we never change the transaction format from here on out :)

n-hutton (changes requested):
Looks great so far

libs/core/include/core/byte_array/const_byte_array.hpp
ejfitzgerald: I don't want to fix #257 on this commit.

This mask is a GCC platform issue and occurs in a bit of weird inlining corner case. Essentially it is a GCC 7 internal call inside `memcpy` is signed. Switching this to a signed integer does not fix the issue because the compiler complains at the type

This is non-great, but have run out of better ideas to solve this issue - The current solution works because during the inline process the compiler can see this value is limited to a 63bit number

libs/ledger/include/ledger/chain/v2/address.hpp
ejfitzgerald: I hadn't come across them before interesting...

libs/ledger/src/chain/v2/transaction_builder.cpp
ejfitzgerald: I did think of that, in general I didn't include it because I wanted a hard fail when you try and create it. Remember this code is mostly just for tests and client code. For the general version this will always be done with the transaction serialiser

libs/ledger/src/chain/v2/transaction_serializer.cpp
ejfitzgerald: Yes, as I eluded to in the summary. I plan to refactor this out in the future (probably making a MsgPack serialiser). This is why I made this generic templated version. When I do that and it becomes clear which things will be publicly visible, it makes sense to refactor that accordingly. For the moment it is scoped to this file to ensure minimal coupling

libs/ledger/src/chain/v2/transaction_serializer.cpp
ejfitzgerald: Yeah I guess it needs a name

libs/ledger/src/chain/v2/transaction_builder.cpp
ejfitzgerald: There is however, that will be caught (with a null tx ptr) once the user completes the build step

libs/ledger/src/chain/v2/transaction_builder.cpp
ejfitzgerald: Also the general idea is that sort of by definition all transactions will be well formed, otherwise the builder or serializer should not emit them

libs/ledger/include/ledger/chain/v2/transaction_layout.hpp
ejfitzgerald: Honestly, I am not sure if we have a convention, happy to start one. I kept this separate for readability, but happy to change

libs/ledger/include/ledger/chain/v2/transaction_layout.hpp
ejfitzgerald: Happy to change, I wanted a name that did not class with Summary and reflected the fact that this is principally used for scheduling, idea?

libs/ledger/include/ledger/chain/v2/address.hpp
ejfitzgerald: I tend to agree, however, in Ethereum and others there is a-lot of inertia around the idea of creating an address. Happy to take suggestions on naming here - I am terrible at them really.

ejfitzgerald:
@n-hutton I think that I have addresses all your concerns. Can you make sure that I haven't missed anything

n-hutton (approved):

PR #886, opened 2019-04-19
Feature/main chain improvements

bipll:
This commit adds the tree of paths to MainChain. Block removal and tips indexing changed accordingly.

[Still a work in progress, perhaps.]


libs/vectorise/include/vectorise/memory/vector_slice.hpp
```diff
@@ -95,7 +96,7 @@ class VectorSlice
     assert(pointer_ != nullptr);
     if (pointer_)
     {
-      std::memset(pointer_, 0, padded_size() * sizeof(Type));
+      std::fill(pointer_, pointer_ + padded_size(), T());
```
n-hutton: Looks better although this technically might not set all zero anymore right, depending on the T constructor - what's the case for char()?

libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -192,6 +215,7 @@ class MainChain
   TipsMap          tips_;          ///< Keep track of the tips
   HeaviestTip      heaviest_;      ///< Heaviest block/tip
   LooseBlockMap    loose_blocks_;  ///< Waiting (loose) blocks
+  mutable Trails   trails_;        ///< The tree of paths from genesis down to every bloc
```
n-hutton: I understand that this can be used to look forward in the chain right?

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -128,97 +100,96 @@ MainChain::BlockPtr MainChain::GetHeaviestBlock() const
...
+  {
+    trails_.emplace(body.previous_hash, body.hash);
+    return true;
+  }
+  return false;
+}
+
+bool MainChain::UncacheBlock(BlockHash const &hash, bool root) const
+{
+  auto chainLink{block_chain_.find(hash)};
```
n-hutton: style: we avoid using camelCase

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -128,97 +100,96 @@ MainChain::BlockPtr MainChain::GetHeaviestBlock() const
...
+  {
+    if (range.first->second == hash)
+    {
+      trails_.erase(range.first);
+      break;
+    }
+  }
+}
+
+MainChain::BlockMap::iterator MainChain::UncacheBlock(BlockHash const &             hash,
```
n-hutton: I would put the two UncacheBlock functions together as they have the same name

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -128,97 +100,96 @@ MainChain::BlockPtr MainChain::GetHeaviestBlock() const
...
+  auto const &body{block->body};
+  if (block_chain_.emplace(body.hash, block).second)
+  {
+    trails_.emplace(body.previous_hash, body.hash);
+    return true;
+  }
+  return false;
+}
+
+bool MainChain::UncacheBlock(BlockHash const &hash, bool root) const
```
n-hutton: I would default root here since it's a bit confusing otherwise

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -128,97 +100,96 @@ MainChain::BlockPtr MainChain::GetHeaviestBlock() const
...
+    // take one step back and leave no trace
+    RemoveStepBack(hash, chainLink->second);
+  }
+
+  for (auto range{trails_.equal_range(hash)}; range.first != range.second;
+       range.first = trails_.erase(range.first))
+  {
+    UncacheBlock(range.first->second, false);
+  }
+  return block_chain_.erase(chainLink);
```
n-hutton: Is this iterator that's returned actually used anywhere? It seems like the two UncacheBlock functions could be one

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -128,97 +100,96 @@ MainChain::BlockPtr MainChain::GetHeaviestBlock() const
...
+}
+
 /**
  * Removes the block (and all blocks ahead of it) from the chain
  *
- * @param hash The has to be removed
+ * @param hash The hash to be removed
  * @return True if successful, otherwise false
  */
```
n-hutton, line 173: Doesn't this PR close issue 666?

libs/ledger/src/chain/main_chain.cpp
n-hutton: It's also unclear what ```root``` is supposed to mean - it seems to me like you remove the first hash in the trail, remove references to it backwards, then walk forwards on everything else?

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1128,70 +1084,54 @@ bool MainChain::DetermineHeaviestTip()
...
-  BlockHashSet tips{};
-  for (auto const &block : block_list)
+  for (auto const &pathNode : trails_)
   {
-    tips.erase(block->body.previous_hash);
-
-    // only non-loose blocks are considered tips
-    if (!block->is_loose)
+    // Find path tree leaves, they will constitute the new tips
+    if (trails_.find(pathNode.second) == trails_.end())
```
n-hutton: Style: camelCase etc.

libs/ledger/include/ledger/chain/main_chain.hpp
n-hutton: Suggestion: rename trails_ to forward_references_

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1128,70 +1084,54 @@ bool MainChain::DetermineHeaviestTip()
...
+    heaviest_.weight = largestWeight;
+    return true;
   }
 
-  // clear the existing tips
-  tips_ = std::move(new_tips);
-
-  // finally update the heaviest tip
-  return DetermineHeaviestTip();
+  return false;
```
n-hutton: this should be impossible, right?

libs/vectorise/include/vectorise/memory/vector_slice.hpp
n-hutton: this might not default construct to zero if  it's not a primitive (athough it does look better). Probably fine though. Do we know what char constructs to?

n-hutton (changes requested):
I think the changes need a doxygen and inline comments - it's hard to follow what is going on. Also, it might be more readable to avoid thing.first->second etc. and rather split it out and name it with auto: ``` auto hash_from_to_pair = thing.first; auto hash_to_remove = hash_from_to_pair->second; ``` - just a thought.

n-hutton:
I almost forgot: I don't think the vendor/benchmark submodule should be changing here

libs/vectorise/include/vectorise/memory/vector_slice.hpp
ejfitzgerald: I would not make this change to be honest, I agree it is nice, however, I believe `std::fill` is not as performant as `memset`. That would need to be checked. In general this class should only be used with Pod data types

libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -203,7 +227,7 @@ class MainChain
  * @return: bool whether the starting hash referred to a valid block on a valid chain
  */
 template <typename T>
-bool MainChain::StripAlreadySeenTx(BlockHash starting_hash, T &container) const
+inline bool MainChain::StripAlreadySeenTx(BlockHash starting_hash, T &container) const
```
ejfitzgerald: Sort of a style thing but we are not adding `inline` to templates (because they are implicitly `inline`

ejfitzgerald (changes requested):

libs/ledger/src/chain/main_chain.cpp
bipll: Yes. `root` should probably be better called `erase_trace_back` or something, this is merely an indicator that the backstep should be removed as well, which is only done for the root deleted block, as trails to its progeny are much simpler deleted.

libs/ledger/src/chain/main_chain.cpp
bipll: Most likely, that is why I didn't specify a particular issue no. in PR's title. I will add `closes` clauses to the body.

libs/vectorise/include/vectorise/memory/vector_slice.hpp
bipll: This is the difference between default and value-initialization: objects of simple types are set to zero. Thus `char()` is `'\0'` and `bool()` is actually `false`.

bipll: I've made changes, should look better now, IMO.

libs/ledger/src/chain/main_chain.cpp
bipll: Yes, it reiterates the (im)possible false return from `DetermineHeaviestTip()`.

Issue #873, opened 2019-04-16
Transaction sync fixes


wkaluza:
Transaction syncing fails when attempting to cache transactions. Re-enable caching and rewrite syncing algorithm to improve performance.

n-hutton:
This bug can be replicated in the following way:

check out fetch-ledger nhutton fork:

nathan/bug/transaction_sync_cache

check out ledger-api-py nhutton fork (https://github.com/n-hutton/ledger-api-py/):

nathan/bug/transaction_sync_cache

https://github.com/n-hutton/ledger-api-py/

Build constellation.

In one folder, start a constellation node:

```
rm *.db; ./constellation -port 8000 -lanes 1 -standalone -block-interval 1000
```

Wait until it is mining blocks. 

Send TXs to the node using the ledger api (this needs installing, use the README for this)

```
python3 ./examples/send_transactions_in_bulk.py
```

Once they are sent and executed, start the second node (in a different folder):

```
rm *.db; ./constellation -port 8020 -lanes 1 -peers tcp://127.0.0.1:8001
```

You should see the message being printed by the first node: ``` [ 2019-04-17 00:15:22.527, #10 WARN  :             ObjectStoreSyncProtocol ] Returning object count of: 0 ``` which is incorrect. This behaviour appears to have either been fixed or hidden in subsequent commits, so the file libs/storage/include/storage/transient_object_store.hpp has been cherry picked from the time the bug was first seen. This file appears to cause it.

The transient object store acts as a wrapper around the object store and holds recent transactions in a buffer - ed wrote this I believe.

Note, the way I would probably debug this is using breakpoints in clion and stepping through the code. If you do this, clion will build constellation in debug mode which changes mutexes in the code to a mutex that will SIGKILL if it's locked for too long (a useful thing to find in debug mode). To disable this so it doesn't trigger during stepping through code, change the following in mutex.hpp: 

```
#ifdef NDEBUG
using Mutex = ProductionMutex;
#else
using Mutex = ProductionMutex;
#endif
```

PR #887, opened 2019-04-23
Fix race conditions in TxSyncStoreProtocol

wkaluza:
Working towards resolving #873 Nathan and I found some race conditions. These should now be fixed. Work on #873 continues.


libs/ledger/include/ledger/storage_unit/tx_list.hpp
```diff
@@ -0,0 +1,28 @@
...
+//
+//------------------------------------------------------------------------------
+
+#include "ledger/chain/transaction.hpp"
+#include <vector>
+
+namespace fetch {
+namespace ledger {
+
+using TxList = std::vector<fetch::ledger::UnverifiedTransaction>;
```
ejfitzgerald: There are actually a number of places where we are using a vector of Transaction. Does it make sense to locate this type alias in the same header as `UnverifiedTransaction` (which will be removed shortly.

libs/storage/include/storage/transient_object_store.hpp
```diff
@@ -104,6 +106,38 @@ class TransientObjectStore
...
+template <typename O>
+std::size_t TransientObjectStore<O>::Size() const
+{
+  FETCH_LOCK(cache_mutex_);
+
+  return archive_.size() + cache_.size();
+}
+
+template <typename O>
+ledger::TxList TransientObjectStore<O>::PullSubtree(byte_array::ConstByteArray const &rid,
```
ejfitzgerald, line 118: I think that longer term it makes sense to move this class from the storage library into the ledger library. There are too many cross dependencies i.e. TxList. CMake handles that at the moment and everything resolves at link time but it is non great. I don't suggest that we do this now, however, can you raise an issue to track this please.

libs/ledger/src/storage_unit/transaction_store_sync_protocol.cpp
```diff
@@ -112,8 +111,7 @@ uint64_t TransactionStoreSyncProtocol::ObjectCount()
 {
   FETCH_LOCK(cache_mutex_);
 
-  // TODO(private issue 502): Improve transient object store interface
-  return store_->archive().size();
+  return store_->Size();
```
ejfitzgerald, line 114: 👍 👍 👍

libs/storage/include/storage/transient_object_store.hpp
```diff
@@ -305,7 +353,6 @@ bool TransientObjectStore<O>::Confirm(ResourceID const &rid)
 template <typename O>
 bool TransientObjectStore<O>::GetFromCache(ResourceID const &rid, O &object)
```
ejfitzgerald, line 357: Worth a quick note in the docs to say that the user is expected to lock the `cache_mutex_` from outside now? (+ the next 2 functions)

ejfitzgerald (changes requested):
Looks good to me. I only have a small request to update some docs and a small comment. Nice one

ejfitzgerald (approved):

PR #888, opened 2019-04-23
(issue #799) VM: Persistent Map implementation

### LINKED ISSUE #799: VM: Persistent Map implementation

pbukva:
Closes #799


libs/ledger/tests/chaincode/smart_contract_tests.cpp
```diff
@@ -69,14 +61,11 @@ class SmartContractTests : public ContractTest
 protected:
   void CreateContract(std::string const &source)
   {
-    // calculate the digest of the contract
-    auto const contract_digest = ToBase64(DigestOf(source));
-
     // generate the smart contract instance for this contract
-    contract_ = std::make_unique<SmartContract>(source);
-
+    auto contract = std::make_shared<SmartContract>(source);
```
n-hutton: why not inline?

libs/ledger/tests/chaincode/smart_contract_tests.cpp
```diff
@@ -551,4 +540,96 @@ TEST_F(SmartContractTests, CheckBasicTokenContract)
   }
 }
 
+TEST_F(SmartContractTests, CheckPersistentMapSetAndQuery)
+{
+  std::string const contract_source = R"(
+    @action
+    function test_persistent_map()
+      var state = PersistentMap<String, Int32>("value");
```
n-hutton, line 392: out of interest, what other than string and primitives is supported?

libs/ledger/tests/chaincode/smart_contract_tests.cpp
```diff
@@ -551,4 +540,96 @@ TEST_F(SmartContractTests, CheckBasicTokenContract)
...
+  // check the registered handlers
+  auto const transaction_handlers = contract_->transaction_handlers();
+  ASSERT_EQ(1u, transaction_handlers.size());
+  EXPECT_TRUE(IsIn(transaction_handlers, "test_persistent_map"));
+
+  // check the query handlers
+  auto const query_handlers = contract_->query_handlers();
+  ASSERT_EQ(2, query_handlers.size());
+
+  // define our what we expect the values to be in our storage requests
```
n-hutton: suggest : remove ``` our ```

libs/ledger/tests/chaincode/smart_contract_tests.cpp
```diff
@@ -551,4 +540,96 @@ TEST_F(SmartContractTests, CheckBasicTokenContract)
...
+
+  // query "bar"
+  {
+    Variant request = Variant::Object();
+    Variant response;
+    EXPECT_EQ(SmartContract::Status::OK, SendQuery("query_bar", request, response));
+
+    // check the response is as we expect
+    ASSERT_TRUE(response.Has("result"));
+    EXPECT_EQ(response["result"].As<int32_t>(), 30);
```
n-hutton: 👍

libs/vm/include/vm/defs.hpp
```diff
@@ -903,20 +903,23 @@ struct Variant
...
-struct TemplateParameter1 : public Variant
-{
-  using Variant::Variant;
-};
-
-struct TemplateParameter2 : public Variant
-{
-  using Variant::Variant;
-};
+using TemplateParameter  = TemplateParameterT<0>;
```
n-hutton: can we have a comment as to why there are 3 template parameter types?

libs/vm/include/vm/persistent_map.hpp
```diff
@@ -0,0 +1,164 @@
...
+    case TypeIds::String:
+      key = key_v.Get<Ptr<String>>();
+      break;
+
+    case TypeIds::Address:
+      key = key_v.Get<Ptr<Address>>()->AsBase64String();
+      break;
+
+    default:
+      RuntimeError("Unexpected type of key value. It must be either String or Address.");
```
n-hutton, line 77: Aren't primitives handled? Does this mean that for now, primitives are using the State class and non-primitives are using this StateMap?

libs/vm/include/vm/persistent_map.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+    Variant &  key_v = Pop();
+    auto const key{ExtractKey(key_v)};
+    key_v.Reset();
+    if (!key)
+    {
+      RuntimeError("Extraction of key value failed.");
+      return;
+    }
+    // std::cout << "Resource would be: " << key->str << std::endl;
```
n-hutton: dead code

libs/vm/include/vm/persistent_map.hpp
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    Variant &value_v = Pop();
+    if (value_type_ != value_v.type_id)
+    {
+      RuntimeError("Incorrect value type for PersistentMap<...> type.");
+    }
+    else
+    {
+      // std::cout << "Resource would be: " << key->str << std::endl;
```
n-hutton: dead code etc.

libs/vm/include/vm/persistent_map.hpp
```diff
@@ -0,0 +1,164 @@
...
+      auto state{
+          IState::ConstructIntrinsic(vm_, TypeIds::IState, value_type_, key, TemplateParameter{})};
+      state->Set(TemplateParameter{value_v});
+    }
+    value_v.Reset();
```
n-hutton: doesn't look like this variant reset is neccessary

libs/vm/include/vm/persistent_map.hpp
```diff
@@ -0,0 +1,164 @@
...
+#include "vm/vm.hpp"
+
+namespace fetch {
+namespace vm {
+
+class IPersistentMap : public Object
+{
+public:
+  // Factory
+  static Ptr<IPersistentMap> Constructor(VM *vm, TypeId type_id, Ptr<Object> const &name);
```
n-hutton: this is declared only while other fns are defined also - suggest to define all fns for this class and PersistentMap at the bottom of the file as Ed usually does

libs/vm/include/vm/persistent_map.hpp
```diff
@@ -0,0 +1,164 @@
...
+  {}
+
+  ~IPersistentMap() override = default;
+
+protected:
+  Ptr<String> name_;
+  TypeId      value_type_;
+};
+
+class PersistentMap : public IPersistentMap
```
n-hutton, line 50: Not sure how this works - seems to involve pushing and popping to the vm in some way - could do with commenting

libs/vm/include/vm/state.hpp
```diff
@@ -224,11 +256,18 @@ inline Ptr<IState> IState::Construct(VM *vm, TypeId type_id, Args &&... args)
   {
     return new State<Address>(vm, type_id, value_type_id, std::forward<Args>(args)...);
   }
-  default:
-  {
-    throw std::runtime_error("Unsupported State type");
-  }
   }  // switch
+
+  vm->RuntimeError("Unsupported state type");
```
n-hutton: suggest: Unsupported type used when accessing state

apps/vm-lang/main.cpp
```diff
@@ -329,5 +329,10 @@ int main(int argc, char **argv)
     state_map.SaveToFile(data_path.c_str());
   }
 
+  if (!success)
+  {
+    std::cout << error << std::endl;
```
wkaluza, line 338: Shouldn't this be

```suggestion
    std::cerr << error << std::endl;
```

?

libs/vm/src/defs.cpp
```diff
@@ -112,159 +112,159 @@ size_t Object::GetHashCode()
...
 void Object::PushElement(TypeId /* element_type_id */)
 {
-  RuntimeError("operator not implemented");
+  RuntimeError("PushElement operator not implemented");
 }
 
 void Object::PopToElement()
 {
-  RuntimeError("operator not implemented");
+  RuntimeError("PopToElement operator not implemented");
```
n-hutton: 👍

n-hutton (changes requested):

libs/ledger/tests/chaincode/smart_contract_tests.cpp
```diff
@@ -551,4 +540,96 @@ TEST_F(SmartContractTests, CheckBasicTokenContract)
...
+      .WillRepeatedly(Return(fetch::storage::Document{expected_value1}));
+  EXPECT_CALL(*storage_, Get(expected_resource2))
+      .Times(2)
+      .WillRepeatedly(Return(fetch::storage::Document{expected_value2}));
+
+  // send the smart contract an "increment" action
+  EXPECT_EQ(SmartContract::Status::OK,
+            SendActionWithParams("test_persistent_map", {"value.bar", "value.foo"}));
+
+  // query "foo"
```
wkaluza: Could we deduplicate this and the following block? E.g. extract to a function. Looks like they have most of their code in common.

libs/vm/include/vm/defs.hpp
```diff
@@ -903,20 +903,23 @@ struct Variant
...
 
-struct TemplateParameter : public Variant
+template <std::size_t ID = 0>
+struct TemplateParameterT : public Variant
 {
   using Variant::Variant;
+  TemplateParameterT(TemplateParameterT const &) = default;
+  TemplateParameterT(TemplateParameterT &&)      = default;
+  TemplateParameterT(Variant const &other)
+    : Variant{other} {};
```
wkaluza: Unnecessary semicolon, ditto below.
```suggestion
    : Variant{other} {}
```

libs/vm/include/vm/defs.hpp
```diff
@@ -903,20 +903,23 @@ struct Variant
...
+  TemplateParameterT(Variant &&other)
+    : Variant{std::move(other)} {};
```
wkaluza:
```suggestion
    : Variant{std::move(other)} {}
```

wkaluza (changes requested):
Just a few small niggles from me.

libs/ledger/tests/chaincode/smart_contract_tests.cpp
pbukva: The `PersistentMap<K,T>` class is just thin wrapper around already existing `State<T>` class, and forwards all the work down to that class. Thus implication is that it can support only what `State<T>` supports, what is that:

 * `T` (value type) can be only primitive numeric types + this PR extends `State<T>` class with support for `String` and `Address`
 * `K` (key type) can be `String` + this PR adds support for `Address`. This is because value of key (of `K` type) is simply used to be concatenated with `name` parameter of `State<T>(name : String)` constructor.

libs/ledger/tests/chaincode/smart_contract_tests.cpp
pbukva: This is because I need to have access to `SmartContract` type on order to have access to `contract_digest()` method 2 lines bellow.
The `contract_` class data member is of `Contract` type which is **base** type and does **not** have `contract_digest()` method.

libs/vm/include/vm/defs.hpp
pbukva: This was here before, I just changed the internals.

libs/vm/include/vm/persistent_map.hpp
pbukva: This is because o readability: Implementation of this method is bigger then other methods(constructors) which basically just empty shells.
This way you nicely actually see interface of the class without being "polluted" by large implementation details which are done somewhere else.

You can see the same approach on many other places in our code base (e.g. `state.hpp`, etc. ...).

libs/vm/include/vm/persistent_map.hpp
pbukva: This is just for KEY type.
The `PersistentMap<KEY, T>` is just thin wrapper around `State<T>(name)` where we **restricted** `name` to be of `String` type (+ I extend it by `Address` type, which is, at the end, just converted to `String` using Base64 encoding).
In our concept, value of `key` from `PeristentMa`  will be just concatenated with `name` (`name = name + "." + key`).

libs/vm/include/vm/persistent_map.hpp
pbukva: It is, based on current concept of VM (specifically how VM handles stack).

libs/vm/include/vm/state.hpp
pbukva: This is original code, I just relocated it f lines bellow.
The current message in the code is quite on spot - it says that `T` type of `State<T>` is unsupported. In this context the `T` type really is what whole `State<T>` stands for, and thus current formulation is really good (it is good enough to be self-explaining and mainly it is short).

If you want to be very explicit with the message, than the better formulation could be "Unsupported type passed as template parameter to State<...>", but it is longer so not as elegant as the current one.

apps/vm-lang/main.cpp
ejfitzgerald: I am sure that I made this change to `std::cerr` previously? 😕

libs/core/include/core/logger.hpp
```diff
@@ -433,7 +433,7 @@ class LogWrapper
   }
 
   template <typename... Args>
-  void DebugWithName(char const *name, Args const &... args)
+  void DebugWithName(char const *name, Args &&... args)
```
ejfitzgerald, line 436: If you are making this change here, it also make sense to change the other logging lmethods

libs/ledger/tests/chaincode/mock_storage_unit.hpp
```diff
@@ -21,7 +21,7 @@
 
 #include <gmock/gmock.h>
 
-class MockStorageUnit final : public fetch::ledger::StorageUnitInterface
+class MockStorageUnit : public fetch::ledger::StorageUnitInterface
 {
 public:
   MockStorageUnit()
```
ejfitzgerald, line 27: After all our discussions I though you were parameterising the `ON_CALL` setup for this mock object?

libs/storage/include/storage/resource_mapper.hpp
```diff
@@ -157,9 +157,8 @@ class ResourceAddress : public ResourceID
 public:
   explicit ResourceAddress(byte_array::ConstByteArray const &address)
     : ResourceID(crypto::Hash<crypto::SHA256>(address))
-  {
-    address_ = address;
-  }
+    , address_{address}
+  {}
```
ejfitzgerald, line 161: 👍

libs/vm/include/vm/defs.hpp
ejfitzgerald: Yeah I agree this needs to information I am not even sure what the score is here. I would recommend that we don't do this in this review and wait for @robertdickson latest and greatest VM release

libs/vm/include/vm/persistent_map.hpp
ejfitzgerald: String keys makes sense for the short to long term, it is a limitation but follows the existing limitations of the object.

libs/vm/src/defs.cpp
ejfitzgerald: 👍 👍

ejfitzgerald (changes requested):

libs/storage/include/storage/resource_mapper.hpp
wkaluza: Doesn't this mean that address_ may be const now? A const ConstByteArray, even?

libs/ledger/tests/chaincode/smart_contract_tests.cpp
n-hutton: makes sense

libs/vm/include/vm/persistent_map.hpp
n-hutton: could you expand on that? It looks like I misread: I thought the variant would destruct here but actually Pop returns a reference? Shouldn't it be Peek instead?

libs/vm/include/vm/state.hpp
n-hutton: I would actually prefer your longer message :)

libs/core/include/core/logger.hpp
pbukva: ?
I do not follow.
All other logging methods are alredy using moving semantics (and they must, based on concept we use all over logging) - this is just fix. Without this fix compilation fails when you set logging severity level at compile time do `debug`.

libs/ledger/tests/chaincode/mock_storage_unit.hpp
pbukva: This change here (dropping `final`) is for something else. It is for enabling to make `::testing::StrictMock<MockStorageUnit>` out of this class.

During our discussion, one important fact got lost - If expectation is **not** set(decalred) for specific call(s), then yes you will get message printed out, but that is it, no error, test will **NOT** fail.

* And this is specially critical in CI.
* Or when people are runnig whole test suite locally using `./scripts/ci-tool.py` where you won't see any printout except slim ctest single-line-per-test overview.
* It also cirical when developing when you can see the printout from the tests run  - however, the reality is that if the printout is long (e.g. a lot of tests executed or a lot of **normal/debug** printout is spit), such warning messages will be scrolled outside of your terminal view (and drowned in between other stuff = hard to catch by eye when scrolling through) with test results summary showing all passed.

That is where `::testing::StrictMock<MockStorageUnit>` comes in - it will ensure that test run will actually FAIL if there is unmet expectation.

libs/storage/include/storage/resource_mapper.hpp
pbukva: yes, but only partially -  yes for `ConstByteArray` but no for `const`.

The reasons being:
 * because of suporting **deserialisation**
 * as the implication of above, this class needs **non**-parametric constructor + this class is dervived from base class which has such non-param. constructor as well due to same reasons.

libs/vm/include/vm/defs.hpp
pbukva: 1) First of all, this does **not** change anything on API level - very same types existed before this change

2) I did this change in order to be able to construct `TemplateTypeX` instance from it's base `Variant` class intance. The `using Variant::Variant;` (forwarding consturctors to derived class) doesn't do justice to original intention, it is actually necessary to define `TemplateTypeX::TemplateTypeX(Variant const &)` and `TemplateTypeX::TemplateTypeX(Variant &&)` constructors so one can actually use the types as they we originally intended to be used but it didn't work out because constructor forwarding in C++11,14 does not work that simple.
So obviously as implication of that fact & that we have 3 of such `TemplateTypeX` types, I didn't want to duplicate the code in 3 different places = > I created single general class.

3) **Purpose** of these types are to denote template parameters of parametrised types in our VM type-system. As the things stand now on conceptual/design level, our VM type-system only supports parametrised types with **max** 2 template type parameters (for example `Array<type1>`, `Map<type1, type2>`, but something like `MyNewType<type1, type2, TYPE3>` is **not** supported yet - please see the limitation in source code - the `Analyser::ConvertType(...)` method (in `libs/vm/src/analyser.cpp` file) + definitions of **TypeId** constants (in the `libs/vm/include/vm/typeids.hpp`).

The general templated class `TemplateTypeT<...>` can be actually simplified by dropping 2 lines carying unnecessary declarations of default constructors.

libs/vm/include/vm/persistent_map.hpp
troels-im: The `Pop` function takes an element of the stack and hence name is correct. `Reset` nulls the smart pointer, also correct. I think this is a non issue?

pbukva: The reason for this is that, when VM is interpreting the code, it creates stack of variables used in the code in the exactly predefined order by the etch code (e.g. it smashes in to the stack return value of the function, function input parameters, internal variables inside of function, etc. ...).
Once the stack is initialised with variables for the **current** frame of execution, you can then accessthe stack in native VM code by using `Pop()` and `Push()`.

==> It is essential to understand here that every operation execution (e.g. calling etch function or operator like +,-, [...], etc. ...) is done as **separate frame** - like it would be a function call (actualy exactly as it is in C++), thus previously defined variables are "copyed" and passed to the actual frame as **new** variable (though VM use `Variant` for this purpose so for `object` types variant internally uses its own `vm::Ptr<T>` breed of smart pointer with reference counting). 
This is very essential, because now we have **at least** 2 "copies" of original variable (actually, to be precise, as many copies as number(depth) of nested frames the variable needs to be passed through).
==> And so obvious implication here is that when we are leaving the frame, we need to indicate that this specific(current) "copy" of the variable is not more necessary (the current frame does **not** need it anymore) - this is relevant only for `object` (or better o say `refernce`) types - look in to `Variant::Reset()` - is checks if it's value it represents is object type and if it is then it is held internally as smart pointer `Ptr<object>`, and Variant:Reset simply calls Ptr::Reset() to decrement the reference count.
There was made the design decission that `Pop()` and `Push()` will return **reference** 
 to variant instance on stack (`Variant&`) where the bovious C++ implication is that destructor is **not** called for reference variable => we need to call `Variant::Reset()` manually. Yes we can create dedicated guard class (like for `std::mutex` for example), which would do this for us.



Pop is used for accessing variable which conceputially represent  the "right" side operand in an operation (nicely depicted on example `a = b;` where `b` variable is the "right" operand (or input), so in this case it is expected that value of that "right" operand variable has been already set by previous operation (previous etch code) and so we are using it as input and so one returned by `Pop()` index in the stack is decremented by one (that variable becomes no more necessary, though still left existing in the stack). 
The `Push()` is used for getting reference to "left" side operand (the `a` variable in our example) - in this case it is expected that value of that variable will be modified (as result of the operation).

Whole this Pop/Push concept is quite a bit convoluted, and non-intuitive.

This is all over the VM code, if I'm going to comment it here in this piece of code, then what about all other places.

Please don't take me wrong, I'm not saying VM code wouldn't need (quite) a few hints in form of either using better naming of methods (= what would give more intuitive/common-sense reading of the code) or adding comments in to the code.
But if we are boing to talk about adding comments here simply does not make sense considering how the rest of the VM written.

robertdickson: Peter, it's not un-intuitive. Push() pushes, Pop() pops, they both return references so you can work with the pushed or popped element and reset when done with it. The alternative would be doing move constructors constantly for no benefit. The stack frame already exists and elements are constructed in-place for maximum effeciency. But in any event, Push() and Pop() are private to the VM where they are optimised for fast operation -- they have leaked out into code you are worrying about only because setting up indexing on classes was not complete previously. In the new version of the VM that will land soon, there is no longer any need for code that is external to the VM to be doing Pop() as that is handled for you in the same style as it is done for Module-created free functions, say.

ejfitzgerald:
Closing for action next week

PR #892, opened 2019-04-24
Refactor TransientObjectStore to internally use StateMachine

wkaluza:
State management in TransientObjectStore was done using a rather cluttered loop and a worker thread. This PR is a refactoring to a finite state machine driven by a reactor owned by LaneService.


cmake/BuildTools.cmake
```diff
@@ -119,7 +119,7 @@ function(add_fetch_test name library directory)
     if ("SLOW" IN_LIST ARGV)
       set(test_label "Slow")
       fetch_warning("Slow Test: ${name}")
-    elseif("INTEGRAION" IN_LIST ARGV)
+    elseif("INTEGRATION" IN_LIST ARGV)
```
ejfitzgerald, line 122: 👍

libs/storage/include/storage/transient_object_store.hpp
```diff
@@ -141,30 +179,98 @@ ledger::TxList TransientObjectStore<O>::PullSubtree(byte_array::ConstByteArray c
...
+typename TransientObjectStore<O>::Phase TransientObjectStore<O>::OnPopulating()
 {
-  thread_ = std::make_shared<std::thread>([this] { ThreadLoop(); });
+  assert(extracted_count < batch_size_);
+
+  // ensure the write count is reset
+  written_count = 0;
+
+  // attempt to extract an element in the confirmation queue
+  bool const extracted = confirm_queue_.Pop(rids[extracted_count], max_wait_interval_);
```
ejfitzgerald: So one consideration here is that this call will block for up to `max_wait_interval_` which is current 2000ms. This is fine when we have a dedicate thread for this procedure, however, now that this is a state machine, this is being run by the reactor. By default the reactor only has a single thread. This means that all other state machines attached to the reactor would be affected (this isn't the case in currently but probably will be in the future).

I think that I would probably rework this timeout maybe even set it to `0`. You can also use the `Delay` method on the state machine in order to postpone the next state execution by a period of time.

ejfitzgerald (changes requested):

libs/storage/include/storage/transient_object_store.hpp
n-hutton: Oh, that's partly my fault : I told wkaluza that the reactor would have a low number of threads, not one

libs/storage/include/storage/transient_object_store.hpp
```diff
@@ -188,42 +189,55 @@ typename TransientObjectStore<O>::Phase TransientObjectStore<O>::OnPopulating()
   // ensure the write count is reset
   written_count = 0;
 
-  // attempt to extract an element in the confirmation queue
-  bool const extracted = confirm_queue_.Pop(rids[extracted_count], max_wait_interval_);
-
-  // update the index if needed
-  if (extracted)
+  while (true)
```
ejfitzgerald, line 192: Don't change this now, but in future I would recommend forever loops done as `for(;;)` the reason is that on some compilers they will complain that the loop condition is always true 🤦‍♂️

libs/storage/include/storage/transient_object_store.hpp
```diff
@@ -188,42 +189,55 @@ typename TransientObjectStore<O>::Phase TransientObjectStore<O>::OnPopulating()
...
   {
-    ++extracted_count;
-  }
+    // attempt to extract an element in the confirmation queue
+    bool const is_end_of_sequence =
```
ejfitzgerald: This name is confusing and in the end of sequence case this will actually be `false` so that will be confusing. Unless I am missing something  it is probably worth a second look

libs/storage/include/storage/transient_object_store.hpp
```diff
@@ -188,42 +189,55 @@ typename TransientObjectStore<O>::Phase TransientObjectStore<O>::OnPopulating()
...
+        return Phase::Writing;
+      }
+      else
+      {
+        break;
+      }
+    }
   }
 
+  state_machine_->Delay(std::chrono::milliseconds(1000u));
```
ejfitzgerald, line 226: I think that I would only execute the delay in the "end of sequence case"

ejfitzgerald (changes requested):

ejfitzgerald (approved):

PR #895, opened 2019-04-25
Fix for sfinae in math::log

pbukva:


markos (approved):
Looks good and works with the FixedPoint PR.

PR #897, opened 2019-04-25
Math tests rework

baykaner:
- re-enabling previously disabled kernels to be used in upcoming SIMD implementations

- simplifies coverage report generation by:
    - consolidating math tests into one target
    - adding a test that includes everything in the math library


libs/math/tests/math/all_tests/include_math_lib/include_math_lib_for_coverage.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include "math/trigonometry.hpp"
+#include "math/type.hpp"
+
+#include <gtest/gtest.h>
+
+// This dummy test includes everything in the math library to ensure
+// that coverage reports also show functionality that is completely
+// untested
+TEST(dummy_test, dummy_test)
+{}
```
pierre-wilmot-fetch-ai: As far as my understanding of compillers goes, this will not have the desired effect.

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
```diff
@@ -0,0 +1,237 @@
...
+    auto tmp_val_2 = gen.AsDouble();
+
+    a = TypeParam(tmp_val_1);
+    b = TypeParam(tmp_val_2);
+    fetch::math::Add(a, b, ret);
+    EXPECT_EQ(ret, a + b);
+
+    EXPECT_EQ(fetch::math::Add(a, b), a + b);
+  }
+}
```
pierre-wilmot-fetch-ai, line 141: IMO this kind of tests is not doing the job properly.
There are many situations where things can go wrong:

+ Test is never ran with negative values.
+ If your random generator starts spitting out 0 everytime, You'r only ever testing for 0 + 0 = 0 (happens for interger since all the generated values are [0-1])
+ It's very likely that your operator+ will be implemented as a call to fetch::Math::Add, so your efctivly testing that whatever comes out of fetch::math::Add is equal to whatever comes out of fetch::math::Add, without checking that the actual value is right 
+ Probably slightly inneficient to run in a loop when sample ~10 pair of number would work

I'd recommend explicitly testing values with:
Positive + Positive
Positive + Negatice
Negative + Positive
Negative + Negative
VeryLarge + VerySmall (permutation of positive / negative)
VerySmall  + VeryLarge (permutation of positive / negative)
VeryLarge + VeryLarge (permutation of positive / negative)
VerySmall + VerySmall (permutation of positive / negative)

This appies for all the test in this file

libs/math/tests/math/all_tests/combinatorics/combinatorics.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+#include <gtest/gtest.h>
+#include <iomanip>
+#include <iostream>
+
+#include "math/combinatorics.hpp"
+//#include <math/tensor.hpp>
+//
+// using namespace fetch::math::combinatorics;
+//
```
pierre-wilmot-fetch-ai: Commented out code to re-enable or remove

libs/math/include/math/kernels/standard_functions.hpp
```diff
@@ -0,0 +1,632 @@
...
+    y = std::remquo(x);
+  }
+};
+
+template <typename type>
+struct Fma
+{
+  void operator()(type const &x, type &y) const
+  {
+    y = std::fma(x);
```
pierre-wilmot-fetch-ai: Looking at the docs, the signature for fma doesn't match : https://en.cppreference.com/w/cpp/numeric/math/fma
I assume all the function in this code are unused / uncompilled / unetested.

Please remove for now and add back in a future PR along with tests (preferably do that when we actually need those).

pierre-wilmot-fetch-ai (changes requested):

libs/math/tests/math/all_tests/include_math_lib/include_math_lib_for_coverage.cpp
baykaner: I did try it out and it seemed like it worked. Honestly I don't like this at all, but I'm not aware of a better way to do this: @n-hutton had any brilliant ideas since we last spoke about this?

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
baykaner: excellent points; I'll rework it

libs/math/include/math/kernels/standard_functions.hpp
baykaner: yup, just FYI I'm pretty sure some of these kernels were added by a script, hence why sometimes the parameters don't match and you're right clearly many are not yet tested and should be.
I'll review these and add tests or remove as necessary

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
```diff
@@ -0,0 +1,585 @@
...
+    b = TypeParam(100 - i);
+    fetch::math::Add(a, b, ret);
+    EXPECT_EQ(ret, 100);
+    EXPECT_EQ(fetch::math::Add(a, b), 100);
+  }
+
+  // test a few big random numbers
+  for (std::size_t i{0}; i < 100; ++i)
+  {
+    // values range from 0 - std::numeric_limits<Type>::max()
```
pierre-wilmot-fetch-ai, line 90: This should actually lead to overflow and fail 25% of the time.
Let's run it to make sure it does (as a sanity check) and fix the test

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
```diff
@@ -0,0 +1,585 @@
...
+    b = TypeParam{100 - i};
+    fetch::math::Add(a, b, ret);
+    EXPECT_EQ(ret, 100);
+    EXPECT_EQ(fetch::math::Add(a, b), 100);
+  }
+
+  // test a few big numbers
+  for (int i{0}; i < 100; ++i)
+  {
+    // values range from - half max to max
```
pierre-wilmot-fetch-ai: Range in comment doesn't reflect the code.
[-max/2 , max/2]

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
```diff
@@ -0,0 +1,585 @@
...
+
+    fetch::math::Add(a, b, ret);
+    EXPECT_EQ(ret, TypeParam(100.0));
+    EXPECT_EQ(fetch::math::Add(a, b), TypeParam(100.0));
+  }
+
+  // test a few big numbers
+  for (int i{0}; i < 100; ++i)
+  {
+    // values range from - half max to max
```
pierre-wilmot-fetch-ai: same

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
```diff
@@ -0,0 +1,585 @@
...
+
+  // test a few big random numbers
+  for (std::size_t i{0}; i < 100; ++i)
+  {
+    // values range from 0 - std::numeric_limits<Type>::max()
+    a = TypeParam(gen.AsDouble() * double(max_val));
+    b = TypeParam(gen.AsDouble() * double(max_val));
+
+    fetch::math::Subtract(a, b, ret);
+    EXPECT_EQ(ret, a - b);
```
pierre-wilmot-fetch-ai, line 229: This will break if the random generator gives b > a
(Please run it to make sure it does and then fix the test)

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
```diff
@@ -0,0 +1,585 @@
...
+    b = TypeParam{100 + i};
+    fetch::math::Subtract(b, a, ret);
+    EXPECT_EQ(ret, 100);
+    EXPECT_EQ(fetch::math::Subtract(b, a), 100);
+  }
+
+  // test a few big numbers
+  for (int i{0}; i < 100; ++i)
+  {
+    // values range from - half max to max
```
pierre-wilmot-fetch-ai: Same

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
```diff
@@ -0,0 +1,585 @@
...
+    b = TypeParam(100);
+    fetch::math::Multiply(b, a, ret);
+    EXPECT_EQ(ret, 100 * i);
+    EXPECT_EQ(fetch::math::Multiply(b, a), 100 * i);
+  }
+
+  // test a few big random numbers
+  for (std::size_t i{0}; i < 100; ++i)
+  {
+    // values range from 0 - std::numeric_limits<Type>::max()
```
pierre-wilmot-fetch-ai, line 352: That will definitly overflow

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
```diff
@@ -0,0 +1,585 @@
...
+class FundamentalOperatorsNonIntTypeTest : public ::testing::Test
+{
+};
+
+static fetch::random::LinearCongruentialGenerator gen;
+
+using MyUnsignIntegralTypes = ::testing::Types<std::uint32_t, std::uint64_t>;
+TYPED_TEST_CASE(FundamentalOperatorsUIntTypeTest, MyUnsignIntegralTypes);
+
+using MySignedIntegralTypes = ::testing::Types<std::int32_t, std::int64_t>;
```
pierre-wilmot-fetch-ai, line 50: Lets add a fetch::fixed_point::FixedPoint<32, 0> in there for fun 😝 
Hopefully shouldn't create problem

pierre-wilmot-fetch-ai (changes requested):

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
baykaner: So, I'm assuming that by overflow you're talking about a loss of precision (since we're talking about multiplying doubles).

I don't really see what you mean by failing 25% of the time?
for example lets print some values:
std::cout << std::setprecision(20) << "max_val: " << max_val << std::endl;
>> 4294967295
std::cout << std::setprecision(20) << "double(max_val): " << double(max_val) << std::endl;
>> 4294967295
no problem for unsigned int.
For unsigned int int there is actually a loss of precision:
18446744073709551615
18446744073709551616
This means there's an approximately 1/184467440737095516.16 chance that one of my 100 iterations will overflow (i.e. if gen.AsDouble() returns 1.0, we will get a = 0). In fact, that will happen precisely never because the random generator is implemented to have a fixed seed; but yep it's not ideal.

However, since all that does is set the values a and b on which we perform arithmetic; I don't see how that will result in a failed test?

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
baykaner: I mean, of course I already ran all the tests and they pass.
Also, why should it fail? we get wrap around. my understanding is that this is well defined behaviour, and that's what we are getting for a-b where b > a:
https://stackoverflow.com/questions/7221409/is-unsigned-integer-subtraction-defined-behavior

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
baykaner: I don't think that's implemented yet, but @markos would know best

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
pierre-wilmot-fetch-ai: You're sampling a uniform distribution over range [0, MAX]

You have 50% chance of getting a in range [MAX/2, MAX]
You have 50% chance of getting b in range [MAX/2, MAX]
=> You have 25% chance of getting both a and b in range [MAX/2, MAX]
if a > MAX/2 and b > MAX/2, then a+b > MAX. <= That's an overflow

And if I'm mistaken, the 25% probability holds for any combination of two values
As long as you sample as follow :
a > (1 - X) * MAX && b > (X * MAX)  for any x in [0, 1]
You'll overflow.

But please fix the build and actually run the test

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
baykaner: see response to first comment

libs/math/tests/math/all_tests/fundamental_operators/fundamental_operators.cpp
pierre-wilmot-fetch-ai: Thinking about, it, test will pass regardless, since both side of the equality will overflow equally, and the equlity relation will still hold.

baykaner: yes, that was my point when I said: "However, since all that does is set the values a and b on which we perform arithmetic; I don't see how that will result in a failed test?"; probably didn't explain it well

pierre-wilmot-fetch-ai (changes requested):
Please remove libs/math/include/math/kernels/standard_functions.hpp as it's mostly  not correct and not used.

pierre-wilmot-fetch-ai (dismissed):

libs/math/tests/CMakeLists.txt
```diff
@@ -10,18 +10,4 @@ include(${FETCH_ROOT_CMAKE_DIR}/BuildTools.cmake)
...
-add_fetch_test(math_gtest_gtest fetch-math math/gtest SLOW)
-add_fetch_test(math_matrix_gtest fetch-math math/matrix_operations)
-add_fetch_test(math_activation_gtest fetch-math math/ml_activation_functions)
-add_fetch_test(math_loss_gtest fetch-math math/ml_loss_functions)
-add_fetch_test(math_serializers_gtest fetch-math math/serializers)
-add_fetch_test(math_tensor_gtest fetch-math math/tensor)
-add_fetch_test(math_trigonometry_gtest fetch-math math/trigonometry)
-add_fetch_test(math_blas_gtest fetch-math math/linalg/blas)
-add_fetch_test(math_tensor_iterator fetch-math math/tensor_iterator)
+add_fetch_test(math_gtest fetch-math math/all_tests)
```
wkaluza: I can live with consolidating test suites in general, but I'm opposed to putting the SLOW test in with the others - the whole test suite takes 5.3s to run after this change, of which 5.2s is exp_gtest.testing_exp. Slow tests should be separate and marked as such in the interest of builds failing fast.

libs/math/include/math/statistics/distributions.hpp
```diff
@@ -1,57 +0,0 @@
-#pragma once
```
wkaluza: Why axe this file?

libs/math/include/math/type.hpp
```diff
@@ -1,85 +0,0 @@
-#pragma once
```
wkaluza, line 1: Why delete this?

libs/math/tests/math/all_tests/include_math_lib/include_math_lib_for_coverage.cpp
wkaluza: I don't think this file is useful unless CI runs a script verifying that all headers in the math directory are indeed included. The script should be simple enough to write and I can help with integrating it with the Jenkins build. Without automatic verification, someone will inevitably create a new header and forget to add it here.

This is separate from Pierre's point on how well this trick fulfills its stated purpose.

wkaluza (changes requested):

libs/math/tests/CMakeLists.txt
baykaner: okay cool - I believe that is really a benchmark masquerading as a test. I'll look into it.

baykaner: Okay on further inspection I think the test could use a little reworking, however, I believe "SLOW" is supposed to related to tests that are supposed to indicate tests that are really benchmarks or should otherwise be moved somewhere else. In this case these are genuine tests;

libs/math/include/math/statistics/distributions.hpp
baykaner: TGamma and LGamma not being tested or used anywhere yet. We'll resurrect them in the future if we find we need them

libs/math/include/math/type.hpp
baykaner: as above not currently used functions. Probably we will need to revive these ones in the not too distant future

libs/math/tests/CMakeLists.txt
wkaluza: We need to take care of naming separately, because the "SLOW" category is referred to elsewhere as integration tests, which is wrong in some respects.

The SLOW tests are actually tests - the utility of defining them separately here is that Jenkins can run cheap tests first and get near-instantaneous feedback on 90% of the codebase, then slightly slower tests, then the really slow ones. That's what I meant by failing fast. This test may well need reworking (I haven't looked), but I believe that's out of scope for this PR. IMO the best thing you can do is leave this one test alone and leave it declared as SLOW and only consolidate the others.

libs/math/tests/math/all_tests/include_math_lib/include_math_lib_for_coverage.cpp
wkaluza: As we discussed earlier today, if you're not keen on automating keeping this file up-to-date, then I would not bother checking this file in, but instead generate the coverage report which lists files that have no coverage and file an issue listing them. Then delete this file, as it will have served its purpose.

wkaluza (dismissed):

wkaluza (approved):

PR #899, opened 2019-04-25
initial POC - tests single node and TX sync with two nodes

n-hutton:
Initial POC for high level integration tests - look at the YAML file for an idea of what should be happening


scripts/run_integration_test.py
```diff
@@ -0,0 +1,399 @@
...
+
+    def trigger(self):
+        self._callback()
+
+class TestInstance():
+    """
+    Sets up an instance of a test, containing references to started nodes and other relevant data
+    """
+
+    _number_of_nodes  = 0
```
ejfitzgerald: As discussed, these probably should be in the `__init__` method unless you want them to be `static`

scripts/ci-tool.py
```diff
@@ -170,19 +169,22 @@ def build_project(project_root, build_root, options):
...
     if not os.path.isdir(build_root):
         raise RuntimeError('Build Root doesn\'t exist, unable to test project')
 
-    # clear all the data files which might be hanging around
-    for root, _, files in os.walk(build_root):
-        for path in fnmatch.filter(files, '*.db'):
-            data_path = os.path.join(root, path)
-            print('Removing file:', data_path)
-            os.remove(data_path)
+    clean_files(build_root);
```
ejfitzgerald: semi colon not required

scripts/run_integration_test.py
```diff
@@ -0,0 +1,399 @@
...
+
+        if self._watchdog:
+            self._watchdog.stop()
+
+def extract(test, key, expected = True, expect_type = None, default = None):
+    """
+    Convenience function to remove an item from a YAML string, specifying the type you expect to find
+    """
+    try:
+        result = test[key]
```
ejfitzgerald: Think it would be cleaner to make this a `if key in test` then you can get rid of the exception handling

scripts/run_integration_test.py
```diff
@@ -0,0 +1,399 @@
...
+            # create and send the transaction to the ledger, capturing the tx hash
+            tx = tokens.wealth(identity.private_key_bytes, balance)
+
+            tx_and_identity.append((tx, identity, balance))
+
+            output("Created wealth with balance: ", balance)
+
+        # Conditionally verify TXs get executed
+        # TODO(HUT): This won't work if the node isn't a miner, I suspect
+        if False:
```
ejfitzgerald: This should work even in the case of the non-miner. The only difference is that you will not get a `Mined` status if you are not a mining node

scripts/run_integration_test.py
```diff
@@ -0,0 +1,399 @@
...
+
+        test_instance.stop()
+
+    output("\nAll integration tests have passed :)")
+
+def parse_commandline():
+    parser = argparse.ArgumentParser( description='TODO(HUT): description')
+
+    # Required argument
+    parser.add_argument('build_directory', type=str, help='Location of the build directory relative to current path')
```
ejfitzgerald: Minor - the `type=str` is redundant that is the default

ejfitzgerald (changes requested):
I love it  - I just have a handful of minor python tweaks

scripts/run_integration_test.py
```diff
@@ -0,0 +1,399 @@
...
+    parser.add_argument('yaml_file', type=str, help='Location of the build directory relative to current path')
+    parser.add_argument('constellation_exe', type=str, help='Location of the build directory relative to current path')
```
wkaluza: I think "Location of the build directory" is misleading here, because the constellation executable won't be found in fetch-ledger/build/ unless users had set up symlinks. This should either point directly to the constellation executable file (preferred, given that it's called constellation_exe) or to the build directory (in which case the logic will need adjusting, and the arg should probably be called something like build_dir).

wkaluza: Could the path to constellation and the build dir live in the yaml file? Then there would be only one command line arg to worry about, and YAML allows to put in explanatory comments

scripts/run_integration_test.py
```diff
@@ -0,0 +1,399 @@
+#!/usr/bin/env python3
```
wkaluza: I worry about exception safety in this file. The watchdog won't be able to shut down cleanly when os.exit is called; there's no guarantee that stop() will be called if there's an exception in run_test.

In general I would suggest that raising exceptions be favoured over calling os.exit, and the watchdog should probably be implemented as a context manager. This would ensure exception safety and spare you some of the manual calls to stop() and start(), and probably wouldn't require running the watchdog in a separate thread.

scripts/run_integration_test.py
```diff
@@ -0,0 +1,399 @@
...
+
+    data_loaded = []
+
+    # Read YAML file
+    with open(yaml_file, 'r') as stream:
+        try:
+            data_loaded = yaml.load_all(stream)
+
+            # https://stackoverflow.com/questions/42802058
+            data_loaded = list(data_loaded)
```
wkaluza: I'm not sure you actually need this - the for-loop that comes next should be happy to iterate through the generator without it being turned into a list.

scripts/run_integration_test.py
```diff
@@ -0,0 +1,399 @@
...
+        if command == 'verify_txs':
+            print("verifying txs")
+            verify_txs(parameters, test_instance)
+        else:
+            output("Found unknown command when verifying expectations: {}".format(command))
+            sys.exit(1)
+
+def run_test(build_directory, yaml_file, constellation_exe):
+
+    data_loaded = []
```
wkaluza: See previous comment (in this file, below) - I don't think leaving data_loaded as a generator is a problem here.

scripts/ci-tool.py
n-hutton: hard to leave C++ mental mode :)

scripts/run_integration_test.py
n-hutton: 👍

scripts/run_integration_test.py
n-hutton: 👍

scripts/run_integration_test.py
n-hutton: Great, I assumed the worst initially :)

scripts/run_integration_test.py
n-hutton: 👍

scripts/run_integration_test.py
n-hutton: there doesn't need to be a guarantee that stop() is called - it's kind of the opposite in fact right. I looked at context managers, it's a nice idea and I will do this on __exit__- when the watchdog leaves scope it can call stop(). However, I don't think the watchdog can be achieved without at least another thread: the watchdog is hanging around so that it either does nothing, or smashes everything with a hammer when it times out. If anything in the test is blocking the watchdog cannot do this.

scripts/run_integration_test.py
n-hutton: you're right

scripts/run_integration_test.py
n-hutton: yup

scripts/run_integration_test.py
n-hutton: This help comment is wrong, the arg should and does point directly at the constellation exe as you suggest. Putting the path to constellation in the yaml file doesn't work for this use case because ci builds multiple directories depending on release mode/debug mode etc.

ejfitzgerald:
@n-hutton what is going on with this?

scripts/ci-tool.py
```diff
@@ -248,7 +261,7 @@ def main():
     if args.integration_tests:
         test_project(build_root, 'Slow')
         test_project(build_root, 'Integration')
-
+        test_integration(project_root, build_root)
```
ejfitzgerald: I think this should be in its own category like E2E

Jenkinsfile
```diff
@@ -4,39 +4,41 @@ pipeline {
 
   stages {
 
-    stage('Builds & Tests') {
```
n-hutton: I feel the diff is messing up here - only alwaysPull should be added, nothing moved around

ejfitzgerald (changes requested):

scripts/ci-tool.py
n-hutton: done

ejfitzgerald (approved):

PR #901, opened 2019-04-26
Replace all uses of AtomicStateMachine and remove it

wkaluza:


libs/ledger/include/ledger/storage_unit/transaction_store_sync_service.hpp
```diff
@@ -110,15 +109,21 @@ class TransactionStoreSyncService : public network::AtomicStateMachine<tx_sync::
...
   // We need this for the testing.
   bool IsReady()
   {
     FETCH_LOCK(is_ready_mutex_);
     return is_ready_;
   }
 
+  void Execute()
+  {
+    if (state_machine_->IsReadyToExecute())
```
pbukva, line 122: Just minor comment - in theory this creates potential for "hot" loop if ```state_machine_->IsReadyToExecute()``` returns ```false```.

To mitigate this potential risk, I would suggest to add escape valve in form of ```else{std::this_thread::sleep_for(...);}```

wkaluza: There is already a wait in this execution path - the lane service calls Execute using a "BackgroundedWorkThread" which locks on a condition variable for SYNC_PERIOD_MS (currently set to 500ms) between executions. See void LaneService::Start() in lane_service.cpp

libs/ledger/src/protocols/executor_rpc_client.cpp
```diff
@@ -64,14 +71,17 @@ class ExecutorConnectorWorker : public network::AtomicStateMachine<ExecutorRpcCl
...
     }
     catch (...)
     {
       FETCH_LOG_WARN(LOGGING_NAME, "Work threw.");
       throw;
     }
-    auto r = this->AtomicStateMachine::Get();
+    auto r = state_machine_->state();
 
     switch (r)
```
pbukva, line 85: Current Implementation is just mimicking 1:1 all states of completely unrelated workflow of lower layer (`Promise` class).
This class (higher layer) should define its own sets of states (and related workflow).
I know that it was like that before the refactoring, but I would suggest rework this now.

The implication of this issue, is that, for example the `State::FAILED` is never triggered at all, it just sits here with related callback method as dead code.

I would suggest to clearly state states & workflow which suits this state machine, for example something like:
```
  enum class State
  {
    INITIAL = 0,
    CONNECTING,
    CONNECTED,
    TIMEDOUT
  };
```

NOTE:
Another thing here is (but it is **out** of scope of this PR), that **timeout** scenario shall be handled by lower layer(s) `Promise` (potentially by thread worker class) by exposing API methods to allow to set timeout, which would be then automatically evaluated by these lower layers, where such API would of course offer way for upper layer to be notified when set timeout actually times out (e.g. callback, or getter method, etc. ...).
And similarly, handling of other common states INITIAL, FAILED and FINISHED scenarios could be also relocated to lower layers.

engineerkatie:
Noooooooooooooooo!!!!!!!!!

AtomicStateMachine is the coolest named C++ component ever in the history
of EVER!!!

libs/ledger/src/protocols/executor_rpc_client.cpp
wkaluza: The issue with FAILED state came up this morning - it's a broader problem with the ExecutorRpcClient, in that it isn't code that currently gets used much outside of its own tests. According to @ejfitzgerald, once the time comes to pry Constellation apart, this class will need a rethink, but it's not a priority right now.

libs/ledger/include/ledger/storage_unit/transaction_store_sync_service.hpp
pbukva: Keep in mind, that worker thread type is injected as template parameter (currently set to `BackgroundedWorkThread`), and that  worker thread type implements `Wait(...)` & `Wake(...)` methods, which can be implemented by other Type (if used in template instead) as empty methods (not really doing anything).

libs/ledger/src/protocols/executor_rpc_client.cpp
pbukva: What I'm saying here is that you can (and should) drop failed from `ExecutorRpcClient::State` enum for reasons stated at the top of my comment.

The other things I wrote in NOTE, I clearly said that that is out of scope of this PR.

wkaluza: That's fair enough.

libs/ledger/include/ledger/storage_unit/transaction_store_sync_service.hpp
wkaluza: I see your point, but I don't see how we could prevent future mistakes like that. I think it's fair enough, that when someone changes the worker thread to one with different behaviour, the resulting system will behave differently. Hopefully we'll be guarded against surprises like that as we build up our test coverage.

libs/ledger/include/ledger/protocols/executor_rpc_client.hpp
```diff
@@ -56,11 +55,10 @@ class ExecutorRpcClient : public ExecutorInterface
   std::shared_ptr<Client> client;
 
   explicit ExecutorRpcClient(NetworkManager const &tm, Muddle &muddle)
-    : network_manager_(tm)
-  {
-    client_ = std::make_shared<Client>("R:Exec", muddle.AsEndpoint(), Muddle::Address(),
-                                       SERVICE_EXECUTOR, CHANNEL_RPC);
-  }
+    : client_(std::make_shared<Client>("R:Exec", muddle.AsEndpoint(), Muddle::Address(),
```
n-hutton, line 58: Unrelated to you: this doesn't need to be explicit. Also I think rename tm to nm here.

I suspect the order here should be the network manager is created before the client_, but perhaps it doesn't matter. It looks like network_manager_ is unused actually?

libs/ledger/include/ledger/storage_unit/transaction_store_sync_service.hpp
```diff
@@ -51,8 +51,7 @@ enum class State
 };
 }
 
-class TransactionStoreSyncService : public network::AtomicStateMachine<tx_sync::State>,
-                                    public VerifiedTransactionSink
+class TransactionStoreSyncService : public VerifiedTransactionSink
```
n-hutton, line 54: 👍 👍 👍

libs/ledger/include/ledger/storage_unit/transaction_store_sync_service.hpp
```diff
@@ -140,11 +145,21 @@ class TransactionStoreSyncService : public network::AtomicStateMachine<tx_sync::
...
+  State OnInitial();
+  State OnQueryObjectCounts();
+  State OnResolvingObjectCounts();
+  State OnQuerySubtree();
+  State OnResolvingSubtree();
+  State OnQueryObjects();
+  State OnResolvingObjects();
+  State OnTrimCache();
+
+  std::shared_ptr<core::StateMachine<State>> state_machine_;
```
n-hutton: normally style is using StateMachine

libs/ledger/src/protocols/executor_rpc_client.cpp
```diff
@@ -64,78 +70,72 @@ class ExecutorConnectorWorker : public network::AtomicStateMachine<ExecutorRpcCl
   {
     try
     {
-      this->AtomicStateMachine::Work();
+      if (state_machine_->IsReadyToExecute())
```
n-hutton, line 73: I am not particularly familiar with the executor_rpc_client stuff. This seems like the state machine isn't following the usual reactor pattern and is instead being driven here - is that correct?

libs/ledger/src/storage_unit/lane_service.cpp
```diff
@@ -171,9 +171,10 @@ void LaneService::Start()
 
   tx_sync_service_->Start();
 
-  // TX Sync service
-  workthread_ = std::make_shared<BackgroundedWorkThread>(
-      &bg_work_, "BW:LS-" + std::to_string(cfg_.lane_id), [this]() { tx_sync_service_->Work(); });
+  // TX Sync service - attach to reactor once #892 is merged
```
n-hutton, line 174: Ah, this sort of answers comment above

libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
```diff
@@ -64,25 +66,36 @@ const std::size_t TransactionStoreSyncService::BATCH_SIZE = 30;
...
+                                  &TransactionStoreSyncService::OnResolvingSubtree);
+  state_machine_->RegisterHandler(State::QUERY_OBJECTS, this,
+                                  &TransactionStoreSyncService::OnQueryObjects);
+  state_machine_->RegisterHandler(State::RESOLVING_OBJECTS, this,
+                                  &TransactionStoreSyncService::OnResolvingObjects);
+  state_machine_->RegisterHandler(State::TRIM_CACHE, this,
+                                  &TransactionStoreSyncService::OnTrimCache);
+
+  state_machine_->OnStateChange([](State new_state, State /* old_state */) {
+    (void)new_state;
```
n-hutton: There is a macro for this: FETCH_UNUSED(x)

libs/network/include/network/generics/atomic_state_machine.hpp
```diff
@@ -1,154 +0,0 @@
-#pragma once
```
n-hutton, line 1: 💯

libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
```diff
@@ -95,272 +108,255 @@ TransactionStoreSyncService::~TransactionStoreSyncService()
...
+  }
+
+  if (counts.failed)
+  {
+    FETCH_LOG_WARN(LOGGING_NAME, "Lane ", cfg_.lane_id, ": ", "Failed promises: ", counts.failed);
   }
-  };
 
-  if (result)
+  return State::TRIM_CACHE;
```
n-hutton, line 349: Looks good, the github diff tool has broken at the lots of changes here so I'll maybe test your branch over the weekend, to see if the subtree sync still functions (not sure there is a unit test for this atm).

n-hutton (changes requested):
See last comment

libs/ledger/include/ledger/storage_unit/transaction_store_sync_service.hpp
n-hutton: I wonder if this BackgroundedWorkThread is 'scheduled for demolition' as Ed would put it

n-hutton: Ed: ya

libs/ledger/include/ledger/protocols/executor_rpc_client.hpp
wkaluza: Good spot. I'll remove the NetworkManager member and leave the constructor explicit.

libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
wkaluza: I'll look into writing some tests if they're not in place already.

n-hutton (approved):
LGTM

PR #907, opened 2019-04-26
Bootstrap V2

ejfitzgerald:
- Updated prover interface to enable it to be thread safe
- Minor updates to `JsonHttpClient` class
- Update the bootstrap monitor to be able to reflect the newly updated bootstrap server


apps/constellation/bootstrap_monitor.cpp
```diff
@@ -31,64 +33,139 @@ using variant::Variant;
...
+ */
+struct Attestation
+{
+  ConstByteArray const public_key;
+  ConstByteArray const nonce{GenerateNonce()};
+  ConstByteArray const challenge;
+  ConstByteArray const signature;
+
+  /**
+   * Build an attenstation based on the
```
n-hutton: incomplete comment

apps/constellation/bootstrap_monitor.cpp
```diff
@@ -31,64 +33,139 @@ using variant::Variant;
...
 
-const char *               BOOTSTRAP_HOST = "bootstrap.economicagents.com";
-const uint16_t             BOOTSTRAP_PORT = 80;
+const char *               BOOTSTRAP_HOST = "https://bootstrap.economicagents.com";
 const std::chrono::seconds UPDATE_INTERVAL{30};
 constexpr char const *     LOGGING_NAME = "bootstrap";
 
-}  // namespace
+/**
+ * Helper object containing all the fields required to make the attenstation
```
n-hutton: spelling: attenstation

apps/constellation/bootstrap_monitor.cpp
```diff
@@ -31,64 +33,139 @@ using variant::Variant;
...
+  , host_name_(std::move(host_name))
+  , token_(std::move(token))
+{
+  // register the state machine handlers
+  state_machine_->RegisterHandler(State::Notify, this, &BootstrapMonitor::OnNotify);
+}
+
+BootstrapMonitor::~BootstrapMonitor()
+{
+  state_machine_->Reset();
```
n-hutton, line 134: Minor: I feel both these are unneccessary right

apps/constellation/bootstrap_monitor.cpp
```diff
@@ -31,64 +33,139 @@ using variant::Variant;
...
+  state_machine_.reset();
+}
+
+bool BootstrapMonitor::DiscoverPeers(UriList &peers, std::string const &external_address)
+{
+  FETCH_LOG_INFO(LOGGING_NAME, "Bootstrapping network node @ ", BOOTSTRAP_HOST);
+
+  // query our external address if non has been provided
```
n-hutton: one has been provided?

apps/constellation/bootstrap_monitor.cpp
```diff
@@ -227,15 +350,17 @@ bool BootstrapMonitor::NotifyNode()
...
-  while (running_)
-  {
-    // periodically request peers, this allows the bootstrap node to see that we are still alive
-    NotifyNode();
+  State next_state{State::Notify};
 
-    std::this_thread::sleep_for(UPDATE_INTERVAL);
-  }
+  // phone home to notify
+  NotifyNode();
```
n-hutton, line 359: really great stuff, there is some nitty gritty details that I am missing but it's laid out well + commented to understand easily

apps/constellation/constellation.cpp
```diff
@@ -356,6 +358,17 @@ void Constellation::Run(UriList const &initial_peers)
 
     FETCH_LOG_DEBUG(LOGGING_NAME, "Still alive...");
     std::this_thread::sleep_for(std::chrono::milliseconds{500});
+
+    // detect the first time that we have fully synced
+    if (BlockCoordinator::State::SYNCHRONIZED == block_coordinator_.GetStateMachine().state() &&
+        (!ready))
+    {
+      // attach the bootstrap monitor (if one exists) to the reactor
+      reactor_.Attach(bootstrap_monitor);
```
n-hutton, line 373: What happens if it doesn't exist?

I like this ready loop - I can see that it will only connect + get peers after everything  is set up + syncronised. Presumably this means there can be races though if I use this syncro state for determining nodes' readyness as it will not be syncro immediately after connecting the bootstrap monitor

apps/constellation/main.cpp
```diff
@@ -162,12 +162,13 @@ struct CommandLineArguments
...
+    p.add(args.network_name,              "network",               "The network name to join",                                                      std::string{});
+    p.add(args.cfg.interface_address,     "interface",             "The address of the network interface to be used",                               std::string{"127.0.0.1"});
     p.add(args.cfg.block_interval_ms,     "block-interval",        "Block interval in milliseconds",                                                uint32_t{DEFAULT_BLOCK_INTERVAL});
     p.add(args.token,                     "token",                 "The authentication token to be used with bootstrapping the client",             std::string{});
     p.add(args.external_address,          "external",              "This node's global IP address",                                                 std::string{});
     p.add(bootstrap_address,              "bootstrap",             "Src address for network bootstrap",                                             std::string{});
+    p.add(args.discoverable,              "discoverable",          "The client is will to be listed as accepting connections to bootstrap",         false);
```
n-hutton: The client is willing to be ...

apps/constellation/main.cpp
```diff
@@ -224,17 +225,17 @@ struct CommandLineArguments
...
-
-        if (args.external_address.empty())
-        {
-          args.external_address = bootstrap->external_address();
-        }
+        args.external_address = args.cfg.interface_address = bootstrap->external_address();
+      }
+      else
+      {
+        // if we have enabled bootstrap and the process fails we must simply exit
```
n-hutton, line 237: I would suggest a log message unless it's in DiscoverPeers

libs/core/include/core/threading/synchronised_state.hpp
```diff
@@ -17,6 +17,8 @@
 //
 //------------------------------------------------------------------------------
 
+#include "core/mutex.hpp"
```
n-hutton, line 20: is this needed here?

libs/crypto/include/crypto/ecdsa.hpp
```diff
@@ -70,63 +71,61 @@ class ECDSASigner : public Prover
...
-    signature_ = Signature::Sign(private_key_, text);
-    return true;
-  }
+    ConstByteArray signature{};
 
-  Identity identity() const final override
-  {
-    return Identity(PrivateKey::ecdsa_curve_type::sn, public_key());
-  }
+    // sign the message in a thread safe way
```
n-hutton, line 97: In what way was it not thread safe before? Surely only one thing is signing it at once

libs/crypto/include/crypto/prover.hpp
```diff
@@ -26,16 +26,35 @@ namespace crypto {
...
 {
 public:
-  using byte_array_type = byte_array::ConstByteArray;
+  using ConstByteArray = byte_array::ConstByteArray;
 
+  // Construction / Destruction
+  Prover()          = default;
+  virtual ~Prover() = default;
+
+  /// @name Prover Interface
```
n-hutton, line 35: del?

libs/http/src/json_client.cpp
```diff
@@ -55,6 +56,50 @@ uint16_t MapPort(JsonClient::ConnectionMode mode)
...
 
+/**
+ * Create the JSON client from a specified URL
+ *
+ * @param url The URL to use for the client
+ * @return The constructed JSON Client
+ */
+JsonClient JsonClient::CreateFromUrl(std::string const &url)
+{
+  // define the URL pattern that we are matching against
```
n-hutton, line 67: A bit unclear at this point -  this third nested capture group I'm fuzzy on

libs/http/src/json_client.cpp
```diff
@@ -55,6 +56,50 @@ uint16_t MapPort(JsonClient::ConnectionMode mode)
...
+
+  ConnectionMode mode{ConnectionMode::HTTP};
+  if (scheme == "https")
+  {
+    mode = ConnectionMode::HTTPS;
+  }
+
+  if (port.matched)
+  {
+    // convert the port from
```
n-hutton: incomplete comment

libs/http/src/json_client.cpp
```diff
@@ -55,6 +56,50 @@ uint16_t MapPort(JsonClient::ConnectionMode mode)
...
+ *
+ * @param url The URL to use for the client
+ * @return The constructed JSON Client
+ */
+JsonClient JsonClient::CreateFromUrl(std::string const &url)
+{
+  // define the URL pattern that we are matching against
+  std::regex const url_pattern{R"(^(https?)://([\w\.-]+)(?::(\d+))?$)"};
+
+  // perform the match
```
n-hutton, line 70: unhelpful comment

libs/http/tests/CMakeLists.txt
```diff
@@ -13,4 +13,4 @@ setup_compiler()
 #-------------------------------------------------------------------------------
 # Test Targets
 #-------------------------------------------------------------------------------
-add_fetch_test(http_gtest fetch-http .)
+add_fetch_test(fetch-http-tests fetch-http .)
```
n-hutton, line 16: inconsistency in test and benchmark naming. Can we have all tests and benchmarks start test_XXX and bench_XXX?

libs/http/tests/json_client_tests.cpp
```diff
@@ -0,0 +1,76 @@
...
+  JsonClient client{JsonClient::CreateFromUrl("https://bar.bar.foo")};
+
+  // try and access the
+  auto const *https_client = dynamic_cast<HttpsClient const *>(&client.underlying_client());
+  ASSERT_NE(https_client, nullptr);  // check
+  EXPECT_EQ(https_client->host(), "bar.bar.foo");
+  EXPECT_EQ(https_client->port(), 443);
+}
+
+TEST(JsonClientTests, CheckBuildHttpsWithPort)
```
n-hutton, line 62: 👍 👍 👍

libs/network/include/network/muddle/packet.hpp
```diff
@@ -364,9 +364,13 @@ inline Packet::BinaryHeader Packet::StaticHeader() const noexcept
 inline void Packet::Sign(crypto::Prover &prover)
 {
   SetStamped();
-  if (prover.Sign((serializers::ByteArrayBuffer() << StaticHeader() << payload_).data()))
+
+  auto const signature =
```
n-hutton, line 368: I'd really like to explore this/muddle at some point

n-hutton (changes requested):
Extremely minor comments

apps/constellation/bootstrap_monitor.cpp
ejfitzgerald: Not currently but you are right it should be. We need to move the reset to the `StateMachine` destructor. For now this is the pattern to ensure safety. I have creates #909 to track this

apps/constellation/bootstrap_monitor.cpp
ejfitzgerald: done

apps/constellation/bootstrap_monitor.cpp
ejfitzgerald: done

apps/constellation/bootstrap_monitor.cpp
ejfitzgerald: done

apps/constellation/main.cpp
ejfitzgerald: Yes you are correct the exact reason for the failure is printed inside discover peers. I felt more comfortable putting the `std::exit(1)` in this `main.cpp` that putting it into the library code.

This is obviously a blunt instrument at the moment and it makes sense to provide a more graceful failure in the future.

libs/core/include/core/threading/synchronised_state.hpp
ejfitzgerald: Yes because we use the `FETCH_LOCK` macro in this class. It hasn't been an issue previously because of the "include the world" header policy in the other pieces of the code base 😁

libs/crypto/include/crypto/ecdsa.hpp
ejfitzgerald: Nope it was just basically handled by not really allowing copies of the private key. The issue with the interface is that to make a signature you had to make 2 calls:

*Old Way*

```c++
prover->Sign(msg); // calculate digest and sign the message (store the signature internally)

auto const signature = prover->signature(); // retrieve the signature
```

There was no internal locking and therefore there was a race condition on the internal signature storage.

The new way is conservative and locks access to the private key internals. This could be relaxed in the future, but would want to look into a little more of the OpenSSL internals to double check. In any case it is 100% thread safe and the interface is much cleaner:

*New Way*

```c++
auto const signature = prover->Sign(msg); // no internal storage, simply
                                          // calculate the signature and return it.
```

libs/http/src/json_client.cpp
ejfitzgerald: Yeah it is a bit of a brain melter. This is the non-capturing group syntax (the `?:` at the start of the capture group) I was talking about the other day. This section of the regex:

```regex
^(https?)://([\w\.-]+)(?::(\d+))?
                      ^        ^
                      |________|
```
Is just describing that when the matcher is trying to collect the port that will be specified with a colon followed by the port i.e.

https://foo.bar.baz
https://foo.bar.baz **:3454** 

We just specifying that the match needs to contain both this colon and the series of numbers in order to be valid (to avoid a case of something like: `http://foo.bar.baz:`). However, the non-capturing group syntax tells the regex engine that we don't want this as a capture - instead we nest that capture group for the port. The non-capture group is necessary because we are using the trailing `?` character to signal that the group is optional.

I would recommend having a play on my favourite website: https://regex101.com/ with this and it will *(hopefully)* make more sense.

All in all this ensures the following code works:

```c++
  // extract the matches
  auto const &scheme = match[1];
  auto const &host   = match[2];
  auto const &port   = match[3];
```

libs/http/src/json_client.cpp
ejfitzgerald: You and I can get into a religious war about comments on Monday 😁 . I would not say this comment is *unhelpful*. In the context of the function the comments are just laying out the stages of the processing.

I would argue that only unhelpful comment is one that bears no relation to the what is going on.

libs/http/tests/CMakeLists.txt
ejfitzgerald: I completely agree about consistency - again this is something where you and I have different opinions about what it should be. Let's discuss with the team and nail in down in stone. 

*As an aside this `-` separated style was the original but was changed by emumba*

libs/crypto/include/crypto/prover.hpp
ejfitzgerald: It is not stranded is does match with the other `/// @}` doxygen tag. I have separated to make a little clearer

apps/constellation/constellation.cpp
ejfitzgerald: Yes an no. I have updated the comment at this point of the code to be very explicit. The get peers actually happens before this, however, what this delays is the notify() call to the bootstrap server. This triggers listing our node as available.

Flipped around, when we make the initial get peers call. We are only peers which have fully connected beforehand. You are right this does lead to a potential initial startup condition that needs to be managed, however, provided that we have at least one node bootstrapped we should be fine.

Your comments did lead me to thinking and I have made this particular section of the code clearer and the mining and bootstrapping condition stricter.

I believe part of this comment is also directed at the recent integration test work that you have been doing. I think these two REST calls are probably what you are after for that:

- https://github.com/uvue-git/fetch-ledger/blob/develop/apps/constellation/health_check_http_module.hpp#L42
- https://github.com/uvue-git/fetch-ledger/blob/develop/libs/network/include/network/p2pservice/p2p_http_interface.hpp#L87

libs/http/src/json_client.cpp
n-hutton: Maybe a comment on a valid match pattern

n-hutton (approved):

pbukva (approved):
LGTM

PR #910, opened 2019-04-27
Bug/772 stdset of byte arrays

### LINKED ISSUE #772: ConstByteArray doesn't work with std::set for some reason

bipll:
All tests (byte array and smart contracts) seem to work just nice with std::set. Whatever behaviour led to #772 was probably caused by something else.

Closes #772


libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -64,13 +64,20 @@ ConstByteArray GenerateDigest(std::string const &source)
  */
 void ValidateAddressesInParams(Transaction const &tx, vm::ParameterPack const &params)
 {
-  // TODO(issue 772): This doesn't work with a set
-  std::unordered_set<ConstByteArray> valid_addresses;
+  std::set<ConstByteArray> valid_addresses;
 
+  std::cerr << "Validating " << tx.signatures().size() << " addresses\n";
```
n-hutton: Don't think we want to have these cerr here - use FETCH_LOG_DEBUG if that's desired

n-hutton (changes requested):

n-hutton (approved):
lgtm

PR #912, opened 2019-04-29
Public Test Net Updates

ejfitzgerald:
- Quick fix for HEAD block size issue
- Fix HTTP empty buffer crash
- Restrict miner
- Minor logging updates
- Environment variable configuration support
- P2P manifest fix

### DIFF
--- a/apps/constellation/main.cpp
+++ b/apps/constellation/main.cpp
@@ -36,15 +36,15 @@
 #include "fetch_version.hpp"
 
 #include <csignal>
+#include <cstdlib>
 #include <fstream>
 #include <iomanip>
 #include <iostream>
+#include <sstream>
 #include <stdexcept>
 #include <string>
 #include <system_error>
 #include <vector>
-#include <cstdlib>
-#include <sstream>
 
 namespace {
 
@@ -99,7 +99,6 @@ void UpdateConfigFromEnvironment(bool &value, char const *name)
     }
   }
 }
-
 
 // REVIEW: Move to platform
 uint32_t Log2(uint32_t value)
@@ -442,7 +441,7 @@ struct CommandLineArguments
       }
 
       manifest = std::move(local_manifest);
-      present = true;
+      present  = true;
     }
 
     return present;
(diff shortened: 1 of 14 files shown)


apps/constellation/main.cpp
```diff
@@ -385,6 +495,9 @@ ProverPtr GenerateP2PKey()
 {
   static constexpr char const *KEY_FILENAME = "p2p.key";
```
wkaluza, line 496: Maybe combine KEY_FILENAME and key_path into a single std::string variable? Given that we still need to access key_path.c_str() later, we might as well.

Relatedly, is key_path actually a path? Or should it be called a file name?

apps/constellation/main.cpp
```diff
@@ -183,6 +228,33 @@ struct CommandLineArguments
     // parse the args
     p.Parse(argc, argv);
 
+    // if the user has specified any environment variables these always take precedence
```
pbukva, line 231: Should this be other way around? As per well esstablished approach, command-line arguments should take precedence over Env. variables (or config files in that matter).

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -489,11 +489,14 @@ void MainChain::RecoverFromFile(Mode mode)
   if (Mode::CREATE_PERSISTENT_DB == mode)
   {
     block_store_->New("chain.db", "chain.index.db");
+    head_store_.open("chain.head.db",
```
wkaluza, line 492: Does this file not potentially leak the head_store_?

pbukva (approved):
LGTM

apps/constellation/main.cpp
ejfitzgerald: It is a path, it just can't be a folder path.

ejfitzgerald: I tend to agree about the structure I will raise an issue to cover this and the @pbukva comments

apps/constellation/main.cpp
ejfitzgerald: Raise #915 to track this

Issue #896, opened 2019-04-25
Unit test shouldn't test for executions speed


pierre-wilmot-fetch-ai:
Unit test shouldn't test for executions speed, that's not what they are for.
This can cause random failures that are not related to the code (Slow Jenkins, as here Unit test shouldn't test for executions speed // Slow execution due to running from docker -> makes it impossible to generate coverage reports as test binary return with non 0 code).

We should add benchmark for the complex part of the code, and track benchmark results accross commits -- Jenkins as a plugin for that, talked to @chr15murray about it.

Currently, the file causing trouble is https://github.com/uvue-git/fetch-ledger/blob/develop/libs/math/tests/math/tensor/tensor_indexing.cpp

ejfitzgerald:
@baykaner @pierre-wilmot-fetch-ai do we have an eta for this "fix". If not might I suggest that the test in question is disabled for the short term.

(This is causing failures for me)

baykaner:
It's trivially fixable, but the deeper question is how to get what we want; i.e. both:
1. no random failures due to slow execution.
2. a robust way to prevent/flag up changes that impact performance unacceptably.
Maybe we can set up a way for Jenkins to send out warning emails whenever a benchmark ends up being slower than some manually pre-specified timings? It's not as robust as a failing test, but it might also be okay.

In the mean-time I'll open a PR to remove the execution times checks.

baykaner:
submitted https://github.com/uvue-git/fetch-ledger/pull/913

ejfitzgerald:
@baykaner I think that this is probably closed - there is an open issue on the DevOps backlog to have a complete system for running and keeping the benchmark data.

If you agree can you please close

PR #913, opened 2019-04-29
tensor execution time

baykaner:
lets leave open #896 until we get Jenkins running the benchmarks


pierre-wilmot-fetch-ai (changes requested):
Probably safe to mesure in Nanosecond here (otherwise every timimg will report 0)

pierre-wilmot-fetch-ai (approved):

PR #914, opened 2019-04-29
Build OpenSSL from source during configure step on Mac and Linux

wkaluza:


n-hutton:
Correct me if I am wrong, but does this not mean that the CI will have to build openssl for every container?

wkaluza:
@n-hutton That's correct. That said, with the current container setup, CMake will find the desired version 1.1.1b and skip the build.

wkaluza:
To test what building OpenSSL would do to our build times, I pointed the submodule at the tip of the 1.1.1-stable branch, so that CMake would not be satisfied with version 1.1.1b and actually perform the build.

Looking at BlueOcean, it seems that the full Configure-make-make test-make install takes around 8 minutes in a 'ledger' container:
https://jenkins.economicagents.com/blue/organizations/jenkins/fetch-ledger/detail/PR-914-head/2/pipeline/
https://jenkins.economicagents.com/blue/organizations/jenkins/fetch-ledger/detail/PR-914-merge/25/pipeline/

A good middle ground, I feel, would be to let containers continue to build OpenSSL for themselves (see link below), but have the option to easily build locally. The main win from doing this is that we would be using the same version of OpenSSL throughout the development process.
https://github.com/uvue-git/docker-images/blob/master/fetch-ledger-develop-image/Dockerfile#L46

wkaluza:
I reverted this to where it will build the same version as the one installed in the Docker container. This way we have an easy method of building the right version of OpenSSL without the undue overhead on the builds. I think this is the pragmatic thing to do.

wkaluza:
Closing for now; we may decide to standardise the OpenSSL version in the future, but currently it's not a priority.

PR #917, opened 2019-04-29
Prevent BlockCoordinator from waiting forever if there's an unknown Tx

wkaluza:
The sleep in the block coordinator unit test is shameful, but it is the only way to make the test pass for now, because the production code relies on a timer which is not easily mocked out. I'll follow up with injecting a fake clock which we can advance on demand in the tests.

### DIFF
--- a/libs/ledger/src/chain/block_coordinator.cpp
+++ b/libs/ledger/src/chain/block_coordinator.cpp
@@ -134,8 +134,6 @@ void BlockCoordinator::TriggerBlockGeneration()
 
 BlockCoordinator::State BlockCoordinator::OnReloadState()
 {
-  State next_state{State::RESET};
-
   // if no current block then this is the first time in the state therefore lookup the heaviest
   // block
   if (!current_block_)
@@ -162,13 +160,11 @@ BlockCoordinator::State BlockCoordinator::OnReloadState()
     }
   }
 
-  return next_state;
+  return State::RESET;
 }
 
 BlockCoordinator::State BlockCoordinator::OnSynchronizing()
 {
-  State next_state{State::SYNCHRONIZING};
-
   // ensure that we have a current block that we are executing
   if (!current_block_)
   {
@@ -208,7 +204,7 @@ BlockCoordinator::State BlockCoordinator::OnSynchronizing()
     if (GENESIS_DIGEST == previous_hash)
     {
       // once we have got back to genesis then we need to start executing from the beginning
-      next_state = State::PRE_EXEC_BLOCK_VALIDATION;
+      return State::PRE_EXEC_BLOCK_VALIDATION;
     }
     else
     {
@@ -227,7 +223,7 @@ BlockCoordinator::State BlockCoordinator::OnSynchronizing()
   else if (current_hash == last_processed_block)
   {
     // the block coordinator has now successfully synced with the chain of blocks.
-    next_state = State::SYNCHRONIZED;
+    return State::SYNCHRONIZED;
   }
   else
   {
@@ -295,16 +291,16 @@ BlockCoordinator::State BlockCoordinator::OnSynchronizing()
 
     // update the current block and begin scheduling
     current_block_ = next_block;
-    next_state     = State::PRE_EXEC_BLOCK_VALIDATION;
+
+    return State::PRE_EXEC_BLOCK_VALIDATION;
   }
 
-  return next_state;
+  return State::SYNCHRONIZING;
 }
 
 BlockCoordinator::State BlockCoordinator::OnSynchronized(State current, State previous)
 {
   FETCH_UNUSED(current);
-  State next_state{State::SYNCHRONIZED};
 
   // ensure the periodic print is not trigger once we have synced
   syncing_periodic_.Reset();
@@ -312,7 +308,7 @@ BlockCoordinator::State BlockCoordinator::OnSynchronized(State current, State pr
   // if we have detected a change in the chain then we need to re-evaluate the chain
   if (chain_.GetHeaviestBlockHash() != current_block_->body.hash)
   {
-    next_state = State::RESET;
+    return State::RESET;
   }
   else if (mining_ && mining_enabled_ && (Clock::now() >= next_block_time_))
   {
@@ -329,7 +325,7 @@ BlockCoordinator::State BlockCoordinator::OnSynchronized(State current, State pr
     current_block_.reset();
 
     // trigger packing state
-    next_state = State::PACK_NEW_BLOCK;
+    return State::PACK_NEW_BLOCK;
   }
   else if (State::SYNCHRONIZING == previous)
   {
@@ -338,7 +334,7 @@ BlockCoordinator::State BlockCoordinator::OnSynchronized(State current, State pr
                    " prev: ", ToBase64(current_block_->body.previous_hash), ")");
   }
 
-  return next_state;
+  return State::SYNCHRONIZED;
 }
 
 BlockCoordinator::State BlockCoordinator::OnPreExecBlockValidation()
@@ -414,8 +410,6 @@ BlockCoordinator::State BlockCoordinator::OnPreExecBlockValidation()
 
 BlockCoordinator::State BlockCoordinator::OnWaitForTransactions()
 {
-  State next_state{State::WAIT_FOR_TRANSACTIONS};
-
   // if the transaction digests have not been cached then do this now
   if (!pending_txs_)
   {
@@ -455,7 +449,7 @@ BlockCoordinator::State BlockCoordinator::OnWaitForTransactions()
     // clear the pending transaction set
     pending_txs_.reset();
 
-    next_state = State::SCHEDULE_BLOCK_EXECUTION;
+    return State::SCHEDULE_BLOCK_EXECUTION;
   }
   else
   {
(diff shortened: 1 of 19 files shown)


libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -70,9 +71,9 @@ BlockCoordinator::BlockCoordinator(MainChain &chain, ExecutionManagerInterface &
   , block_difficulty_{block_difficulty}
   , num_lanes_{num_lanes}
   , num_slices_{num_slices}
-  , tx_wait_periodic_{TX_SYNC_NOTIFY_INTERVAL}
-  , exec_wait_periodic_{EXEC_NOTIFY_INTERVAL}
-  , syncing_periodic_{NOTIFY_INTERVAL}
+  , tx_wait_periodic_{std::chrono::milliseconds{1000}}
```
ejfitzgerald: I think that I would have still kept these constants. They provide a little more meaning when reading the code (maybe the names should be improved)

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -92,20 +93,13 @@ BlockCoordinator::BlockCoordinator(MainChain &chain, ExecutionManagerInterface &
   state_machine_->RegisterHandler(State::RESET,                        this, &BlockCoordinator::OnReset);
   // clang-format on
 
-  // for debug purposes
-#if 0
```
ejfitzgerald, line 96: I would keep this in - This is actually very useful when debugging the block coordinator

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -208,7 +198,7 @@ BlockCoordinator::State BlockCoordinator::OnSynchronizing()
     if (GENESIS_DIGEST == previous_hash)
     {
       // once we have got back to genesis then we need to start executing from the beginning
-      next_state = State::PRE_EXEC_BLOCK_VALIDATION;
+      return State::PRE_EXEC_BLOCK_VALIDATION;
```
ejfitzgerald, line 202: This is a quite a large function I know, however, in general I prefer to try and keep a single return point in a function. This actually makes debugging in the future much easier.

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -257,7 +246,7 @@ BlockCoordinator::State BlockCoordinator::OnSynchronizing()
       FETCH_LOG_DEBUG(LOGGING_NAME, "Sync: Common Parent: ", ToBase64(common_parent->body.hash));
       FETCH_LOG_DEBUG(LOGGING_NAME, "Sync: Next Block...: ", ToBase64(next_block->body.hash));
 
-      // calculate a percentage syncronisation
+      // calculate a percentage synchronisation
```
ejfitzgerald, line 250: 👍

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -781,51 +760,42 @@ bool BlockCoordinator::ScheduleBlock(Block const &block)
...
 {
-  ExecutionStatus status{ExecutionStatus::ERROR};
-
   // based on the state of the execution manager determine
   auto const execution_state = execution_manager_.GetState();
 
   // map the raw executor status into our simplified version
   switch (execution_state)
   {
+  default:
```
ejfitzgerald: The reason for not including a default here and having a single return path is to trigger the compiler to generate a warning when a specific case is not handled. I think I would revert back this block of changes

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -846,78 +816,56 @@ void BlockCoordinator::UpdateTxStatus(Block const &block)
 
 char const *BlockCoordinator::ToString(State state)
 {
-  char const *text = "Unknown";
-
   switch (state)
   {
+  default:
```
ejfitzgerald: Same comment from above, here

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -846,78 +816,56 @@ void BlockCoordinator::UpdateTxStatus(Block const &block)
...
-  return text;
 }
 
 char const *BlockCoordinator::ToString(ExecutionStatus state)
 {
-  char const *text = "Unknown";
-
   switch (state)
   {
+  default:
```
ejfitzgerald: Same comment from above, here

ejfitzgerald (changes requested):

libs/ledger/src/chain/block_coordinator.cpp
wkaluza: I found it quite difficult to follow, is all - the advantage of early returns to readability is that I can disregard any code in a branch that comes after a return statement. With next_state being assigned and reassigned, I have to follow the function every time top to bottom to establish what's being returned in various circumstances. For short functions this is not a problem, but, as you say, this function is not short.

libs/ledger/src/chain/block_coordinator.cpp
wkaluza: To my mind, it duplicates the production code - if someone is debugging and wants to see every state change in the log, they can equally comment out the conditional  `if (periodic_print_.Poll())`

libs/ledger/src/chain/block_coordinator.cpp
wkaluza: That's sound, I'll revert.

libs/ledger/src/chain/block_coordinator.cpp
wkaluza: That's fair enough.

wkaluza:
@ejfitzgerald Ping? How strongly do you feel about the remaining two points?

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -412,9 +403,24 @@ BlockCoordinator::State BlockCoordinator::OnPreExecBlockValidation()
...
+      // Assume block was invalid and discard it
+      chain_.RemoveBlock(current_block_->body.hash);
+
+      return State::RESET;
+    }
+  }
+  else if (previous != current)
+  {
+    // Only just started waiting for transactions - reset timeout
+    wait_for_tx_timeout_ = std::chrono::milliseconds(300u);
```
ejfitzgerald: I suggest that we make this a constant and keep it with the others at the top of the file. Also I would make it larger that 300ms. Probably a better timeout would be 60 seconds. This is quite a destructive step and should be a last resort.

In the final version we probably need something flow like the following:

1. Wait for a period of time for the transactions to sync normally across the network
2. After a threshold maybe something like 30secs (guess) deed that there has been an error and explicitly trigger the request for the pending transactions
3. After this time interval (and/or the status from the explicit NOT_FOUND responses from step 2) - remove the block from the chain

libs/ledger/src/chain/block_coordinator.cpp
ejfitzgerald: Looking at this function it already has multiple return points so its not really a n issue.

A more general point I would make from this review is that there are a number of changes that seem to have been made for superficial/style reasons. I guess there is an argument about consistency across the file, however, in most cases I think I would caution against it as it mostly adds noise to the review.

libs/ledger/src/chain/block_coordinator.cpp
ejfitzgerald: Fair enough

ejfitzgerald (changes requested):

ejfitzgerald:
Sorry busy couple of days - I have a few additional comments, however, in general they simply boil down to the fact that we probably want t timeout longer that 300ms

libs/ledger/src/chain/block_coordinator.cpp
wkaluza: Fair enough, I'll revert those changes.

libs/ledger/src/chain/block_coordinator.cpp
wkaluza: Step 2. is in the works and will be a separate PR. For now I'll extract the constant and set it to 60 seconds.

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -475,30 +480,35 @@ BlockCoordinator::State BlockCoordinator::OnWaitForTransactions(State current, S
...
 {
+  State next_state{State::RESET};
+
   // schedule the current block for execution
   if (ScheduleCurrentBlock())
   {
     exec_wait_periodic_.Reset();
 
-    return State::WAIT_FOR_EXECUTION;
+    next_state = State::WAIT_FOR_EXECUTION;
```
ejfitzgerald: My comment was a more general one with these changes as an example. I was not expecting you to change them. Lets make sure we have a catch up today.

ejfitzgerald (dismissed):

libs/ledger/src/chain/block_coordinator.cpp
wkaluza: Sorry, I misunderstood. I'll wait with merging this PR for now

ejfitzgerald (approved):

PR #918, opened 2019-04-29
Fix the diamond graph bug [close #916][close #890]

### LINKED ISSUE #890: Cache logic in the ML graph causes trouble
The current caching logic in the ML graph is causing problems when the compute graph has multiple branches merging.

We curently reallocate output buffers when the input size changes.
Problem happens when we have a graph like 

```
A -- B -- C \
             D -- E
G -- H -- I /
```

Setting input on A will trigger buffer reallocation chain A -> B -> C -> D
When reaching D, the compute output size will need size information from C and I.
Problem is at this point I doesn't have that information.

If we start by setting G, the problem is the same with branches reversed.
The current (ugly) workaround is to call setInput with the recompute_cache flag set To false, but that doesn't even solve all case (would work for example graph)

### LINKED ISSUE #916: Diamond shaped ops graph cannot be evaluated
Diamond shaped graph cannot be evaluated.

Example:
tmp1=input+input
tmp2=input+input
output=tmp1+tmp2

Attached code that can be used for reproduction:
[diamond.zip](https://github.com/uvue-git/fetch-ledger/files/3127534/diamond.zip)

pierre-wilmot-fetch-ai:


libs/ml/include/ml/node.hpp
```diff
@@ -51,6 +51,14 @@ class NodeInterface
 template <class T, class O>
 class Node : public NodeInterface<T>, public O
 {
+private:
+  enum class CachedOutputStatus
+  {
+    ValidCache,
```
baykaner: should be SCREAMING_SNAKE_CASE

libs/ml/include/ml/node.hpp
```diff
@@ -61,7 +69,7 @@ class Node : public NodeInterface<T>, public O
   Node(std::string const name, Params... params)
     : O(params...)
     , name_(std::move(name))
-    , cached_output_present_(false)
+    , cached_output_present_(CachedOutputStatus::ChangedSize)
```
baykaner: rename cached_output_present_ to cached_output_state_ since it no longer indicates the presence of something.

libs/ml/include/ml/node.hpp
```diff
@@ -81,8 +89,17 @@ class Node : public NodeInterface<T>, public O
   {
     std::vector<std::reference_wrapper<const ArrayType>> inputs = GatherInputs();
     FETCH_LOG_INFO("ML_LIB", "Evaluating node [", name_, "]");
-    if (!cached_output_present_)
+    if (cached_output_present_ != CachedOutputStatus::ValidCache)
     {
+      if (cached_output_present_ == CachedOutputStatus::ChangedSize)
+      {
+        std::vector<std::reference_wrapper<const ArrayType>> inputs = GatherInputs();
```
baykaner: redefinition of inputs specified on line 90 - presumably this doesn't even compile?

libs/ml/include/ml/node.hpp
```diff
@@ -81,8 +89,17 @@ class Node : public NodeInterface<T>, public O
...
+        auto output_size = this->ComputeOutputShape(inputs);
```
baykaner: rename to output_shape, since thats what it is right?

libs/ml/include/ml/node.hpp
```diff
@@ -81,8 +89,17 @@ class Node : public NodeInterface<T>, public O
...
+        if (cached_output_.shape() != output_size)
+        {
+          cached_output_ = ArrayType(output_size);
```
baykaner: I guess this should probably be a Reshape call rather than a new constructor?

baykaner (changes requested):

libs/ml/include/ml/node.hpp
baykaner: I think you may have misunderstood, it should be like this:
enum class CachedOutputStatus
{
  VALID_CACHE,
  CHANGED_SIZE,
  CHANGED_CONTENT
}

baykaner (changes requested):

baykaner (approved):

PR #919, opened 2019-04-29
Adding `Rand(min, max)` function to VM

pbukva:
The function represents random generator with Uniform
distribution for all avalilable primitive numeric types.

### DIFF
--- a/libs/vm_modules/include/vm_modules/math/random.hpp
+++ b/libs/vm_modules/include/vm_modules/math/random.hpp
@@ -1,0 +1,59 @@
+#pragma once
+//------------------------------------------------------------------------------
+//
+//   Copyright 2018-2019 Fetch.AI Limited
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


libs/vm_modules/include/vm_modules/math/random.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include "math/meta/math_type_traits.hpp"
+#include "vm/vm.hpp"
+
+#include <random>
+
+namespace fetch {
+namespace vm_modules {
+
+template <typename T>
+fetch::meta::IfIsInteger<T, T> Rand(fetch::vm::VM *, T const &a = T{0}, T const &b = T{100})
```
ejfitzgerald, line 29: Not sure that is make sense to have this default

ejfitzgerald (changes requested):

ejfitzgerald (approved):

libs/vm_modules/include/vm_modules/math/random.hpp
pbukva: For me it felt like the best option for integer - parallel with 100%.
But I don’t think it even matters, since I do not think that default values are propagated to VM.

PR #921, opened 2019-04-29
Prevent undefined behaviour in Rand

wkaluza:
uniform_int_distribution yields undefined behaviour if the first range bound is larger than the second
https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution/uniform_int_distribution

Likewise, uniform_real_distribution misbehaves if it's called with equal lower and upper bound
https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution/uniform_real_distribution

This PR attempts to treat the float and int overload uniformly and allows end-users to call Rand(a, b) regardless of whether a < b, a == b or a > b. We may not wish to be this permissive, but in the short term this prevents the UB.

### DIFF
--- a/libs/vm_modules/include/vm_modules/math/random.hpp
+++ b/libs/vm_modules/include/vm_modules/math/random.hpp
@@ -20,6 +20,7 @@
 #include "math/meta/math_type_traits.hpp"
 #include "vm/vm.hpp"
 
+#include <algorithm>
 #include <random>
 
 namespace fetch {
@@ -28,19 +29,27 @@ namespace vm_modules {
 template <typename T>
 fetch::meta::IfIsInteger<T, T> Rand(fetch::vm::VM *, T const &a = T{0}, T const &b = T{100})
 {
-  std::random_device               rd;
-  std::mt19937_64                  mt(rd());
-  std::uniform_int_distribution<T> dist{a, b};
-  return dist(mt);
+  if (a == b) {
+    return a;
+  }
+
+  std::random_device rd;
+  std::mt19937_64    mt(rd());
+
+  return a < b ? std::uniform_int_distribution<T>{a, b}(mt) : std::uniform_int_distribution<T>{b, a}(mt);
 }
 
 template <typename T>
 fetch::meta::IfIsFloat<T, T> Rand(fetch::vm::VM *, T const &a = T{.0}, T const &b = T{1.0})
 {
-  std::random_device                rd;
-  std::mt19937_64                   mt(rd());
-  std::uniform_real_distribution<T> dist{a, b};
-  return dist(mt);
+  if (a == b) {
+    return a;
+  }
+
+  std::random_device rd;
+  std::mt19937_64    mt(rd());
+
+  return a < b ? std::uniform_real_distribution<T>{a, b}(mt) : std::uniform_real_distribution<T>{b, a}(mt);
 }
 
 static void CreateRand(std::shared_ptr<fetch::vm::Module> module)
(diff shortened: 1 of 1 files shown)


pbukva:
Honestly, I would not bother to do this. For the same reason/ideology as it is has not been done in STL C++ implementation of `std::uniform_int_distribution(...)` - we can not (and should not) solve all non-sensical input data setups which programmers should detect/check before calling functionality.

wkaluza:
I'm not worried about the users' scripts not working correctly so much as, if the UB causes our VM to crash, we cannot guarantee a clean shutdown. We should either sanitize users' input (probably impossible in this case), or we should handle stupid things they do at runtime. "Handle" here could well mean emitting an error and shutting down, but this should happen on our terms.

libs/vm_modules/include/vm_modules/math/random.hpp
```diff
@@ -28,19 +29,31 @@ namespace vm_modules {
 template <typename T>
 fetch::meta::IfIsInteger<T, T> Rand(fetch::vm::VM *, T const &a = T{0}, T const &b = T{100})
 {
-  std::random_device               rd;
-  std::mt19937_64                  mt(rd());
-  std::uniform_int_distribution<T> dist{a, b};
-  return dist(mt);
+  if (a == b)
```
ejfitzgerald: It makes more sense to me to have the following and simply catch the error case:

```suggestion
  if (a >= b)
  {
    vm->RuntimeError("Invalid random range requested");
    return T{0};
  }
```

ejfitzgerald (changes requested):

n-hutton:
I agree with this point - allowing users to create undefined behaviour is extremely bad as the network will fork. I wonder when this function will ever be called though - for the regular use case this would also fork the network as nodes would disagree on the result

pbukva (approved):
LGTM

pbukva:
> I agree with this point - allowing users to create undefined behaviour is extremely bad as the network will fork. I wonder when this function will ever be called though - for the regular use case this would also fork the network as nodes would disagree on the result

@n-hutton This function was intended as part of Etch language in general, where Etch language (and its VM) can be used in standalone mode in separation from ledger - the same way as python.

**However, you made very good point @n-hutton** - this actually kicks door open for **new attack vector**:   If smart contract developer uses such non-deterministic functionality in smart contract implementation then it looks like it will have fatal consequences for the whole network (with ability to effectivelly shutdown the whole network), because if someone keeps commiting transactions which trigger the affected contract code, then **none** of mined/generated blocks announced to the network will be accepted by rest of the network since such blocks are essentially unverifiable.

This is not just Rand, but potentially any functionality of VM with potence of resulting in to non-deterministic-behaviour - for example using floating point types.

**I'm starting to think it would be better to drop this Rand(...) function form Etch.**
If we want to keep such functionality in, we should extend VM with support for mechanism of decorating such methods as non-determinstic, so that such non-determinstic methods won't compile **IF** VM is executed in **deterministic** mode (what ledger would of course use as mandatory  execution mode).

On top of this, we should come up with ideas how to protect ledger against such non-deterministic contracts (no matter what the reason for non-deterministic behahiour is).

wkaluza:
@pbukva @n-hutton An alternative would be to use a Fetch PRNG implementation (so behaviour does not depend on stdlib implementation details, which depend on the platform and may change over time) and always seed the PRNG. Maybe we could use the contract's hash as seed, to provide some variability? We could also allow/force users to seed the generator themselves, by exposing another Etch function.

There's no reason Etch users couldn't perform statistical simulations a'la R, as long as the results are deterministic. We would have to make it clear in the docs that multiple calls to Rand will produce an identical sequence of outputs when a given contract is run multiple times. Users who only care about having a uniformly distributed sequence as input for some computation would not mind.

It's a matter for debate how useful Rand would be for the community at large, because statistics-savvy users may prefer to use a trusted PRNG to generate their own input sequence.

Just to circle back to the original point, I think this discussion reinforces the case that we must, in fact, predict and prevent users from being able to trigger UB by submitting suitably fabricated contracts, as they would also yield non-deterministic results.

wkaluza:
@ejfitzgerald I tweaked the changes compared with your original suggestion. Are you happy with this going in?

ejfitzgerald (approved):

PR #928, opened 2019-05-01
trigonometry math & vm implementations & tests

baykaner:
- re-introduced previous trigonometry kernels
- implemented math library trig functions calling trig kernels (including double casting based FixedPoint interfaces) for trig functions
- implemented tests for trig functions
- implemented vm bindings to trig functions

Issues raised:
927 - atan2 not implemented
925 - standard kernels need testing & benchmarking (no longer relevant)
926 - trigonometry functions need benchmarking (closed by this PR also)


libs/math/benchmark/trigonometry/trigonometry.cpp
```diff
@@ -0,0 +1,268 @@
...
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include <iomanip>
+#include <iostream>
```
wkaluza: You need neither iomanip nor iostream in this file.

libs/math/include/math/kernels/sign.hpp
```diff
@@ -0,0 +1,39 @@
...
+//
+//------------------------------------------------------------------------------
+
+namespace fetch {
+namespace math {
+namespace kernels {
+
+template <typename vector_register_type>
+struct Sign
+{
```
wkaluza, line 26: Why is `y` computed this way? Doesn't it mean that you can't short-circuit the greater-than and less-than case if x happens to equal zero? Or is this faster than having `if...else if...else`?

libs/math/include/math/trigonometry.hpp
```diff
@@ -57,19 +80,41 @@ fetch::math::meta::IfIsMathArray<ArrayType, ArrayType> Sin(ArrayType const &x)
...
+  using Type = typename ArrayType::Type;
+  ASSERT(ret.size() == x.size());
+  kernels::Cos<double> c;
+  auto                 x_it = x.cbegin();
+  auto                 rit  = ret.begin();
+  double               casted_x;
+  double               casted_ret;
+  while (x_it.is_valid())
+  {
+    // TODO(800) - native fixed point implementation required - casting to double will not be
```
wkaluza, line 110: ...allowed?

libs/math/include/math/trigonometry.hpp
```diff
@@ -83,23 +128,44 @@ fetch::math::meta::IfIsMathArray<ArrayType, ArrayType> Cos(ArrayType const &x)
...
+  using Type = typename ArrayType::Type;
+  ASSERT(ret.size() == x.size());
+  kernels::Tan<double> s;
+  auto                 x_it = x.cbegin();
+  auto                 rit  = ret.begin();
+  double               casted_x;
+  double               casted_ret;
+  while (x_it.is_valid())
+  {
+    // TODO(800) - native fixed point implementation required - casting to double will not be
```
wkaluza, line 159: as above

libs/math/include/math/trigonometry.hpp
```diff
@@ -110,23 +176,44 @@ fetch::math::meta::IfIsMathArray<ArrayType, ArrayType> Tan(ArrayType const &x)
...
+  using Type = typename ArrayType::Type;
+  ASSERT(ret.size() == x.size());
+  kernels::ASin<double> s;
+  auto                  x_it = x.cbegin();
+  auto                  rit  = ret.begin();
+  double                casted_x;
+  double                casted_ret;
+  while (x_it.is_valid())
+  {
+    // TODO(800) - native fixed point implementation required - casting to double will not be
```
wkaluza, line 208: SO. MUCH. DUPLICATION.

Probably worth refactoring this in future.

libs/math/include/math/trigonometry.hpp
```diff
@@ -137,23 +224,44 @@ fetch::math::meta::IfIsMathArray<ArrayType, ArrayType> ASin(ArrayType const &x)
...
+  using Type = typename ArrayType::Type;
+  ASSERT(ret.size() == x.size());
+  kernels::ACos<double> s;
+  auto                  x_it = x.cbegin();
+  auto                  rit  = ret.begin();
+  double                casted_x;
+  double                casted_ret;
+  while (x_it.is_valid())
+  {
+    // TODO(800) - native fixed point implementation required - casting to double will not be
```
wkaluza, line 257: truncated comment, as above

libs/math/include/math/trigonometry.hpp
```diff
@@ -314,23 +464,45 @@ fetch::math::meta::IfIsMathArray<ArrayType, ArrayType> TanH(ArrayType const &x)
...
+  using Type = typename ArrayType::Type;
+  ASSERT(ret.size() == x.size());
+  kernels::ASinH<double> s;
+  auto                   x_it = x.cbegin();
+  auto                   rit  = ret.begin();
+  double                 casted_x;
+  double                 casted_ret;
+  while (x_it.is_valid())
+  {
+    // TODO(800) - native fixed point implementation required - casting to double will not be
```
wkaluza, line 503: truncated comment

libs/math/include/math/trigonometry.hpp
```diff
@@ -341,23 +513,45 @@ fetch::math::meta::IfIsMathArray<ArrayType, ArrayType> ASinH(ArrayType const &x)
...
+  using Type = typename ArrayType::Type;
+  ASSERT(ret.size() == x.size());
+  kernels::ACosH<double> s;
+  auto                   x_it = x.cbegin();
+  auto                   rit  = ret.begin();
+  double                 casted_x;
+  double                 casted_ret;
+  while (x_it.is_valid())
+  {
+    // TODO(800) - native fixed point implementation required - casting to double will not be
```
wkaluza, line 553: comment

libs/math/include/math/trigonometry.hpp
```diff
@@ -368,27 +562,373 @@ fetch::math::meta::IfIsMathArray<ArrayType, ArrayType> ACosH(ArrayType const &x)
...
+  using Type = typename ArrayType::Type;
+  ASSERT(ret.size() == x.size());
+  kernels::ATanH<double> s;
+  auto                   x_it = x.cbegin();
+  auto                   rit  = ret.begin();
+  double                 casted_x;
+  double                 casted_ret;
+  while (x_it.is_valid())
+  {
+    // TODO(800) - native fixed point implementation required - casting to double will not be
```
wkaluza, line 603: comment

libs/math/include/math/trigonometry.hpp
```diff
@@ -368,27 +562,373 @@ fetch::math::meta::IfIsMathArray<ArrayType, ArrayType> ACosH(ArrayType const &x)
...
+template <typename Type>
+fetch::math::meta::IfIsArithmetic<Type, Type> CosH(Type const &x)
+{
+  Type ret;
+  CosH(x, ret);
+  return ret;
+}
+
+/**
+ * maps every element of the array x to ret = ATan(x)
```
wkaluza: to ret = TanH(x), surely?

libs/math/tests/math/trigonometry/trig.cpp
```diff
@@ -0,0 +1,572 @@
...
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include <iomanip>
+#include <iostream>
```
wkaluza: I don't think these belong here.

libs/vm_modules/include/vm_modules/core/print.hpp
```diff
@@ -43,64 +43,61 @@ out << byte_array::ToHex(s->byte_array()) << std::endl;
 }
 */
```
wkaluza: I'm not comfortable with the changes to this file - the print functions should go to the VM for an output device, as etch doesn't know about standard output. I'd rather someone more VM savvy ok'd this, as the previous version of the file makes more sense to me.

libs/vm_modules/include/vm_modules/math/abs.hpp
```diff
@@ -30,7 +30,7 @@ template <typename T>
 fetch::math::meta::IfIsMath<T, T> Abs(fetch::vm::VM *, T const &a)
 {
   T x = T(a);
-  fetch::math::Abs(x);
+  fetch::math::Abs(x, x);
```
wkaluza, line 33: Why copy a into x? Zero-initializing x and running `fetch::math::Abs(a, x);` makes more sense to me here. Or can T be an array type?

libs/vm_modules/include/vm_modules/math/trigonometry.hpp
```diff
@@ -0,0 +1,194 @@
...
+namespace fetch {
+namespace vm_modules {
+
+/**
+ * method for taking the sine of a value
+ */
+template <typename T>
+fetch::math::meta::IfIsMath<T, T> Sin(fetch::vm::VM *, T const &a)
+{
+  T x = T(a);
```
wkaluza, line 32: Same question as with Abs, what is the reason for all the copies?

wkaluza (changes requested):

libs/math/include/math/kernels/sign.hpp
baykaner: as discussed; i think this is because SIMD methods have to be branchless - so you can OR a bunch of conditions like this into one operation (as you said, you've always got to make all the evaluations)

libs/vm_modules/include/vm_modules/core/print.hpp
baykaner: As discussed, if fetch has a builtin print function then this isn't necessary. Looking at vm/examples/01_basic_vm/ though there's a static function implemented and registered in the example; since I don't want to do this for every vm_module example, I thought it made more sense to have a general print vm_module to register. If the print function is builtin and this isn't necessary however then we should delete this vm_module; but I'm not sure that's a job for this PR; this PR made these changed just because the vm changed in upstream/develop while this was still waiting for review and these changes were necessary to avoid breaking.

libs/vm_modules/include/vm_modules/math/abs.hpp
baykaner: IfIsMath does allow T to be an array type; so I can't zero initialise.
I guess we could get a performance boost by making multiple specialised implementations here; but i think we leave that for another PR

libs/vm_modules/include/vm_modules/math/trigonometry.hpp
baykaner: same answer

wkaluza (approved):

PR #930, opened 2019-05-01
Embeddings fix

baykaner:
Previous implementation made multiple copies of entire tensor. This implementation just copies the slice out, does the work, and copies the slice back in.

This is extremely fast because instead of 3 copies of tensors of size VocabSize * EmbeddingsSize, we only have to make 4 copies of data of size 1 * EmbeddingsSize.

### DIFF
--- a/libs/ml/include/ml/ops/weights.hpp
+++ b/libs/ml/include/ml/ops/weights.hpp
@@ -168,6 +168,15 @@ public:
     return *this->output_;
   }
 
+  /**
+   * Returns a copy of embeddings gradients for enquiry
+   * @return
+   */
+  ArrayType Gradients()
+  {
+    return gradient_accumulation_->Copy();
+  }
+
   static constexpr char const *DESCRIPTOR = "Weights";
 
 private:
--- a/libs/ml/include/ml/ops/embeddings.hpp
+++ b/libs/ml/include/ml/ops/embeddings.hpp
@@ -96,7 +96,7 @@ public:
     }
     return {ArrayType(errorSignal.shape())};
   }
-  
+
   virtual void Step(typename T::Type learningRate)
   {
     if (updated_rows_.size() > 0)
--- a/libs/ml/include/ml/ops/embeddings.hpp
+++ b/libs/ml/include/ml/ops/embeddings.hpp
@@ -112,7 +112,7 @@ public:
         gradient_accumulation_slice.InlineMultiply(-learningRate);
         output_slice.InlineAdd(gradient_accumulation_slice);
 
-        this->gradient_accumulation_->Slice(r).Assign(gradient_accumulation_slice);
+        this->gradient_accumulation_->Slice(r).Assign(ArrayType::Zeroes(gradient_accumulation_slice.shape()));
         this->output_->Slice(r).Assign(output_slice);
       }
       updated_rows_.clear();
--- a/libs/ml/include/ml/ops/embeddings.hpp
+++ b/libs/ml/include/ml/ops/embeddings.hpp
@@ -105,7 +105,7 @@ public:
     {
       // get the relevant slice from gradients and embeddings
       auto grad_slice = this->gradient_accumulation_->Slice(r);
-      auto out_slice = this->output_->Slice(r);
+      auto out_slice  = this->output_->Slice(r);
 
       embedding_slice = out_slice.Copy();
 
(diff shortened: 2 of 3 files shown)


pierre-wilmot-fetch-ai:
Embeddings::Step role is to copy add the error signal to the embedding matrix.
In other word, copy values from an existing source buffer (this->gradient_accumulation_) to an existing destination buffer (this->output_).

In this updated piece of code we get : 

- L104 : Call to Copy : New memory allocation + data copy (that is iteration over src and dst buffer)
- L105 : Call to Copy : New memory allocation + data copy (that is iteration over src and dst buffer)

- L109 : Call to Copy : New memory allocation + data copy (that is iteration over src and dst buffer)
- L110 : Call to Copy : New memory allocation + data copy (that is iteration over src and dst buffer)

- L109 : Tensor::Operator= : Will trigger destructor for Tensor::data_
- L110 : Tensor::Operator= : Will trigger destructor for Tensor::data_

- L112 : InlineMultiply(Scalar) : Iteration over one array
- L113 : InlineAdd(Array) : Iteration over two arrays

- L115 : Assign(Array) : Iteration over 2 arrays
- L116 : Assign(Array) : Iteration over 2 arrays

- L120 : Exiting scope, calling destructor for temporatry objects  gradient_accumulation_slice & output_slice


This is a somewhat good example of why I've been complaining so much about the new tensor interface.

\+ At no point you zero out your gradient accumulation buffer (usually leads to a diverging loss)

For an example of how to do it more efficiantly, please see : https://github.com/pierre-wilmot-fetch-ai/fetch-ledger/blob/W2V/libs/ml/include/ml/ops/averaged_embeddings.hpp

baykaner:
This raises a good point though - none of our tests caught the fact that I didn't zero out. I think I'll add a test for that

pierre-wilmot-fetch-ai:
Embeddings::Step should be a zero allocation operation, all the buffers you need already exist.
You still have 3 allocations (repeated in a loop) : 

+ L110 explicit call to Copy
+ L113 explicit call to Copy
+ L116 allocation of a temporary through ArrayType::Zeroes

And you're still iterating over your arrays way more time than you need.

You should have only (updated_rows_.size() * EMBEDDING_SIZE * 2) call to iterator++;
If I'm not mistaken, you currently have (updated_rows_.size() * EMBEDDING_SIZE * 12)

12 = 
2 on L110 in Copy +
2 on L113 in Copy +
1 on L113 in InlineMultiply(Scalar) +
2 on L113 in InlineSubtract(Array) +
1 on L116 in ArrayType::Zeroes +
2 on L116 in Assign +
2 on L117 in Assign

Even though it is possible to do it right with the current Tensor implementation, using the iterators properly, I'm gonna say it one more time : 
**THE TENSOR SLICE INTERFACE DESIGN IS BAD AND LEADS TO BAD CODE.**

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -168,6 +168,15 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<T>
     return *this->output_;
   }
 
+  /**
+   * Returns a copy of embeddings gradients for enquiry
+   * @return
+   */
+  ArrayType Gradients()
```
pierre-wilmot-fetch-ai, line 175: Let's try to avoid adding function that are used only for unit test purposes.
You can chack that gradients are zeroed out properlly by calling step twice in your test and making sure the output doesn't change after second call.

pierre-wilmot-fetch-ai (changes requested):

libs/ml/include/ml/ops/weights.hpp
baykaner: I disagree.

The machine learning library is intended to be a general purpose tool. This means that A 3rd party should be able to use it to do any reasonable things even if we don't do it anywhere except in a unit test. I think getting a copy of the gradients of a node is a totally reasonable thing to do if, for example, you're debugging or experimenting with your own model/architecture (e.g. suppose you're wondering why the model doesn't perform as well as you'd hoped; you investigate and see some of the gradients are exploding and decide you want to add gradient clipping, or some other regularization).

I'm open to arguments about implementational details, though; it's not just weights that needs this method, so maybe this belongs in Trainable as well/instead; or somewhere else. For now I'm happy with this

baykaner:
redundant as changed merged into #983

PR #932, opened 2019-05-01
Private Network Support

### LINKED ISSUE #931: Allowing Private networks

ejfitzgerald:
Implement a concept of the network node to control when/how the miner identity is checked.

-standalone now means that the network is expected to be of size 1
-private-network now means that the the network is expected to be of size > 1

Absence of both flags (default) means that it is a public network

Fixes #931

### DIFF
--- a/apps/constellation/main.cpp
+++ b/apps/constellation/main.cpp
@@ -312,7 +312,9 @@ struct CommandLineArguments
     // configure the network mode
     if (standalone_flag && private_flag)
     {
-      std::cout << "Invalid configuration, unable to be in both the standalone and private network mode" << std::endl;
+      std::cout
+          << "Invalid configuration, unable to be in both the standalone and private network mode"
+          << std::endl;
       std::exit(1);
     }
     else if (standalone_flag)
@@ -330,7 +332,8 @@ struct CommandLineArguments
       // bootstrapping is not allowed in a non-public network
       if (NetworkMode::PUBLIC_NETWORK != args.cfg.network_mode)
       {
-        std::cout << "Unable to use bootstrapping servers with a private or standalone network" << std::endl;
+        std::cout << "Unable to use bootstrapping servers with a private or standalone network"
+                  << std::endl;
         std::exit(1);
       }
 
--- a/apps/constellation/main.cpp
+++ b/apps/constellation/main.cpp
@@ -229,30 +229,30 @@ struct CommandLineArguments
     bool private_flag{false};
 
     // clang-format off
-    p.add(args.port,                      "port",                  "The starting port for ledger services",                                         DEFAULT_PORT);
-    p.add(args.cfg.num_executors,         "executors",             "The number of executors to configure",                                          DEFAULT_NUM_EXECUTORS);
-    p.add(num_lanes,                      "lanes",                 "The number of lanes to be used",                                                DEFAULT_NUM_LANES);
-    p.add(args.cfg.num_slices,            "slices",                "The number of slices to be used",                                               DEFAULT_NUM_SLICES);
-    p.add(raw_peers,                      "peers",                 "The comma separated list of addresses to initially connect to",                 std::string{});
-    p.add(args.cfg.db_prefix,             "db-prefix",             "The directory or prefix added to the node storage",                             std::string{"node_storage"});
-    p.add(args.network_name,              "network",               "The network name to join",                                                      std::string{});
-    p.add(args.cfg.interface_address,     "interface",             "The address of the network interface to be used",                               std::string{"127.0.0.1"});
-    p.add(args.cfg.block_interval_ms,     "block-interval",        "Block interval in milliseconds",                                                uint32_t{DEFAULT_BLOCK_INTERVAL});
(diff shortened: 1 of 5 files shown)


apps/constellation/main.cpp
```diff
@@ -220,9 +249,10 @@ struct CommandLineArguments
...
     p.add(args.cfg.peers_update_cycle_ms, "peers-update-cycle-ms", "How fast to do peering changes",                                                uint32_t{0});
-    p.add(args.cfg.disable_signing,       "disable-signing",       "Do not sign outbound packets or verify those inbound, in trusted network",      bool{});
-    p.add(args.cfg.sign_broadcasts,       "sign-broadcasts",       "Sign and verify broadcast packets",                                             bool{});
-    p.add(args.cfg.standalone,            "standalone",            "Expect the node to run in on its own (useful for testing and development)",     false);
+    p.add(args.cfg.disable_signing,       "disable-signing",       "Do not sign outbound packets or verify those inbound, in trusted network",      false);
+    p.add(args.cfg.sign_broadcasts,       "sign-broadcasts",       "Sign and verify broadcast packets",                                             false);
+    p.add(standalone_flag,                "standalone",            "Expect the node to run in on its own (useful for testing and development)",     false);
```
wkaluza: "to be run in on its own" - remove "in"

apps/constellation/main.cpp
```diff
@@ -462,19 +528,13 @@ struct CommandLineArguments
     s << "external address..........: " << args.external_address << '\n';
     s << "db-prefix.................: " << args.cfg.db_prefix << '\n';
     s << "interface.................: " << args.cfg.interface_address << '\n';
-    s << "mining....................: " << ((args.cfg.block_interval_ms > 0) ? "Yes" : "No")
-      << '\n';
+    s << "mining....................: " << ToYesNo(args.cfg.block_interval_ms > 0) << '\n';
```
wkaluza, line 531: :+1:

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -433,5 +453,20 @@ MainChainRpcService::State MainChainRpcService::OnSynchronised(State current, St
...
+
+  // evaluate if this mining node is correct
+  bool const is_valid_miner =
+      (Mode::PUBLIC_NETWORK == mode_) ? crypto::IsFetchIdentity(block.body.miner) : true;
+  if (is_valid_miner && block.proof())
+  {
+    block_valid = true;
+  }
+
+  return block_valid;
```
wkaluza, line 468: How about `return is_valid_miner && block.proof();`? That's all the conditional says in this case.

apps/constellation/main.cpp
wkaluza: I'd tack on "Incompatible with -private-network" to the description.

apps/constellation/main.cpp
```diff
@@ -220,9 +249,10 @@ struct CommandLineArguments
...
-    p.add(args.cfg.disable_signing,       "disable-signing",       "Do not sign outbound packets or verify those inbound, in trusted network",      bool{});
-    p.add(args.cfg.sign_broadcasts,       "sign-broadcasts",       "Sign and verify broadcast packets",                                             bool{});
-    p.add(args.cfg.standalone,            "standalone",            "Expect the node to run in on its own (useful for testing and development)",     false);
+    p.add(args.cfg.disable_signing,       "disable-signing",       "Do not sign outbound packets or verify those inbound, in trusted network",      false);
+    p.add(args.cfg.sign_broadcasts,       "sign-broadcasts",       "Sign and verify broadcast packets",                                             false);
+    p.add(standalone_flag,                "standalone",            "Expect the node to run in on its own (useful for testing and development)",     false);
+    p.add(private_flag,                   "private-network",       "Expect the node is to be run as part of a private network (disables bootsrap)", false);
```
wkaluza: I'd tack on "Incompatible with -standalone" to the description.

wkaluza: Also, "bootsrap" => "bootstrap"

wkaluza (changes requested):
Looks fine, minor niggles only.

libs/ledger/src/protocols/main_chain_rpc_service.cpp
ejfitzgerald: You are of course right. In general, I think this sort of style is easier to read actually and it makes extending the logic in the future marginally easier. I did a double check in [godbolt](https://godbolt.org/) with the following and there is no difference at `-O3` with [example.cpp.txt](https://github.com/uvue-git/fetch-ledger/files/3137109/example.cpp.txt)

apps/constellation/main.cpp
ejfitzgerald: Haha my english is terrible - whoops good spot

apps/constellation/main.cpp
ejfitzgerald: ✅

libs/ledger/src/protocols/main_chain_rpc_service.cpp
wkaluza: That's all right then.

wkaluza (approved):

PR #933, opened 2019-05-01
State Machine Updates

### LINKED ISSUE #909: Rework StateMachine pattern
The current state machine requires that the user calls `Reset` in the desctructor of the parent class. This pattern could be improved to help with developer productivity

ejfitzgerald:
Update the state machine primitive so that the callbacks are removed on destruction. This means that the parent classes can simply destruct normally.

Closes #909


wkaluza (changes requested):
Looks fine, but can Reset() not be private now? Or, better yet, inlined in the destructor and removed entirely?

ejfitzgerald:
I was thinking along those lines too, but thought that it might be useful in the future. I have updated it so that it is private (so that it is a simple change if it is needed in the future).

wkaluza (approved):

PR #935, opened 2019-05-02
skipgram fixes & optimisations

baykaner:
- bugfixing and optimisations to Skipgram dataloader. Including also the embeddings fix from previous PR, the training SGNS W2V with text8 locally is now at around 1400 words/sec (compared with TF tutorial singly-threaded ~4000, and TF singly-threaded tutorial optimised ~9000).

- changed hyper-params to be in line with equivalent TF tutorial
- changed file reading to allow testing with text8 file


baykaner:
work continued in PR 978

PR #938, opened 2019-05-03
Jenkinsfile refactoring

wkaluza:
This is a rewrite of the ledger Jenkinsfile from the current Declarative Pipeline syntax to a dialect called Scripted Pipeline. This resulted in a shorter file which has less duplication and should be easier to maintain as it grows.

As this really is a complete rewrite, the diff view is of limited use for the purposes of review. I recommend that you read the file starting at the main() function, and I would be happy to sit down with the team and explain how Scripted and Declarative Pipeline code maps to each other, and discuss why I'm recommending a switch to Scripted Pipeline in the first place.

Two functional changes that I introduced along the way:

1. A 2h timeout for the parallel builds and a 3h total timeout for the entire run. I had one run freeze on me during the weekend. It is currently still running, with the clock indicating 30h. This will have to be stopped manually by someone with Admin Superpowers. I think it's good practice to have some sort of timeout on any but the most trivial builds to prevent these sort of hangups that require human intervention to resolve.
https://jenkins.economicagents.com/blue/organizations/jenkins/fetch-ledger/detail/PR-938-merge/85/pipeline/

2. The static analysis no longer runs in a 'ledger' (high-load) container, as I found that this does not actually decrease build times.

Another change that would occur has to do with the BlueOcean pipeline visualization: due to an active bug, circles representing sequential stages are displayed as a single circle when processing Scripted Pipeline. Only the visualization is affected: all logs and build information are accessible as normal.
https://issues.jenkins-ci.org/browse/JENKINS-53162

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -95,27 +95,29 @@ def create_build(Platform platform, Configuration config)
   return {
     stage(suffix) {
       node(HIGH_LOAD_NODE_LABEL) {
-        stage("SCM ${suffix}") {
-          checkout scm
-        }
+        timeout(120) {
+          stage("SCM ${suffix}") {
+            checkout scm
+          }
 
-        withEnv(environment()) {
-          DOCKER_IMAGE.inside {
-            stage("Build ${suffix}") {
-              sh "./scripts/ci-tool.py -B ${config.label}"
-            }
+          withEnv(environment()) {
+            DOCKER_IMAGE.inside {
+              stage("Build ${suffix}") {
+                sh "./scripts/ci-tool.py -B ${config.label}"
+              }
 
-            stage("Unit Tests ${suffix}") {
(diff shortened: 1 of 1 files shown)


chr15murray:
A note on Point 2:
We discussed this earlier but sharing for anyone else reading. 

Ledger nodes are default at the moment so not specifying a label for an agent currently results in the workload running on a ledger node. This will change once we have the ledger label propagated to all active ledger branches.

wkaluza:
@ejfitzgerald @chr15murray I checked the build logs and using a specific tag instead of 'latest' seems to have solved the problem. The digest for all Docker containers used during the build is the same.

ejfitzgerald (approved):
Looks good to me

chr15murray (approved):
All looks good to me and the test runs look consistent with what we have discussed
👍

PR #947, opened 2019-05-06
RPC request for unknown Txs

### LINKED ISSUE #873: Transaction sync fixes

wkaluza:
Closes #873

### DIFF
--- a/libs/ledger/include/ledger/chain/mutable_transaction.hpp
+++ b/libs/ledger/include/ledger/chain/mutable_transaction.hpp
@@ -32,6 +32,7 @@
 #include <functional>
 #include <set>
 #include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
 namespace fetch {
@@ -582,6 +583,8 @@ void Deserialize(T &serializer, Signatory &b)
 {
   serializer >> b.first >> b.second;
 }
+
+using TxDigestSet = std::unordered_set<TransactionSummary::TxDigest>;
 
 }  // namespace ledger
 }  // namespace fetch
--- a/libs/ledger/src/storage_unit/storage_unit_client.cpp
+++ b/libs/ledger/src/storage_unit/storage_unit_client.cpp
@@ -460,7 +460,7 @@ void StorageUnitClient::IssueCallForMissingTxs(TxDigestSet const &tx_set)
   for (auto const &hash : tx_set)
   {
     ResourceID resource{hash};
-    lanes_of_interest[LookupAddress(resource)].insert(resource);
+    lanes_of_interest[LookupAddress(resource)].insert(std::move(resource));
   }
 
   for (auto const &lane_resources : lanes_of_interest)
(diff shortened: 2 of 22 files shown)


libs/ledger/include/ledger/chain/mutable_transaction.hpp
```diff
@@ -583,5 +584,7 @@ void Deserialize(T &serializer, Signatory &b)
   serializer >> b.first >> b.second;
 }
 
+using TxDigestSet = std::unordered_set<TransactionSummary::TxDigest>;
```
ejfitzgerald, line 587: 👍

libs/ledger/src/storage_unit/storage_unit_client.cpp
```diff
@@ -453,6 +454,24 @@ bool StorageUnitClient::HasTransaction(ConstByteArray const &digest)
   return present;
 }
 
+void StorageUnitClient::IssueCallForMissingTxs(TxDigestSet const &tx_set)
+{
+  std::map<Address, std::unordered_set<ResourceID>> lanes_of_interest;
+  for (auto const &hash : tx_set)
+  {
+    ResourceID resource{hash};
+    lanes_of_interest[LookupAddress(resource)].insert(resource);
```
ejfitzgerald: Minor: Move here a little more efficient

libs/ledger/src/storage_unit/transaction_finder_protocol.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include "ledger/storage_unit/transaction_finder_protocol.hpp"
+
+namespace fetch {
+namespace ledger {
+
+/**
+ * Create a transaction store sync protocol
+ *
+ * @param p The protocol id for this service
```
ejfitzgerald: Looks like this documentation doesn't belong here.

On another note, I think that new files that we create we should probably try and document them

libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
```diff
@@ -302,8 +304,25 @@ TransactionStoreSyncService::State TransactionStoreSyncService::OnQueryObjects()
     return State::QUERY_OBJECTS;
   }
 
+  std::vector<ResourceID> rids;
+  ResourceID              rid;
+  while (tx_finder_protocol_->Pop(rid))
```
ejfitzgerald: It would be my preference to add a hard limit to the number of transactions that can be requested in a single iteration. In addition I would used this number to preallocate the `rids` vector

libs/ledger/tests/chain/fake_storage_unit.cpp
```diff
@@ -129,7 +132,8 @@ FakeStorageUnit::Hash FakeStorageUnit::CurrentHash()
 
 FakeStorageUnit::Hash FakeStorageUnit::LastCommitHash()
 {
-  assert(state_history_stack_.size() != 0);
+  assert(!state_history_stack_.empty());
```
ejfitzgerald, line 135: 👍

ejfitzgerald (changes requested):

ejfitzgerald (approved):

PR #949, opened 2019-05-07
Tensor concat

baykaner:
implementation that allows concatenation of tensors along a specified axis + tests & benchmarking


libs/math/include/math/tensor.hpp
```diff
@@ -2165,17 +2168,104 @@ template <typename T, typename C>
...
+ * sum of tensor sizes at that axis
+ * @param tensors
+ * @param axis
+ * @returnf
+ */
+template <typename T, typename C>
+typename Tensor<T, C>::SelfType Tensor<T, C>::Concat(std::vector<SelfType> &tensors,
+                                                     SizeType const         axis)
+{
+  // cant concatenate a single tensor
```
wkaluza, line 2196: Why not? Couldn't we just return the single tensor at this point? Then we have ASSERT(!tensors.empty()) here.

libs/math/tests/math/tensor/tensor_concat.cpp
```diff
@@ -0,0 +1,273 @@
...
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "math/tensor.hpp"
+#include <gtest/gtest.h>
+
+using namespace std;
+using namespace std::chrono;
```
wkaluza: Both of these are unnecessary, I think.

libs/math/tests/math/tensor/tensor_concat.cpp
```diff
@@ -0,0 +1,273 @@
...
+
```
wkaluza: First of all - thank goodness those tests are here. :)

The initial two tests in this file and the latter four have a lot in common. I'd be happier if this were deduplicated by pulling out common functions and constants to the fixture where possible. If the functions have sensible names, the test cases themselves would become more meaningful, short and punchy.

wkaluza (changes requested):
Looks good bar niggles.

libs/math/include/math/tensor.hpp
baykaner: okay - that should read "shouldnt" concatenate a single tensor. Don't forget this is a static function; we're not concatenate other tensors onto this one; we're concatenating a vector of tensors into a single Tensor. Whilst we could easily compute this for a vector of size() == 1 (i.e. return tensors[0];), I suspect that a user calling this function with a vector of length 1 has probably made a mistake and it's better to flag up that mistake

libs/math/tests/math/tensor/tensor_concat.cpp
baykaner: 👍

libs/math/include/math/tensor.hpp
wkaluza: Fair enough. I think there are ways we could be more explicit from an API design point of view, but as a first cut this is ok. Crashing the application seemed a disproportionate way to handle a case that we can handle gracefully, is all.

ejfitzgerald: Just to wade in a little more on this issue. It seems like a shame to create a new vector of tensors for the input to this function. This will simply be a source of more copies.

Does it make sense therefore to use std::initialiser_list<SelfType const *> const &

ejfitzgerald (changes requested):

libs/math/include/math/tensor.hpp
wkaluza: In any case, if we end up keeping this overload, tensors should be a const reference.

baykaner: @ejfitzgerald as discussed I agree it's a shame to create a vector of tensors, but for compatibility with ml ops (which are required to have a std::vector<> as input under current specifications) I don't see an obvious way around it.

baykaner: @wkaluza constness should be fixed now.

libs/ml/include/ml/ops/concatenate.hpp
```diff
@@ -0,0 +1,84 @@
...
+namespace ml {
+namespace ops {
+
+template <class T>
+class Concatenate : public fetch::ml::ElementWiseOps<T>
+{
+public:
+  using ArrayType    = T;
+  using DataType     = typename ArrayType::Type;
+  using ArrayPtrType = std::shared_ptr<ArrayType>;
```
wkaluza: You don't seem to use ArrayPtrType anywhere.

wkaluza: Or DataType for that matter.

libs/ml/include/ml/ops/concatenate.hpp
```diff
@@ -0,0 +1,84 @@
...
+  virtual ~Concatenate() = default;
+
+  /**
+   * concatenates multiple input tensors into one
+   */
+  virtual ArrayType Forward(std::vector<std::reference_wrapper<ArrayType const>> const &inputs,
+                            ArrayType &                                                 output)
+  {
+    std::vector<ArrayType> tensors;
+    for (std::size_t i = 0; i < inputs.size(); ++i)
```
wkaluza: Can you use a range-for loop here? for (auto const& input : inputs) { ... } ?

wkaluza (changes requested):
More niggles, minor this time.

libs/ml/include/ml/ops/concatenate.hpp
baykaner: 👍

wkaluza (approved):

ejfitzgerald (approved):

PR #950, opened 2019-05-07
Feature/re enable caching plus merkle

n-hutton:


ejfitzgerald:
Can you profile the public test net sync time with this combination of fixes

ejfitzgerald:
What is the score with this. If this will not land soon then I would suggest closing this PR and reopening  once you have got to the bottom of the test net issue.

PR #953, opened 2019-05-07
improve scripts

n-hutton:
Minor improvements to the end to end tests, improved fail messages, removed the expectation state as per Ed's suggestion.

I added support for pickling TXs and recovering them. They are not used since I am not sure we actually want to do this.

Note: Look at ``` DISABLED - Transactions sync - corner case with connection order ``` in the yaml - this may be a bug


scripts/end_to_end_test/end_to_end_test.yaml
```diff
@@ -0,0 +1,83 @@
...
+        Send lots of transactions to a node
+
+setup_conditions:
+    test_name : "miner_recieves_more_tx"
+    number_of_nodes: 1
+    mining_nodes: [0]
+    max_test_time: 60
+
+steps:
+    - send_txs : { "name": "100_identities", "amount": 100, "nodes" : [0] }
```
n-hutton, line 80: Sadly this takes a fair amount of time to generate the TXs - I was considering pickling them but this merits discusison.

wkaluza (approved):

PR #954, opened 2019-05-07
Correct Reverse RPC Muddle example

ejfitzgerald:


ejfitzgerald:
FYI: @troelsfr

libs/network/examples/reverse_rpc/client.cpp
```diff
@@ -65,78 +69,82 @@ class AEA
...
-    for (;;)
-    {
-    }
+    FETCH_LOG_INFO(LOGGING_NAME, "Registering node...FAILED");
+    return EXIT_FAILURE;
   }
+  FETCH_LOG_INFO(LOGGING_NAME, "Registering node...complete");
 
-  tm.Stop();
+  // run the main server loop while we have directly connected peers
```
wkaluza, line 141: Pretend to run? I'm guessing the sleep is there to simulate a workload.

libs/network/examples/reverse_rpc/client.cpp
```diff
@@ -65,78 +69,82 @@ class AEA
...
-  client_muddle->Start({});
 
-  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  // create the muddle and attach all the RPC services
+  auto muddle = Muddle::CreateMuddle(NetworkId{"TEST"}, tm);
+  auto client = std::make_shared<Client>("Client", muddle->AsEndpoint(), Muddle::Address(),
+                                         SERVICE_TEST, CHANNEL_RPC);
+
+  // register the RPC server
+  auto server = std::make_shared<Server>(muddle->AsEndpoint(), SERVICE_TEST, CHANNEL_RPC);
```
wkaluza: There is no reason to make server a shared_ptr here; an instance on the stack would look better.

libs/network/examples/reverse_rpc/server.cpp
```diff
@@ -17,126 +17,139 @@
...
+  using RpcServerPtr = std::shared_ptr<Server>;
+
   ClientRegister    client_register_;
+  AEAToNodeProtocol aea_to_node_protocol_{&client_register_};
+  RpcServerPtr      rpc_server_{};
 };
 
 int main()
 {
+  // create and start the
```
wkaluza: ...network manager? :)

libs/network/examples/reverse_rpc/client.cpp
```diff
@@ -27,36 +26,41 @@
...
-const int CHANNEL_RPC  = 1;
+static char const *LOGGING_NAME = "RPC-Client";
 
 class AEA
 {
 public:
-  std::string SearchFor(std::string val)
-  {
-    std::cout << "Searching for " << val << std::endl;
+  using StringList = std::vector<std::string>;
```
wkaluza: Can we call this something more generic like Strings or StringCollection? We shouldn't call it a list unless it's actually a list, that's just misleading.

wkaluza (changes requested):
LGTM apart from changes indicated.

A general remark, out of scope for this PR - shouldn't we put examples in a namespace? Either jointly as fetch::examples, or something more granular like fetch::network::examples. We're doing this work in the open, and example code should be, well, exemplary. Maybe create an issue to track this for later?

libs/network/examples/reverse_rpc/client.cpp
ejfitzgerald: Actually no. In the Reverse RPC case the client is actually running the RPC server. This is why this loop is more like a monitoring loop. When no one is connected (because the server has disconnected or similar) the monitor loop will exit and tear down.

libs/network/examples/reverse_rpc/client.cpp
ejfitzgerald: You are correct. I was just following what was there, but that is not a good reason to keep it the same 😁

libs/network/examples/reverse_rpc/server.cpp
ejfitzgerald: Done

libs/network/examples/reverse_rpc/client.cpp
ejfitzgerald: Fair point

ejfitzgerald:
@wkaluza - Yes at some point we need to do a bit of an overhaul of all the examples in the project. They originally where a mechanism for testing smaller subsets of the functionality. I think we should probably reorganise and remove the old ones.

In general, we also need to think about a nice way to test these. Since this is a good example that has gone stale over time.

wkaluza (approved):

PR #955, opened 2019-05-07
Python code style

wkaluza:
Enforce Python code style automatically. Reformat repo to follow new style. Manual fixes were applied in places where the formatter got confused.

The file `libs/python/setup.py.in` got pretty mangled by the formatter so I had to refactor it. It does not appear to be used anywhere. There are a few other files that are suspect and potentially deserve deletion. @baykaner and @pierre-wilmot-fetch-ai Could you have a look at the 4 files under libs/python/examples in this PR and let me know if they are of use to anyone?


scripts/ci-tool.py
```diff
@@ -74,22 +78,26 @@ def create_junit_format(output_path, data):
...
         group_size = len(group_data)
         group_time = sum(map(lambda x: x[3], group_data))
-        group_failures = sum(map(lambda x: 0 if x[4] == 'passed' else 1, group_data))
+        group_failures = sum(
+            map(lambda x: 0 if x[4] == 'passed' else 1, group_data))
 
         all_tests += group_size
         all_time += group_time
-        all_failures ++ group_failures
+        all_failures + + group_failures
```
ejfitzgerald: This looks like a bug! 😨

wkaluza: Aye, there's a few others I need to look at, mostly to do with autopep8 rearranging imports in a potentially breaking way. I'll have gone through them by Friday. Then we can add Python to the CI style-check.

Our CMake scripts also deserve cleanup, but there's a Python package for that, too. That's for next week.

n-hutton:
Do we want to make python style enforcement part of the CI? It might be worth waiting to do this until our CI automatically creates a linted branch that it will merge for you which has been suggested

wkaluza:
@n-hutton Linters can break code, even clang-format can be tricked into doing it. I think we should integrate changes manually until we know where the limits are.

Also, in our workflow letting CI automatically merge linting changes would mean giving Jenkins write permission for our GitHub profiles. I'm not thrilled about that thought, unless there would be a way to create an access token that's only valid for uvue-git forks, or something along those lines.

n-hutton (commented):
Looks fine to me - I don't write much python code though so I defer to Khan or Ed maybe

scripts/apply_style.py
```diff
@@ -154,19 +190,17 @@ def postprocess_file(filename):
         destination.writelines(postprocess_contents(contents))
 
 
-def project_sources(project_root):
-
+def project_sources(PROJECT_ROOT):
```
ejfitzgerald: This looks broken

apps/demoswarm/runswarm.py
```diff
@@ -24,36 +24,70 @@
 class RunSwarmArgs(object):
```
ejfitzgerald, line 24: FYI: This is one of the files the can be removed on the next pass

ejfitzgerald (dismissed):

scripts/apply_style.py
wkaluza: Ah, yes, good spot.

apps/demoswarm/runswarm.py
wkaluza: :+1: :gun: take no prisoners

pierre-wilmot-fetch-ai:
Files in libs/python/examples are just examples of how to use the library, they do not do anything specific, probably belongs into the documentation rather than here.

Feel free to delete them 😄

wkaluza:
> Feel free to delete them

Music to my ears. :smile:

baykaner (approved):

ejfitzgerald (approved):

PR #956, opened 2019-05-07
Address Sanitiser Fixes

ejfitzgerald:


scripts/ci-tool.py
```diff
@@ -223,6 +228,8 @@ def test_project(build_root, label):
         sys.exit(exit_code)
 
 def test_end_to_end(project_root, build_root):
+    import run_end_to_end_test
```
ejfitzgerald, line 225: This means that you only need to have `fetch.cluster` installed when running the end 2 end tests

libs/storage/tests/gtest/mmap_random_access_stack_tests.cpp
```diff
@@ -270,7 +270,7 @@ TEST(DISABLED_mmap_random_access_stack, set_bulk)
   }
 }
 
-TEST(mmap_random_access_stack, file_writing_and_recovery)
+TEST(mmap_random_access_stack, DISABLED_file_writing_and_recovery)
```
ejfitzgerald, line 273: Have raised #958 to track this issue. Basically the MMP RAS is mostly broken. Since we are not using it raise a bug and move on. 

At this point I am half temped to simply remove it in one commit which we can revert it at another date. @n-hutton thoughts?

n-hutton: I don't think it's so heavily broken. But yes, I don't disagree with removing it and also some of the other random access stacks that are legacy/unused. I can do this if you like.

scripts/ci-tool.py
n-hutton: This does mean that the test only breaks when this import fails on merge-pr which could be annoying

libs/storage/include/storage/key.hpp
```diff
@@ -43,18 +43,25 @@ namespace storage {
 template <std::size_t BITS = 256>
 struct Key
 {
+  using Block = uint64_t;
```
n-hutton, line 46: where have these changes come from?

libs/core/src/json/document.cpp
```diff
@@ -399,7 +399,7 @@ void JSONDocument::Tokenise(ConstByteArray const &document)
       ++pos;
       break;
     case '}':
-      if (brace_stack_.back() != '}')
+      if (brace_stack_.empty() || brace_stack_.back() != '}')
```
n-hutton, line 402: is any of this related to clean shutdown of constellation?

n-hutton (approved):

PR #966, opened 2019-05-08
Testnet http bug

wkaluza:
This adds a test for preventing a segfault in HTTPRequest when the buffer is empty, which had been observed in the test net. This test and the associated fix will then be cherry-picked to the release branch.


ejfitzgerald (dismissed):

ejfitzgerald (dismissed):

n-hutton (approved):
lgtm

Issue #843, opened 2019-04-11
Add proper support for NaN/Infinity/etc in FixedPoint<> class


markos:
Add proper support for NaN/Infinity/etc in FixedPoint<> class

PR #968, opened 2019-05-08
Feature/better fixed point sqrt

### LINKED ISSUE #629: Fix Fixed Point CheckNoRounding
Current implementation is wrong

markos:
* Add working Sqrt() FixedPoint implementation
* Fix bugs in corner cases of Exp(), Log(), Pow(), add respective test cases.
* Provide constants for tolerance, decimal_digits for FixedPoint.
* Fix some testcases with hardcoded tolerance values.
* Add initial NaN support
* Some style fixes

Closes: #629 , #843


libs/math/include/math/meta/math_type_traits.hpp
```diff
@@ -131,6 +131,30 @@ template <typename DataType, typename ReturnType>
 using IfIsMathNonFixedPointArray =
     IfIsNotFixedPoint<typename DataType::Type, IfIsMathArray<DataType, ReturnType>>;
 
+template <typename DataType>
+fetch::meta::IfIsFixedPoint<DataType, DataType> static tolerance()
```
baykaner: Okay, so a couple things:

First either these functions should live in math/base_types.hpp OR the functions in there (NumericMax, NumericMin, NumericLowest) should be moved here: I guess I don't really have any opinion on which happens; but it feels like they should all live together to me.

Also! You are correct to make these snake_case functions, but I was previously incorrect to make NumericMax etc. PascalCase (since they're cheap!). Would you mind snake_case ifying those functions when you move them all to the same place please?

libs/math/include/math/meta/math_type_traits.hpp
```diff
@@ -131,6 +131,30 @@ template <typename DataType, typename ReturnType>
...
+}
+
+template <typename DataType>
+fetch::meta::IfIsFixedPoint<DataType, DataType> static minimum_tolerance()
+{
+  return DataType::CONST_SMALLEST_FRACTION;
+}
+
+template <typename DataType>
+fetch::meta::IfIsFloat<DataType, DataType> static minimum_tolerance()
```
baykaner: I'm probably being dense, but can you explain the difference between a tolerance and a minimum tolerance please? (and then write it in a function comment :-P )

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -311,74 +272,143 @@ class FixedPoint : public BaseFixedpointType
...
-  ////////////////////////////
+  ///////////////////////////////////////////////////
+  /// comparison operators for FixedPoint objects ///
+  ///////////////////////////////////////////////////
+
+  constexpr bool operator==(FixedPoint const &o) const
+  {
+    if (isNaN(*this) || isNaN(o))
+    {
+      return false;
```
baykaner, line 330: what if they're both NaN? aren't they equal in that case? :-p it might be moot point

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -311,74 +272,143 @@ class FixedPoint : public BaseFixedpointType
...
+    }
+    else
+    {
+      return (data_ < o.Data());
+    }
+  }
+
+  template <typename OtherType>
```
baykaner: why is there a template here?
is this operator tested anywhere? if so doesn't the compiler complain it can't figure out what the template is meant to be? if not, shouldn't we test this?

libs/core/include/core/fixed_point/fixed_point.hpp
```diff
@@ -683,22 +743,102 @@ class FixedPoint : public BaseFixedpointType
...
   }
 
-  static constexpr FixedPoint Log10(const FixedPoint &x)
+  static constexpr FixedPoint Log10(FixedPoint const &x)
   {
     return Log2(x) / CONST_LOG210;
   }
 
-  static constexpr FixedPoint Sqrt(const FixedPoint &x)
+  static constexpr int Reduce_Sqrt(FixedPoint &x)
```
baykaner: strange naming convention here. I'm guessing this should be ReduceSqrt ?

baykaner: would it be a good idea to wrap this in an anonymous namespace (or a namespace called details/implementations or similar)? or at least move this implementation into the private part of fixed point?

baykaner (changes requested):

libs/core/include/core/fixed_point/fixed_point.hpp
markos: by definition a NaN != NaN (yes, I just recently learned that myself), so I cannot check or assume equality, so I have to check both.

baykaner: aah!! IEEE this is witchcraft!
but I guess it makes sense when the whole point of the NaN is to flag up "something went seriously wrong"

libs/math/include/math/meta/math_type_traits.hpp
markos: I agree, and I'm going to do the changes, but first let's agree on which file they should be pleaced :)

libs/math/include/math/meta/math_type_traits.hpp
markos: yeah, that's ambiguous, one is the type's tolerance (minimum_tolerance) and the other the precision of most of the function approximations (Exp/Pow/Log/etc) as they are not accurate down to the last digit for performance reasons. Initially I had implemented the first, but then it became apparent that I needed something more.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: that is a leftover by mistake, apparently the compiler didn't mind. fixed.

libs/core/include/core/fixed_point/fixed_point.hpp
markos: yup, it will be extra fun trying to get this working in SIMD also :)

libs/core/include/core/fixed_point/fixed_point.hpp
markos: Fixed both locally, will send commit soon.

libs/math/include/math/meta/math_type_traits.hpp
markos: perhaps a solution might be to rename minimum_tolerance that refers to the type to maximum_precision (though I also don't feel great about this) and leave tolerance() as it is.

libs/math/include/math/meta/math_type_traits.hpp
baykaner: If a solution still hasn't been agreed upon I recommend you move them to math_type_traits.hpp on the basis that NumericMax and NumericMin are traits of types.

libs/math/include/math/meta/math_type_traits.hpp
baykaner: How about you rename minimum_tolerance to "tolerance", and you rename the other one to function_tolerance, functional_tolerance, or result_tolerance or something similar; at least then it would be doing what the name says

libs/math/include/math/meta/math_type_traits.hpp
markos: moved to math_type_traits.hpp.

libs/math/include/math/meta/math_type_traits.hpp
markos: I checked again where it was used, and the "minimal_tolerance" was only used in the TSNE class in this particular form and values (1e-12 for float/double). So instead of making it generic, I just renamed it to tsne_tolerance and put it in the TSNE class.

baykaner (dismissed):

baykaner (approved):

n-hutton (approved):

PR #969, opened 2019-05-09
Add sigmoid benchmark

pierre-wilmot-fetch-ai:
Adding some benchmark to see if it's worth optimising the sigmoid a bit.
The Google Word2Vec code does precompute the values for exponenetial, may be worth doing it too.

Curent version loops over the array 2 times : 
- To compute the sigmoid
- To clamp all values in range [epsilon, 1 - epsilon]

Current benchmark values : 

Screenshot 2019-05-09 at 11 28 57

### DIFF
--- a/libs/ml/benchmark/CMakeLists.txt
+++ b/libs/ml/benchmark/CMakeLists.txt
@@ -10,5 +10,5 @@ include(${FETCH_ROOT_CMAKE_DIR}/BuildTools.cmake)
 # Compiler Configuration
 setup_compiler()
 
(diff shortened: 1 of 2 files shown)


baykaner (changes requested):
That looks fine but would mind also adding the equivalent benchmark for fetch::math::Sigmoid to math/benchmarks.

This is where Troels' ApproxExp function (math/approx_exp.hpp) could probably be useful (seems like it already has a test and benchmark, but might need some updating).

baykaner:
closing and picking up work to go forward in #1068

PR #970, opened 2019-05-09
Code2vec dataloader

juergendkeba:
Implemented a dataloder for loading data in the format proposed in the code2vec python implementation


libs/ml/examples/code2vec_dataloading/main.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "math/tensor.hpp"
+#include "ml/dataloaders/code2vec_context_loaders/context_loader.hpp"
+#include <fstream>
+#include <iostream>
+
+using DataType  = uint64_t;
+using ArrayType = fetch::math::Tensor<DataType>;
+using SizeType  = fetch::math::Tensor<DataType>::SizeType;
+
+std::string readFile(std::string const &path)
```
baykaner: style: ReadFile

libs/ml/examples/code2vec_dataloading/main.cpp
```diff
@@ -0,0 +1,61 @@
...
+int main(int ac, char **av)
+{
+
+  if (ac < 2)
+  {
+    std::cerr << "Usage: " << av[0] << " INPUT_FILES_TXT" << std::endl;
+    return 1;
+  }
+
+  fetch::ml::dataloaders::C2VLoader<std::tuple<ArrayType, ArrayType, ArrayType>, SizeType> cloader(
```
baykaner: maybe add something like:

using LabelType = SizeType;
using ContextPathType = std::tuple<ArrayType, ArrayType, ArrayType>;

fetch::ml::dataloaders::C2VLoader<ContextPathType, SizeType> cloader(...

libs/ml/examples/code2vec_dataloading/main.cpp
```diff
@@ -0,0 +1,61 @@
...
+      20);
```
baykaner: please add something like:

#define MAX_CONTEXTS 20;

cloader(MAX_CONTEXTS);

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
```diff
@@ -0,0 +1,451 @@
...
+  umap_str_int GetUmapPathToIdx();
+  umap_str_int GetUmapWordToIdx();
+
+  umap_str_int GetCounterFunctionNames();
+  umap_str_int GetCounterPaths();
+  umap_str_int GetCounterWords();
+  std::vector<std::pair<std::tuple<SizeType, SizeType, SizeType>, LabelType>> data;
+
+private:
+  SizeType iterator_position_get_next_context;
```
baykaner: style:
all of these private variables need a trailing _

baykaner: it's also a good idea in general to give a default initialisation for debugging purposes - usually best to put something that will stand out if something has gone wrong. For example for the SizeTypes you might do this:
SizeType iterator_position_get_next_context_{fetch::math::NumericMax<SizeType>()};
that way if you're debugging and you see one of these values has a ridiculously large number you'll know it was never assigned when it should have been.

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
```diff
@@ -0,0 +1,451 @@
...
+  /**
+   * @brief Adds (raw) data in the c2v format and creates the lookup maps
+   *
+   * @param text
+   */
+  void AddData(std::string const &text);
+
+  void createIdxUMaps();
+
+  umap_int_str GetUmapIdxToFunctionName();
```
baykaner: all of these Get something functions need to be renamed as follows:
First, I think we can drop the Get if the functions are named well it should still be clear what they do.
Second, these getters should all be snake_case; this is the exception to the rule about functions being PascalCase. If they're extremely fast (i.e. they just return a reference to an object, or they just return a copy of an integer) they should be snake_case. This was confusing for me at first, but is actually very helpful when optimising since you can see what code is safe to put inside a loop and what will be costly.

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
```diff
@@ -0,0 +1,451 @@
...
+  umap_int_str GetUmapIdxToWord();
+
+  umap_str_int GetUmapFunctioNameToIdx();
+  umap_str_int GetUmapPathToIdx();
+  umap_str_int GetUmapWordToIdx();
+
+  umap_str_int GetCounterFunctionNames();
+  umap_str_int GetCounterPaths();
+  umap_str_int GetCounterWords();
+  std::vector<std::pair<std::tuple<SizeType, SizeType, SizeType>, LabelType>> data;
```
baykaner, line 126: I think you need a typedef or two, something like:
using ContextPathIdxType = typename std::tuple<SizeType, SizeType, SizeType>;
using DataPairType = typename std::pair<ContextPathIdxType, LabelType>;

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
```diff
@@ -0,0 +1,451 @@
...
+   * @param counter unordered map with counts of words
+   * @param name_to_idx unordered map for storing the mapping string->numeric
+   * @param idx_to_name unordered map for storing the mapping numeric->string
+   */
+  static void createIdxUMapsFromCounter(umap_str_int &counter, umap_str_int &name_to_idx,
+                                        umap_int_str &idx_to_name);
+
+  /**
+   * @brief Add value to the unordered map for counting strings. If the provided string exists, the
+   * counter is set +1, otherwise its set to 1
```
baykaner, line 159: should this say -1?

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
```diff
@@ -0,0 +1,451 @@
...
+
+  /**
+   * @brief Creates an unordered map for hashing strings from a counter (unordered map counting the
+   * occurences of words in the input)
+   *
+   * @param counter unordered map with counts of words
+   * @param name_to_idx unordered map for storing the mapping string->numeric
+   * @param idx_to_name unordered map for storing the mapping numeric->string
+   */
+  static void createIdxUMapsFromCounter(umap_str_int &counter, umap_str_int &name_to_idx,
```
baykaner, line 154: Expensive functions are supposed to be PascalCase not camelCase

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
```diff
@@ -0,0 +1,451 @@
...
+                                        umap_int_str &idx_to_name);
+
+  /**
+   * @brief Add value to the unordered map for counting strings. If the provided string exists, the
+   * counter is set +1, otherwise its set to 1
+   *
+   * @param umap the unordered map where the counts are stored
+   * @param word the string to be counted.
+   */
+  static void addValueToCounter(umap_str_int &umap, std::string word);
```
baykaner, line 164: please check all function casing

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
baykaner: I think the counter can be a const & right?

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
baykaner: word can be a std::string const & right?

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
```diff
@@ -0,0 +1,451 @@
...
+
+  /**
+   * @brief Adds a string to the unordered maps for hashing and outputs the hashing index
+   *
+   * @param input input string
+   * @param name_to_idx unordered map for mapping string->numeric
+   * @param idx_to_name unordered map for mapping numeric->string
+   * @return SizeType index of the string in the unordered maps
+   */
+  SizeType addToIdxUMaps(std::string input, umap_str_int &name_to_idx, umap_int_str &idx_to_name);
```
baykaner: can input be a std::string const & ?

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
```diff
@@ -0,0 +1,451 @@
...
+  ContextLabelPair GetNextContext();
+
+  ContextTensorsLabelPair GetNext();
+
+  /**
+   * @brief Gets the number of feature/target pairs
+   *
+   * @return std::uint64_t
+   */
+  std::uint64_t Size() const;
```
baykaner: when you override a virtual method you should put override at the end of the declaration.

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
```diff
@@ -0,0 +1,451 @@
...
+
+  /**
+   * @brief Indicates if the GetNextContext() generater is at its end
+   *
+   * @return true
+   * @return false
+   */
+  bool IsDone() const;
```
baykaner: override

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
baykaner: incidentally, Size and IsDone should both be lower case but I understand that this is wrong in the DataLoader from which you inherit. I think leave it and I'll fix it in another PR

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
```diff
@@ -0,0 +1,451 @@
...
+   * @param input input string
+   * @param name_to_idx unordered map for mapping string->numeric
+   * @param idx_to_name unordered map for mapping numeric->string
+   * @return SizeType index of the string in the unordered maps
+   */
+  SizeType addToIdxUMaps(std::string input, umap_str_int &name_to_idx, umap_int_str &idx_to_name);
+};
+
+template <typename DataType, typename LabelType>
+void C2VLoader<DataType, LabelType>::AddData(std::string const &c2v_input)
```
baykaner, line 187: function comment please

libs/ml/include/ml/dataloaders/code2vec_context_loaders/context_loader.hpp
```diff
@@ -0,0 +1,451 @@
...
+{
+  auto return_value = this->data[this->iterator_position_get_next_context];
+  this->iterator_position_get_next_context += 1;
+  return return_value;
+}
+
+template <typename DataType, typename LabelType>
+typename C2VLoader<DataType, LabelType>::ContextTensorsLabelPair
+C2VLoader<DataType, LabelType>::GetNext()
+{
```
baykaner, line 246: let's have a conference call and discuss this function, since . this seems to be where most of the work is done.
I'd like you to talk me through what's happening in this function and we can discuss design choices.

It looks like a lot of work is happening here, but in practice a training loop is always going to look something like this:

// one-time setup code
while (not_last_epoch)
{
  data_pair = dataloader.next();

  model.set_inputs(data_pair);
  pred = model.predict();
  loss = criterion(pred, gt);
  model.backward();
}

that means that if we can make design choices that frontload the work into that one-time setup code so that next() is very cheap, we're going to massively speed up training time.

It's not important at this stage to do any extreme optimisation, but I suspect that there are probably a few basic design choice changes we can make here to massively speed up training; but I'd like you to talk me through this function so we can figure that out.

baykaner (changes requested):

baykaner (approved):

PR #971, opened 2019-05-09
Remove dead benchmark

wkaluza:
Is this file worth anything? It seems it's been forgotten since November - can someone determine if it's reusable, or if it is preferable to scrap it and start over. It uses obsolete APIs and includes non-existent headers, so it cannot be built without changes.


libs/ledger/benchmark/tx_syncro.cpp
```diff
@@ -1,353 +0,0 @@
-//------------------------------------------------------------------------------
```
ejfitzgerald, line 1: There should be an issue to reinstate this benchmark. I am not sure that we want to delete it actually

ejfitzgerald (changes requested):

libs/ledger/benchmark/tx_syncro.cpp
wkaluza: I created issue #972 and pasted the code there for the record.

ejfitzgerald (approved):

PR #975, opened 2019-05-10
Miscellaneous changes

wkaluza:
This week I bring you the usual mix of removed typos, unused headers, constants and functions, some refs made const, some goto-fail errors waiting to happen, and a little Python context manager magic sprinkled in for good measure.


libs/http/include/http/request.hpp
```diff
@@ -76,11 +76,6 @@ class HTTPRequest
     return query_;
   }
 
-  std::size_t const &header_length() const
```
ejfitzgerald: I would keep this, while it is not used still a useful to keep

libs/ledger/include/ledger/protocols/main_chain_rpc_service.hpp
```diff
@@ -98,23 +98,16 @@ class MainChainRpcService : public muddle::rpc::Server,
 
   void BroadcastBlock(Block const &block);
 
-  State state() const
```
ejfitzgerald: I suspect there are going to be a few of these to will abbrev with "keep"

libs/metrics/include/metrics/metrics.hpp
```diff
@@ -39,47 +39,28 @@ class Metrics
   // Singleton instance
   static Metrics &Instance();
 
-  // Construction / Destruction
```
ejfitzgerald: Keep

libs/metrics/include/metrics/metrics.hpp
```diff
@@ -39,47 +39,28 @@ class Metrics
...
   Metrics(Metrics const &) = delete;
   Metrics(Metrics &&)      = delete;
-  ~Metrics()
```
ejfitzgerald, line 45: Keep - want to have the ability to replace the metric handler

libs/metrics/src/metrics.cpp
```diff
@@ -22,6 +22,14 @@
 namespace fetch {
 namespace metrics {
 
+Metrics::Metrics() = default;
```
ejfitzgerald: Keep in header file

libs/metrics/src/metrics.cpp
```diff
@@ -22,6 +22,14 @@
...
+
+Metrics::~Metrics()
```
ejfitzgerald: Lets keep the original

libs/metrics/include/metrics/metrics.hpp
```diff
@@ -39,47 +39,28 @@ class Metrics
...
-  {
-    RemoveMetricHandler();
-  }
 
   // Configuration
   void ConfigureFileHandler(std::string filename);
 
   void RecordMetric(ConstByteArray const &identifier, Instrument instrument, Event event,
-                    Timestamp const &timestamp = Clock::now())
```
ejfitzgerald: These were all in the header for a reason so that they get inlined (wherever possible). This was in an effort to minimise the noise from the creating the first timestamp

ejfitzgerald (changes requested):
There are lots of good bits here. I think the metric changes are unnecessary and there are some unused functions that have been removed that remove features which is probably non-great

libs/http/include/http/request.hpp
wkaluza: That seems contradictory but ok.

libs/ledger/include/ledger/protocols/main_chain_rpc_service.hpp
wkaluza: Done

libs/metrics/include/metrics/metrics.hpp
wkaluza: Done

libs/metrics/include/metrics/metrics.hpp
wkaluza: Done

libs/metrics/src/metrics.cpp
wkaluza: 👍

libs/metrics/src/metrics.cpp
wkaluza: 👍

libs/metrics/include/metrics/metrics.hpp
wkaluza: Eh? I understand the inlining bit, but what noise are we talking about?

libs/metrics/include/metrics/metrics.hpp
wkaluza: I reverted most changes to this file, but I'd quite like to make the destructor private, as it underlines that this class is self-owning (more so than a private ctor alone).

ejfitzgerald (approved):

baykaner (approved):

PR #977, opened 2019-05-10
Feature/sync performance improvements2

n-hutton:
Reduce syncing times by 100x - path to common ancestor was being called unneccessarily often during block coord syncing.

### DIFF
--- a/libs/ledger/include/ledger/chain/main_chain.hpp
+++ b/libs/ledger/include/ledger/chain/main_chain.hpp
@@ -118,8 +118,9 @@ public:
   BlockHash GetHeaviestBlockHash() const;
   Blocks    GetHeaviestChain(uint64_t limit = UPPER_BOUND) const;
   Blocks    GetChainPreceding(BlockHash at, uint64_t limit = UPPER_BOUND) const;
-  bool      GetPathToCommonAncestor(Blocks &blocks, BlockHash tip, BlockHash node,
-                                    uint64_t limit = UPPER_BOUND, BehaviourWhenLimit behaviour = BehaviourWhenLimit::RETURN_MOST_RECENT) const;
+  bool      GetPathToCommonAncestor(
+           Blocks &blocks, BlockHash tip, BlockHash node, uint64_t limit = UPPER_BOUND,
+           BehaviourWhenLimit behaviour = BehaviourWhenLimit::RETURN_MOST_RECENT) const;
   /// @}
 
   /// @name Tips
@@ -247,8 +248,9 @@ bool MainChain::StripAlreadySeenTx(BlockHash starting_hash, T &container) const
   {
     ++blocks_checked;
 
-    // Traversing the chain fully is costly: break out early if we know the transactions are all duplicated (or empty)
-    if(transactions_to_check.size() == transactions_duplicated.size())
+    // Traversing the chain fully is costly: break out early if we know the transactions are all
+    // duplicated (or empty)
+    if (transactions_to_check.size() == transactions_duplicated.size())
     {
       break;
     }
--- a/libs/ledger/include/ledger/chain/main_chain.hpp
+++ b/libs/ledger/include/ledger/chain/main_chain.hpp
@@ -132,7 +132,7 @@ public:
   /// @name Missing / Loose Management
   /// @{
   BlockHashSet GetMissingTips() const;
-  BlockHashs   GetMissingBlockHashes(std::size_t limit = UPPER_BOUND) const;
+  BlockHashs   GetMissingBlockHashes(uint64_t limit = UPPER_BOUND) const;
   bool         HasMissingBlocks() const;
   /// @}
 
--- a/libs/ledger/include/ledger/chain/main_chain.hpp
+++ b/libs/ledger/include/ledger/chain/main_chain.hpp
@@ -247,6 +247,7 @@ bool MainChain::StripAlreadySeenTx(BlockHash starting_hash, T &container) const
   {
     ++blocks_checked;
 
+    // Traversing the chain fully is costly: break out early if we know the transactions are all duplicated (or empty)
     if(transactions_to_check.size() == transactions_duplicated.size())
     {
       break;
--- a/libs/ledger/include/ledger/chain/main_chain.hpp
+++ b/libs/ledger/include/ledger/chain/main_chain.hpp
@@ -90,6 +90,15 @@ public:
     LOAD_PERSISTENT_DB
   };
 
+  // When traversing the chain and returning a subset due to hitting a limit,
+  // either return blocks closer to genesis (least recent in time), or
+  // return closer to head (most recent)
+  enum class BehaviourWhenLimit
+  {
+    RETURN_MOST_RECENT = 0,
+    RETURN_LEAST_RECENT
+  };
+
   // Construction / Destruction
   explicit MainChain(Mode mode = Mode::IN_MEMORY_DB);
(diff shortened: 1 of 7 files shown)


libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -238,6 +248,13 @@ bool MainChain::StripAlreadySeenTx(BlockHash starting_hash, T &container) const
   {
     ++blocks_checked;
 
+    // Traversing the chain fully is costly: break out early if we know the transactions are all
+    // duplicated (or empty)
+    if (transactions_to_check.size() == transactions_duplicated.size())
```
n-hutton, line 253: this means mainchain when mining empty blocks doesn't take +30s (does though for non empty blocks of course)

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -39,6 +39,7 @@ const std::chrono::milliseconds EXEC_NOTIFY_INTERVAL{500};
 const std::chrono::seconds      NOTIFY_INTERVAL{10};
 const std::chrono::seconds      WAIT_BEFORE_ASKING_FOR_MISSING_TX_INTERVAL{30};
 const std::chrono::seconds      WAIT_FOR_TX_TIMEOUT_INTERVAL{30};
+const uint32_t                  THRESHOLD_FOR_FAST_SYNCING{100u};
```
n-hutton, line 42: basically, hold onto path to common ancestor blocks when there are over 100 (otherwise behave statelessly)

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -327,7 +329,27 @@ bool MainChain::GetPathToCommonAncestor(Blocks &blocks, BlockHash tip, BlockHash
       }
 
       // left side always loaded into output queue as we traverse
-      blocks.push_back(left);
+      // blocks.push_back(left);
+      res.push_back(left);
+
+      if (behaviour == BehaviourWhenLimit::RETURN_LEAST_RECENT)
```
ejfitzgerald: 👍 👍 👍 - May a recommend refactoring this slightly to be a switch statement without a default case. Then this way when if another mode is added we will get a compiler warning

libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -107,10 +116,11 @@ class MainChain
   /// @{
   BlockPtr  GetHeaviestBlock() const;
   BlockHash GetHeaviestBlockHash() const;
-  Blocks    GetHeaviestChain(uint64_t limit = ALL) const;
-  Blocks    GetChainPreceding(BlockHash at, uint64_t limit = ALL) const;
-  bool      GetPathToCommonAncestor(Blocks &blocks, BlockHash tip, BlockHash node,
-                                    uint64_t limit = ALL) const;
+  Blocks    GetHeaviestChain(uint64_t limit = UPPER_BOUND) const;
```
ejfitzgerald, line 119: This is only a sticky plaster - if you are making this change then I think that you should enforce this upper bound by adding something like the following to the  bodies of these functions. 

```cpp
limit = std::min(limit, UPPER_BOUND);
```

(Maybe it is there already and I haven't seen it)

libs/ledger/include/ledger/chain/main_chain.hpp
ejfitzgerald: 👍

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -361,6 +383,9 @@ bool MainChain::GetPathToCommonAncestor(Blocks &blocks, BlockHash tip, BlockHash
     }
   }
 
+  blocks.resize(res.size());
```
ejfitzgerald, line 395: I think I would also just update the return type for this function. This steps seems more expensive than we would ideally want (even though we are just manipulating smart pointers).

If this turns out to be super annoyance we can live with the copying but would be a shame

ejfitzgerald (changes requested):

libs/ledger/src/chain/main_chain.cpp
n-hutton: lets chat today? tbh it's a little bit of an annoyance to either update all blocks type or have this fn an exception

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -423,15 +460,17 @@ MainChain::BlockHashSet MainChain::GetMissingTips() const
  * @param maximum The specified maximum number of blocks to be returned
  * @return The generated array of missing hashes
  */
-MainChain::BlockHashs MainChain::GetMissingBlockHashes(std::size_t maximum) const
+MainChain::BlockHashs MainChain::GetMissingBlockHashes(uint64_t limit) const
 {
+  uint64_t limit_here = MainChain::UPPER_BOUND;
+  limit               = std::min(limit, limit_here);
```
ejfitzgerald:
```suggestion
  limit  = std::min(limit, uint64_t{MainChain::UPPER_BOUND});
```

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -294,8 +301,10 @@ MainChain::Blocks MainChain::GetChainPreceding(BlockHash start, uint64_t limit)
  * @return true if successful, otherwise false
  */
 bool MainChain::GetPathToCommonAncestor(Blocks &blocks, BlockHash tip, BlockHash node,
-                                        uint64_t limit) const
+                                        uint64_t limit, BehaviourWhenLimit behaviour) const
 {
+  uint64_t limit_here = MainChain::UPPER_BOUND;
+  limit               = std::min(limit, limit_here);
```
ejfitzgerald:
```suggestion
  limit = std::min(limit, uint64_t{MainChain::UPPER_BOUND});
```

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -245,6 +250,8 @@ MainChain::Blocks MainChain::GetHeaviestChain(uint64_t limit) const
  */
 MainChain::Blocks MainChain::GetChainPreceding(BlockHash start, uint64_t limit) const
 {
+  uint64_t limit_here = MainChain::UPPER_BOUND;
+  limit               = std::min(limit, limit_here);
```
ejfitzgerald:
```suggestion
  limit = std::min(limit, uint64_t{MainChain::UPPER_BOUND});
```

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -228,6 +229,10 @@ bool MainChain::RemoveBlock(BlockHash hash)
  */
 MainChain::Blocks MainChain::GetHeaviestChain(uint64_t limit) const
 {
+  // Note: min needs a reference to something, so this is a workaround since UPPER_BOUND is a
+  // constexpr
+  uint64_t limit_here = MainChain::UPPER_BOUND;
+  limit               = std::min(limit, limit_here);
```
ejfitzgerald:
```suggestion
  limit = std::min(limit, uint64_t{MainChain::UPPER_BOUND});
```

ejfitzgerald (changes requested):
Super not a fan of `limit_here` can you see if my suggestions work?

ejfitzgerald (dismissed):

ejfitzgerald (approved):

Issue #666, opened 2019-02-19
Improve Remove Block Performance


ejfitzgerald:
The current block removal implementation suffers from poor performance

n-hutton:
I would suggest the following:

Write a benchmark that tests the function ```   bool        RemoveBlock(BlockHash hash) ``` of main_chain.hpp/cpp - probably best to use the in memory only mode (no writing to disk)

There are already a few benchmarks for the main chain in ``` ~/repos/fetch-ledger/libs/ledger/benchmark/main_chain_benchs.cpp ```

This will probably have O(N^2) performance or something like that.

Review the internal workings of mainchain, and rewrite the remove algorithm to be less ugly (Ed wrote it :) )

Hints:

The way the main chain works is that it keeps track of all known blocks and tries to link them back to genesis. It attempts to know the 'heaviest' valid chain (where to be valid it must connect to genesis). When file writing is enabled, it writes blocks that are almost certain into the object store (almost certain is block head - confirmation blocks, in bitcoin this would be 6 blocks).

Occasionally, the block coordinator realises that it needs to remove a block - perhaps it can't find a TX that the block references. The removal of this block will neccessarily cause all of the blocks upwards from that to become 'loose' (not connected to genesis). Since those blocks are a hashed linked list, they should also be removed (not sure if that happens or not).

PR #979, opened 2019-05-11
Feature/666 main chain forward refs

bipll:
This commit adds forward refs to main chain and improves the performance of `RemoveBlock()` and `ReindexTips()`.

It is also incompatible with previous versions in chain DB binary format.

It also fixes, in a way, the build if compiling with `FETCH_COMPILE_LOGGING_LEVEL=debug`, which, as it turns out, has completely diverged from the default logging level. Perhaps, not the logging logic itself, fully (few cases were not automatically instantly clear), but the build is working.

Solves: #666


ejfitzgerald (changes requested):
In general looks good, however, I don't really think it is a good idea to include the next block hash in with the main block information. Since the block serialiser is both used for block storage in the main chain but also block transfer over the network. Receiving a next block hash doesn't really make much sense for the client since it would at a minimum need to null it out.

The way I was thinking about this (and would be interested in your thoughts) is that in general the existing structure works well for the 90% case. This sync case is more of an interesting side note. This is why I was thinking that conceptually this requirement is more like an additional index in a database. This is why I was initially suggesting adding a new instance of a file storage object to sit alongside the existing one. In the simplest version of this the you just stop the Hash -> Hash forward references. Since as we discussed the a working assumption with the main chain is that once you have flushed to disk you assume that the chain is linear.

I was envisaging that when doing this process, that loop could be updated to ensure that the forward references were written to the new index (file object store). Some additional care might be needed here and it might be worthwhile to add a few unit tests around to see if you can exercise some corner cases.

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -130,94 +195,96 @@ MainChain::BlockPtr MainChain::GetHeaviestBlock() const
...
+  // mark hash as being invalidated
+  invalidated_blocks.insert(hash);
+
+  // first remove all the progeny recursively;
+  // this way any hash that could potentially stem from this one, reference tree-wise, is completely
+  // wiped out
+  auto children{references_.equal_range(hash)};
+  for (auto child{children.first}; child != children.second; ++child)
+  {
+    RemoveTree(child->second, invalidated_blocks);
```
ejfitzgerald: Feels like we should not do this with recursion. While in practise we might not hit the limit at the moment, I am worried about the case when a large loose side chain is added

ejfitzgerald (changes requested):
Looks good to me as a first stage. I just have one comment about the recursive loop.

Also would be good to add some new methods / tests for these changes - not sure if you were planning on doing that as a separate PR

libs/ledger/src/chain/main_chain.cpp
bipll: I have added more thorough tests for block removal. In fact, functionally, this shouldn't change the observable behaviour at all, only improve it, so unsure whether it can be properly tested (other than consistency/benchmarking).

libs/core/include/core/logger.hpp
```diff
@@ -433,7 +434,7 @@ class LogWrapper
   }
 
   template <typename... Args>
-  void DebugWithName(char const *name, Args const &... args)
+  void DebugWithName(char const *name, Args &&... args)
```
ejfitzgerald: Worth updating the other handlers?

ejfitzgerald (commented):
Looks good, however, I think there will be a number of merge conflicts with the recent changes merged into develop

PR #983, opened 2019-05-12
Feature/blas refactor

troels-im:
Reimplementing partial BLAS support for tensor class.


libs/math/benchmark/basic_math/exp_bench.cpp
```diff
@@ -29,14 +29,16 @@ static void BM_ApproxExpImplementation(benchmark::State &state)
...
 
   fetch::math::ApproxExpImplementation<N, C> fexp;
-  double                                     x = (double)state.range(0);
-  double                                     result;
+  double                                     x      = 0.1;  //(double)state.range(0);
+  double                                     result = 0;
   for (auto _ : state)
   {
     // Single iteration is too small to get accurate benchmarks.
+    x += 0.1;
```
baykaner, line 37: I think we need to do something like this:

state.PauseTiming();
x+= 0.1;
for (int i = 0; i < 1000; i++)
{
  x += 0.0001;
  state.ResumeTiming();
  auto tmp = fexp(x);
  state.PauseTiming();
  result += tmp;
}
otherwise we'll be including all those other operations in the benchmarking which is presumably not the goal?

libs/math/include/math/linalg/blas/gemm_nn_novector.hpp
```diff
@@ -0,0 +1,56 @@
...
+namespace fetch {
+namespace math {
+namespace linalg {
+
+template <typename S>
+class Blas<S, Signature(_C <= _alpha, _A, _B, _beta, _C),
+           Computes(_C <= _alpha * _A * _B + _beta * _C), platform::Parallelisation::NOT_PARALLEL>
+{
+public:
+  using type               = S;
```
baykaner: style: Type should be PascalCase

libs/math/include/math/linalg/blas/gemm_nn_vector.hpp
```diff
@@ -0,0 +1,56 @@
...
+namespace fetch {
+namespace math {
+namespace linalg {
+
+template <typename S>
+class Blas<S, Signature(_C <= _alpha, _A, _B, _beta, _C),
+           Computes(_C <= _alpha * _A * _B + _beta * _C), platform::Parallelisation::VECTORISE>
+{
+public:
+  using type               = S;
```
baykaner: Instead of adding the same review comment on every file I'll just say that all of these files need to replace type with Type.

libs/math/include/math/linalg/blas/swap_all.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+#include "math/linalg/blas/base.hpp"
+#include "math/linalg/prototype.hpp"
+#include "math/tensor.hpp"
+
+namespace fetch {
+namespace math {
+namespace linalg {
+
+template <typename S, uint64_t V>
```
baykaner, line 59: style std::uint64_t

libs/math/include/math/linalg/blas/trmv_n.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * following Python code:
+ *
+ * import numpy as np
+ * import copy
+ *
+ * def trmv_n(A, x, n):
+ *   leny = A.shape[0]
+ *   lenx = A.shape[1]
+ *
+ *   return x
```
baykaner: does this function it really do what this python code says? Seems unlikely since the python code just returns the input x.

libs/math/include/math/matrix_operations.hpp
```diff
@@ -60,10 +60,25 @@ inline void Min(ArrayType const &array, typename ArrayType::Type &ret)
...
+  {
+    ret =
+        obj1.data().in_parallel().Reduce(memory::TrivialRange(0, obj1.size()),
                                          [](typename ArrayType::VectorRegisterType const &a,
                                             typename ArrayType::VectorRegisterType const &b) ->
                                          typename ArrayType::VectorRegisterType { return a * b; });
+  }
+  else
+  {
+    std::cout << "WAS HERE??" << std::endl;
```
baykaner: remove debugging code

libs/math/include/math/matrix_operations.hpp
```diff
@@ -60,10 +60,25 @@ inline void Min(ArrayType const &array, typename ArrayType::Type &ret)
...
+    // TODO: Vectorise in the general case
```
baykaner: this TODO needs to be attached to an issue number

libs/math/include/math/matrix_operations.hpp
```diff
@@ -60,10 +60,25 @@ inline void Min(ArrayType const &array, typename ArrayType::Type &ret)
...
+    auto it1 = obj1.cbegin();
+    ret      = 1;
```
baykaner: this will be an implicit cast from int to Type; I don't think that will work for FixedPoint Type. Use:
typename ArrayType::Type{1}

libs/math/include/math/matrix_operations.hpp
```diff
@@ -60,10 +60,25 @@ inline void Min(ArrayType const &array, typename ArrayType::Type &ret)
...
+    while (it1.is_valid())
+    {
+      ret *= (*it1);
```
baykaner, line 77: infinite loop; you forgot to increment it1.
++it1;

This implies this function is not yet covered by a test :-(

libs/math/include/math/matrix_operations.hpp
```diff
@@ -113,7 +128,8 @@ meta::IfIsMathArray<ArrayType, void> BooleanMask(ArrayType const &input_array,
     ++rit;
   }
 
-  ret.LazyResize(counter);
+  // TODO(tfr): Is this correct - that does not look right!
```
baykaner: I noticed a few things about this actually:
1. Yup, probably should be Resize in case the user wants to pass in a small tensor that will get resized up (not actually possible with current ASSERT checking ret.size())
2. the function Asserts that ret.size() == sum(Mask); but probably the user of the function shouldn't be required to do that calculation and that Assert should go away.

libs/math/include/math/matrix_operations.hpp
```diff
@@ -140,9 +156,10 @@ void Scatter(ArrayType &input_array, ArrayType const &updates,
 
   auto     indices_it = indices.begin();
   SizeType update_idx{0};
+
   while (indices_it != indices.end())
   {
-    input_array[input_array.ComputeIndex(*indices_it)] = updates[update_idx];
+    input_array.data()[input_array.ComputeIndex(*indices_it)] = updates[update_idx];
```
baykaner, line 160: why make this change?
operator[] just calls data[] anyway right?

libs/math/include/math/tensor.hpp
```diff
@@ -69,13 +69,23 @@ class Tensor
...
+  using ConstSliceIteratorType = ConstTensorSliceIterator<T, typename SelfType::ContainerType>;
+  using SizeType               = fetch::math::SizeType;
+  using SizeVector             = fetch::math::SizeVector;
 
   static constexpr char const *LOGGING_NAME = "Tensor";
 
+  enum
+  {
+    LOG_PADDING = 3,
+    PADDING     = 1ull << LOG_PADDING
```
baykaner: is it worth defining PADDING as:
SizeType(1) << LOG_PADDING instead?
since there are operations using PADDING that are mixed with SizeType.
As long as SizeType is defined as a std::uint64_t it doesn't matter; but if it's ever changed  those operations will start implicitly casting.

libs/math/include/math/tensor.hpp
```diff
@@ -164,36 +174,103 @@ class Tensor
...
+  Tensor FillArange(Type const &from, Type const &to);
 
   static SelfType UniformRandom(SizeType const &N);
   static SelfType UniformRandomIntegers(SizeType const &N, int64_t const &min, int64_t const &max);
   SelfType &      FillUniformRandom();
   SelfType &      FillUniformRandomIntegers(int64_t const &min, int64_t const &max);
   static SelfType Zeroes(SizeVector const &shape);
   static SelfType Ones(SizeVector const &shape);
-  SizeType        ComputeIndex(SizeVector const &indices) const;
+  SizeType        ComputeIndex(SizeVector indices) const;
```
baykaner: I'm confused about this.
Why do a vector copy followed by a std::move?
Why not just have this a const &?

libs/math/include/math/tensor.hpp
```diff
@@ -164,36 +174,103 @@ class Tensor
...
+  SelfType  Transpose(SizeVector &new_axes) const;
+  SelfType &Squeeze();
+  SelfType &Unsqueeze();
+
+  /////////////////////////
+  /// memory management ///
+  /////////////////////////
+
+  // New interface
+  bool Resize(SizeVector const &shape, bool copy = false)
```
baykaner: Lets maintain having function declarations only in the class body, and then having the full definition elsewhere. Tensor.hpp is big and hard to read even if you already know what you're looking at, and it helps a lot fo have clusters of function names when you're not already familiar with the class.

libs/math/include/math/tensor.hpp
```diff
@@ -164,36 +174,103 @@ class Tensor
...
+  SelfType  Transpose() const;  // TODO (private 867)
+  SelfType  Transpose(SizeVector &new_axes) const;
+  SelfType &Squeeze();
+  SelfType &Unsqueeze();
+
+  /////////////////////////
+  /// memory management ///
+  /////////////////////////
+
+  // New interface
```
baykaner: doxygen style function comments needed

libs/math/include/math/tensor.hpp
```diff
@@ -164,36 +174,103 @@ class Tensor
...
+  bool Resize(SizeVector const &shape, bool copy = false)
+  {
+    Tensor old_tensor = *this;
```
baykaner: I think it's a bit confusing that we have both Tensor which is defined as:
template <typename DataType, typename ContainerType>
class Tensor;
In meta/math_type_traits.hpp
AND
using SelfType = Tensor<T, C>
I GUESS these always have identical templates in our code; but I also guess that technically it could be possible to write code where Tensor has a T & C that is not the same as SelfType's T & C.
Presumably this is not a deliberate design aspect? I think we should stick to SelfType everywhere if we can.

libs/math/include/math/tensor.hpp
```diff
@@ -164,36 +174,103 @@ class Tensor
...
+
+    SizeType newsize = SelfType::PaddedSizeFromShape(shape);
```
baykaner: newsize should be new_size

libs/math/include/math/tensor.hpp
```diff
@@ -164,36 +174,103 @@ class Tensor
...
+    size_          = SelfType::SizeFromShape(shape);  // Note: differs from newsize
+    padded_height_ = PadValue(shape[0]);
+    UpdateStrides();
+
+    // Effectively a reshape
+    if (copy && (size_ == old_tensor.size()))
+    {
+      auto it  = begin();
+      auto oit = old_tensor.begin();
+      assert(it.size() == oit.size());
```
baykaner: prefer ASSERT over assert

libs/math/include/math/tensor.hpp
```diff
@@ -390,13 +453,68 @@ class Tensor
     }
   }
 
+  /////////////////////////////
+  /// Convinience functions ///
```
baykaner: typo "Convenience"

libs/math/include/math/tensor.hpp
```diff
@@ -390,13 +453,68 @@ class Tensor
...
+  }
+
+  SizeType width() const
+  {
+    return (shape_.size() > 1 ? shape_[1] : 1);
+  }
+
+  SizeType depth() const
+  {
+    return (shape_.size() > 1 ? shape_[2] : 1);
```
baykaner: you mean:
return (shape_.size() > 2 ? shape_[2] : 1);

libs/math/include/math/tensor.hpp
```diff
@@ -390,13 +453,68 @@ class Tensor
...
+  {
+    return padded_height_;
+  }
+
+  constexpr SizeType padding()
+  {
+    return PADDING;
+  }
+
+  static SizeType PadValue(SizeType size)
```
baykaner, line 478: would be nice to have a function comment here that explains this function returns the smallest size that is a multiple of PADDING and that first size inside it

libs/math/include/math/tensor.hpp
```diff
@@ -446,15 +566,18 @@ class Tensor
...
     return index;
   }
+  */
 
-  SizeType ComputeColIndex(SizeVector const &indices) const
+  template <typename S>
+  fetch::meta::IfIsUnsignedInteger<S, SizeType> ComputeColIndex(std::vector<S> const &indices) const
   {
-    SizeType index  = 0;
+    assert(indices.size() > 0);
```
baykaner, line 562: ASSERT

libs/math/include/math/tensor.hpp
```diff
@@ -922,9 +1039,19 @@ typename Tensor<T, C>::Type Tensor<T, C>::operator()(SizeType const &index) cons
...
+  if (shape_.size() == 1)
+  {
+    return data_[n];
+  }
+
+  SizeType j = static_cast<SizeType>(n) / height();
+  SizeType i = static_cast<SizeType>(n) - j * height();
+
+  assert(i + padded_height_ * j < padded_size());
+  return data_[i + padded_height_ * j];
```
baykaner, line 1041: I might be mistaken, but this looks like a function that doesn't work for NDArrays for N>=3
is that correct?

libs/math/include/math/tensor.hpp
```diff
@@ -1361,9 +1482,10 @@ typename Tensor<T, C>::SelfType Tensor<T, C>::Transpose(SizeVector &new_axes) co
 template <typename T, typename C>
 typename Tensor<T, C>::SelfType &Tensor<T, C>::Squeeze()
 {
-  ASSERT(shape_.at(0) == 1);
-  shape_.erase(shape_.begin());
-  UpdateStrides();
+  auto shape = shape_;  // TODO: Make last dimension for efficiency
```
baykaner: raise an issue and mark TODO with the issue number

libs/math/include/math/tensor_iterator.hpp
```diff
@@ -61,57 +41,29 @@ class TensorIterator
...
   TensorIterator(TensorIterator &&other)                 = default;
   TensorIterator &operator=(TensorIterator &&other) = default;
 
-  /**
-   * Iterator for more interesting ranges
-   * @param array the Tensor to operate upon
-   * @param step the from,to,and step range objects
-   */
-  TensorIterator(TensorType &array, std::vector<std::vector<SizeType>> const &step)
+  TensorIterator(TensorType &array, SizeType position)
```
baykaner, line 52: would be nice to have a doxygen style function comment that explains that (presumably) this constructor lets you iterator starting at a point that is not begin()

libs/math/include/math/tensor_iterator.hpp
```diff
@@ -61,57 +41,29 @@ class TensorIterator
...
     : array_(array)
-  {
-    std::vector<std::vector<SizeType>> step{};
-    for (auto i : array.shape())
-    {
-      step.push_back({0, i, 1});
-    }
-    Setup(step, array_.shape());
-  }
+  {}
```
baykaner, line 40: I think you don't use any vectors here anymore right?
So you can remove #include <vector>

libs/math/tests/math/linalg/blas/gemm_nn_novector.cpp
```diff
@@ -0,0 +1,312 @@
...
+#include "math/linalg/blas/base.hpp"
+#include "math/linalg/blas/gemm_nn_novector.hpp"
+#include "math/linalg/prototype.hpp"
+#include "math/tensor.hpp"
+
+using namespace fetch;
+using namespace fetch::math;
+using namespace fetch::math::linalg;
+
+TEST(blas_DGEMM, blas_gemm_nn_novector1)
```
baykaner, line 30: is it worth adding any new tests that handle tensors with dimension greater than 2?
Tensor<type>::FromString() doesn't work for >= 3 dimensional tensors :-(

libs/math/tests/math/tensor_iterator/iterator.cpp
```diff
@@ -0,0 +1,238 @@
...
+  for (SizeType i = 0; i < 4; ++i)
+  {
+    for (SizeType j = 0; j < 6; ++j)
+    {
+      for (SizeType k = 0; k < 7; ++k)
+      {
+        for (SizeType l = 0; l < 6; ++l)
+        {
+          std::vector<SizeType> idxs = {i, j, k, l};
+          ASSERT_TRUE(int(ret.Get(idxs)) == int(array.Get(idxs)));
```
baykaner: ASSERT_EQ is nicer for debugging :-)

libs/ml/include/ml/ops/activations/relu.hpp
```diff
@@ -56,14 +56,22 @@ class Relu : public fetch::ml::ElementWiseOps<T>
     ASSERT(inputs.size() == 1);
     ASSERT(inputs[0].get().shape() == errorSignal.shape());
 
-    ArrayType return_signal = errorSignal.Copy();
-    for (SizeType i{0}; i < inputs.front().get().size(); ++i)
+    ArrayType const &input         = inputs.front().get();
+    ArrayType        return_signal = errorSignal.Copy();
```
baykaner: style: error_signal

libs/ml/include/ml/ops/activations/relu.hpp
```diff
@@ -56,14 +56,22 @@ class Relu : public fetch::ml::ElementWiseOps<T>
...
-      if (inputs.front().get()[i] <= DataType(0))
+      if (*it1 <= DataType(0))
       {
-        return_signal.data().Set(i, DataType(0));
+        *it2 = DataType(0);
       }
+      ++it1;
+      ++it2;
     }
+
```
baykaner, line 87: Seems to me that doing a full copy, followed by checking every element in this way is neither the most efficient or the most clear implementation.
The more efficient implementation is to just do the iterator while loop and have an else clause which sets *it2 = *it1
Probably we should add a math function called Threshold like this:
void Threshold(ArrayType const& x, Type const val, ArrayType const& ret)
{
  ASSERT(x.shape() == ret.shape())
  auto it1 = x.cbegin();
  auto rit = ret.begin();

  while (it1.is_valid())
  {
    *rit = (*it1 > val) ? *it1 : x;
    ++it1;
    ++rit;
    ++rit;
  }
}

libs/ml/include/ml/ops/convolution_1d.hpp
```diff
@@ -265,7 +265,9 @@ void Convolution1D<ArrayType>::ReverseFillVerticalStride(ArrayType &      input,
                                                          SizeType const   input_channels,
                                                          SizeType const   kernel_height)
 {
-  SizeType j_s = 0;                                      // stride height iterator
+  SizeType j_s = 0;  // stride height iterator
+  assert(input.shape().size() == 3);
```
baykaner, line 269: ASSERT

libs/ml/include/ml/ops/convolution_2d.hpp
```diff
@@ -211,6 +211,11 @@ std::vector<ArrayType> Convolution2D<ArrayType>::Backward(
   ReverseFillVerticalStride(kernel_error, error2, output_channels, input_channels, kernel_height,
                             kernel_width);
 
+  for (auto const &e : input_error)
```
baykaner: leftover debugging code

baykaner: probably my fault :-p

libs/vectorise/include/vectorise/memory/vector_slice.hpp
```diff
@@ -100,7 +99,6 @@ class VectorSlice
 
   void SetPaddedZero()
   {
-    assert(pointer_ != nullptr);
     if (pointer_)
     {
       std::memset(static_cast<void *>(pointer_ + size()), 0,
```
baykaner, line 104: you have changes to mio and benchmark submodules again. This problem keeps happening when merging from very old branches.

baykaner (changes requested):

libs/math/benchmark/basic_math/exp_bench.cpp
troels-im: Don't think that the benchmark as it is and was written makes much sense. The old one was dominated by other operations than the actual exponential. You suggestion will probably not make the test more accurate as `fexp` is a total of two instructions and one need to be careful in the design test. I am leaving this for now as it is not priority atm.

libs/math/include/math/linalg/blas/gemm_nn_novector.hpp
troels-im: Fixed.

libs/math/include/math/matrix_operations.hpp
troels-im: No it does not. It remaps the index such that only data inside the Tensor can be used with this operator.

libs/math/include/math/tensor.hpp
troels-im: Yes, I agree - should be const ref.

libs/math/include/math/tensor.hpp
troels-im: Fixed.

libs/math/include/math/tensor.hpp
troels-im: I agree - changed!

libs/math/include/math/tensor.hpp
troels-im: Yes.

libs/math/include/math/tensor.hpp
troels-im: Since when do we use all caps assert?

libs/math/include/math/tensor.hpp
troels-im: No, it works for all dimensions. We only do padding on first dimension.

libs/math/tests/math/linalg/blas/gemm_nn_novector.cpp
troels-im: Blas only works for vectors and matrices, hence no test for dimensions above 2.

libs/math/include/math/linalg/blas/swap_all.hpp
troels-im: We don't use `std::uint64_t` anywhere else - throughout the library we are consistently using `uint64_t` and we should keep it that way unless there has been some agreement to change this.

libs/math/include/math/matrix_operations.hpp
troels-im: Your point 1 will not work as you resize _after_ setting the values. This would seg fault if a small tensor is passed to the routine.

libs/math/include/math/tensor.hpp
troels-im: Sure.

libs/math/include/math/linalg/blas/swap_all.hpp
troels-im: Just double checked this one: `uint64_t` is correct and `std::uint64_t` is not.

libs/ml/include/ml/ops/activations/relu.hpp
troels-im: Yes, that is a good suggestion. Feel free to add a ticket for it. Meanwhile, I've made it such that it does not copy, but just allocate.

libs/vectorise/include/vectorise/memory/vector_slice.hpp
troels-im: Not sure what you mean and what this has to do with vector slice.

libs/math/include/math/linalg/blas/swap_all.hpp
baykaner: huh, my bad, guess I was mistaken!

libs/vectorise/include/vectorise/memory/vector_slice.hpp
baykaner: It has nothing to do with vector slice but github doesn't let me comment on the submodules files changed. Take a look at the list of files changed; right at the bottom; the last two are changes to vendor/benchmarks and vendor/mio; I think you don't actually want to push those changes; it's just something leftover from the old branch I think

libs/math/include/math/matrix_operations.hpp
baykaner: That's true. In which case I probably used the LazyResize here because it's marginally cheaper than the Resize and knew it would always be safe in this context. But since we have an intention to remove LazyResize from public anyway the change is fine with me.

libs/math/include/math/tensor.hpp
baykaner: well, I it's not very important, but I tend to prefer using our ASSERT macro since I don't see any downsides to using it and it sometimes makes life a little easier when debugging (since you can more easily comment out code without warnings).

baykaner: Now I'm giving it more consideration I'm in two minds about whether we should still have this macro though; since it can also hide poor design choices

libs/vectorise/include/vectorise/memory/vector_slice.hpp
troels-im: OK. How do I fix this? I did `git submodule update --recursive` and nothing happened.

libs/ml/include/ml/ops/embeddings.hpp
```diff
@@ -101,8 +101,12 @@ class Embeddings : public fetch::ml::ops::Weights<T>
   {
     for (auto const &r : updated_rows_)
     {
-      auto gradient_accumulation_slice = this->gradient_accumulation_->Slice(r).Tensor();
-      auto output_slice                = this->output_->Slice(r).Tensor();
+      // TODO (private issue 997): The net result of the previous operations
+      // was a shallow copy of the tensor. This does not look like it was the intended
+      // behaviour.
+      (void)(r);
```
baykaner: I submitted a PR two weeks ago that gives the intended behaviour. Please take a look at https://github.com/uvue-git/fetch-ledger/pull/930
If you agree with the solution you approve that PR or merge or cherry pick it into your branch?

baykaner (changes requested):

baykaner (approved):

PR #984, opened 2019-05-12
VM intermediate representation

troels-im:


libs/vm/include/vm/matrix.hpp
```diff
@@ -167,56 +174,92 @@ struct Matrix : public IMatrix
...
-    if ((lhs_rows != rhs_rows) || (lhs_columns != rhs_columns))
+    if (lhs_columns != rhs_rows)
     {
       RuntimeError("invalid operation");
       return;
     }
-    lhs->matrix.InlineSubtract(rhs->matrix);
+    Ptr<Matrix> m = AcquireMatrix(vm_, type_id_, element_type_id_, lhs_rows, rhs_columns);
+    // TODO(tfr): use blas
+    TODO_FAIL("Use BLAS TODO");
```
ejfitzgerald, line 237: Not really sure if we want TODO_FAILS right in the middle here. Does it just make sense to set a RuntimeError on the VM

libs/vm/include/vm/common.hpp
```diff
@@ -17,15 +17,230 @@
...
+static TypeId const UInt16         = 7;
+static TypeId const Int32          = 8;
+static TypeId const UInt32         = 9;
+static TypeId const Int64          = 10;
+static TypeId const UInt64         = 11;
+static TypeId const Float32        = 12;
+static TypeId const Float64        = 13;
+static TypeId const PrimitiveMaxId = 13;
+static TypeId const String         = 14;
+static TypeId const Address        = 15;
```
ejfitzgerald, line 61: Out of interest why does the `Address` get a static `TypeId` and `State` does not? Or am I missing something

ejfitzgerald (approved):
Makes some level of sense to me. I have a couple of questions more for my on curiosity. Difficult to review 100% when such a large code base. Since this blocks a number of things I suggest that we simply get it merged in and deal with the fallout

libs/vm/include/vm/matrix.hpp
troels-im: No, agreed - we should use runtime error in the VM.

libs/vm/include/vm/common.hpp
troels-im: I suspect that it is because these are the types which has builtin literals. As such, Address should not be there in the first place since it does not have a literal. Rather, it should have been added as a module extension (unless there is a good reason why this was brought into the core of the VM).

libs/vm/include/vm/matrix.hpp
troels-im: I can change this tonight, but it might be worthwhile to wait for @baykaner  to land the blas functionality and just fix this properly.

libs/vm/src/ir.cpp
```diff
@@ -0,0 +1,219 @@
...
+  {
+    Reset();
+    Clone(other);
+  }
+  return *this;
+}
+
+IR &IR::operator=(IR &&other)
+{
+  if (&other != this)
```
wkaluza, line 50: I'm not sure you need to check this in a move assignment. Also, I'd consider making this operator noexcept.

PR #1002, opened 2019-05-15
add error checking and make the file continue rather than break if a …

n-hutton:
…test times out or returns an error code


ejfitzgerald (approved):

ejfitzgerald (approved):
You need to work on your PR titles

PR #1003, opened 2019-05-15
CMake autoformatting

wkaluza:
This would enforce a CMake code style in CI using the Python module cmake_format.


CMakeLists.txt
```diff
@@ -1,13 +1,14 @@
-################################################################################
+#
```
ejfitzgerald, line 1: Seems a shame to loose these headers but whatever

CMakeLists.txt
```diff
@@ -21,36 +22,42 @@ option(FETCH_ENABLE_COVERAGE        "Enable coverage generation"        OFF)
...
 
 # custom string options
-set(FETCH_DEBUG_SANITIZER "" CACHE STRING "The clang based sanitizer to be enabled")
-set(FETCH_COMPILE_LOGGING_LEVEL "info" CACHE STRING "The precompiled logging level for the project")
-
-# target architecture (to be replaced with automatic detection)
-## sysctl -a | grep machdep.cpu.features
+set(FETCH_DEBUG_SANITIZER
+    ""
+    CACHE STRING "The clang based sanitizer to be enabled")
```
ejfitzgerald: This is a good example of where after this change it looks worse. Can we not configure the cmake formatter in someway to adjust the line length. I don't really think we need to limit ourselves to something that is 80 characters in the build system.

CMakeLists.txt
```diff
@@ -21,36 +22,42 @@ option(FETCH_ENABLE_COVERAGE        "Enable coverage generation"        OFF)
...
-set (FETCH_ROOT_CMAKE_DIR ${FETCH_ROOT_DIR}/cmake)
-set (FETCH_ROOT_VENDOR_DIR ${FETCH_ROOT_DIR}/vendor)
+set(FETCH_ROOT_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
+set(FETCH_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+set(FETCH_ROOT_CMAKE_DIR ${FETCH_ROOT_DIR}/cmake)
+set(FETCH_ROOT_VENDOR_DIR ${FETCH_ROOT_DIR}/vendor)
 
 # Default to a release build
-if (NOT CMAKE_BUILD_TYPE)
+if(NOT CMAKE_BUILD_TYPE)
```
ejfitzgerald: I sort of preferred the space separator here, not super strong though. If it is something we can configure that would be good.

CMakeLists.txt
wkaluza: I increased the line width to 100 chars which took care of the worst offenders. I would ask that we don't go any higher, as that can be hard to read on a portrait monitor with a big font (like the one I need to use).

CMakeLists.txt
wkaluza: Yup, that's easily done. We get treasures like `endif ()` as part of the package, though.

libs/core/tests/CMakeLists.txt
```diff
@@ -16,5 +16,11 @@ add_fetch_test(byte_array_gtest fetch-core byte_array)
 add_fetch_test(random_generator_gtest fetch-core random)
 add_fetch_test(json_gtest fetch-core json/gtest/)
 add_fetch_test(core_encode_decode_gtest fetch-core tokenizer/gtest/)
-add_fetch_test(containers-tests fetch-core containers/ SLOW)
-add_fetch_test(sync_gtest fetch-core sync/ SLOW)
+add_fetch_test(containers-tests
```
ejfitzgerald, line 19: Lets make an issue to create a `add_fetch_slow_test` wrapper to avoid these in the future

ejfitzgerald (approved):

Issue #894, opened 2019-04-25
tests & benchmarking for remaining matrix ops 


baykaner:
tests and benchmarks:
- argmax
- sum
- reducesum
- reducemean
- PeakToPeak
- Dot
- DotTranspose
- TransposeDot
- DynamicStitch
- Concat

All tests done in PR #1004 
Benchmarks done in PR #1027

PR #1004, opened 2019-05-15
Tests/tests for matrix ops

MissingNO57:
tests & benchmarking for remaining matrix ops #894


libs/math/include/math/matrix_operations.hpp
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#pragma once
```
baykaner: once is probably enough ;-)

libs/math/include/math/matrix_operations.hpp
```diff
@@ -296,6 +297,18 @@ void Max(ArrayType const &array, typename ArrayType::SizeType const &axis, Array
     }
   }
 }
+
+template <typename ArrayType>
+ArrayType Max(ArrayType const &array, typename ArrayType::SizeType const &axis)
```
baykaner: hmm.... I think this interface will throw errors if we ever try to make a Tensor<SizeTypes>, which DOES actually seem like something we would like to do.

One easy solution would be to only have the 'in place' Max axis function; i.e.:
void ArrayType Max(ArrayType const& array, SizeType const &axis, Type &ret);

libs/math/include/math/matrix_operations.hpp
```diff
@@ -386,6 +399,17 @@ void Min(ArrayType const &array, typename ArrayType::SizeType const &axis, Array
   }
 }
 
+template <typename ArrayType>
+ArrayType Min(ArrayType const &array, typename ArrayType::SizeType const &axis)
```
baykaner: Same as previous comment - will fail for Tensor<SizeType>

Lets only have the in-place interface

libs/math/include/math/matrix_operations.hpp
```diff
@@ -501,29 +532,64 @@ ArrayType ReduceSum(ArrayType const &obj1, SizeType axis)
...
-void PeakToPeak(ArrayType arr)
+typename ArrayType::Type PeakToPeak(ArrayType const &array)
 {
-  return Max(arr) - Min(arr);
+  return Max(array) - Min(array);
+}
+
+template <typename ArrayType>
+void PeakToPeak(ArrayType const &array, typename ArrayType::Type &ret)
+{
```
baykaner, line 569: These two PeakToPeak implementations are in the wrong order; in your case, the inplace PeakToPeak is calling the return type PeakToPeak; but this means you're copying the return Type out; so it's not really an inplace operation anymore.

What you want is for the implementation to be inside the inplace operation, and the return type operation should set up a return object and call the inplace implementation.

Also these implementations are kind of inefficient because Max and Min each has to iterate over the whole tensor.

baykaner (changes requested):

baykaner (approved):

Issue #948, opened 2019-05-07
REstore Ops::ComputeOutputShape() constness


pierre-wilmot-fetch-ai:
This was removed in commit : https://github.com/uvue-git/fetch-ledger/commit/9d695b574b4a83bd15239fe8bb8553729bb30187 and shoiuld be added back.

PR #1006, opened 2019-05-15
Make ComputeShape Const Again

MissingNO57:
REstore Ops::ComputeOutputShape() constness #948


pierre-wilmot-fetch-ai (approved):
Good, we should just be consistent with the use of override and use it everywhere

PR #1024, opened 2019-05-17
C2v doc

juergendkeba:
Short summary of the c2v + possible extensions


docs/source/ml/c2v.rst
```diff
@@ -0,0 +1,77 @@
+# Code2Vec
+
+# Input files
+The content of the input file(s) is assumed to have the following syntax:
```
baykaner, line 4: this means to train on new data one needs to first use a parser to convert from code -> this input format. Can you include a link to the python repo that already has this parser for python programming language please?

Another custom parser would be required for other languages.

docs/source/ml/c2v.rst
```diff
@@ -0,0 +1,77 @@
...
+```
+fct_1 sourceword_1a,path_1a,targetword_1a sourceword_1b,path_1b,targetword_1b ...
```
baykaner: maybe point that the end token required for each line is just a \n?

docs/source/ml/c2v.rst
```diff
@@ -0,0 +1,77 @@
...
+
+
+# Data loading
+The dataloader class loads in the data and prepares it such that it can be fed into the graph.
+During training, a fixed number of contexts for a function has to be read in (`MAX_CONTEXTS`). If less contexts are available, dummy contexts are created.
+(The elements of the dummy context get the index 0 in the path and word corpus)
+
+The method `GetNext()` retrieves a pair of
+```
+(fct_i, (sourceword_tensor_ij, path_tensor_ij, targetword_tensor_ij))
```
baykaner: I believe you have accidentally reversed the order here?

baykaner: but I could be wrong :-p

docs/source/ml/c2v.rst
```diff
@@ -0,0 +1,77 @@
...
+
+
+Legend:
+i denotes the index of the samples (i.e. i runs from 0 to MAX_CONTEXTS)
+j denotes the index of a context embedding (i.e. j runs from 0 to 3*EMBEDDING_SIZE)
+k denotes the index of an embedding (i.e. j runs from 0 to EMBEDDING_SIZE)
+m denotes the index of the function name vocab
+\vec SE_{i} (\vec PE_{i}, \vec FE_{i}) gives the representation of the word/path/.. of sample i
+
+
```
baykaner, line 45: subheading "model architecture"

docs/source/ml/c2v.rst
```diff
@@ -0,0 +1,77 @@
...
+\vec CV_{i} = (\vec SE_{i}, \vec PE_{i}, \vec SE_{i})
+
+3. Combined Context Vector: Transpose CV, multiply with FC, Applying Tanh
+CC_{k, i} = tanh * (FC_{k, j} CV_{j, i}
+
+4. Context with attention: Transposition of CC, Multiplication with AV
+CA_{i,1} = \sum_{} CC_{i,k} AV_{k,1}
+
+5. Attenion weights: Transposition, Softmax normalisation
+AW_{1,i} = SoftMax(CA{1,i})
```
baykaner, line 58: question:
why do we do transpose, softmax, then transpose back, and multiply?
Is the softmax meant to be row-wise rather than element-wise or something?

docs/source/ml/c2v.rst
```diff
@@ -0,0 +1,77 @@
...
+
+8. Normalised prediction: Transpose, Softmax
+PR_{1,m} = Softmax{UP_{1,m}}
+
+9. Cross entropy loss
+L = CrossEntropy(PR_{1,m}, GT_{1,m})
+
+
+# Extensions: 
+ - In the TF implemention, the SE, PE, FE are initialised with a variance_scaling_initializer. AV and FC with a glorot_uniform_initializer.
```
baykaner, line 74: would be worth stating explicitly what we are doing instead (i.e. initialize all trainable params with glorot uniform)

docs/source/ml/c2v.rst
```diff
@@ -0,0 +1,77 @@
...
+ - Adding a Softmax function to the fetch lib, which can normalise along any axis. (Currently it works only for axis 0, that's the reason for the many transposition operations)
```
baykaner, line 75: okay then this is as I thought. Incidentally. that softmax operation exists already :-p
It probably didn't when you started building the example though.
You can just construct Softmax(axis) and it should work

docs/source/ml/c2v.rst
```diff
@@ -0,0 +1,77 @@
...
+ - In the data loader, the whole "corpus" (i.e. the umaps etc.) are created based on a c2v file. In the TF implementation, this corpus is created once, and then persisted, such that it can be reused next time (e.g. for model evaluation)
+ - Bringing the AST extraction into fetch
```
baykaner, line 77: may be worth just adding a sentence to the very start of the docs making it totally clear; this model does not take a string of code as input! An AST parser is needed to convert that into the input format

docs/source/ml/c2v.rst
```diff
@@ -0,0 +1,77 @@
...
+ - Extension to "real" batch training, s.t. contexts of more than one function can be digested simultaneously. (I.e. it would require, that several tensors get a new dimension, and all operations can handly the extra dim)
```
baykaner: maybe you can explain just a tiny bit more here? Is the new dimensions a batch dimension? Do you just mean, implementing true batch training through all ML ops? or is there something more specific that would need to be done for c2v regarding the multiple contexts per function

baykaner (changes requested):

docs/source/ml/c2v.rst
juergendkeba: whoops :-D Thx!!!!! Good eye!! ;-)

docs/source/ml/c2v.rst
juergendkeba: Exactly! That was the problem! If we could apply the softmax along any axis, then the code could be shortened!

((A good check that the transposition is needed is to not include the transposition beforehand... then you see, that nothings trains anymore, cause the output vector of the softmax is then something like [1.0, 1.0, 1.0, 1.0] ))

docs/source/ml/c2v.rst
juergendkeba: ^^ i tried it, but https://github.com/uvue-git/fetch-ledger/blob/982a856b21066934910e5b45f8f590ede7182071/libs/math/include/math/ml/activation_functions/softmax.hpp#L81 doesnt let me :-D

docs/source/ml/c2v.rst
juergendkeba: kk :-) i try to be a bit more verbose ;-) but yep, you're right! by that i meant, that every operation should be able to handle a new batch dimension! i try to give an example in the doc!

baykaner (approved):

PR #1027, opened 2019-05-17
Added benchmarks for matrix ops

MissingNO57:
tests & benchmarking for remaining matrix ops #894


libs/math/benchmark/matrix_ops/matrix_ops.cpp
```diff
@@ -261,13 +351,360 @@ void BM_Maximum(benchmark::State &state)
...
+    ->Unit(benchmark::kMillisecond);
+
+template <class T, int H, int W>
+void BM_ReduceSum(benchmark::State &state)
+{
+  using SizeType = fetch::math::SizeType;
+  for (auto _ : state)
+  {
+    state.PauseTiming();
+    fetch::math::Tensor<T> t(std::vector<std::uint64_t>{H, W});
```
baykaner: an alternative would be to instantiate these tensors once before you start the loop, rather than pausing timings

libs/math/benchmark/matrix_ops/matrix_ops.cpp
```diff
@@ -261,13 +351,360 @@ void BM_Maximum(benchmark::State &state)
...
+    ->Unit(benchmark::kMillisecond);
+
+template <class T, int H, int W>
+void BM_ReduceMean(benchmark::State &state)
+{
+  using SizeType = fetch::math::SizeType;
+  for (auto _ : state)
+  {
+    state.PauseTiming();
+    fetch::math::Tensor<T> t(std::vector<std::uint64_t>{H, W});
```
baykaner: as above

libs/math/benchmark/matrix_ops/matrix_ops.cpp
```diff
@@ -261,13 +351,360 @@ void BM_Maximum(benchmark::State &state)
...
+BENCHMARK_TEMPLATE(BM_ReduceMean, fetch::fixed_point::FixedPoint<32, 32>, 1024, 1024)
+    ->Unit(benchmark::kMillisecond);
+
+template <class T, int C, int H, int W>
+void BM_PeakToPeak(benchmark::State &state)
+{
+  for (auto _ : state)
+  {
+    state.PauseTiming();
+    fetch::math::Tensor<T> t(std::vector<std::uint64_t>{C, H, W});
```
baykaner: please always use SizeType when you mean SizeType. If we ever decide to change the underlying type of what a SizeType is I don't want other code to break. The right way to do it is to include math base_types if you need to and just:
using SizeType = fetch::math::SizeType;
in some cases you may also like to use: 
using SizeVector = fetch::math::SizeVector;

libs/math/benchmark/matrix_ops/matrix_ops.cpp
```diff
@@ -261,13 +351,360 @@ void BM_Maximum(benchmark::State &state)
...
+BENCHMARK_TEMPLATE(BM_PeakToPeak, fetch::fixed_point::FixedPoint<32, 32>, 256, 256, 256)
+    ->Unit(benchmark::kMillisecond);
+
+template <class T, int H, int W>
+void BM_Dot(benchmark::State &state)
+{
+  for (auto _ : state)
+  {
+    state.PauseTiming();
+    fetch::math::Tensor<T> t(std::vector<std::uint64_t>{H, W});
```
baykaner: SizeVectors. initialise them outside the loop

libs/math/benchmark/matrix_ops/matrix_ops.cpp
```diff
@@ -261,13 +351,360 @@ void BM_Maximum(benchmark::State &state)
...
+    ->Unit(benchmark::kMillisecond);
+
+template <class T, int H, int W>
+void BM_DotTranspose(benchmark::State &state)
+{
+  for (auto _ : state)
+  {
+    state.PauseTiming();
+    fetch::math::Tensor<T> t(std::vector<std::uint64_t>{H, W});
+    fetch::math::Tensor<T> ret(std::vector<std::uint64_t>{H, W});
```
baykaner: same comment

baykaner (changes requested):

baykaner (approved):

PR #1030, opened 2019-05-18
Injectable Clocks + Remove Sleeps from BC Tests

ejfitzgerald:
New time library (`moment`) has `DeadlineTimer` a replacement for `FutureTimepoint`.

Update block coordinator tests to remove sleeps


libs/ledger/tests/chain/block_coordinator_tests.cpp
```diff
@@ -1014,21 +1033,21 @@ TEST_F(NiceMockBlockCoordinatorTests, UnknownTransactionDoesNotBlockForever)
...
   // Time out wait to request Tx from peers
-  std::this_thread::sleep_for(std::chrono::seconds(31u));
+  clock_->Advance(std::chrono::seconds(31u));
 
-  Advance();
+  ASSERT_TRUE(RemainsOn(State::WAIT_FOR_TRANSACTIONS));
 
-  // Time out wait for Tx - block should be invalidated at this point
-  std::this_thread::sleep_for(std::chrono::seconds(31u));
+  ASSERT_EQ(State::WAIT_FOR_TRANSACTIONS, block_coordinator_->GetStateMachine().state());
```
wkaluza: This is redundant given the previous assertion.

libs/moment/include/moment/clock_interfaces.hpp
```diff
@@ -0,0 +1,97 @@
...
+ * @param duration The offset to be applied to the clock
+ */
+template <typename R, typename P>
+void AdjustableClockInterface::Advance(std::chrono::duration<R, P> const &duration)
+{
+  AddOffset(std::chrono::duration_cast<Duration>(duration));
+}
+
+using ClockPtr           = std::shared_ptr<ClockInterface>;
+using AdjustableClockPtr = std::shared_ptr<AdjustableClockInterface>;
```
wkaluza, line 94: I think these should be unique_ptrs. That would also let GetClock and GetAdjustableClock return non-owning raw pointers or references, and ownership would be clearly confined to the clock store rather than shared.

libs/moment/include/moment/detail/steady_clock.hpp
```diff
@@ -0,0 +1,52 @@
...
+namespace moment {
+namespace detail {
+
+class SteadyClock : public ClockInterface
+{
+public:
+  // Construction / Destruction
+  SteadyClock()                    = default;
+  SteadyClock(SteadyClock const &) = delete;
+  SteadyClock(SteadyClock &&)      = delete;
```
wkaluza, line 34: Why not moveable?

libs/moment/src/clocks.cpp
```diff
@@ -0,0 +1,136 @@
...
+AdjustableClockPtr CreateAdjustableClock(char const *name, ClockType type)
+{
+  std::string const clock_name{name};
+
+  FETCH_LOCK(clock_store_lock);
+
+  // create the new clock
+  auto clock = CreateAdjustable(type);
+
+  // add the clock to the store if one was successfully creates
```
wkaluza: created

libs/moment/src/clocks.cpp
```diff
@@ -0,0 +1,136 @@
...
+  if (clock)
+  {
+    clock_store[clock_name] = clock;
```
wkaluza, line 129: This would overwrite an existing clock of the same name. Is that something we would ever want? Ditto in GetClock

libs/moment/src/clocks.cpp
```diff
@@ -0,0 +1,136 @@
...
+  auto it = clock_store.find(name);
+  if (it != clock_store.end())
+  {
+    return it->second;
+  }
+
+  // create the new clock
+  auto clock = CreateNormal(default_type);
+
+  // add the clock to the store if one was successfully creates
```
wkaluza: created

libs/moment/src/clocks.cpp
```diff
@@ -0,0 +1,136 @@
...
+  AdjustableClockPtr clock{};
+
+  switch (type)
+  {
+  case ClockType::STEADY:
+    clock = std::make_shared<detail::AdjustableClock<detail::SteadyClock>>();
+    break;
+  }
+
+  return clock;
```
wkaluza, line 74: I would make this an assert(false) and return an instance (rather than a pointer) - that way clients of Get*Clock won't have to worry about them returning null. Ditto in CreateNormal. The compiler will warn us if we add another ClockType.

libs/moment/src/clocks.cpp
```diff
@@ -0,0 +1,136 @@
...
+}
+}  // namespace
+
+/**
+ * Create or lookup a requested clock
+ *
+ * @param name The name of the clock
+ * @param default_type The default type of the clock
+ * @return The requested clock if successful, otherwise a null pointer
```
wkaluza, line 83: If we stored unique_ptrs in the clock store and returned references from here, client's would not have to worry about the return being null.

wkaluza (dismissed):
Well, this is embarrassing. You beat me to it.

I left a few comments regarding clock ownership and passing around references rather than pointers, but I can live with these changes in their current shape (get rid of the redundant assert though). I leave it up to you whether you want to implement my suggestions or not.

libs/moment/include/moment/clock_interfaces.hpp
ejfitzgerald: The design sort of relies of shared ownership. An adjustable clock is simple a wrapper around the  existing clock plus an ownership.

In addition the design allows for the clocks to be replaced in the case of multiple calls to `CreateAdjustableClock`. Maybe we can discuss more in person when I am in the office

libs/moment/include/moment/detail/steady_clock.hpp
ejfitzgerald: In general, my default pattern is to disable both the movable and copy semantics of a new type. If needed these can be added but it gives a nod to the developer that this was never particularly intended to moved and copied

wkaluza (approved):

PR #1031, opened 2019-05-18
Tensor View, Vectorise refactor and CBOW example

troels-im:
This PR contains a number of changes:
- Refactoring vectorise
- Implementing Tensor view
- Moving CBOW across using graph to define computation.


libs/math/include/math/tensor_declaration.hpp
```diff
@@ -0,0 +1,34 @@
...
+//
+//------------------------------------------------------------------------------
+
+#include "vectorise/memory/array.hpp"
+#include "vectorise/memory/shared_array.hpp"
+
+namespace fetch {
+namespace math {
+
+template <typename DataType, typename ContainerType = memory::SharedArray<DataType>>
```
baykaner, line 26: presumably the tensor declaration ought to live in math/base_types.hpp?
Is it only here so that you don't always have to include vectorise? or is there some other reason?

libs/meta/include/meta/is_iterable.hpp
```diff
@@ -0,0 +1,40 @@
...
+//------------------------------------------------------------------------------
+
+#include <type_traits>
+
+namespace fetch {
+namespace meta {
+namespace detail {
+
+template <typename T>
+auto IsIterableImplementation(int)
```
baykaner, line 27: this implementation is pretty interesting.
Do we plan to use it for anything outside of the math library?
If not it maybe should be moved to fetch/math/meta/math_type_traits.hpp

libs/math/include/math/math_vector_support.hpp
```diff
@@ -0,0 +1,58 @@
...
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include <type_traits>
+
+namespace fetch {
+namespace math {
+
+template <typename T>
```
baykaner: I do understand what's going on here, but I can't help but think it would be nice to have either:
1. some comment at the top explaining that this is doing SFINAE that ensures only float & double math operations call vectorised blas implementations; or
2. an equivalent comment in docs/source/math/index.rst explaining that vectorisation is handled via SFINAE checks that can be found in math/math_vector_support.hpp

libs/math/include/math/matrix_operations.hpp
```diff
@@ -20,16 +20,26 @@
...
+#include "math/linalg/blas/gemm_tn_vector.hpp"
+#include "math/tensor_declaration.hpp"
+
+#include "math/linalg/prototype.hpp"
+
 #include "math/base_types.hpp"
 #include "math/comparison.hpp"
 #include "math/fundamental_operators.hpp"  // add, subtract etc.
+#include "math/math_vector_support.hpp"
+#include "math/tensor_declaration.hpp"
```
baykaner: you've included tensor declaration twice. Also whether tensor_declaration should exist or whether it should just live inside base_types

libs/math/include/math/matrix_operations.hpp
```diff
@@ -187,7 +197,8 @@ meta::IfIsMathArray<ArrayType, void> Product(ArrayType const &array1, T &ret)
   }
 }
 
-template <typename T, typename C, typename = std::enable_if_t<meta::IsArithmetic<T>>>
+template <typename T, typename C /*template<class> class C*/,
```
baykaner, line 198: is this template comment doing something useful?

libs/math/include/math/matrix_operations.hpp
```diff
@@ -787,17 +800,20 @@ fetch::math::meta::IfIsMathArray<ArrayType, void> DotTranspose(ArrayType const &
...
   {
-    for (size_t j(0); j < B.shape()[0]; ++j)
-    {
-      for (size_t k(0); k < A.shape()[1]; ++k)
-      {
-        ret.At(i, j) += A.At(i, k) * B.At(j, k);
-      }
-    }
-  }
+    OPTIMISATION_FLAGS = HasVectorSupport<Type>::value ? platform::Parallelisation::VECTORISE
```
baykaner: is there some way we can avoid all the repetition of this enum? For example what if we define something like this at the top of the file:
template <typename Type>
struct Optimisation
{
  enum
  {
    OPTIMISATION_FLAGS = HasVectorSupport<Type>::value ? platform::Parallelisation::VECTORISE : platform::Parallelisation::NOT_PARALLEL
  }
}
and then we just invoke the Optimisation::OPTIMISATION_FLAS when we need it.
Does something like this work or is there no way to do something like this without suffering an overhead?

libs/math/include/math/meta/math_type_traits.hpp
```diff
@@ -99,11 +97,14 @@ struct IsMathImpl<int, ReturnType>
 {
   using Type = ReturnType;
 };
-template <typename DataType, typename ContainerType, typename ReturnType>
+
+template <typename DataType, typename ContainerType /*template<class> class ContainerType*/,
```
baykaner, line 137: does this template comment do something?

libs/math/include/math/tensor_declaration.hpp
```diff
@@ -0,0 +1,34 @@
...
+class Tensor;
+
+// TODO: Migrate to:
+// template <typename DataType, template<class> class ContainerType = memory::SharedArray>
```
baykaner, line 30: issue number for this TODO?

libs/math/include/math/tensor.hpp
```diff
@@ -296,10 +298,8 @@ class Tensor
   template <typename OtherType>
   Tensor operator/=(OtherType const &other);
 
-  Tensor &DotTranspose(Tensor const &A, Tensor const &B, Type alpha = 1.0, Type beta = 0.0);
-  Tensor &TransposeDot(Tensor const &A, Tensor const &B, Type alpha = 1.0, Type beta = 0.0);
-  Type    Sum() const;
-
+  // TODO: Make free functions
```
baykaner, line 299: issue number for this todo?

libs/math/include/math/tensor.hpp
```diff
@@ -317,20 +317,57 @@ class Tensor
...
 
   ConstSliceType Slice(SizeType i, SizeType axis = 0) const;
   TensorSlice    Slice(SizeType i, SizeType axis = 0);
 
+  //////////////
+  /// Views  ///
+  //////////////
+  TensorView<Type, ContainerType> View()
+  {
+    SizeType N     = shape_.size() - 1;
```
baykaner, line 330: it looks like there's an inconsistency in the concept of width used in tensor.
Calling tensor.width() gives you shape()[1].size()
But here you're using width to mean shape()[N].size()

libs/math/include/math/tensor.hpp
```diff
@@ -317,20 +317,57 @@ class Tensor
...
+  TensorView<Type, ContainerType> const View() const
+  {
+    SizeType N     = shape_.size() - 1;
+    SizeType width = shape_[N] * stride_[N] / padded_height_;
+    return TensorView<Type, ContainerType>(data_, height(), width);
+  }
+
+  TensorView<Type, ContainerType> View(SizeType index)
+  {
+    SizeType N                = shape_.size() - 1 - 1;
```
baykaner, line 344: I think this needs a comment explaining what's going on here. The fact that you've put - 1 - 1 presumably implies that each 1 indicates something conceptually significant, and these should be spelt out in the comment.

Also, what happens when we do this:
Tensor t(1);
auto view = t.View(0);

I think it crashes right? size == 1, therefore SizeType N = 1 - 1 - 1 = std::numeric_limits<unit64_t>::max(), therefore the next line that invokes stride_[N] is a bad access. Unless I misunderstand something?

libs/math/tests/math/tensor/view_test.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+};
+
+using MyTypes = ::testing::Types<int32_t, uint32_t, int64_t, uint64_t, float, double,
+                                 fetch::fixed_point::FixedPoint<32, 32>>;
+TYPED_TEST_CASE(TensorViewTests, MyTypes);
+
+using namespace fetch::math;
+
+TYPED_TEST(TensorViewTests, size_test)
```
baykaner, line 37: Are tensor views compatible with 3, 4, or higher dimensional tensors? If so it would be nice to have a test of this case to see how that works exactly.

libs/math/include/math/tensor.hpp
```diff
@@ -317,20 +317,57 @@ class Tensor
...
+    SizeType dimension_length = (N == 0 ? padded_height_ : shape_[N]);
+    SizeType volume           = dimension_length * stride_[N];
+    SizeType width            = volume / padded_height_;
+    SizeType offset           = volume * index;
+    return TensorView<Type, ContainerType>(data_, height(), width, offset);
+  }
+
+  TensorView<Type, ContainerType> View(std::vector<SizeType> indices)
+  {
```
baykaner, line 363: Let's have function comment for at least this view and the one above that just at least states the use case.
Also can we keep the function declarations here and move the implementations out of the class body as with the other functions.

libs/math/tests/math/linalg/blas/gemm_nn_novector.cpp
```diff
@@ -77,32 +78,33 @@ TEST(blas_DGEMM, blas_gemm_nn_novector2)
...
-  Tensor<Type> A = Tensor<Type>::FromString(R"(
+  Tensor<Type>     A_tensor = Tensor<Type>::FromString(R"(
   	0.13949386065204183 0.29214464853521815;
  0.3663618432936917 0.45606998421703593;
  0.7851759613930136 0.19967378215835974
   	)");
-
-  Tensor<Type> B = Tensor<Type>::FromString(R"(
+  TensorView<Type> A        = A_tensor.View();
+  Tensor<Type>     B_tensor = Tensor<Type>::FromString(R"(
```
baykaner, line 87: not very important and a bit pedantic but this A_tensor and B_tensor stuff is weird style

libs/math/tests/math/matrix_operations/matrix_operations.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include "core/random/lcg.hpp"
 #include "math/fixed_point/fixed_point.hpp"
+#include "math/math_vector_support.hpp"
```
baykaner: maybe this math vector support SFINAE should live in math/meta/math_type_traits.hpp with all the other math SFINAE?

libs/math/tests/math/tensor/view_test.cpp
baykaner: is it worth also having a test which checks what happens when we misuse tensor views? e.g. something like:

Tensor t();
EXPECT_THROW(auto view = t.View({1, 2, 3}));

baykaner (changes requested):
You have an issue with the submodules again. You'll need to git checkout upstream/develop -- vendors to fix it

libs/meta/include/meta/is_iterable.hpp
troels-im: We might use it outside the math library and it is not math specific, so the right place to keep it is in the `meta` library.

libs/math/include/math/math_vector_support.hpp
troels-im: It is very simple: This is just a register for those types for which we provide vectorised operations. I can add a comment.

libs/math/include/math/matrix_operations.hpp
troels-im: Yes, it is keeping a code snippet that we want to migrate to, but we are not quite ready yet.

libs/math/tests/math/linalg/blas/gemm_nn_novector.cpp
troels-im: Do you prefer `tensor_A` instead?

libs/math/tests/math/matrix_operations/matrix_operations.cpp
troels-im: It's not SFINAE, but sure we can move it to `type_traits`.

libs/math/include/math/matrix_operations.hpp
troels-im: Yes, it is useful for a future migration we will make of the tensor lib.

libs/math/include/math/tensor_declaration.hpp
troels-im: You get recursion problems if you try to put it there.

libs/math/src/math/linalg/blas/gemm_nn_vector.cpp
```diff
@@ -19,32 +19,33 @@
...
+  if (alpha == static_cast<Type>(0.0))
   {
-    if (beta == 0.0)
+    if (beta == static_cast<Type>(0.0))
     {
       for (j = 0; j < c.width(); ++j)
       {
 
-        VectorRegisterType vec_zero(0.0);
+        VectorRegisterType vec_zero(static_cast<Type>(0.0));
```
markos, line 48: I'd refrain from using vec_* variables as these are used for VSX (IBM Power) vector intrinsics and we would have a conflict when porting to that arch. In particular vec_zero() just returns a zeroed vector. We can change those later of course, but it would be easier to avoid those right from the start.

libs/math/src/math/linalg/blas/gemm_nn_vector.cpp
```diff
@@ -77,17 +78,17 @@ void Blas<S, Signature(_C <= _alpha, _A, _B, _beta, _C),
   for (j = 0; j < c.width(); ++j)
   {
     std::size_t l;
-    if (beta == 0.0)
+    if (beta == static_cast<Type>(0.0))
     {
 
-      VectorRegisterType vec_zero(0.0);
+      VectorRegisterType vec_zero(static_cast<Type>(0.0));
```
markos, line 84: same comment for vec_zero.

libs/math/src/math/linalg/blas/gemm_nt_vector.cpp
```diff
@@ -77,17 +78,17 @@ void Blas<S, Signature(_C <= _alpha, _A, _B, _beta, _C),
   for (j = 0; j < c.width(); ++j)
   {
     std::size_t l;
-    if (beta == 0.0)
+    if (beta == static_cast<Type>(0.0))
     {
 
-      VectorRegisterType vec_zero(0.0);
+      VectorRegisterType vec_zero(static_cast<Type>(0.0));
```
markos, line 84: same comment for vec_zero.

libs/math/src/math/linalg/blas/gemm_tn_vector.cpp
```diff
@@ -19,33 +19,34 @@
...
+  if (alpha == static_cast<Type>(0.0))
   {
-    if (beta == 0.0)
+    if (beta == static_cast<Type>(0.0))
     {
       for (j = 0; j < c.width(); ++j)
       {
 
-        VectorRegisterType vec_zero(0.0);
+        VectorRegisterType vec_zero(static_cast<Type>(0.0));
```
markos, line 49: again the same comment for vec_zero applies.

libs/math/src/math/linalg/blas/gemm_tn_vector.cpp
```diff
@@ -19,33 +19,34 @@
...
   if ((c.height() == 0) ||
-      ((c.width() == 0) || (((alpha == 0.0) || (a.height() == 0)) && (beta == 1.0))))
+      ((c.width() == 0) || (((alpha == static_cast<Type>(0.0)) || (a.height() == 0)) &&
+                            (beta == static_cast<Type>(1.0)))))
   {
     return;
   }
 
-  if (alpha == 0.0)
+  if (alpha == static_cast<Type>(0.0))
```
markos, line 42: regarding the vectorised form, branching on the different values of alpha/beta should not be needed, it would actually add a few more extra cycles. We should just calculate the result immediately.

libs/math/src/math/linalg/blas/gemm_nn_vector.cpp
troels-im: What would be your preferred naming for these variables? `fvreg_` (fetch vector register)?

markos: anything that lowers the probability of a name clash. In that aspect, fvreg_ is fine and understandable.

libs/math/src/math/linalg/blas/gemm_tn_vector.cpp
troels-im: Branching is done as per the original Fortran implementation. We can improve this if needed, but I suspect this implementation will be faster in the special case of `alpha=0` as it does not include unnecessary multiplications.

libs/math/src/math/linalg/blas/gemm_nn_vector.cpp
troels-im: It suffices to comment on these kind of issues once. Please make a list with suggested name changes and I will implement it.

libs/math/tests/math/linalg/blas/gemm_nn_novector.cpp
baykaner: it's just the mixing of cases and underscores that's weird; so a_tensor, tensor_a indicate at a glance their nature as types which.
TensorView<Type> A should probably be TensorView<Type> a also; but you're presumably using the maths style of using a capitalised letter to indicate a matrix. not important.

PR #1032, opened 2019-05-20
Normalize header use: <iomanip>

wkaluza:
We have a few dozen files in our codebase that needlessly include <iomanip>, and three files that only include it indirectly. This is now fixed.

### DIFF
--- a/apps/constellation/main.cpp
+++ b/apps/constellation/main.cpp
@@ -38,7 +38,6 @@
 #include <csignal>
 #include <cstdlib>
 #include <fstream>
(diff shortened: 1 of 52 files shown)


ejfitzgerald (approved):

n-hutton (approved):
Did you notice these, or did you run some tool?

wkaluza:
> Did you notice these, or did you run some tool?

Not a tool, unfortunately - I noticed that we were including `<iomanip>` in files that had no business including it, so I pruned them manually when I had a moment.

PR #1034, opened 2019-05-20
VM: Universal Serialisation Support

ejfitzgerald:
Added the ability for all object types to be serialised to and from the state object. This is implemented with a general object level interface for serialisation.

This change does however couple the VM with `fetch-core` since the same serialisation logic has been used.


pbukva (approved):
LGTM
One comment - `State<T>` is accepting to `T` type exclusivelly `String`.

PR #1041, opened 2019-05-21
Changed Graph implementation and added Optimizers

MissingNO57:


libs/ml/tests/ml/ops/graph.cpp
```diff
@@ -49,53 +51,82 @@ TEST(graph_test, node_placeholder)
...
   g.SetInput("Input", data);
-  fetch::math::Tensor<int> prediction = g.Evaluate("Relu");
+  ArrayType prediction = g.Evaluate("Relu");
 
   // test correct values
   ASSERT_TRUE(prediction.AllClose(gt));
 }
 
-TEST(graph_test, getStateDict)
+TYPED_TEST(GraphTest, getStateDict)
```
baykaner: getStateDict is wrong style

libs/ml/tests/ml/ops/graph.cpp
```diff
@@ -49,53 +51,82 @@ TEST(graph_test, node_placeholder)
...
-  fetch::ml::Graph<fetch::math::Tensor<float>> g;
+  using ArrayType = TypeParam;
+  using SizeType  = typename TypeParam::SizeType;
+
+  fetch::ml::Graph<ArrayType> g;
 
-  g.template AddNode<fetch::ml::ops::PlaceHolder<fetch::math::Tensor<float>>>("Input", {});
-  g.template AddNode<fetch::ml::layers::SelfAttention<fetch::math::Tensor<float>>>(
-      "SelfAttention", {"Input"}, 50u, 42u, 10u);
+  g.template AddNode<fetch::ml::ops::PlaceHolder<ArrayType>>("Input", {});
```
baykaner, line 93: extend this test. AddNode returns the true name of the nodes added as strings. Therefore you should Add two nodes specifying the same name, e.g. something like this:

std::string sa_1 = g.template AddNode<fetch::ml::layers::SelfAttention<ArrayType>>("SelfAttention", {"Input"}, 50u,  "SelfAttention", {"Input"}, 50u, 42u, 10u);
std::string sa_2 = g.template AddNode<fetch::ml::layers::SelfAttention<ArrayType>>("SelfAttention", {"Input"}, 50u,  "SelfAttention", {"Input"}, 50u, 42u, 10u);

EXPECT_NEQ(sa_1, sa_2);
EXPECT_EQ(sa_1, "SelfAttention");
EXPECT_EQ(sa_2, "SelfAttention_1");

libs/ml/tests/ml/ops/graph.cpp
```diff
@@ -49,53 +51,82 @@ TEST(graph_test, node_placeholder)
...
+  std::string output_name =
+      g.template AddNode<fetch::ml::ops::Add<ArrayType>>(name + "_Add3", {add1_name, add2_name});
+
+  // Evaluate
+  g.SetInput(input_name, data);
+  TypeParam output = g.Evaluate("Diamond_Add3");
+
+  // Test correct values
+  ASSERT_EQ(output.shape(), data.shape());
+  ASSERT_TRUE(output.AllClose(gt, static_cast<DataType>(1e-5f), static_cast<DataType>(1e-5f)));
```
baykaner, line 178: as discussed, need to test backward pass; that's where the biggest danger lies.

libs/ml/tests/ml/ops/graph.cpp
```diff
@@ -49,53 +51,82 @@ TEST(graph_test, node_placeholder)
...
+}
```
baykaner, line 310: I think we should add another test that checks forward pass ordering for a slightly more complicated scenario. Something a bit like this:
X1 -> (op_1: Square) 
X2 -> (op_2: X1 * X2) -> (op_3: op2 - op1) -> Y

This kind of scenario checks that forward and back passing is happening in the correct order. In your diamond graph example, the order of the forward pass is irrelevant as long as it all happens (because addition is fully commutative).

baykaner:
Oh and there's one more type of test we need to add. Suppose we have a graph like this:
X1 -> (op_1: Square)
X2 -> (op_2: X1 * X2) -> (op_3: op2 - op1) -> Y

We need to run a forward pass. Then try to evaluate the individual nodes. Then change the value of only X2; When we call g.Evaluate, the graph needs to recompute some ops. We need to verify this actually happens (and the output Y changes as expected)

libs/ml/include/ml/ops/subtract.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+  virtual ArrayType Forward(std::vector<std::reference_wrapper<ArrayType const>> const &inputs,
+                            ArrayType &                                                 output)
+  {
+    (void)output;
+    ASSERT(inputs.size() == 2);
+    ASSERT(inputs.at(0).get().size() == inputs.at(1).get().size());
+    ASSERT(output.shape() == this->ComputeOutputShape(inputs));
+
+    auto output_it  = output.begin();
```
baykaner: I don't think you should be implementing this with iterators. I think you should just call fetch::math::Subtract and trust that it will use the correct underlying implementation (whether that's tensor iterators, or blas or whatever)

libs/ml/include/ml/ops/subtract.hpp
```diff
@@ -0,0 +1,78 @@
...
+      ++a_it;
+      ++b_it;
+    }
+
+    return output;
+  }
+
+  virtual std::vector<ArrayType> Backward(
+      std::vector<std::reference_wrapper<const ArrayType>> const &inputs,
+      ArrayType const &                                           errorSignal)
```
baykaner: style. error_signal

libs/ml/include/ml/ops/subtract.hpp
```diff
@@ -0,0 +1,78 @@
...
+namespace ops {
+
+template <class T>
+class Subtract : public fetch::ml::ElementWiseOps<T>
+{
+public:
+  using ArrayType      = T;
+  using DataType       = typename ArrayType::Type;
+  using ArrayPtrType   = std::shared_ptr<ArrayType>;
+  using ConstSliceType = typename ArrayType::ConstSliceType;
```
baykaner: remove constslicetype

libs/ml/include/ml/ops/subtract.hpp
```diff
@@ -0,0 +1,78 @@
...
+namespace ml {
+namespace ops {
+
+template <class T>
+class Subtract : public fetch::ml::ElementWiseOps<T>
+{
+public:
+  using ArrayType      = T;
+  using DataType       = typename ArrayType::Type;
+  using ArrayPtrType   = std::shared_ptr<ArrayType>;
```
baykaner: do you use ptr type?

libs/ml/include/ml/ops/subtract.hpp
```diff
@@ -0,0 +1,78 @@
...
+  using ConstSliceType = typename ArrayType::ConstSliceType;
+  Subtract()           = default;
+  virtual ~Subtract()  = default;
+
+  virtual ArrayType Forward(std::vector<std::reference_wrapper<ArrayType const>> const &inputs,
+                            ArrayType &                                                 output)
+  {
+    (void)output;
+    ASSERT(inputs.size() == 2);
```
baykaner: lets move away from using these assert macros from now onwards (also remove the related include)

baykaner (changes requested):

libs/ml/include/ml/graph.hpp
```diff
@@ -282,5 +301,57 @@ void Graph<ArrayType>::LoadStateDict(struct fetch::ml::StateDict<ArrayType> cons
...
+  for (auto const &t : trainable_)
+  {
+    ret.push_back(t.second->GetWeights());
+  }
+  return ret;
+}
+
+/**
+ * Assigns all trainable parameters to vector of ArrayType for exporting and serialising
+ * @return ret is vector containing all weights values
```
baykaner: this comment is a copy paste of the one for the above function.

libs/ml/include/ml/meta/ml_type_traits.hpp
```diff
@@ -36,12 +39,27 @@ constexpr bool IsTrainable = std::is_base_of<fetch::ml::ops::Trainable<T>, Opera
...
 constexpr bool IsNotTrainable = !IsTrainable<T, OperationType>;
 
+template <typename T, typename OperationType>
+constexpr bool IsGraph = std::is_base_of<fetch::ml::Graph<T>, OperationType>::value;
+
+template <typename T, typename OperationType>
+constexpr bool IsNotGraph = !IsGraph<T, OperationType>;
+
+template <typename T, typename OperationType>
+constexpr bool IsNotGraphOrTrainable = !IsGraph<T, OperationType> && !IsTrainable<T, OperationType>;
```
baykaner: this would be slightly more readable if it were expressed as:
IsNotGraph<T, OperationType> && IsNotTrainable<T, OperationType>

libs/ml/include/ml/ops/loss_functions/criterion.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include "ml/ops/ops.hpp"
+#include <cassert>
+
+namespace fetch {
+namespace ml {
+namespace ops {
+
+template <class T>
+class Criterion
```
baykaner, line 28: I think just delete this and have the cost functions inherit from Ops

libs/ml/include/ml/ops/loss_functions/cross_entropy.hpp
```diff
@@ -42,40 +43,34 @@ class CrossEntropy
   CrossEntropy()          = default;
   virtual ~CrossEntropy() = default;
 
+  /**
+   * @param inputs vector of 2 one-hot Tensors
```
baykaner, line 47: looking at the math cross entropy loss function there are two possibilities:
1. we're dealing with one-hot vectors, in which case n_classes is just inferred from the shape
2. we're dealing with a two-class binary logistic regression problem (i.e. non one-hot)
either way you don't need to pass n_classes on to the math function. just don't provide an n_classes input. Also we should probably remove n_classes as a parameter to the math function since it's only used for asserts

libs/ml/include/ml/ops/loss_functions/cross_entropy.hpp
```diff
@@ -42,40 +43,34 @@ class CrossEntropy
...
-      fetch::math::Subtract(ret, inputs[1], ret);
-      fetch::math::Multiply(ret, inputs[0], ret);
-    }
-    else if (inputs[0].shape().size())  // one-hot
-    {
-      ret = fetch::math::Softmax(inputs[0], 0);
-      fetch::math::Divide(inputs[1], ret, ret);
-      fetch::math::Multiply(DataType(-1), ret, ret);
-    }
+    ArrayType ret = fetch::math::Softmax(inputs[0], 0);
```
baykaner: this change is incorrect. You should still permit the binary logistic regression use case

baykaner: either that or implement a binary logistic regression op as a totally different op; which maybe is preferable actually

libs/ml/include/ml/ops/loss_functions/softmax_cross_entropy.hpp
```diff
@@ -43,18 +44,14 @@ class SoftmaxCrossEntropy
   virtual typename ArrayType::Type Forward(std::vector<ArrayType> const &inputs)
   {
     // third term may be present for specifying n_classes
-    assert((inputs.size() == 2) || ((inputs.size() == 3) && (inputs.at(2).size() == 1)));
+    assert(inputs.size() == 2);
```
baykaner, line 47: same as cross entropy comments

libs/ml/include/ml/ops/subtract.hpp
```diff
@@ -0,0 +1,63 @@
...
+  virtual std::vector<ArrayType> Backward(VecTensorType const &inputs,
+                                          ArrayType const &    error_signal)
+  {
+    (void)inputs;
+
+    assert(inputs.size() == 2);
+    assert(inputs.at(0).get().size() == inputs.at(1).get().size());
+    assert(error_signal.size() == inputs.at(1).get().size());
+
+    return {error_signal, fetch::math::Multiply(error_signal, DataType{-1})};
```
baykaner, line 55: I suspect this would be equivalent to return {error_signal, -error_signal}

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -50,9 +50,13 @@ class Trainable
   using ArrayType    = T;
   using ArrayPtrType = std::shared_ptr<ArrayType>;
 
-  virtual void                           Step(typename T::Type learningRate) = 0;
-  virtual struct fetch::ml::StateDict<T> StateDict() const                   = 0;
-  virtual void LoadStateDict(struct fetch::ml::StateDict<T> const &dict)     = 0;
+  virtual void                           Step(typename T::Type learningRate)         = 0;
```
baykaner, line 53: learning_rate

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -95,6 +99,20 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<T>
     this->output_->InlineAdd(*gradient_accumulation_);
     // Major DL framework do not do that, but as I can't think of any reason why, I'll leave it here
```
baykaner, line 100: remove this comment since it's not adding anything useful

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -95,6 +99,20 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<T>
...
     // for convenience. Remove if needed -- Pierre
+    ResetGradients();
+  }
+
+  virtual void ApplyGradient(ArrayType const &grad)
+  {
```
baykaner, line 106: is the intention to keep both Step and ApplyGradient?
If so I would have one call the other; since Step also multiplies by learning rate why not have Step defined as:
Multiply(gradient_accumulation_, learning_rate_, gradient_accumulation);
ApplyGradient(gradient_accumulation_)

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -95,6 +99,20 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<T>
     this->output_->InlineAdd(*gradient_accumulation_);
     // Major DL framework do not do that, but as I can't think of any reason why, I'll leave it here
     // for convenience. Remove if needed -- Pierre
+    ResetGradients();
+  }
+
+  virtual void ApplyGradient(ArrayType const &grad)
```
baykaner, line 105: add a function comment for this and also for Step explaining how they differ

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -172,7 +190,7 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<T>
    * Returns a copy of embeddings gradients for enquiry
```
baykaner, line 190: this is not actually the embeddings.hpp so this comment looks like a copy paste error

libs/ml/include/ml/optimization/adagrad_optimizer.hpp
```diff
@@ -0,0 +1,85 @@
...
+    SizeType i{0};
+    for (auto &grad : gradients)
+    {
+      auto git = grad.begin();
+      auto cit = cache_[i].begin();
+      while (git.is_valid())
+      {
+        *cit += (*git) * (*git);
+        // 1e-8 is added to prevent division by 0
+        *git = -this->learning_rate_ * ((*git) / fetch::math::Sqrt(*cit + DataType{1e-8f}));
```
baykaner, line 70: I wonder if there's a more optimized implementation of this. Doing a square root in a loop like this will be expensive

libs/ml/include/ml/optimization/adagrad_optimizer.hpp
```diff
@@ -0,0 +1,85 @@
...
+  using SizeType  = typename ArrayType::SizeType;
+
+  AdaGradOptimizer(std::shared_ptr<Graph<T>> graph, std::string const &input_node_name,
+                   std::string const &output_node_name, DataType const &learning_rate)
+    : Optimizer<T, C>(graph, input_node_name, output_node_name, learning_rate)
+  {
+    auto weights = this->graph_->GetWeights();
+    for (auto &wei : weights)
+    {
+      this->cache_.push_back(ArrayType(wei.shape()));
```
baykaner, line 49: emplace_back?

libs/ml/include/ml/optimization/optimizer.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+  Optimizer(std::shared_ptr<Graph<T>> graph, std::string const &input_node_name,
+            std::string const &output_node_name, DataType const &learning_rate)
+    : graph_(graph)
+    , input_node_name_(input_node_name)
+    , output_node_name_(output_node_name)
+    , learning_rate_(learning_rate)
+  {}
+
+  DataType DoBatch(ArrayType &data, ArrayType &labels)
```
baykaner, line 48: this is not a very descriptive name and has no function comment

libs/ml/include/ml/optimization/optimizer.hpp
```diff
@@ -0,0 +1,77 @@
...
+  {
+    DataType loss{0};
+    SizeType n_data = data.shape().at(0);
+
+    for (SizeType step{0}; step < n_data; ++step)
+    {
+      auto cur_input = data.Slice(step).Copy();
```
baykaner, line 55: this copying is not very optimal, and not really what we're going for with batch operations. What we really want to do with batch operations is wrap up data into a tensor with the final dimension being batch, then set that whole tensor as input, and then have each operation in the graph have a ForwardBatch and BackwardBatch implementation.
Then label_pred and cur_label will both be tensors of size {1, batch_size}

libs/ml/include/ml/optimization/momentum_optimizer.hpp
```diff
@@ -0,0 +1,81 @@
...
+    auto weights = this->graph_->GetWeights();
+    for (auto &wei : weights)
+    {
+      this->momentum_.push_back(ArrayType(wei.shape()));
+    }
+    ResetMomentum();
+  }
+
+private:
+  void ApplyGradients()
```
baykaner, line 56: function comments would be nice.

libs/ml/include/ml/optimization/adam_optimizer.hpp
```diff
@@ -0,0 +1,113 @@
...
+      DataType mt;
+      DataType vt;
+      while (git.is_valid())
+      {
+        *cit = (beta1_t_ * (*cit)) + ((DataType{1} - beta1_t_) * (*git));
+        mt   = *cit / (DataType{1} - beta1_t_);
+        *mit = (beta2_t_ * (*mit)) + ((DataType{1} - beta2_t_) * (*git) * (*git));
+        vt   = *mit / (DataType{1} - beta2_t_);
+        // 1e-4 is added to prevent division by 0
+        *git = -this->learning_rate_ * mt / (fetch::math::Sqrt(vt) + DataType{1e-4f});
```
baykaner, line 81: the value 1e-4f should be a member variable called epsilon and can be set by the user on construction

libs/ml/include/ml/optimization/adam_optimizer.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+template <class T, class C>
+class AdamOptimizer : public Optimizer<T, C>
+{
+public:
+  using ArrayType = T;
+  using DataType  = typename ArrayType::Type;
+  using SizeType  = typename ArrayType::SizeType;
+
+  AdamOptimizer(std::shared_ptr<Graph<T>> graph, std::string const &input_node_name,
```
baykaner, line 41: it would be nice to reference the author:
https://arxiv.org/abs/1412.6980

libs/ml/include/ml/optimization/rmsprop_optimizer.hpp
```diff
@@ -0,0 +1,88 @@
...
+    SizeType i{0};
+    for (auto &grad : gradients)
+    {
+      auto git = grad.begin();
+      auto cit = cache_[i].begin();
+      while (git.is_valid())
+      {
+        *cit = decay_rate_ * (*cit) + (DataType{1} - decay_rate_) * (*git) * (*git);
+        // 1e-8 is added to prevent division by 0
+        *git = -this->learning_rate_ * ((*git) / fetch::math::Sqrt(*cit + DataType{1e-8f}));
```
baykaner, line 72: again have a member variable epsilon

libs/ml/include/ml/optimization/rmsprop_optimizer.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include "math/standard_functions/sqrt.hpp"
+#include "ml/graph.hpp"
+#include "ml/ops/loss_functions/criterion.hpp"
+#include "ml/optimization/optimizer.hpp"
+
+namespace fetch {
+namespace ml {
+
+template <class T, class C>
+class RMSPropOptimizer : public Optimizer<T, C>
```
baykaner, line 35: reference hinton: https://www.cs.toronto.edu/~tijmen/csc321/slides/lecture_slides_lec6.pdf

libs/ml/tests/ml/ops/loss_functions/cross_entropy.cpp
```diff
@@ -104,44 +104,6 @@ TYPED_TEST(CrossEntropyTest, one_dimensional_forward_test)
   EXPECT_NEAR(double(op.Forward({data1, data2})), double(0.893887639), 3e-7);
 }
 
-TYPED_TEST(CrossEntropyTest, non_one_hot_forward_test)
```
baykaner: reinstate this test; it's a legitimate use-case and can work even without specifying number of classes with small changes as discussed above

libs/ml/tests/ml/ops/loss_functions/cross_entropy.cpp
```diff
@@ -221,33 +183,3 @@ TYPED_TEST(CrossEntropyTest, one_dimensional_backward_test)
   EXPECT_TRUE(op.Backward({data1, data2})
                   .AllClose(gt, typename TypeParam::Type(1e-5), typename TypeParam::Type(1e-5)));
 }
-
-TYPED_TEST(CrossEntropyTest, non_one_hot_dimensional_backward_test)
```
baykaner: reinstate

libs/ml/tests/ml/optimization/optimizers.cpp
```diff
@@ -0,0 +1,332 @@
...
+  std::string act_name    = g->template AddNode<fetch::ml::ops::Relu<TypeParam>>("", {fc1_name});
+  std::string output_name = g->template AddNode<fetch::ml::layers::FullyConnected<TypeParam>>(
+      "FC2", {act_name}, hidden_size, output_size);
+
+  ////////////////////////////////////////
+  /// DEFINING DATA AND LABELS FOR XOR ///
+  ////////////////////////////////////////
+
+  TypeParam data{{4, 1}};
+  data.Set(0, 0, DataType(1));
```
baykaner: a lot of repeated code. Make a single function that returns this data as necessary and call it repeatedly.

libs/ml/tests/ml/optimization/optimizers.cpp
```diff
@@ -0,0 +1,332 @@
...
+  /////////////////////
+  /// TRAINING LOOP ///
+  /////////////////////
+
+  DataType loss;
+  for (std::size_t i = 0; i < n_batches; ++i)
+  {
+    loss = optimizer.DoBatch(data, gt);
+  }
+  EXPECT_NEAR(static_cast<double>(loss), 9.3365340689197183, 1e-5);
```
baykaner: these tests only check that the training converges, which is important. But they don't offer a numerical test for what the updated gradients should be after one step. I think it would be good to also have that kind of test

baykaner (changes requested):

libs/ml/include/ml/ops/loss_functions/criterion.hpp
MissingNO57: But we will have to change Ops interface first, because current one cannot be used for that.

Ops:
virtual void                   Forward(VecTensorType const &inputs, ArrayType &output)      = 0;
  virtual std::vector<ArrayType> Backward(VecTensorType const &inputs,
                                          ArrayType const &    error_signal) 

Criterion:
  virtual typename ArrayType::Type Forward(std::vector<ArrayType> const &inputs)  = 0;
  virtual ArrayType                Backward(std::vector<ArrayType> const &inputs) = 0;

baykaner: ah, yep you're right. leave it as is then.

PR #1043, opened 2019-05-21
Add shadowing warning flag, and fix shadowing bugs

n-hutton:
Fixes found when turning on -Wshadow - also force naming convention for fetch tests


libs/core/include/core/logger.hpp
```diff
@@ -787,6 +787,31 @@ extern log::details::LogWrapper logger;
 
 #endif
 
+#define ERROR_BACKTRACE_STRING ERROR_BACKTRACE_STRING_START(0)
```
n-hutton, line 790: scheduled for demolition - will replace with backward-cpp trace library

libs/storage/include/storage/random_access_stack.hpp
```diff
@@ -52,7 +52,7 @@ namespace storage {
  * The header for the stack optionally allows arbitrary data to be stored, which can be useful to
  * the user
  */
-template <typename T, typename D = uint64_t, typename STREAM = std::fstream>
```
n-hutton, line 55: This, and the test are not related but are minor - reverse this dependency injection from emumba a while back (also deletes disabled tests)

libs/math/include/math/ml/loss_functions/cross_entropy.hpp
```diff
@@ -41,14 +41,14 @@ typename ArrayType::Type CrossEntropyLoss(
     ArrayType const &x, ArrayType const &y,
     typename ArrayType::SizeType n_classes = typename ArrayType::SizeType(2))
 {
-  using DataType = typename ArrayType::Type;
-  using SizeType = typename ArrayType::SizeType;
+  using DataType     = typename ArrayType::Type;
+  using SizeTypeHere = typename ArrayType::SizeType;
```
baykaner: ?? why have you made this change>?

ejfitzgerald (changes requested):
Looks like a load of random binaries added to this, probably should sort out before merging

libs/ml/include/ml/layers/skip_gram.hpp
```diff
@@ -93,9 +93,9 @@ class SkipGram : public Layer<T>
     {
       ArrayType slice_input   = inputs.front().get().Slice(b).Copy();
       ArrayType slice_context = inputs.back().get().Slice(b).Copy();
-      ArrayType output(ComputeOutputShape({slice_input, slice_context}));
-      this->Forward({slice_input, slice_context}, output);
-      results.push_back(output);
+      ArrayType output_here(ComputeOutputShape({slice_input, slice_context}));
```
baykaner, line 96: terrible name but sure that's fine I guess since this function needs reworking anyway

baykaner (approved):
looks good for the parts of the code I own

PR #1045, opened 2019-05-21
Use Git to print diff in style check

wkaluza:
The script apply_style now always runs in 'fix' mode, i.e. it will
traverse the directory tree and reformat files that do not conform to
the style.

For use on the build server, the -d/--diff argument may be used. The
script will then call out to Git in order to generate the diff of the
changes. If there are changes, the script will print the diff and exit
with an error.

### DIFF
--- a/scripts/apply_style.py
+++ b/scripts/apply_style.py
@@ -235,20 +235,15 @@ def format_cpp(args):
 def format_python(args):
     jobs_arg = ['-j {}'.format(args.jobs)]
 
-    autopep8_cmd = ['autopep8',    '.',
-
-                    '--in-place', '--recursive',
-                    '--exclude',
-
-                    'vendor'] + jobs_arg
+    autopep8_cmd = ['autopep8', '.', '--in-place', '--recursive',
+                    '--exclude', 'vendor'] + jobs_arg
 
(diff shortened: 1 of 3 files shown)


scripts/apply_style.py
```diff
@@ -249,70 +218,50 @@ def apply_style_to_file(source_path):
...
 
     # TODO(WK) Make multilanguage reformatting concurrent
+    output('Formatting C/C++ ...')
     format_cpp(args)
+    output('Formatting Python ...')
     format_python(args)
+    output('Done.')
+
+    if args.diff:
+        diff = get_diff()
```
ejfitzgerald, line 259: Is there a non-zero exit on diff output here?

wkaluza: I'm not sure if you are asking about the script or `git diff` invoked in `get_diff()`.

* The exit code of the script is non-zero if we're in diff mode and the diff is non-empty, i.e. if the reformatting resulted in changes. This will let the script fail the Jenkins build when running CI. If we're not in diff mode, but just a developer wanting to bring their local copy into line with the style, then the exit code is 0 regardless of whether there have been changes made or not.

* `git diff` on the other hand, returns a zero exit code upon successful execution, regardless of whether it finds any changes or not.

n-hutton (approved):

PR #1051, opened 2019-05-22
fetch::math::Clamp change with tests

MissingNO57:


libs/math/include/math/standard_functions/clamp.hpp
```diff
@@ -35,8 +35,35 @@ namespace math {
...
-  Max(x, min, ret);
-  Min(ret, max, ret);
+  assert(min <= max);
+
+  if (x <= min)
+  {
+    ret = min;
+    return;
+  }
+  if (x >= max)
```
baykaner: if you do:
if ()
else if ()
else ()
you can ditch the return statements

libs/math/include/math/standard_functions/clamp.hpp
```diff
@@ -35,8 +35,35 @@ namespace math {
 template <typename Type>
 void Clamp(Type const &x, Type const &min, Type const &max, Type &ret)
 {
-  Max(x, min, ret);
-  Min(ret, max, ret);
+  assert(min <= max);
+
+  if (x <= min)
```
baykaner: put implementations inside an anonymous namespace

libs/math/include/math/standard_functions/clamp.hpp
```diff
@@ -71,13 +100,20 @@ meta::IfIsMathArray<ArrayType, ArrayType> Clamp(ArrayType const &
...
 {
   ArrayType ret{array.shape()};
-  auto      ret_it = ret.begin();
-  for (auto &e : array)
+  Clamp(array, min, max, ret);
+  return ret;
+}
+
+template <typename ArrayType>
+meta::IfIsMathArray<ArrayType, void> Clamp(typename ArrayType::Type const &min,
```
baykaner, line 59: while we're modifying the clamp file, let's remove all return type interfaces (since we're migrating everything to return parameter style functions)

baykaner (changes requested):

baykaner (approved):

PR #1052, opened 2019-05-22
IWYU fixes

wkaluza:
This is a subset of corrections applied by IWYU ( https://github.com/include-what-you-use/include-what-you-use ) when ran on our codebase. Additionally, there are some cosmetic fixes mixed in that I applied as I went.


ejfitzgerald:
I am sure that @n-hutton went down this path a little, if you haven't already done so, it would be worth a conv. with him.

wkaluza:
@ejfitzgerald We spoke already - in fact, it was Nathan who made me aware of IWYU in the first place. Preliminary conclusions are that IWYU is not mature enough to serve full-time as a CI check, but I think I can wrangle some useful stuff out of it. In particular, it is quite good at hinting which header includes *might* be replaceable with forward declarations. It's a lot of data to sift through, and we have a busy period coming up, so I may set this aside for a time, but I'll keep the branch alive.

PR #1054, opened 2019-05-23
Vm functor invoke

robertdickson:


libs/vm/include/vm/module.hpp
```diff
@@ -216,6 +217,34 @@ class Module
...
+  /*
+  // Create a free function that takes an Int32 and a Float32 and returns a Float64
+  auto lambda = [](fetch::vm::VM*, int32_t i, float f)
+  {
+    return double(float(i) + f);
+  };
+  module.CreateFreeFunction("myfunc", lambda);
+  */
+  template <typename Functor>
+  void CreateFreeFunction(std::string const &name, Functor const &functor)
```
bipll: `Functor &&functor` perhaps?

libs/vm/include/vm/module/base.hpp
```diff
@@ -228,5 +247,24 @@ struct IndexedValueSetter<Type, std::tuple<InputTypes...>, OutputType>
...
 };
 
+template <typename T>
+struct FunctorReturnTypeExtractor : public FunctorReturnTypeExtractor<decltype(&T::operator())>
+{
+};
+template <typename Functor, typename ReturnType, typename... Ts>
+struct FunctorReturnTypeExtractor<ReturnType (Functor::*)(Ts...) const>
+{
+  using type = ReturnType;
```
bipll: `std::result_of`/`std::invoke_result`?

libs/vm/include/vm/module/base.hpp
```diff
@@ -193,6 +204,14 @@ struct MakeParameterType<T, typename std::enable_if_t<fetch::vm::IsPtr<T>::value
   using type = T const &;
 };
 
+template <typename T, typename... Ts>
+struct RemoveFirstType;
```
bipll: This should belong to a deeper root library, like meta.

libs/vm/include/vm/module.hpp
robertdickson: Agreed.

libs/vm/include/vm/module/base.hpp
robertdickson: Agreed, there are a few more tuple ones in there that are generically useful.

libs/vm/include/vm/module/base.hpp
robertdickson: std::result_of<Functor(Ts...)>::type is superior, then let's delete FunctorReturnTypeExtractor

libs/vm/include/vm/module.hpp
wkaluza: I picked up a part of this task from Robert. The universal refs are done.

libs/vm/include/vm/module/base.hpp
wkaluza: Done.

libs/vm/include/vm/module/base.hpp
wkaluza: I have found no convenient way of working `std::result_of` into the current solution. Annoyingly, not only does it seem to not handle void functions neatly (rather than setting `type` to `void`, it simply does not set it at all), but also it requires the parameter pack to work. The easiest way to get at the param pack in this code is through `decltype(&T::operator())`, which also gives you the return type for free, thus obviating the need for `std::result_of`.

It's very likely that I am missing something and `std::result_of` can be worked into the current scheme - I would welcome suggestions.

wkaluza: @bipll @robertdickson In light of the above, could you say a bit more about what benefits std::result_of would give us in this case?

PR #1056, opened 2019-05-26
Add basic utility functions to String

### LINKED ISSUE #965: VM: Basic String Manipulation Functions
Should be able to perform simple operations like:

- [ ] trim
- [ ] find
- [ ] sub-string etc

wkaluza:
This adds the following member functions to `vm::String`:

* find
* length
* reverse
* substr
* trim

Please note that `reverse()` and `trim()` mutate their instance String. That's easily changed if we would rather they return a new instance.

Closes #965.

### DIFF
--- a/libs/vm/include/vm/string.hpp
+++ b/libs/vm/include/vm/string.hpp
@@ -20,6 +20,7 @@
 #include "vm/variant.hpp"
 
 #include <cstddef>
+#include <cstdint>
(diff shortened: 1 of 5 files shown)


ejfitzgerald (approved):

ejfitzgerald:
@uvue-git/ledger-docs FYI

PR #1057, opened 2019-05-26
Add basic utility functions to Array

### LINKED ISSUE #812: VM: Common Array options
The following should be possible when working with arrays:

### Initialisation
```c#

// value initialisation
var arr = [1,2,3,4]; // should deduce an array 4 element array with type of Int32

// range initialisation
var arr = [1u; 5]; // should decude an array of 5 elements all set with the value 1 and of type UInt32
```

### Appending / Removal (Back)

```c#
var arr = [1,2,3,4];

// add the value 5 to the array
arr.append(5); // now should contain [1,2,3,4,5]

// remove a value from the array
var result1 = arr.pop_back(); // no args mean size 1, arr = [1,2,3,4], result1 = 5

// removal of multiple values
var result2 = arr.pop_back(2); // remove multiple elements arr = [1,2], result2 = [3,4]

// extending of array
arr.extend([1,2,3]); // arr = [1,2,1,2,3]

```

### Corresponding Queue like functionality

```c#
var arr = [1,2,3,4];

// remove a value from the array
var result1 = arr.pop_front(); // no args mean size 1, arr = [2,3,4,5], result1 = 1

// removal of multiple values
var result2 = arr.pop_front(2); // remove multiple elements arr = [4,5], result2 = [2,3]
```

wkaluza:
This adds the following member functions to `vm::Array`:

* `append`
* `pop_back()` and `pop_back(int)`
* `pop_front()` and `pop_front(int)`
* as an optional extra, `reverse()`.

TODO: `extend()`

Closes #812.

### DIFF
--- a/libs/vm/include/vm/array.hpp
+++ b/libs/vm/include/vm/array.hpp
@@ -89,7 +89,7 @@ struct Array : public IArray
       return {};
     }
 
-    auto element = GetIndexedValue(AnyInteger(elements.size() - 1u, TypeIds::UInt64))
+    auto element = GetIndexedValue(AnyInteger(elements.size() - 1u, TypeIds::Int32))
                        .template Move<TemplateParameter>();
 
     elements.pop_back();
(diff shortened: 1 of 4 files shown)


ejfitzgerald (approved):
I approve, but I think that this might change in the future when we add iterators. For the moment this makes sense.

ejfitzgerald:
@uvue-git/ledger-docs FYi

wkaluza:
> I approve, but I think that this might change in the future when we add iterators. For the moment this makes sense.

Sure thing. Note that this did not contain `extend()`, as I have had significant difficulties with getting it to work. If I don't work it out by Tuesday, I may need to pair up with someone.

PR #1058, opened 2019-05-26
Simplify test labels

### LINKED ISSUE #973: "No tests were found!!!" in "integration" test output
For example go here:
https://jenkins.economicagents.com/blue/organizations/jenkins/fetch-ledger/detail/develop/68/pipeline/46

and inspect one of the `ci-tool.py -I` logs. Underneath the tests (which are being executed), CTest complains that there were no tests. This should be investigated, as it suggests that we are misusing the runner in some way.

### LINKED ISSUE #1065: Create add_fetch_slow_test CMake function
See discussion underneath PR #1003.

wkaluza:
This simplifies the way we use CTest labels.

* By default tests are now unlabelled rather than labelled 'Normal'
* The 'Integration' label has been removed as it was not used. This fixes #973.
* Accordingly, the `ci-tool.py` `-I` switch has been renamed from to `-S` (for 'Slow')
* The way we `ci-tool.py` passes labels to CTest has changed: in `-T` mode it will run all tests *except* those labelled 'Slow'. In `-S` mode, it will *only* run tests labelled 'Slow'. This, I think, is safer than relying on all fast tests bearing the 'Normal' label and reduces the risk of introducing a test that CI never runs. (Yes, those are the sort of things that keep me awake at night.)

Fixes #1065.


ejfitzgerald (commented):
Let's have a chat about this on Tuesday. My only real comment is about consistent naming of these sort of tests in the project. I kind of liked:

- Unit
- Integration (not used currently I know)
- End to End Tests

The "Slow" tests instead where only ever supposed to be a short term label. They are a mix of integration and/or benchmarking and/or bad tests but basically they were intended as a placeholder so that we would tackle them one at a time

wkaluza:
I'm not married to calling them Slow rather than Integration, but Slow has the benefit of being true. Right now, most of our 'integration' tests are, as you say, a mish-mash of other stuff. Rather than pretend this isn't the case, I'd rather give this category of tests a name they deserve: one that explains why they stand apart from the other Google Tests, and why we run them separately during builds. The Slow label may have been intended to be short-lived, but it doesn't seem like it will be going away any time soon. Let's talk over tea on Tuesday if there's time.

What I feel the most strongly about here is getting rid of the Normal label.

NB, I've been experimenting with running CTest with the -jN switch, and the results are encouraging. It can shave five minutes of a Debug Slow test run. The problem, of course, is our flaky network_test, which consistently fails on the build server if I try this. Work in progress.

Issue #566, opened 2019-01-22
JSON: Fix conversion to/from integral and floating point types (minimize risk of value overflow) 


pbukva:
As first step it is necessary to fix existing conversion functions for integral and floating point numbers from/to JSON, the current implementation uses `std::atoi(...)` & `std::atof(...)` routines for this job, what makes json extremely vulnerable to value overflow).

PR #1059, opened 2019-05-26
Replace atoi/atof with safer strtol/strtod

### LINKED ISSUE #729: Replace atoi in ConstByteArray
Would suggest to replace atoi with atol in the constbytearray implementation:

int AsInt() const { std::string const value = static_cast<std::string>(*this); return atol(value.c_str()); // atoi(value.c_str()); }

wkaluza:
Fixes #729 and #566 .

### DIFF
--- a/libs/core/src/json/document.cpp
+++ b/libs/core/src/json/document.cpp
@@ -68,7 +68,7 @@ void JSONDocument::ExtractPrimitive(Variant &variant, JSONToken const &token,
       errno = 0;
       FETCH_LOG_ERROR(LOGGING_NAME, "Failed to convert str=", str, " to integer");
 
-      throw std::domain_error("Failed to convert str=" + str + " to integer");
+      throw JSONParseException("Failed to convert str=" + str + " to integer");
     }
     success = true;
     break;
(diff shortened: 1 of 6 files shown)


ejfitzgerald (changes requested):
While you are making this change could you also update the JSON library too

ejfitzgerald (dismissed):

libs/core/src/json/document.cpp
```diff
@@ -58,16 +61,29 @@ void JSONDocument::ExtractPrimitive(Variant &variant, JSONToken const &token,
     break;
 
   case NUMBER_INT:
-    // TODO(private issue #566): the `std::strtoll(...)` should be used here instead (converting to
-    // `long long`)
-    variant = std::atoll(document.char_pointer() + token.first);
+    str     = document.char_pointer() + token.first;
```
ejfitzgerald, line 64: Is this extra copy necessary in this case? Also we should raise a tracking issue to update all of these to `std::from_chars` once we start using C++17

wkaluza: It isn't! Good catch. I'll create that issue.

ejfitzgerald (approved):

PR #1062, opened 2019-05-27
Correct Tx Integer Parsing for values > 32bit

ejfitzgerald:
- Fix issue with encoding and decoding values > 32bit
- Extract out integer encoding logic into own header file (in order to write tests)
- Add a test suite to specifically test the integer encoding scheme

### DIFF
--- a/libs/ledger/include/ledger/chain/transaction_encoding.hpp
+++ b/libs/ledger/include/ledger/chain/transaction_encoding.hpp
@@ -63,11 +63,17 @@ meta::IfIsInteger<T, fetch::byte_array::ConstByteArray> EncodeInteger(T value)
     else
     {
       // determine the corresponding number of bytes required to encode this value
-      uint64_t const bits_to_encode = platform::ToLog2(abs_value);
+      uint64_t const bits_to_encode  = 64ull - platform::CountLeadingZeroes64(abs_value);
       uint64_t const bytes_to_encode = (bits_to_encode + 7u) >> 3u;
-      uint8_t const log2_bytes_required = static_cast<uint8_t>(
-        platform::ToLog2(bytes_to_encode)
-      );
+
+      uint8_t log2_bytes_required{3};
+      for (uint8_t limit = 4u; limit; limit >>= 1u)
+      {
+        if (bytes_to_encode <= limit)
+        {
+          --log2_bytes_required;
+        }
+      }
 
       // calculate the actual number of bytes that will be required
       std::size_t const bytes_required = 1u << log2_bytes_required;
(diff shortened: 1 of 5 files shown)


libs/ledger/include/ledger/chain/transaction_encoding.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+    if (is_signed && (abs_value <= 0x1F))
+    {
+      encoded.Resize(1);
+      encoded[0] = 0xE0 | static_cast<uint8_t>(abs_value);
+    }
+    else
+    {
+      // determine the corresponding number of bytes required to encode this value
+      uint64_t const bits_to_encode  = 64ull - platform::CountLeadingZeroes64(abs_value);
```
ejfitzgerald, line 66: This is basically the area of the code which had the logical flaw

libs/ledger/include/ledger/chain/transaction_encoding.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+      uint64_t partial_value{0};
+      for (std::size_t index = (1u << log2_value_length) - 1u;; --index)
+      {
+        // read the next byte
+        uint8_t encoded_byte{0};
+        buffer.ReadBytes(&encoded_byte, 1);
+
+        // build up the partial value
+        partial_value |= static_cast<uint64_t>(encoded_byte) << (index * 8u);
```
ejfitzgerald, line 183: This also had a logical flaw here

wkaluza (approved):

libs/ledger/include/ledger/chain/transaction_encoding.hpp
```diff
@@ -0,0 +1,206 @@
...
+{
+  using fetch::byte_array::ByteArray;
+
+  bool const is_signed = meta::IsSignedInteger<T> && (value < 0);
+
+  ByteArray encoded;
+  if (!is_signed && (value <= T{0x7f}))
+  {
+    encoded.Resize(1);
+    encoded[0] = static_cast<uint8_t>(value & 0x7f);
```
pbukva, line 52: Nitpicking - the ` & 0x7f` masking should not be necessary (based on `if(...)` context)

libs/ledger/include/ledger/chain/transaction_encoding.hpp
```diff
@@ -0,0 +1,206 @@
...
+  }
+  else
+  {
+    uint64_t const abs_value = ToU64(value);
+
+    if (is_signed && (abs_value <= 0x1F))
```
pbukva, line 58: The `is_signed` can be dropped based on context (`else {...}`)

PR #1066, opened 2019-05-27
Fix Reactor Busy Waits

ejfitzgerald:
Fix the busy waits that were caused by the `BlockCoordinator` and `MainChainRpcService` busy waiting on the idle state.

In addition also remove a busy wait for the MuddleRpcClient Background Worker

### DIFF
--- a/libs/network/src/muddle/rpc/client.cpp
+++ b/libs/network/src/muddle/rpc/client.cpp
@@ -137,14 +137,18 @@ void Client::BackgroundWorker()
 
   std::list<MuddleEndpoint::Response> pending_promises;
 
+  std::size_t idle_loops{0};
   while (running_)
   {
+    bool was_idle{true};
+
     // attempt to extract and promises from the queue
     {
       std::unique_lock<std::mutex> lk(promise_queue_lock_);
       if (!promise_queue_.empty())
       {
         pending_promises.splice(pending_promises.end(), promise_queue_);
+        was_idle = false;
       }
     }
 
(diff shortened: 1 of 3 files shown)


wkaluza (dismissed):

libs/network/src/muddle/rpc/client.cpp
```diff
@@ -180,16 +180,13 @@ void Client::BackgroundWorker()
...
 
-      // if we are sustained a period of idleness then we should sleep the thread
-      if (idle_loops >= 3)
-      {
-        std::this_thread::sleep_for(std::chrono::milliseconds{100});
-      }
+    // if we are in sustained a period of idleness then we should sleep the thread
+    if (consecutive_idle_loops >= 3)
+    {
+      std::this_thread::sleep_for(std::chrono::milliseconds{100});
```
wkaluza, line 189: Shouldn't `consecutive_idle_loops` be reset to zero here?

wkaluza (changes requested):

libs/network/src/muddle/rpc/client.cpp
ejfitzgerald: That wasn't my intention, no. As soon as this loop becomes non-idle then the count is reset. This will mean that in the idle case it will have a poll interval of 100ms.

Please note that in general, if there is no pending promises, this code will never execute because the thread should be pending on the `promise_queue_cv_`. After it has woken up from this event it will reset this count.

wkaluza (approved):

PR #1068, opened 2019-05-28
benchmarks for math and ml activation functions

baykaner:
picking up and extending work from PR #969


MissingNO57 (approved):
LGTM, just wait for green tick.

PR #1079, opened 2019-05-29
Fix integer parse areas for values >= 32bits

ejfitzgerald:


libs/ledger/include/ledger/chain/transaction_encoding.hpp
```diff
@@ -0,0 +1,206 @@
...
+  if (!is_signed && (value <= T{0x7f}))
+  {
+    encoded.Resize(1);
+    encoded[0] = static_cast<uint8_t>(value & 0x7f);
+  }
+  else
+  {
+    uint64_t const abs_value = ToU64(value);
+
+    if (is_signed && (abs_value <= 0x1F))
```
pbukva, line 58: Is signed check should not be necessary, given the `else` context.

wkaluza (approved):

libs/ledger/include/ledger/chain/transaction_encoding.hpp
ejfitzgerald: Nope, still necessary because it might be unsigned but greater that 7bits in length

PR #1080, opened 2019-05-29
Implemented Optimizers and updated Graph

MissingNO57:


libs/ml/include/ml/graph.hpp
```diff
@@ -77,6 +84,18 @@ class Graph : public ops::Trainable<T>
     trainable_[name] = op;
   }
 
+  template <class OperationType>
+  meta::IfIsGraph<ArrayType, OperationType, void> AddTrainable(
```
baykaner, line 82: lets move the implementations out of the class body. Also one of your AddTrainable functions is missing comments

libs/ml/include/ml/graph.hpp
```diff
@@ -282,5 +301,65 @@ void Graph<ArrayType>::LoadStateDict(struct fetch::ml::StateDict<ArrayType> cons
...
+ * @return ret is vector containing values for all weights
+ */
+template <typename ArrayType>
+std::vector<ArrayType> Graph<ArrayType>::GetWeights() const
+{
+  std::vector<ArrayType> ret;
+
+  for (auto const &t : trainable_)
+  {
+    ret.push_back(t.second->GetWeights());
```
baykaner: emplace_back

libs/ml/include/ml/graph.hpp
```diff
@@ -282,5 +301,65 @@ void Graph<ArrayType>::LoadStateDict(struct fetch::ml::StateDict<ArrayType> cons
...
+ * @return ret is vector containing all gradient values
+ */
+template <typename ArrayType>
+std::vector<ArrayType> Graph<ArrayType>::GetGradients() const
+{
+  std::vector<ArrayType> ret;
+
+  for (auto const &t : trainable_)
+  {
+    ret.push_back(t.second->Gradients());
```
baykaner: emplace_back

libs/ml/include/ml/graph.hpp
```diff
@@ -282,5 +301,65 @@ void Graph<ArrayType>::LoadStateDict(struct fetch::ml::StateDict<ArrayType> cons
...
 }
 
+/**
+ * Assigns all trainable weights parameters to vector of ArrayType for exporting and serialising
+ * @return ret is vector containing values for all weights
+ */
+template <typename ArrayType>
+std::vector<ArrayType> Graph<ArrayType>::GetWeights() const
+{
+  std::vector<ArrayType> ret;
```
baykaner, line 261: point of clarification - GetWeights is not used repeatedly in a training loop (i.e. for the optimiser) correct?
because this implementation is expensive of course

libs/ml/include/ml/graph.hpp
```diff
@@ -282,5 +301,65 @@ void Graph<ArrayType>::LoadStateDict(struct fetch::ml::StateDict<ArrayType> cons
...
+}
+
+/**
+ * Add gradient values to weight for each trainable
+ * @param grad vector of gradient values for each trainable stored in ArrayType
+ */
+template <typename ArrayType>
+void Graph<ArrayType>::ApplyGradients(std::vector<ArrayType> &grad)
+{
+  typename ArrayType::SizeType i = 0;
```
baykaner: here you're mixing an iterator with indexing. probably better to use two iterators: i.e.
auto grad_it = grad.begin();
for(auto const &t : trainable_)
{
  t.second->ApplyGradient(*grad_it);
  ++grad_it;
}

libs/ml/include/ml/ops/loss_functions/criterion.hpp
```diff
@@ -0,0 +1,44 @@
...
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "ml/ops/ops.hpp"
+#include <cassert>
```
baykaner: do you need this include? i don't see an assert

libs/ml/include/ml/ops/multiply.hpp
```diff
@@ -46,27 +46,19 @@ class Multiply : public fetch::ml::ElementWiseOps<T>
...
   /**
    * elementwise multiplication gradient is:
    * f'(input0)=input0*error_signal
    * f'(input1)=input1*error_signal
    */
-  virtual std::vector<ArrayPtrType> Backward(VecTensorType const &inputs, ArrayPtrType error_signal)
+  virtual std::vector<ArrayType> Backward(VecTensorType const &inputs,
+                                          ArrayType const &    error_signal)
   {
+    (void)inputs;
```
baykaner: is this necessary? you do use inputs right?

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -89,12 +93,26 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<T>
     return false;
   }
 
-  virtual void Step(typename T::Type learningRate)
+  virtual void Step(typename T::Type learning_rate)
   {
-    this->gradient_accumulation_->InlineMultiply(-learningRate);
+    this->gradient_accumulation_->InlineMultiply(-learning_rate);
     this->output_->InlineAdd(*gradient_accumulation_);
     // Major DL framework do not do that, but as I can't think of any reason why, I'll leave it here
```
baykaner: remove comment

libs/ml/include/ml/optimization/adagrad_optimizer.hpp
```diff
@@ -0,0 +1,91 @@
...
+public:
+  using ArrayType = T;
+  using DataType  = typename ArrayType::Type;
+  using SizeType  = typename ArrayType::SizeType;
+
+  AdaGradOptimizer(std::shared_ptr<Graph<T>> graph, std::string const &input_node_name,
+                   std::string const &output_node_name, DataType const &learning_rate)
+    : Optimizer<T, C>(graph, input_node_name, output_node_name, learning_rate)
+  {
+    auto weights = this->graph_->GetWeights();
```
baykaner: this is kind of an expensive operation when all we really want to do is get the shapes of these objects. may be we should have a GetWeightShapes?

libs/ml/include/ml/optimization/adagrad_optimizer.hpp
```diff
@@ -0,0 +1,91 @@
...
+    SizeType i{0};
+    for (auto &grad : gradients)
+    {
+      auto git = grad.begin();
+      auto cit = cache_[i].begin();
+      while (git.is_valid())
+      {
+        *cit += (*git) * (*git);
+        // 1e-8 is added to prevent division by 0
+        *git = -this->learning_rate_ * ((*git) / fetch::math::Sqrt(*cit + DataType{1e-8f}));
```
baykaner: epsilon value should be a member variable and an input parameter to constructor - not hard-coded

libs/ml/include/ml/optimization/adagrad_optimizer.hpp
```diff
@@ -0,0 +1,91 @@
...
+//
+//------------------------------------------------------------------------------
+
+#include "math/standard_functions/sqrt.hpp"
+#include "ml/graph.hpp"
+#include "ml/ops/loss_functions/criterion.hpp"
+#include "ml/optimization/optimizer.hpp"
+
+namespace fetch {
+namespace ml {
```
baykaner: maybe we wrap these in optimisers namespace?

libs/ml/include/ml/optimization/adagrad_optimizer.hpp
```diff
@@ -0,0 +1,91 @@
...
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "math/standard_functions/sqrt.hpp"
+#include "ml/graph.hpp"
+#include "ml/ops/loss_functions/criterion.hpp"
+#include "ml/optimization/optimizer.hpp"
```
baykaner: btw optimizer is american-english. optimiser is british-english. I believe our code style is supposed to be british-english

libs/ml/include/ml/optimization/adam_optimizer.hpp
```diff
@@ -0,0 +1,118 @@
...
+      DataType mt;
+      DataType vt;
+      while (git.is_valid())
+      {
+        *cit = (beta1_t_ * (*cit)) + ((DataType{1} - beta1_t_) * (*git));
+        mt   = *cit / (DataType{1} - beta1_t_);
+        *mit = (beta2_t_ * (*mit)) + ((DataType{1} - beta2_t_) * (*git) * (*git));
+        vt   = *mit / (DataType{1} - beta2_t_);
+        // 1e-4 is added to prevent division by 0
+        *git = -this->learning_rate_ * mt / (fetch::math::Sqrt(vt) + DataType{1e-4f});
```
baykaner: dont hard code epsilon

libs/ml/include/ml/optimization/optimizer.hpp
```diff
@@ -0,0 +1,82 @@
...
+  {}
+
+  DataType DoBatch(ArrayType &data, ArrayType &labels)
+  {
+    DataType loss{0};
+    SizeType n_data = data.shape().at(0);
+
+    for (SizeType step{0}; step < n_data; ++step)
+    {
+      auto cur_input = data.Slice(step).Copy();
```
baykaner: this slice.copy is expensive. I recommend that for now you leave this implementation but please raise an issue that this needs to be optimised

libs/ml/include/ml/optimization/rmsprop_optimizer.hpp
```diff
@@ -0,0 +1,94 @@
...
+    SizeType i{0};
+    for (auto &grad : gradients)
+    {
+      auto git = grad.begin();
+      auto cit = cache_[i].begin();
+      while (git.is_valid())
+      {
+        *cit = decay_rate_ * (*cit) + (DataType{1} - decay_rate_) * (*git) * (*git);
+        // 1e-8 is added to prevent division by 0
+        *git = -this->learning_rate_ * ((*git) / fetch::math::Sqrt(*cit + DataType{1e-8f}));
```
baykaner: dont hard code epsilons

libs/ml/include/ml/optimization/rmsprop_optimizer.hpp
```diff
@@ -0,0 +1,94 @@
...
+      this->cache_.push_back(ArrayType(wei.shape()));
+    }
+    ResetCache();
+  }
+
+private:
+  void ApplyGradients()
+  {
+
+    std::vector<ArrayType> gradients = this->graph_->GetGradients();
```
baykaner: this looks like a needless expense. Apply Gradients is getting called repeatedly in the training loop, and you are instantiating a vector inside GetGradients, then copying it out into another newly instantiated vector in ApplyGradients. At the very least we can make gradients_ a member variable. this applies for all your optimisers

libs/ml/include/ml/optimization/rmsprop_optimizer.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+    // Do operation with gradient
+    SizeType i{0};
```
baykaner: why index with i when you can iterate?
auto cached_weight_it = cache_.begin();
for ()
{
  ...
  auto cit = (*cached_weight_it).begin();
...
}

libs/ml/include/ml/optimization/rmsprop_optimizer.hpp
```diff
@@ -0,0 +1,94 @@
...
+    for (auto &grad : gradients)
+    {
+      auto git = grad.begin();
+      auto cit = cache_[i].begin();
+      while (git.is_valid())
+      {
+        *cit = decay_rate_ * (*cit) + (DataType{1} - decay_rate_) * (*git) * (*git);
```
baykaner: It might be nice to have a comment somewhere explaining this line, or at the very least break it into several lines.
Also you're reusing that DataType{1} - might as well make it a member variable

baykaner (changes requested):

libs/ml/include/ml/graph.hpp
MissingNO57: No, it is here just for test purposes, but can be used for something else in the future.

libs/ml/include/ml/optimization/rmsprop_optimizer.hpp
MissingNO57: In Graph, I would like to replace  std::unordered_map<std::string, TrainablePtrType>  with  std::vector<TrainablePtrType> and just return const pointer to that class, but it would completely break StateDict, so I would like to do that as part of future StateDict refinement process.

libs/ml/include/ml/optimization/optimizer.hpp
MissingNO57: Slice should give you something that will behave exactly like Tensor rather than TensorSlice.

libs/ml/include/ml/ops/weights.hpp
MissingNO57: Comment removed, but I will remove Step as part of refactorization anyway.

libs/ml/include/ml/optimization/adagrad_optimizer.hpp
MissingNO57: I replaced that with Trainable::GetWeights() which will give you constant reference to Trainable.

libs/ml/include/ml/graph.hpp
```diff
@@ -284,5 +251,160 @@ void Graph<ArrayType>::LoadStateDict(struct fetch::ml::StateDict<ArrayType> cons
...
+
+  return ret;
+}
+
+/**
+ * Assigns all trainable pointers to vector for optimiser purpose
+ * @return ret is vector containing pointers to all trainables
+ */
+template <typename ArrayType>
+std::vector<typename std::shared_ptr<fetch::ml::ops::Trainable<ArrayType>>>
```
baykaner, line 397: To improve compute speed we could either take in an empty vector reference as an input parameter (which is a bit of a clumsy interface I guess), or we could use return as a move:
return std::move(ret);

libs/ml/include/ml/ops/loss_functions/cross_entropy.hpp
```diff
@@ -42,21 +43,25 @@ class CrossEntropy
   CrossEntropy()          = default;
   virtual ~CrossEntropy() = default;
 
+  /**
+   * @param inputs vector of 2 one-hot Tensors
```
baykaner: I think this is not true, since this should handle a non-on-hot tensor with two classes as well (i.e. for binary logistic regression) - also same for comment on backward

libs/ml/include/ml/optimisation/adagrad_optimiser.hpp
```diff
@@ -0,0 +1,101 @@
...
+namespace optimisers {
+
+/**
+ * Adaptive Gradient Algorithm optimiser
+ * i.e. Modified stochastic gradient descent with per-parameter learning rate
+ * @tparam T ArrayType
+ * @tparam C CriterionType
+ */
+template <class T, class C>
+class AdaGradOptimiser : public Optimiser<T, C>
```
baykaner, line 36: btw I know these optimisers classes are quite short (100 lines); but I would still like them written with only declarations in the class body and function definitions below. It's still a lot easier to read for most developers if they look and see immediately "okay this class just has a constructor an ApplyGradients(), and a ResetCache(). Plus it maintains consistency with the larger classes

libs/ml/include/ml/optimisation/adam_optimiser.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+    beta1_(beta1)
+    , beta2_(beta2)
+    , beta1_t_(beta1)
+    , beta2_t_(beta2)
+    , epsilon_(epsilon)
+  {
+    for (auto &train : this->graph_trainables_)
+    {
+      this->cache_.push_back(ArrayType(train->GetWeights().shape()));
```
baykaner: should these pushes be emplaces?

libs/ml/include/ml/optimisation/adam_optimiser.hpp
```diff
@@ -0,0 +1,150 @@
...
+    for (auto &moment : this->momentum_)
+    {
+      moment.Fill(DataType{0});
+    }
+    beta1_t_ = beta1_;
+    beta2_t_ = beta2_;
+  }
+
+  std::vector<ArrayType> cache_;
+  std::vector<ArrayType> momentum_;
```
baykaner, line 52: in general it's nicer to have the member variables at the top too (so when you read a function you already know which are member variables); this is made easier when the function definitions are outside of the class since the class body becomes something like this:

class Name
{
  public:
    .. public variables
    .. public method declarations

  private:
    .. private variables
    .. private method declarations
}
.. public definitions
.. private definitions

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -0,0 +1,99 @@
...
+  {
+    graph_trainables_ = graph_->GetTrainables();
+
+    for (auto &train : graph_trainables_)
+    {
+      this->gradients_.push_back(ArrayType(train->GetWeights().shape()));
+    }
+  }
+
+  // Todo 1090: Optimise TensorSlice for graph-feeding without using .Copy
```
baykaner: please write todos likes this: TODO (private 1090)

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -0,0 +1,99 @@
...
+    // Compute and apply gradient
+    ApplyGradients();
+
+    return loss;
+  }
+
+protected:
+  std::shared_ptr<Graph<T>>                                          graph_;
+  CriterionType                                                      criterion_;
+  std::string                                                        input_node_name_;
```
baykaner: nice to add defaults, e.g.:
std::string input_node_name_ = "";
DataType learning_rate_ = NumericMax<DataType>();

baykaner (changes requested):

libs/ml/include/ml/optimisation/optimiser.hpp
MissingNO57: Default for learning rate should be:  DataType const &learning_rate = DataType{0.001f}
And not telling input_node_name shouldn't be permitted.

baykaner (approved):

PR #1086, opened 2019-05-30
Generic way to bind functors to VM callables (wkaluza fork)

wkaluza:


robertdickson:
Agreed about std::result_of, I was under the impression that it was nicer/better but as you have discovered, it isn't. The new FunctorTraits are nice.

robertdickson (approved):
LGTM

PR #1092, opened 2019-05-30
Add automated check for missing include guards

wkaluza:
Yes, we have missing include guards.


scripts/apply_style.py
```diff
@@ -290,14 +318,18 @@ def main():
...
+
     output('Done.')
 
     if args.diff:
         diff = get_diff()
         if diff:
-            print('*' * 80)
-            print(diff)
-            print('*' * 80)
+            output('*' * 80)
```
ejfitzgerald, line 330: One for the todo list - we can replace `output` for `print` provided we run the script with `PYTHONUNBUFFERED=1`

ejfitzgerald (approved):

scripts/apply_style.py
wkaluza: Doesn't that mean that our way is therefore safer? I doubt the performance improvement would be measurable - I expect this script is limited by disk access by a long way.

PR #1093, opened 2019-05-30
Fix for bahviour of default value in State<...> constructor

### LINKED ISSUE #1073: Possible bug: state var assigned and not explicitly .set() usable but not globally persistent
From developer slack user. Surrounding editorial is:

>Here's another one (not sure if this is a bug):
>when the value of a state var is only assigned and not explicitly .set() it is locally useable, but not globally persistent (see test2() getting different value).
>At least docu does not state this. And I think this might lead to much lost money in contracts and would suggest assigned values should be persistent (globally)

Issue could be documentation related. 

[stateSet.txt](https://github.com/uvue-git/fetch-ledger/files/3231475/stateSet.txt)

pbukva:
Closes: #1073

### DIFF
--- a/libs/vm/include/vm/state.hpp
+++ b/libs/vm/include/vm/state.hpp
@@ -120,7 +120,7 @@ inline IoObserverInterface::Status WriteHelper(std::string const &name, Ptr<Obje
 
   // convert the type into a byte stream
   ByteArrayBuffer buffer;
-  if (!val->SerializeTo(buffer))
+  if (!val || !val->SerializeTo(buffer))
   {
     return IoObserverInterface::Status::ERROR;
   }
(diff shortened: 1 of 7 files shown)


libs/vm/include/vm/state.hpp
```diff
@@ -149,10 +149,25 @@ class State : public IState
...
   }
 
-  ~State() override = default;
+  ~State() override
+  {
+    try
+    {
+      FlushIO();
+    }
+    catch (...)
```
ejfitzgerald, line 171: Think that it is worth catching the `std::exception` here and then added the message to the runtime error.

ejfitzgerald:
FYI: @fetchai/docs This new `State` behaviour will need to be updated in the documentation

libs/vm/include/vm/state.hpp
pbukva: Can do, I was originally thinking about it, but over the time I grew dislike of multiple `catch` blocks in name of lazyness and concise code (we will have to have `catch(...)` block anyway as ultimate fallback since we are in destructor).

ejfitzgerald (approved):

ejfitzgerald:
@pbukva I am going to land this because i want to cut the release this morning

PR #1096, opened 2019-05-30
Miscellaneous changes

wkaluza:
Highlights include:

* Four files' worth of dead code in `libs/miner/`
* Run a cleaning step before running end-to-end tests - right now e2e tests rely on the previous testing stage cleaning up as part of its teardown, which is unnecessary coupling and could bite us in the future


ejfitzgerald (changes requested):
Please keep the annealer miner code also can you raise and issue for reinstating the module as an option

ejfitzgerald (approved):

PR #1098, opened 2019-05-30
Feature/fixedpoint trig functions

markos:
Add all major trigonometry functions for FixedPoint, functions include:

Sin(), Cos(), Tan(), ASin(), ACos(), ATan(), ATan2(), SinH(), CosH(), TanH(), ASinH(), ACosH(), ATanH().

Most functions should perform well within the accuracy limits imposed by the datatype (eg. <1e-4 for FixedPoint<16,16>, and <1e-7 for FixedPoint<32,32>), unless noted otherwise.

Remaining issues to be done in a next PR: 
* move the polynomial constants outside the function space
* finish tests for infinity


libs/math/include/math/trigonometry.hpp
```diff
@@ -30,42 +30,20 @@ namespace math {
...
 template <typename ArrayType>
-fetch::math::meta::IfIsMathNonFixedPointArray<ArrayType, void> Sin(ArrayType const &x,
-                                                                   ArrayType &      ret)
+fetch::math::meta::IfIsMathArray<ArrayType, void> Sin(ArrayType const &x, ArrayType &ret)
 {
-  assert(ret.size() == x.size());
-  kernels::Sin<typename ArrayType::Type> s;
-  auto                                   x_it = x.cbegin();
-  auto                                   rit  = ret.begin();
+  ASSERT(ret.size() == x.size());
```
baykaner: ASSERT macro probably not required here - that macro is helpful when a parameter is unused other than in the assertion; but in this case all of your parameters are used so an ordinary lower_case assert is fine

libs/math/tests/math/all_tests/fixed_point/fixed_point.cpp
```diff
@@ -753,6 +753,26 @@ TEST(FixedPointTest, Exponential_16_16)
...
+
+  fp32   step{0.001};
+  fp32   x{-10.0};
+  double max_error = 0, avg_error = 0;
+  size_t iterations = 0;
+  double tolerance  = 2 * (double)fp32::TOLERANCE;
+  for (; x < 5.0; x += step)
+  {
+    fp32   e     = fp32::Exp(x);
+    double r     = std::exp((double)x);
```
baykaner: i know this is a pedantic comment for a test, but static_cast<double>() is preferred over c-style casts like this

libs/math/tests/math/all_tests/fixed_point/fixed_point.cpp
```diff
@@ -753,6 +753,26 @@ TEST(FixedPointTest, Exponential_16_16)
...
+    double delta = std::abs((double)e - r);
+    max_error    = std::max(max_error, delta);
+    avg_error += delta;
+    iterations++;
+  }
+  avg_error /= (double)iterations;
+  EXPECT_NEAR(max_error, 0.0, 10 * tolerance);
```
baykaner, line 774: is there any special significant to this value 10 * tolerance?
Do we have some kind of expected variance characteristic for this test?

libs/math/tests/math/all_tests/fixed_point/fixed_point.cpp
```diff
@@ -786,39 +807,172 @@ TEST(FixedPointTest, Exponential_32_32)
...
+  {
+    fp64   e     = fp64::Exp(x);
+    double r     = std::exp((double)x);
+    double delta = std::abs((double)e - r);
+    max_error    = std::max(max_error, delta);
+    avg_error += delta;
+    iterations++;
+  }
+  avg_error /= (double)iterations;
+  EXPECT_NEAR(max_error, 0.0, 10 * tolerance);
```
baykaner, line 851: thinking more about this max error test: isn't it pretty much guaranteed to have the highest error for the highest value of x? in which case seems like it makes more sense to just check this edge case. but maybe I've misunderstood

libs/math/tests/math/all_tests/fixed_point/fixed_point.cpp
```diff
@@ -786,39 +807,172 @@ TEST(FixedPointTest, Exponential_32_32)
...
+      double delta = std::abs((double)e - r);
+      max_error    = std::max(max_error, delta);
+      avg_error += delta;
+      iterations++;
+    }
+  }
+  avg_error /= (double)iterations;
+  EXPECT_NEAR(max_error, 0.0, tolerance);
+  EXPECT_NEAR(avg_error, 0.0, tolerance);
+  std::cout << "Pow: max error = " << max_error << std::endl;
```
baykaner: I don't think you need to print these things out; the test is testing the bounds, so the cout is only really useful for debugging or deeper investigation. When we run all tests and are looking for the one in a thousand that failed this will tend to fill up the console with irrelevant data to sift through.

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
```diff
@@ -1037,6 +659,2027 @@ FixedPoint<I, F> const FixedPoint<I, F>::CONST_MIN{FixedPoint::FromBase(FixedPoi
...
+ * @tparam T any float/double type
+ * @param n float/double value to set FixedPoint to
+ */
+template <std::uint16_t I, std::uint16_t F>
+template <typename T>
+constexpr FixedPoint<I, F>::FixedPoint(T n, meta::IfIsFloat<T> *)
+  : data_(static_cast<typename FixedPoint<I, F>::Type>(n * ONE_MASK))
+{
+  assert(CheckNoOverflow(n * ONE_MASK));
+  //    assert(CheckNoRounding<T>());
```
baykaner, line 730: leftover code?

baykaner (changes requested):
only minor style issues. Fixed point class looking very beautiful now!

libs/math/tests/math/all_tests/fixed_point/fixed_point.cpp
markos: this was mostly for my own information and to put in the comments of the mathematical functions, so that people have an idea of the expected accuracy, but sure I can comment out or remove from the tests altogether.

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
markos: Well, not sure about rounding. It's a fact that we will have rounding at least for the 2 FP types <16,16> and <32,32>, and not for <64,64> or bigger types. So the CheckNoRounding<T> feels somehow redundant at the moment.

libs/math/include/math/trigonometry.hpp
markos: fixed.

libs/math/tests/math/all_tests/fixed_point/fixed_point.cpp
markos: fixing now.

libs/math/tests/math/all_tests/fixed_point/fixed_point.cpp
markos: the correct thing would be to draw a graph over the whole defined range for each function, but the 10* is just empirical, the average error on all values tests should be ok for most purposes, however there exist arguments for most functions that cause slightly bigger errors. I only wanted to make sure that the max errors were still small and smaller 1e-7 in the case of FP<32,32>, where tolerance is ~1e-10. I hope that explains it better.

libs/math/tests/math/all_tests/fixed_point/fixed_point.cpp
markos: I am not sure what you mean about highest value of x, it may be true for some monotonic functions, but not all, it depends on the approximation used I guess, eg for tan(x), max_error will really skyrocket around PI/2.

libs/math/tests/math/all_tests/fixed_point/fixed_point.cpp
markos: done.

libs/math/tests/math/all_tests/fixed_point/fixed_point.cpp
markos: I will comment out these prints for normal test runs.

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
markos: I will leave it there for now and will revisit CheckNoRounding when/if we tackle larger types like <64,64>

baykaner (approved):

PR #1102, opened 2019-05-31
[VM]: Support for `ShardedState<T>`

### LINKED ISSUE #799: VM: Persistent Map implementation

pbukva:
Resolves:  #799


libs/vm/src/object.cpp
```diff
@@ -116,143 +116,143 @@ std::size_t Object::GetHashCode()
...
 {
-  RuntimeError("operator not implemented");
+  RuntimeError("IsLessThanOrEqual operator not implemented");
   return false;
 }
 
 bool Object::IsGreaterThan(Ptr<Object> const & /* lhso */, Ptr<Object> const & /* rhso */)
 {
-  RuntimeError("operator not implemented");
+  RuntimeError("IsGreaterThan operator not implemented");
```
ejfitzgerald: 👍 👍 👍

libs/vm/src/module.cpp
```diff
@@ -203,11 +204,16 @@ Module::Module()
   address.CreateMemberFunction("signed_tx", &Address::HasSignedTx);
 
   auto istate = GetClassInterface<IState>();
-  istate.CreateConstuctor<Ptr<String>, TemplateParameter>();
-  istate.CreateConstuctor<Ptr<Address>, TemplateParameter>();
+  istate.CreateConstuctor<Ptr<String>, TemplateParameter1>();
```
ejfitzgerald: I think that we should keep the original code here

libs/vm/include/vm/variant.hpp
```diff
@@ -475,37 +475,23 @@ struct Variant
...
-struct AnyInteger : public Variant
-{
-  using Variant::Variant;
-};
-
-struct AnyFloatingPoint : public Variant
-{
-  using Variant::Variant;
-};
+using TemplateParameter1 = VariantT<0>;
```
ejfitzgerald: This is just for a unique value from `typeid` right. Is there any downside from this approach in terms of getting the string type name? Also would it make sense to simple use `__LINE__` for the value here so that the number doesn't need to be incremented?

Or are these questions better directed at @robertdickson

libs/vm/include/vm/state.hpp
```diff
@@ -31,11 +31,11 @@ class IState : public Object
   IState()          = delete;
   virtual ~IState() = default;
 
-  static Ptr<IState> Constructor(VM *vm, TypeId type_id, Ptr<String> const &name,
+  static Ptr<IState> Constructor(VM *vm, TypeId type_id, Ptr<Object> &&name,
                                  TemplateParameter const &value);
 
-  static Ptr<IState> Constructor(VM *vm, TypeId type_id, Ptr<Address> const &address,
-                                 TemplateParameter const &value);
+  //  static Ptr<IState> Constructor(VM *vm, TypeId type_id, Ptr<Object> const &address,
```
ejfitzgerald: Remove the dead code?

apps/vm-lang/main.cpp
```diff
@@ -336,5 +336,10 @@ int main(int argc, char **argv)
     state_map.SaveToFile(data_path.c_str());
   }
 
+  if (!success)
+  {
+    std::cout << error << std::endl;
+  }
+
```
wkaluza: Unreachable code - see line 316.

libs/vm/include/vm/object.hpp
```diff
@@ -209,6 +221,11 @@ class Object
   virtual bool SerializeTo(ByteArrayBuffer &buffer);
   virtual bool DeserializeFrom(ByteArrayBuffer &buffer);
 
+  TypeId getTypeId() const
```
ejfitzgerald: Style: `type_id()`

libs/ledger/tests/chaincode/contract_test.hpp
```diff
@@ -32,32 +32,45 @@
...
+  retval.SetAllOne();
+  return retval;
+}
+
 class ContractTest : public ::testing::Test
 {
+  // ContractTest() = default;
+  // ContractTest(fetch::BitVector shards)
+  //  : shards_{shards}
+  //{}
```
wkaluza: Dead code

libs/ledger/tests/chaincode/contract_test.hpp
```diff
@@ -32,32 +32,45 @@
...
+inline auto FullShards(std::size_t lane_count)
+{
+  fetch::BitVector retval{lane_count};
+  retval.SetAllOne();
+  return retval;
+}
+
 class ContractTest : public ::testing::Test
 {
+  // ContractTest() = default;
```
ejfitzgerald: Dead code

apps/vm-lang/main.cpp
```diff
@@ -336,5 +336,10 @@ int main(int argc, char **argv)
     state_map.SaveToFile(data_path.c_str());
   }
 
+  if (!success)
```
ejfitzgerald: This code will never be called

ejfitzgerald (changes requested):

libs/ledger/tests/chaincode/smart_contract_tests.cpp
```diff
@@ -201,117 +194,14 @@ TEST_F(SmartContractTests, CheckQueryReturnTypes)
...
-    EXPECT_EQ(response["status"].As<ConstByteArray>(), "success");
-  }
+  VerifyQuery("get_bool", true);
+  VerifyQuery("get_int32", int32_t{14});
+  VerifyQuery("get_uint32", uint32_t{15});
+  VerifyQuery("get_int64", int64_t{16});
+  VerifyQuery("get_uint64", uint64_t{17});
+  VerifyQuery("get_float", float{1.0});
+  VerifyQuery("get_double", double{2.0});
+  VerifyQuery("get_string", ConstByteArray{"Why hello there"});
```
wkaluza, line 204: :+1: :+1: :+1:

libs/vm/include/vm/object.hpp
```diff
@@ -87,6 +88,17 @@ struct IsAddress<T, typename std::enable_if_t<std::is_base_of<Address, T>::value
 {
 };
 
+template <typename T, typename = void>
+struct IsString : std::false_type
+{
+};
+template <typename T>
+struct IsString<
+    T, typename std::enable_if_t<std::is_base_of<String, typename std::decay<T>::type>::value>>
```
wkaluza: `is_base_of_v`, `decay_t`?

libs/vm/include/vm/state.hpp
```diff
@@ -31,11 +31,11 @@ class IState : public Object
...
+  //                                 TemplateParameter const &value);
```
wkaluza: dead code?

libs/vm/include/vm/state.hpp
```diff
@@ -128,24 +136,24 @@ inline IoObserverInterface::Status WriteHelper(std::string const &name, Ptr<Obje
...
 {
 public:
   // Construct state object, default argument = get from state DB, initializing to value if not
   // found
-  State(VM *vm, TypeId type_id, TypeId value_type_id, Ptr<String> name,
+  State(VM *vm, TypeId type_id, TypeId value_type_id, Ptr<Object> &&name,
         TemplateParameter const &value)
     : IState(vm, type_id)
-    , name_{std::move(name)}
+    , name_{NameToString(vm, std::move(name))}
```
wkaluza: use `std::forward` with universal ref

libs/vm/include/vm/state.hpp
```diff
@@ -285,28 +291,40 @@ inline Ptr<IState> IState::Construct(VM *vm, TypeId type_id, Args &&... args)
...
+  return ConstructIntrinsic(vm, type_id, value_type_id, std::forward<Args>(args)...);
+}
+
+inline Ptr<IState> IState::Constructor(VM *vm, TypeId type_id, Ptr<Object> &&name,
                                        TemplateParameter const &value)
 {
   if (name)
   {
-    return Construct(vm, type_id, name, value);
+    return Construct(vm, type_id, std::move(name), value);
```
wkaluza: std::forward

libs/vm/include/vm/state.hpp
```diff
@@ -285,28 +291,40 @@ inline Ptr<IState> IState::Construct(VM *vm, TypeId type_id, Args &&... args)
...
+inline std::string IState::NameToString(VM *vm, Ptr<Object> &&name)
 {
-  if (address)
+  switch (name->getTypeId())
   {
-    return Construct(vm, type_id, address->AsString(), value);
+  case TypeIds::String:
+    return dynamic_cast<String const &>(*name).str;
+  case TypeIds::Address:
+    return dynamic_cast<Address &>(*name).AsString()->str;
```
wkaluza: why not const ref? Also, what happens if this throws?

libs/vm/include/vm/state.hpp
```diff
@@ -285,28 +291,40 @@ inline Ptr<IState> IState::Construct(VM *vm, TypeId type_id, Args &&... args)
...
   }
 
-  vm->RuntimeError("Failed to construct State: address is null");
-  return nullptr;
+  vm->RuntimeError(
+      "Unsupported type of `name` parameter for `State<...>(name)` constructor. It must be either "
+      "`String` or `Address` type.");
+  return std::string{};
```
wkaluza: redundant type: I'd prefer `{}` or `""`

libs/vm/include/vm/variant.hpp
```diff
@@ -475,37 +475,23 @@ struct Variant
   }
 };
 
-struct TemplateParameter1 : public Variant
+template <std::size_t ID = 0>
+struct VariantT : public Variant
 {
   using Variant::Variant;
+  VariantT(Variant const &other)
```
wkaluza: Should this be explicit?

libs/vm/include/vm/variant.hpp
```diff
@@ -475,37 +475,23 @@ struct Variant
...
+    : Variant{other} {};
```
wkaluza: redundant semicolon

libs/vm/include/vm/variant.hpp
```diff
@@ -475,37 +475,23 @@ struct Variant
...
+  VariantT(Variant &&other)
+    : Variant{std::move(other)} {};
```
wkaluza: redundant semicolon; use std:;forward

libs/vm/include/vm/variant.hpp
```diff
@@ -475,37 +475,23 @@ struct Variant
   }
 };
 
-struct TemplateParameter1 : public Variant
+template <std::size_t ID = 0>
```
wkaluza: Why default to 0? (as in, why have a default at all?)

wkaluza (changes requested):

libs/vm/src/module.cpp
pbukva: Sure I can do.
I changed it to make the code with intention to make it consistent (with other generic types like Map or PersistentMap).

I would argue we should rather drop the `TemaplateParameter` type alias and use everywhere the indexed type `TemaplateParameter1`.

libs/vm/include/vm/variant.hpp
troelsfr: `VariantT` is capital T capitalised purposefully?

libs/vm/include/vm/variant.hpp
pbukva: All these new types are (by design) meant to be equivalent to `Variant` type *and* at the same time being different between each other from C++ type system perspective.

Types in original implementation were lacking construction from `Variant` base class (note that original implementation used the `using Variant::Variant;` to propagate constructors but it is not enough to achieve that goal), thus it is necessary to add these constructors by hand.
This change avoids unnecessary copy-paste of per each type in original implementation.

I'm really not quite sure about using `__LINE__`.
Reason being that it somewhat masks the original intention, and makes the code dependent on formatting (e.g. in the case if someone actually wants to use `typeid(...).name` to identify the type, then these types could have different names from build to build if there was added new code somewhere above this types are declared (since it will change line numbers)).

libs/vm/include/vm/state.hpp
pbukva: The `std::forward<T>(...)` makes sense only in perfect forwarding case, what requires the method being templated with `T` type as **method** template parameter (**not class** template paremeter).
In this case the method is not templated (we are dealing here with specific (upfront known) type `Ptr<Object>`).

libs/vm/include/vm/state.hpp
pbukva: See my comment https://github.com/fetchai/ledger/pull/1102/files/fc3e625d2d5d0ac68e68ef551e4538d7959f8a53#r289524102

apps/vm-lang/main.cpp
pbukva: I don't even know how this bit of code got here. I suspect it has to be artifact of weird merge of master branch in to resurrected branch of original PR.

Btw, there are other weird pieces of code similar to this one above (lines 327-331) which is not supposed here either.

libs/vm/src/object.cpp
```diff
@@ -116,143 +116,143 @@ std::size_t Object::GetHashCode()
 
 bool Object::IsEqual(Ptr<Object> const & /* lhso */, Ptr<Object> const & /* rhso */)
 {
-  RuntimeError("IsEqual operator not implemented");
+  RuntimeError(std::string(__func__) + "operator not implemented");
```
ejfitzgerald, line 119: Is `__func__` portable? I am sure that we have a system independent macro somewhere for this?

libs/vm/src/parser.cpp
```diff
@@ -27,9 +27,8 @@ namespace fetch {
 namespace vm {
 
 Parser::Parser()
-{
-  template_names_ = {"Matrix", "Array", "Map", "State"};
-}
+  : template_names_{"Matrix", "Array", "Map", "State", "PersistentMap"}
+{}
```
wkaluza, line 31: Could `template_names_` be const now?

libs/vm/src/module.cpp
robertdickson: I think I agree with Peter -- drop the TemplateParameter alias and use TemplateParameter1. The actual name of the type inside the compiler and VM, btw, is "TemplateParameter1".

libs/vm/include/vm/object.hpp
robertdickson: This is GetTypeId() elsewhere in the code

libs/vm/src/module.cpp
robertdickson: ...Actually it's "[TempateParameter1]" but the point still stands.

libs/vm/include/vm/variant.hpp
robertdickson: What's the point of this VariantT? If you want to get the official string name of a type at runtime you can ask the VM for it via VM::GetTypeInfo(type_id).name  or   GetTypeInfo(GetTypeId< T >).name

libs/vm/src/object.cpp
pbukva: Yes, it is portable, because of C99: https://isocpp.org/wiki/faq/cpp11-language-misc#cpp11-c99

libs/vm/include/vm/variant.hpp
pbukva: The very same question can beasked about all default values across the whole project.
Default is default, and zero makes the most sense based on all aspects in given context.
I don't mind really, can be dropped.

libs/vm/include/vm/variant.hpp
pbukva: The way how the code is writtenright now, switching here to the `std::forward<...>` does not make sense unless this is the perfect forwarding.
In order of beying perfect-forwardable, the method/function must templated with one of template parameters beying type of paremeter which is intended to be forwareded.
Otherwise it is just move, because it is explicitly declared that this is cnstructor *rvalue* type .

The constructor would need to be templated, as follow:
```
template<typename T>
VariantT(T &&other) : Variant{std::forward<T>(other)}
{}
```

However, IF we make the constructor templated with perferct forwarding, as drafted in above example, we do not need to have 2 cosntructors (we can drop copy constructor since it will be covered with perfect forwarding constructor).

libs/vm/include/vm/variant.hpp
pbukva: The primary objecive of the `VariantT<...>` is to implement 2 missing constructors from `Variant` (copy & move constructors).
Getting the string represention of the type (it's `name`) was branched-out discussion which derailed us off the primary objective.

robertdickson: But why do you need those 2 constructors? The Variant already has templated Move() and Get() calls to transfer/copy a Variant into a derived-Variant like AnyPrimitive, and such mechanisms are in use, for example, to transfer Variant's on the VM stack in to function parameters, see StackGetter < T >

libs/vm/include/vm/state.hpp
pbukva: Because `AsString()` method is not declared as `const`.
I wanted to change that, however it it not possible to do as easily as ti looks like, since the method uses internal state (`vm_` member) the way that complicate to change the method to `const`.

libs/vm/include/vm/state.hpp
pbukva: ok

libs/vm/include/vm/object.hpp
pbukva: This is all over this file and other files of VM library.
I'm having another branch addressing this and many other issue of this kind.

robertdickson: Addressing what? Is there an issue here?

libs/vm/src/object.cpp
ejfitzgerald: Yep, I know what should happen but I would suggest that almost certainly MSVC will not support this for instance. In any case, it makes sense to me to make a fetch specific macro for this so that when we run into this case, it is a simple update to a single header file.

libs/vm/include/vm/object.hpp
pbukva: No, there is no an issue per say, I used _issues of this kind_ only to refer **verbosity** kind of issues pointed out by the comment on top which triggered this discussion - instead of using unnecesarily verbose form of meta traits like `std::is_base_of<...>::value` and `typename std::decay<...>::type`, we can use less versose  `std::is_base_of_v<...>` and `std::decay_t<...>`,
where in the first case dropped `::value` and in the second case we dropped even more - unnecessary `typename` and `::type`

libs/vm/include/vm/state.hpp
pbukva: Regarding the second comment about _what happens if it throws_: We can ask exactly the same question about almost every line in the VM code - we are calling methods to dependencies - for example serialisation, IO observer, constructors, etc. ..., where almost evry such a call can potentially throw. Even `vm->RuntimeError(...)` can potentially throw.

VM has protection mechanism - the top level try-catch block where the otherwise uncaught exception will be caught and properly handled.

libs/vm/src/object.cpp
pbukva: VS 2015 or later supports `__func__` for C++11+ code.

pbukva: We can add such macro when it will be really needed (it will be easy addition in one header file). However I'm really struggling to understand what we are trying to solve here - the `__func__`  has been made part of C++11 standard - if we are going to question that, than I'm not quite sure how to respond. (Just on side note, **all** relevant C++ compilers I know of do support it)

libs/vm/include/vm/variant.hpp
pbukva: This is just to make types easily constructible from Variant. They are just empty wrappers around `Variant` base type after all.

libs/vm/include/vm/variant.hpp
pbukva: The `Variant` base type does not have any constructor declared as explicit. All these types are, by design, just empty wrappers around `Variant` base type, with the only desired difference (when compared to Variant type itself) that they have different type id in C++ type system.
I can make these constructors explicit, but I don't know what benefit it actually brings.

ejfitzgerald (commented):
Looks okay from my side. Just make sure @robertdickson is happy with it

libs/vm/include/vm/analyser.hpp
```diff
@@ -134,6 +134,7 @@ class Analyser
   TypePtr           matrix_type_;
   TypePtr           array_type_;
   TypePtr           map_type_;
+  TypePtr           persistent_map_type_;
```
robertdickson: This should be changed to sharded_state_type_

libs/vm/include/vm/sharded_state.hpp
```diff
@@ -0,0 +1,53 @@
...
+  // Factory
+  static Ptr<IShardedState> Constructor(VM *vm, TypeId type_id, Ptr<String> const &name);
+  static Ptr<IShardedState> Constructor(VM *vm, TypeId type_id, Ptr<Address> const &name);
+
+  // Construction / Destruction
+  IShardedState(VM *vm, TypeId type_id, Ptr<String> const &name, TypeId value_type)
+    : Object(vm, type_id)
+    , name_{name->str}
+    , value_type_{value_type}
+  {}
```
robertdickson, line 36: Please use value_type_id and value_type_id_ here

libs/vm/src/sharded_state.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+namespace fetch {
+namespace vm {
+namespace {
+class ShardedState : public IShardedState
+{
+public:
+  // Construction / Destruction
+  ShardedState(VM *vm, TypeId type_id, Ptr<String> const &name, TypeId value_type)
+    : IShardedState(vm, type_id, name, value_type)
```
robertdickson: same drill, change to value_type_id

libs/vm/src/sharded_state.cpp
```diff
@@ -0,0 +1,132 @@
...
+  {}
+
+protected:
+  using ByteBuffer = std::vector<uint8_t>;
```
robertdickson: Is this ByteBuffer used anywhere?

libs/vm/src/sharded_state.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+  void SetIndexedValue(Ptr<Address> const &key, TemplateParameter1 const &value) override
+  {
+    if (!key)
+    {
+      RuntimeError("Index is null reference.");
+    }
+    SetIndexedValueInternal(key->AsString(), value);
+  }
+
```
robertdickson, line 70: String key could be null too, no?

libs/vm/src/sharded_state.cpp
```diff
@@ -0,0 +1,132 @@
...
+Ptr<IShardedState> IShardedState::Constructor(VM *vm, TypeId type_id, Ptr<Address> const &name)
+{
+  if (name)
+  {
+    TypeInfo const &type_info     = vm->GetTypeInfo(type_id);
+    TypeId const    value_type_id = type_info.parameter_type_ids[0];
+    return new ShardedState(vm, type_id, name->AsString(), value_type_id);
+  }
+
+  vm->RuntimeError("Failed to construct SharededState instance: the `name` is null reference.");
```
robertdickson: SharededState -- here and above, spelling.

libs/vm/src/sharded_state.cpp
pbukva: Yes, but it is checked for null in side of `SetIndexedValueInternal(...)` by `ComposeFullKey(...)` method.

pbukva:
This is peculiar fail of unit test. I can't reproduce it when running the same configuration in our docker containe or of our fetch-ledger-develop image.

robertdickson (approved):
Looks good

ejfitzgerald (approved):

wkaluza (approved):

PR #1107, opened 2019-06-01
Extend apply_style and refactor for full parallelism

wkaluza:
The main impetus behind this change is the long runtime of the `apply-style.py` script reported by some users. The three key changes which aim to reduce the runtime are the following:

* End-to-end parallelism. Instead of sequentially walking the directory tree to reformat C/C++, then Python, and then CMake files, the list of files to check is composed up front and processed in parallel. The checks and fixes applicable for each file are selected based on the file's name and extension
* Minimised disk access. Each file is read exactly once and written to at most once.
* Git-aware file selection. The `-c`/`--commit` option allows the user to specify a hash or ref, against which the current working copy will be diffed, e.g. `apply_style.py -c upstream/master`. The script will the only process the files which appear in the diff.

As a result, a branch with a hundred changed files can be processed in less than a second on a Linux PC.

The resulting structure is extensible, with clear, sequential transformations applied to each file. To take advantage of that, I pulled in the license check and CMake version check into this script using a fraction of their original code.

I also introduced a new check which ensures that every source file's last line has a line ending ( https://unix.stackexchange.com/a/18789 ), which accounts for the bulk of this PR. This will probably break existing PRs; I will notify ledger devs before merging, so they know to sync with `master`.

To keep things simple, I removed most of the CLI options, leaving only `--jobs` and `--diff` in addition to `--commit`; I suspect most of them were no longer in use.

There is one functional difference between this script and the current version - because we now pass the code to be formatted to `clang-format` using standard input rather than pointing it at a file path, `clang-format` has no way to identify the header which is named identically to the current source file (if any) in order to put it at the top of the include list. Where this affected output, I separated that header by a blank line so it remained at the top of the file.


scripts/apply_style.py
```diff
@@ -89,13 +91,17 @@ def find_clang_format():
...
             '--separate-ctrl-name-with-space',
             '--line-width', '100',
-            '--in-place'
+            '-'
         ],
         'filename_patterns': ('*.cmake', 'CMakeLists.txt')
     }
 }
 
+NUM_FILES_PROCESSEFD_SO_FAR = 0  # Global variable
```
ejfitzgerald: We have generally been using `SCREAMING_SNAKE_CASE` for contants. Consider refactoring to normal `snake_case` to make it more consistent

scripts/apply_style.py
```diff
@@ -104,233 +110,277 @@ def output(text):
...
+
+        increment_file_count()
+    except:
+        output(traceback.format_exc())
+
+
+def process_in_parallel(files_to_process, jobs):
+    output('Processing {} file(s) with {} job(s) ...'
+           .format(TOTAL_FILES_TO_PROCESS, jobs))
+    with ThreadPoolExecutor(max_workers=jobs) as pool:
```
ejfitzgerald: I think you will surprised but I don't think this is going to do what want because of python's GIL. (in the case of the calling external tools it is fine). It would be probably worth switching to `ProcessPoolExecutor`

ejfitzgerald (changes requested):

scripts/apply_style.py
wkaluza: That's done. The performance increase is quite modest, as profiling indicates that about 95% of the time is spent calling out to the external formatters anyway. I had trouble finding a way of synchronising the progress indicators when using `ProcessPoolExecutor`, so I switched to using lower level `multiprocessing` utilities.

n-hutton (commented):
So, if I understand this correctly, there is only one apply style now?

Does this break my script quickly_run?

wkaluza:
> Does this break my script quickly_run?

Oops. Fixed.

n-hutton (approved):
lgtm

ejfitzgerald (approved):

PR #1108, opened 2019-06-01
Add macOS builds

wkaluza:
Due to issue #1192 this is only a Release build.

The flow is the following:
* Each PR receives it's own workspace (checkout) on the mac-mini
* The first build is clean for obvious reasons. This should be relatively quick now that ccache is installed and enabled
* Subsequent builds are incremental (no workspace wipes in between)
* Only merge branches and master are built on the mac-mini, as machine time on it is a scarce resource
* Only fast unit tests are executed post-build (no slow, integration or e2e tests)
* Jenkins should delete the workspace automatically when the PR is closed


ejfitzgerald (approved):
Seems good to me

PR #1110, opened 2019-06-01
Add Array extend member function

wkaluza:
This ties a bow on the non-initialisation portion of #812.

### DIFF
--- a/libs/vm/include/vm/array.hpp
+++ b/libs/vm/include/vm/array.hpp
@@ -19,8 +19,10 @@
 
 #include "vm/vm.hpp"
 
+#include <algorithm>
 #include <cstddef>
 #include <cstdint>
+#include <iterator>
 
 namespace fetch {
 namespace vm {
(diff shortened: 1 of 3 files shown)


libs/vm/include/vm/array.hpp
```diff
@@ -179,6 +182,17 @@ struct Array : public IArray
...
 
+  void Extend(Ptr<IArray> const &other) override
+  {
+    Ptr<Array<ElementType>> const &other_array    = other;
+    auto const &                   other_elements = other_array->elements;
+
+    elements.reserve(elements.size() + other_elements.size());
+
+    auto insert_iterator = std::back_inserter(elements);
+    std::copy(other_elements.cbegin(), other_elements.cend(), insert_iterator);
```
bipll: Why not `elements.insert()`? (Should probably even eliminate the need for explicit `reserve()`.)

bipll (changes requested):

libs/vm/include/vm/array.hpp
wkaluza: Good idea. I'd rather leave the `reserve` call in, though, as cppreference mentions no guarantees that `insert` reserves the required space in advance.

bipll: Well, it mentions that time complexity is linear which implies at-most O(1) reallocations, which in turn implies pre-reservation, but ok.

bipll (approved):

PR #1111, opened 2019-06-01
Fix issues with etch print(); add tests

wkaluza:
The `print` function is extensively used, but has only been implicitly tested so far. This PR creates a dedicated suite of tests and also extends the print module's functionality with additional overloads. This is not all there is to be done on this front, but I think it's a step in the right direction. In particular, stringifying compound types like maps and multidimensional arrays needs further work.

* There was one case in the `print` function's implementation where it wrote directly to `std::cout`, rather than to an output device provided by the VM
* The `printLn` functions were issuing a `std::endl` manipulator to effect a linebreak, which has the side-effect of flushing the output stream's buffer. The non-breaking `print` functions now issue `std::flush` for consistency.
* All integer and Boolean types types are now printable
* The `etch` type Byte is now called `UInt8` to bring it into line with the other integral types; I am under the impression that `Byte` is marked for deprecation, but please correct me if that is not the case
* `print.hpp` contained a lot of implementation which only it cared about - this has been moved to a `.cpp` file.
* Some tests which occurred to me are not currently supported by the language or would significantly increase the scope of this PR. I left them in, disabled and marked with TODOs.


ejfitzgerald (commented):
Does it makes sense to completely remove all references to the Byte object. I think that it is still currently in the analyser for example

wkaluza:
> Does it makes sense to completely remove all references to the Byte object. I think that it is still currently in the analyser for example

@ejfitzgerald Could you point out which bits you mean? As far as I can tell, neither the analyser, tokenizer nor the grammar mention the byte type after these changes.

robertdickson (approved):
LGTM

PR #1120, opened 2019-06-03
[VM] Fix for Variant serialisation issue

### LINKED ISSUE #1118: [VM] Variant serialisation issue exposes VM to undefined behaviour
The `Serialize(...)` & `Deserialize(...)` functions for `vm::Variant` are accessing the `vm::Variant::object` member of **union** disregard of whether the union has been actually initialised/set via `object` member originally.
Beside of the fact that accessing union like this is undefined behavior,  this also opens the the way for possibly executing the `SerialiseTo(...)` & `DeserialiseFrom(...)` methods on `vm::Variant::object` which were not meant to be executed.

pbukva:
Resolves: #1118

### DIFF
--- a/libs/vm/include/vm/variant.hpp
+++ b/libs/vm/include/vm/variant.hpp
@@ -545,15 +545,14 @@ inline void Serialize(ByteArrayBuffer &buffer, Variant const &variant)
     buffer << variant.primitive.f64;
     break;
   default:
-    break;
-  }
-
-  if (variant.object)
-  {
-    if (!variant.object->SerializeTo(buffer))
+    if (variant.object)
     {
-      throw std::runtime_error("Unable to serialize type");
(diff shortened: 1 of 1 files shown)


ejfitzgerald (changes requested):
While I sort of understand this fix, I can't see why the original code will break given correctly formulated variants. Lets have a chat about this fix in the office

pbukva:
The original code accessed the `varian.object` union member for **all** calls (and thus called on it  `Ptr<Object>::bool()` casting operator overload) disregard whether whether the union carries `primitive` or `object` member.
Thus put this in another way, the `varian.object::bool ()` casting operator was being called for all _primitive_ types.

ejfitzgerald (approved):

ejfitzgerald:
@pbukva  - I have merged this so that I can cherry pick to the release branch

PR #1124, opened 2019-06-03
Add installer

theoturner:
Adds installer automating the build steps, including the slightly awkward `DOPENSSL_ROOT_DIR` on macOS. Readme updated accordingly.


install.sh
```diff
@@ -0,0 +1,44 @@
...
+    xcode-select --version || xcode-select --install
+    brew --version || yes | /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+    python3.7 --version || brew install python
+    cmake --version || brew install cmake
+    brew install openssl
+    for version in `ls -t /usr/local/Cellar/openssl/`; do
+        ossl=/usr/local/Cellar/openssl/$version
+        break
+    done
+fi
```
ejfitzgerald: I think that it would be good to catch the `else` case here and exit with an error.

ejfitzgerald (changes requested):
Hi Theo,

A really interesting change. In general I like this approach but I would recommend renaming the file to `quickstart.sh` or something similar. Can you also move the script into the `scripts/` folder.

install.sh
theoturner: Renamed and moved. Catch error if the openssl install is not present. Also added minor change if script run from scripts dir, jump up to root dir and continue as before.

theoturner:
Final commit is merge of 3f8cdb8

ejfitzgerald:
Sorry @theoturner we had a CI issue which is preventing this to be merged. Could you merge in the latest changes from `master` and then we can land this feature.

theoturner:
@ejfitzgerald merged and changed `make -j` to `make -j 4` in the installer.

ejfitzgerald (approved):

ruthbr:
I'd probably just change '(in the case where you have been running a local network)' to just '(if you have been running a local network)'

PR #1127, opened 2019-06-03
Feature/666 main chain forward refs

bipll:
Reopening #979:
- forward references are kept in mainchain's cache and permanent store;
- consequently, block removal from cache and tips reindexing reworked;
------
- core/logger.hpp undergone some refactoring;
- debug build logs were fixed (for that moment).


libs/core/src/logger.cpp
```diff
@@ -28,27 +28,26 @@ log::details::LogWrapper logger;
 
 namespace log {
 Context::Context(void *instance)
+  : details_{std::make_shared<ContextDetails>(instance)}
```
ejfitzgerald, line 31: 👍

ejfitzgerald (dismissed):
Looks good to me lets get the CI appeased and we should get this merged. Nice one!

wkaluza:
@bipll be aware of this week's merge which added timestamps to Block. Several of our tests in the area where you have been seeing failures were overspecified to check that identical, but otherwise unrelated blocks have the same hash. Two blocks that have identical contents will not have the same hash anymore unless their timestamps match, and any assertions to that effect no longer hold true.

ejfitzgerald (approved):

PR #1132, opened 2019-06-04
Sgns simple

baykaner:
- merging cbow_simple with sgns implementation
- saving trained vectors
- implementing analogy evaluation


gheaeckkseqrz:
Not sure if that helps : https://github.com/pierre-wilmot/Word2Vec-CPP
With some explainations : https://gheaeckkseqrz.wordpress.com/2019/05/18/word2vec-from-a-developer-perspective/

This runs at 9 min per epoch, single threaded / no vectorisation.
It's using the Tensor class you don't like, but shouldn't be too hard to adapt.

libs/math/tests/math/all_tests/clustering/knn_test.cpp
```diff
@@ -36,41 +37,13 @@ TYPED_TEST_CASE(ClusteringTest, MyTypes);
 
 TYPED_TEST(ClusteringTest, knn_euclidean_test)
 {
-  using DataType  = typename TypeParam::Type;
   using ArrayType = TypeParam;
   using SizeType  = typename TypeParam::SizeType;
 
-  ArrayType A = ArrayType({4, 4});
+  ArrayType A = ArrayType::FromString("1, 2, 3, 4; 2, 3, 4, 5; -1, -2, -3, -4; -2, -3, -4, -5");
```
MissingNO57, line 43: Is FromString safe now? Shouldn't we use that R() stuff?

libs/math/include/math/clustering/knn.hpp
```diff
@@ -27,36 +27,61 @@ namespace details {
...
+
+  // test vector must be either {1, N} or {N, 1}
+  assert((vec.shape().at(0) == 1) || (vec.shape().at(1) == 1));
+
+  // array must be either {M, N} (if vector shape is {1, N}) or {N, M}
+  assert(((array.shape().at(1) == vec.shape().at(1)) && (vec.shape().at(0) == 1)) ||
+         ((array.shape().at(0) == vec.shape().at(0)) && (vec.shape().at(1) == 1)));
+
+  // vec <1, M>; array <N, M>
+  //  assert(array.shape().at(1) == vec.shape().at(1));
```
MissingNO57: Leftover code

libs/math/include/math/clustering/knn.hpp
```diff
@@ -27,36 +27,61 @@ namespace details {
...
-  for (SizeType i(0); i < array.shape().at(0); ++i)
+  std::vector<std::pair<SizeType, DataType>> ret;
+
+  similarities.reserve(array.shape().at(data_axis));
+
+  // compute distances
+  for (SizeType i = 0; i < array.shape().at(data_axis); ++i)
   {
-    typename ArrayType::Type d = Distance(one_vector, array.Slice(i).Copy());
+    DataType d = Distance(vec, array.Slice(i, data_axis).Copy());
```
MissingNO57, line 66: It's not possible now, but we should get rid of Copy and replace .Slice with .View, but View will need to take offset in order to work same way as Slice. Maybe we will not even need Slice at all.
DataType d=Distance(vec, array.View(data_axis,i));

libs/math/include/math/correlation/cosine.hpp
```diff
@@ -32,26 +32,52 @@ template <typename ArrayType>
...
+  else
+  {
+    dp_ret = fetch::math::TransposeDot(b, b);
+  }
   assert(dp_ret.size() == 1);
   b_r = dp_ret[0];
+  fetch::math::Sqrt(b_r, b_r);
 
-  denom_r = fetch::math::Sqrt(fetch::math::Multiply(a_r, b_r));
+  fetch::math::Multiply(a_r, b_r, denom_r);
```
MissingNO57, line 75: Are you sure that SQRT(a x b) == a x SQRT(b)?
SQRT(4 x 4)=4
4 x SQRT(4)=8

I would suggest
fetch::math::Multiply(a_r, b_r,denom_r);
fetch::math::Sqrt(denom_r, denom_r);

MissingNO57 (changes requested):
I found bug in libs/math/include/math/correlation/cosine.hpp

libs/math/tests/math/all_tests/clustering/knn_test.cpp
baykaner: nope - it's safe to use like this now. FromString was fixed and the raw literals aren't necessary

libs/math/include/math/clustering/knn.hpp
baykaner: suggested change will be part of an upcoming PR as discussed

libs/math/include/math/correlation/cosine.hpp
baykaner: fixed previous wrong implementation

libs/math/include/math/clustering/knn.hpp
```diff
@@ -67,72 +89,78 @@ std::vector<std::pair<typename ArrayType::SizeType, typename ArrayType::Type>> G
 
 }  // namespace details
 
+/**
+ */
```
MissingNO57: Some leftover?

libs/math/tests/math/all_tests/clustering/knn_test.cpp
```diff
@@ -84,41 +57,13 @@ TYPED_TEST(ClusteringTest, knn_euclidean_test)
...
-  v.Set(SizeType{0}, SizeType{0}, DataType(3));
-  v.Set(SizeType{0}, SizeType{1}, DataType(4));
-  v.Set(SizeType{0}, SizeType{2}, DataType(5));
-  v.Set(SizeType{0}, SizeType{3}, DataType(6));
+  ArrayType A = ArrayType::FromString("1, 2, 3, 4; 2, 3, 4, 5; -1, -2, -3, -4; -2, -3, -4, -5");
+  ArrayType v = ArrayType::FromString("3, 4, 5, 6");
 
-  std::vector<std::pair<typename ArrayType::SizeType, typename ArrayType::Type>> output =
-      fetch::math::clustering::KNNCosine(A, v, 4);
+  auto output = fetch::math::clustering::KNNCosine(A, v, 4);
```
MissingNO57: I don't want to be too pedantic, but you are using namespace fetch::math::clustering;

libs/math/include/math/clustering/knn.hpp
MissingNO57: Some leftover?

MissingNO57 (commented):
There are just some minor style issues.

MissingNO57 (approved):
LGTM

Issue #1137, opened 2019-06-04
Specify Memory Requirements


Genysys:
Problem: Unable to build image with docker.

Description:

Docker adding docker images to a repo solves the issue of cross device compatibility. I was trying to build an image for the latest ubuntu image , but could not, even with reallocating 12 GB of memory to the process.

I have eventually had to get the biggest Azure CPU with 256 GB to build it (I know its over kill). 

Would be helpful if there were some guidelines on this.

ejfitzgerald:
Hi,

Sorry to here you are having issues building the ledger. We have seen this issue from time to time. When we have seen this before it was an issue with `make` not detecting the number of CPUs on the system and spinning up a large number of jobs. Can you try the following:

`make -j 4 constellation`

This would limit the build to 4 concurrent jobs.

Hope that helps

PR #1138, opened 2019-06-04
Readme: Limit concurrency of build in instructions

ejfitzgerald:
A few people (#1137) have reported issues with building the ledger especially in cases where they are built in the cloud.

The proposed change limits the stated number of cores for the build. 4 seems a reasonable amount and I think from memory would require around 4GB of memory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ The project uses cmake so you can following formal build procedure of:
 
     cmake ../
 
-    make -j
+    make -j 4
 
 ## Connecting to a test network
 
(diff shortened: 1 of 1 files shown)


ejfitzgerald:
@fetchai/docs can we also make sure this change is also reflected across all our other documentation

devjsc:
Yup

wkaluza (approved):

katharine-murphy:
updated on the community site installation guide (currently, we don't
detail this in the etch documentation).

PR #1142, opened 2019-06-04
Use builder pattern in Module ctor; enforce internal linkage

wkaluza:


libs/vm/src/module.cpp
```diff
@@ -143,72 +149,122 @@ double toFloat64(VM * /* vm */, AnyPrimitive const &from)
   return Cast<double>(from);
 }
 
+class AllClassInterfaces
```
pbukva: It feels like this is unnecessarily complex.
The `ClassInterface<T>` design supports *builder* design pattern, which allows to fully setup the interface using exclusivelly r-value (= no necessity to create l-value variable).

As example, this can be demonstrated on `Array` type:
```cpp
  GetClassInterface<IArray>()
    .CreateConstuctor<int32_t>()
    .CreateMemberFunction("count", &IArray::Count)
    .CreateMemberFunction("append", &IArray::Append)
    .CreateMemberFunction("popBack", &IArray::PopBackOne)
    .CreateMemberFunction("popBack", &IArray::PopBackMany)
    .CreateMemberFunction("popFront", &IArray::PopFrontOne)
    .CreateMemberFunction("popFront", &IArray::PopFrontMany)
    .CreateMemberFunction("reverse", &IArray::Reverse)
    .CreateMemberFunction("extend", &IArray::Extend)
    .EnableIndexOperator<AnyInteger, TemplateParameter>()
    .CreateInstantiationType<Array<bool>>()
    .CreateInstantiationType<Array<int8_t>>()
    .CreateInstantiationType<Array<uint8_t>>()
    .CreateInstantiationType<Array<int16_t>>()
    .CreateInstantiationType<Array<uint16_t>>()
    .CreateInstantiationType<Array<int32_t>>()
    .CreateInstantiationType<Array<uint32_t>>()
    .CreateInstantiationType<Array<int64_t>>()
    .CreateInstantiationType<Array<uint64_t>>()
    .CreateInstantiationType<Array<float>>()
    .CreateInstantiationType<Array<double>>()
    .CreateInstantiationType<Array<Ptr<String>>>();
```

libs/vm/src/module.cpp
```diff
@@ -143,72 +149,122 @@ double toFloat64(VM * /* vm */, AnyPrimitive const &from)
...
+public:
+  ClassInterface<Address> address;
+  ClassInterface<IArray>  iarray;
+  ClassInterface<IMap>    imap;
+  ClassInterface<IMatrix> imatrix;
+  ClassInterface<IState>  istate;
+  ClassInterface<String>  string;
+};
+
+void InstantiateTypes(AllClassInterfaces &interfaces)
```
pbukva: Splitting aspects of type setup from *single* place in to multiple new functions obscures the view on how the type is actually setup.
Before, the type was fully setup on signle place, where developer can clearly see all spects without necessity to dig around to get the full picture.

libs/vm/src/module.cpp
```diff
@@ -143,72 +149,122 @@ double toFloat64(VM * /* vm */, AnyPrimitive const &from)
...
+{
+  // Ordering must match that in AllClassInterfaces
+  return {m->GetClassInterface<Address>(), m->GetClassInterface<IArray>(),
+          m->GetClassInterface<IMap>(),    m->GetClassInterface<IMatrix>(),
+          m->GetClassInterface<IState>(),  m->GetClassInterface<String>()};
+}
+
+}  // namespace
+
 Module::Module()
```
pbukva, line 154: I would suggest to go back to the original concept (put everything back in to this method), but use the *builder* pattern as it was shown in above comment https://github.com/fetchai/ledger/pull/1142/files#r290517903.

pbukva (changes requested):
Seee my inline comments.

libs/vm/src/module.cpp
wkaluza: Context for this change: this came out of my conversation with @robertdickson yesterday. I think the current arrangement is not extensible, as it interleaves registering new type instantiations and creating member functions which may need certain types to have been registered with the VM. For example, with the current shape of the code on `master`, it's impossible to add a `String.split(String) : Array<String>` member function without moving the whole String block below the Array block, where `Array<String>` is instantiated. This kind of connascence is unnecessary, and can be avoided by registering types up front.

The proposed Builder pattern, while pretty, does not solve the problem relevant to this PR.

libs/vm/src/module.cpp
wkaluza: See my answer above.

libs/vm/src/module.cpp
wkaluza: See my answer above.

robertdickson: I think we stay with the original code, but change it to use the superior Builder pattern in general, as Peter suggests, but deal with interdependency by having multiple builder chains where necessary. The present difficulty (String depending on Array < String > ) can be solved by simply moving a GetClassInterface < String >()... chain to the end.

robertdickson: In the general cyclic case, A depends on B, B depends on A, you can create type A, create type B and _then_ define their member functions etc in two subsequent ClassInterface chains.

wkaluza: @robertdickson that's my worry here: eventually in order to account for these interdependencies, we'll have to set up bits of interfaces A and B (instantiations), followed by the rest of A and B (members etc.), and in the general case the `Module()` ctor becomes a big blob of stuff, all in one function. My rationale here was that we can get ahead of that mess and segregate the different parts of the interfaces' setup by category (first create all instantiations of all interfaces, then all members, etc.) rather than by interface (first we setup everything related to String, then everything related to Array...). In the former setup, here we don't have to fiddle with reordering lines within each category, ever again, because the categories are not interdependent, while the interfaces in general are, because instantiations need to be registered before they are used.

Another way to tackle this would be to make function and method definitions lazily validate that all the types had been registered with the VM after Module is constructed, but that's a big gun to roll out for such a small problem.

I'll reorder the String and Array block to make `split()` work and introduce the builder pattern, but I think that instantiating all the interfaces first would be the sustainable way to go.

libs/vm/src/module.cpp
pbukva: You can call `GetClassInterface<Array>()` at any point, disregard whether the `Array` type has been previously setup or not, for example, the following should work just fine:

```cpp
//This can be called anywhere (even before preceding `Array` full/partial setup), and is sort-of equivalent of C++ forward declaration of the class type.
GetClassInterface<IArray>();

//Then you can call do of the `String` type:
GetClassInterface<String>().CreateMemberFunction("split", &String::Split);

...

//Then sometime later you can finish/extend setup of `Array` type:
GetClassInterface<IArray>()
    .CreateConstuctor<int32_t>()
    .CreateMemberFunction("count", &IArray::Count)
    .CreateMemberFunction("append", &IArray::Append)
    .CreateMemberFunction("popBack", &IArray::PopBackOne)
    .CreateMemberFunction("popBack", &IArray::PopBackMany)
    . ...
```

libs/vm/src/module.cpp
pbukva: See my comment https://github.com/fetchai/ledger/pull/1142#discussion_r290641730 above.

libs/vm/src/module.cpp
robertdickson: That's not true I'm afraid. The IArray type has been created before, inside the Analyser.

libs/vm/src/module.cpp
robertdickson: This a solution in search of a problem at the moment....let's move on.

libs/vm/src/module.cpp
pbukva: You can call `GetClassInterface<Array>().CreateConstuctor()` at any point, disregard of whether the `Array` type has been previously setup or not, what makes it behave as sort-of equivalent of C++ *forward declaration* of the class type.

For example, the following should work just fine:

```cpp
//This can be called anywhere (even before preceding `Array` full/partial setup),
// and is sort-of equivalent of C++ forward declaration of the class type.
GetClassInterface<Array>().CreateConstuctor();

//Then you can call do of the `String` type:
GetClassInterface<String>().CreateMemberFunction("split", &String::Split);

...

//Then sometime later you can finish/extend setup of `Array` type:
GetClassInterface<IArray>()
    .CreateConstuctor<int32_t>()
    .CreateMemberFunction("count", &IArray::Count)
    .CreateMemberFunction("append", &IArray::Append)
    .CreateMemberFunction("popBack", &IArray::PopBackOne)
    .CreateMemberFunction("popBack", &IArray::PopBackMany)
    . ...
```

robertdickson: That's not true I'm afraid. The IArray type has been created before, inside the Analyser.

pbukva: I would rather introduce (hopefully) small change in to the `Module`, `Module::ClassInterface<T>` and `Compiler` classes, which would enable registration of the type without necessity/forcing to add methods to that type. This would be sort-of equivalent to the C++ forward declaration.
This can be implemented either directly in the constructor of the `Module::ClassInterface<T>` class (what would make it nice to declare the type by simply calling `GetClassInterface<T>()` function, or it can be made explicit by implementing it in separate dedicated method of the class, for instance in the `Module::ClassInterface<T>::ForwardDeclare()` method, so one would need to call then the `GetClassInterface<T>().ForwardDeclare()`.

Once the above would be done then you can simply `GetClassInterface<Array>()` at any point, disregard of whether the `Array` type has been previously setup or not, for example, the following should work just fine:

```cpp
//This can be called anywhere (even before preceding `Array` full/partial setup),
// and is sort-of equivalent of C++ forward declaration of the class type.
GetClassInterface<Array>();

//Then you can call do of the `String` type:
GetClassInterface<String>().CreateMemberFunction("split", &String::Split);

...

//Then sometime later you can finish/extend setup of `Array` type:
GetClassInterface<IArray>()
    .CreateConstuctor<int32_t>()
    .CreateMemberFunction("count", &IArray::Count)
    .CreateMemberFunction("append", &IArray::Append)
    .CreateMemberFunction("popBack", &IArray::PopBackOne)
    .CreateMemberFunction("popBack", &IArray::PopBackMany)
    . ...
```

robertdickson: You can register a type before defining its methods -- that's creating the class! Module has CreateClassType(). Template creation is not, by design, available to the Module system as it is private to the compiler.

robertdickson: Core primitives, String and the templates are created inside the compiler and are thus already registered.

pbukva: The problem with `Module::CreateClassType(std::string const &name)` method is that it requires the `name` string parameter.
I was looking in to this method originally, and that method felt like good candidate to start off with to implement that _forward declaration_ concept I mentioned above. In this sense I keep advocating to create separate dedicated method called `ForwardDeclare()` (or something like that) in order to signify that programmer can call that method without fear (anywhere and multiple times) that it will do something unexpected/sinister behind the scenes (e.g. creating multiple registration records in VM type-system for the same type, etc. ...)

robertdickson (approved):
LGTM

pbukva (approved):
LGTM

PR #1148, opened 2019-06-06
Block timestamps

wkaluza:
Blocks now have a timestamp counting seconds since the Unix epoch.

While the timestamp is part of the block hash, we *must not* rely on it for any quantitative information. In particular, timestamps are not strongly guaranteed to monotonically increase in line with the blockchain's topology (though they mostly will, barring shenanigans from malicious nodes).

### DIFF
--- a/libs/ledger/src/chain/block.cpp
+++ b/libs/ledger/src/chain/block.cpp
@@ -23,10 +23,9 @@
 #include "crypto/sha256.hpp"
 #include "ledger/chain/constants.hpp"
 
-#include <chrono>
+#include <cstddef>
 #include <cstdint>
-
-using namespace std::chrono;
+#include <ctime>
 
 namespace fetch {
 namespace ledger {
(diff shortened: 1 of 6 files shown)


libs/ledger/src/chain/block.cpp
```diff
@@ -76,5 +82,14 @@ void Block::UpdateDigest()
   proof.SetHeader(body.hash);
 }
 
+void Block::UpdateTimestamp()
+{
+  if (body.previous_hash != GENESIS_DIGEST)
+  {
+    const auto ms_since_epoch = duration_cast<milliseconds>(system_clock::now().time_since_epoch());
```
ejfitzgerald: Aren't Unix timestamps are seconds since epoch?

I am not sure `system_clock` is mandated to use UTC until C++20. Another approach would be to use `std::time()` + `std::gmtime()` + `std::mktime()` etc.

What are your thoughts?

wkaluza: They are seconds, but I thought we'd like millisecond precision on this field.

You're right, it's not mandated. I'll switch to the `<cstdlib>` functions to be safe.

ejfitzgerald (approved):

PR #1156, opened 2019-06-07
Batching at op level

MissingNO57:
Implemented:
BatchwiseFlatten
BatchwiseAdd
SubsetToArray for DataLoader base class
Optimiser parameter restart_after_epoch for loader

Added batch dimension for:
FullyConnected
Convolution1D
Convolution2D
MaxPool1D
MaxPool2D
LeakyReluOp
PRelu
Transpose
Embeddings

Fixed:
Tensor<T, C>::Transpose(SizeVector &new_axes)

Another changes:
Since all Ops supports batching, I removed ElementWiseOp and BatchOp, there are now just Ops.


n-hutton:
Closed due to inactivity

libs/math/include/math/matrix_operations.hpp
```diff
@@ -907,5 +907,63 @@ fetch::math::meta::IfIsMathArray<ArrayType, void> DynamicStitch(ArrayType &
   }
 }
 
+template <typename ArrayType>
+fetch::math::meta::IfIsMathArray<ArrayType, void> BatchwiseFlatten(ArrayType const &A,
+                                                                   ArrayType &      ret)
+{
+  // Test if batch dimensions are equal
```
baykaner: I think two functions should live in the ml library, because the math library has no concept of a "batch dimension", and makes no assumption about the trailing dimension being the "batch dimension"

libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -48,15 +48,15 @@ class FullyConnected : public SubGraph<T>
...
+    std::string flat_input = this->template AddNode<fetch::ml::ops::BatchwiseFlattenOp<ArrayType>>(
+        name + "_Flatten", {input});
     std::string weights =
         this->template AddNode<fetch::ml::ops::Weights<ArrayType>>(name + "_Weights", {});
     std::string weights_matmul = this->template AddNode<fetch::ml::ops::MatrixMultiply<ArrayType>>(
         name + "_MatrixMultiply", {weights, flat_input});
     std::string bias =
         this->template AddNode<fetch::ml::ops::Weights<ArrayType>>(name + "_Bias", {});
-    std::string output = this->template AddNode<fetch::ml::ops::Add<ArrayType>>(
+    std::string output = this->template AddNode<fetch::ml::ops::BatchwiseAddOp<ArrayType>>(
```
baykaner: I think this is the only place you are using the Batchwise Add; it seems like the BatchwiseAdd function can be replaced with a call to tensor_broadcast.hpp (+ a few assertions before the call maybe)

libs/math/include/math/matrix_operations.hpp
```diff
@@ -907,5 +907,63 @@ fetch::math::meta::IfIsMathArray<ArrayType, void> DynamicStitch(ArrayType &
...
+  for (SizeType i{0}; i < batch_size; i++)
+  {
+    auto input_slice  = A.Slice(i, input_batch_dimension);
+    auto output_slice = ret.Slice(i, output_batch_dimension);
+    output_slice.Assign(input_slice);
+  }
+}
+
+template <typename ArrayType>
+fetch::math::meta::IfIsMathArray<ArrayType, void> BatchwiseAdd(ArrayType const &A,
```
baykaner: I think this is a special case of tensor_broadcast.hpp

libs/math/include/math/tensor.hpp
```diff
@@ -1544,7 +1544,14 @@ Tensor<T, C> Tensor<T, C>::Transpose(SizeVector &new_axes) const
   assert(shape_.size() > 1);
   assert(shape_.size() == new_axes.size());
 
-  Tensor ret(shape());
+  SizeVector new_shape;
+  new_shape.reserve(new_shape.size());
+  for (SizeType i{0}; i < shape_.size(); ++i)
+  {
+    new_shape.push_back(shape_[new_axes[i]]);
```
baykaner: prefer new_shape.push_back(shape_.at(new_axes.at(i))

baykaner: also use an iterator:
while (auto & val: new_axes)
{
  new_shape.push_back(shape_.at(val));
}

libs/math/include/math/tensor.hpp
```diff
@@ -1560,7 +1567,7 @@ Tensor<T, C> &Tensor<T, C>::Squeeze()
 {
   // TODO(private issue 998): Make last dimension for efficiency
```
baykaner: remove the TODO if you've fixed this

libs/math/include/math/tensor.hpp
```diff
@@ -1576,7 +1583,7 @@ template <typename T, typename C>
 Tensor<T, C> &Tensor<T, C>::Unsqueeze()
 {
   auto shape = shape_;  // TODO: Make last dimension for efficiency
```
baykaner: todo

libs/ml/include/ml/ops/activations/batchwise_softmax.hpp
```diff
@@ -0,0 +1,135 @@
...
+#include <memory>
+#include <stdexcept>
+#include <vector>
+
+namespace fetch {
+namespace ml {
+namespace ops {
+
+template <class T>
+class BatchwiseSoftmax : public fetch::ml::BatchOps<T>
```
baykaner: It's not clear to me why we need a separate BatchwiseSoftmax and Softmax activation ops. These should be quite easy to unify into a single Softmax op that handles batches including batch size = 1

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -48,7 +48,8 @@ class Optimiser
                SizeType batch_size = 0);
 
   DataType Run(fetch::ml::DataLoader<ArrayType, ArrayType> &loader, SizeType batch_size = 0,
-               SizeType subset_size = fetch::math::numeric_max<SizeType>());
+               bool     restart_after_epoch = false,
```
baykaner: I think we can just drop this parameter and always treat this as false

libs/ml/examples/mnist/main.cpp
```diff
@@ -48,7 +48,7 @@ int main(int ac, char **av)
 {
   DataType learning_rate{0.01f};
   SizeType subset_size{100};
-  SizeType epochs{10};
+  SizeType epochs{30};
```
baykaner: necessary change?

libs/ml/include/ml/clustering/tsne.hpp
```diff
@@ -423,20 +423,26 @@ class TSNE
...
 
         // (Pij-Qij)
-        DataType tmp_val = tmp_p_i_j - tmp_q_i_j;
+        DataType val = p_i_j - q_i_j;
 
         // /(1+||yi-yj||^2)
-        fetch::math::Multiply(num.At(i, j), tmp_val, tmp_val);
+        fetch::math::Multiply(num.At(i, j), val, val);
 
         // tmp_val*(yi-yj), where tmp_val=(Pij-Qij)/(1+||yi-yj||^2)
```
baykaner: remove tmp from comment also

libs/ml/include/ml/clustering/tsne.hpp
```diff
@@ -423,20 +423,26 @@ class TSNE
...
-        tmp_slice += Multiply(tmp_val, (output_matrix.Slice(j).Copy().Squeeze()) -
-                                           (output_matrix.Slice(i).Copy().Squeeze()));
+
+        ArrayType diff = (output_matrix.Slice(j).Copy()) - (output_matrix.Slice(i).Copy());
+
+        auto shape = diff.shape();
+        shape.erase(shape.begin());
+        diff.Reshape(shape);
+
+        tmp_slice += Multiply(val, diff);
```
baykaner: better name for tmp slice?

libs/ml/include/ml/dataloaders/dataloader.hpp
```diff
@@ -31,9 +32,13 @@ class DataLoader
   DataLoader(bool random_mode)
     : random_mode_(random_mode)
   {}
+
+  using SizeType = fetch::math::SizeType;
+
   virtual ~DataLoader()                                         = default;
   virtual std::pair<LabelType, std::vector<DataType>> GetNext() = 0;
```
baykaner: using ReturnType = std::pair<LabelType, std::vector<DataType>>

libs/ml/include/ml/dataloaders/dataloader.hpp
```diff
@@ -42,5 +47,59 @@ class DataLoader
   bool random_mode_ = false;
 };
 
+template <typename LabelType, typename DataType>
+std::pair<LabelType, std::vector<DataType>> DataLoader<LabelType, DataType>::SubsetToArray(
```
baykaner: lets rename this method to something like PrepareBatch

libs/ml/include/ml/dataloaders/mnist_loaders/mnist_loader.hpp
```diff
@@ -44,14 +44,17 @@ class MNISTLoader : public DataLoader<LabelType, T>
     , cursor_(0)
   {
     std::uint32_t recordLength(0);
-    data_        = read_mnist_images(imagesFile, size_, recordLength);
-    labels_      = read_mnist_labels(labelsFile, size_);
-    figure_size_ = 28 * 28;
+    data_          = read_mnist_images(imagesFile, size_, recordLength);
```
baykaner: images_file and record_length

libs/ml/include/ml/dataloaders/mnist_loaders/mnist_loader.hpp
```diff
@@ -100,17 +103,26 @@ class MNISTLoader : public DataLoader<LabelType, T>
...
+      SizeType i{0};
+      auto     it = ret_images.at(0).Slice(index, 2).begin();
+      while (it.is_valid())
       {
-        ret_images.at(0)(j, i) = static_cast<typename T::Type>(data_[i][j]) / typename T::Type(256);
+        *it = static_cast<DataType>(data_[cursor_][i]) / DataType{256};
+        i++;
+        ++it;
       }
+      ret_labels(labels_[cursor_], index) = one;
```
baykaner: use static_cast<typename LabelType::Type>(1)
instead of preallocating one variable (compiler will make this optimisation for you anyway)

libs/ml/include/ml/layers/PRelu.hpp
```diff
@@ -47,7 +47,7 @@ class PRelu : public SubGraph<T>
     std::string alpha =
         this->template AddNode<fetch::ml::ops::Weights<ArrayType>>(name + "_Alpha", {});
 
-    ArrayType alpha_data(std::vector<std::uint64_t>({1, in}));
+    ArrayType alpha_data(std::vector<std::uint64_t>({in, 1}));
```
baykaner: should this be a SizeVector?

libs/ml/include/ml/ops/batchwise_flatten.hpp
```diff
@@ -0,0 +1,79 @@
...
+  BatchwiseFlattenOp()          = default;
+  virtual ~BatchwiseFlattenOp() = default;
+
+  virtual void Forward(VecTensorType const &inputs, ArrayType &output)
+  {
+    ASSERT(inputs.size() == 1);
+    ASSERT(output.shape() == ComputeOutputShape(inputs));
+    input_shape_ = inputs.front().get().shape();
+
+    BatchwiseFlatten(inputs.front().get(), output);
```
baykaner: move the flatten implementation out of the math library and into an anonymous namespace in this file

libs/ml/include/ml/ops/embeddings.hpp
```diff
@@ -53,44 +53,56 @@ class Embeddings : public fetch::ml::ops::Weights<T>
...
+
+    for (SizeType n{0}; n < batch_size; n++)
     {
-      auto tmp  = this->embeddings_output_->Slice(j);
-      auto tmp2 = this->output_->Slice(typename ArrayType::SizeType(i));
-      tmp.Assign(tmp2);
-      j++;
+      for (SizeType i{0}; i < inputs.front().get().shape().at(0); i++)
+      {
+        SizeType e = static_cast<SizeType>(inputs.front().get().At(i, n));
```
baykaner: why the change away from slices to using At() accessors? This is probably slightly slower?

libs/ml/include/ml/ops/embeddings.hpp
```diff
@@ -53,44 +53,56 @@ class Embeddings : public fetch::ml::ops::Weights<T>
...
-      updated_rows_.insert(typename ArrayType::SizeType(double(i)));
-      this->gradient_accumulation_->Slice(typename ArrayType::SizeType(double(i)))
-          .Assign(error_signal.Slice(j));
-      j++;
+      for (SizeType i{0}; i < inputs.front().get().shape().at(0); i++)
+      {
+        SizeType e = static_cast<SizeType>(inputs.front().get().At(i, n));
+        updated_rows_.insert(e);
+
+        for (SizeType j{0}; j < this->gradient_accumulation_->shape().at(1); j++)
```
baykaner: again why not slicing?

libs/ml/include/ml/ops/leaky_relu_op.hpp
```diff
@@ -58,41 +59,59 @@ class LeakyReluOp : public fetch::ml::ElementWiseOps<T>
...
-      if (*input1_it >= zero)
-      {
-        *rs1_it = one;
-        *rs2_it = zero;
-      }
-      else
+      a_size *= inputs.at(0).get().shape().at(i);
+    }
+    ArrayType return_signal2({a_size, 1});
+    return_signal2.Fill(zero);
```
baykaner: I believe tensor construction zeroes out anyway?

libs/ml/include/ml/ops/matrix_multiply.hpp
```diff
@@ -36,36 +37,215 @@ class MatrixMultiply : public fetch::ml::BatchOps<T>
...
-    ASSERT(inputs.size() == 2);
-    ASSERT(inputs.at(0).get().shape().size() == 2);
-    ASSERT(inputs.at(1).get().shape().size() == 2);
-    ASSERT(output.shape() == ComputeOutputShape(inputs));
+    fetch::math::Dot(inputs.at(0).get(), inputs.at(1).get(), output);
+  }
+  // Batchwise 3D @ 3D or broadcast matmul 2D @ 3D, 3D @ 2D
+  else
+  {
+    ASSERT((inputs.at(0).get().shape().size() == 3 || inputs.at(0).get().shape().size() == 2) &&
```
baykaner: lower case asserts are fine here

libs/ml/include/ml/ops/matrix_multiply.hpp
```diff
@@ -36,36 +37,215 @@ class MatrixMultiply : public fetch::ml::BatchOps<T>
...
+           (inputs.at(1).get().shape().size() == 3 || inputs.at(1).get().shape().size() == 2));
+
+    // Get batch size
+    SizeType batch_size;
+    if (inputs.at(0).get().shape().size() == 3)
```
baykaner, line 74: it looks like you assume (but don't check) that inputs.at(0).get().shape().at(2) == inputs.at(1).get().shape().at(2) (in the case of 3D by 3D matmul). worth an assert probably

libs/ml/include/ml/ops/matrix_multiply.hpp
```diff
@@ -36,36 +37,215 @@ class MatrixMultiply : public fetch::ml::BatchOps<T>
...
+    {
+      batch_size = inputs.at(1).get().shape().at(2);
+    }
 
-    fetch::math::Dot(inputs[0].get(), inputs[1].get(), output);
+    // Iterate over batch
+    for (SizeType i{0}; i < batch_size; i++)
+    {
+
+      auto      output_slice        = output.Slice(i, 2);
```
baykaner: you can write these two lines as:
ArrayType output_slice_tensor = output.Slice(i, 2);
because there is an operator= function in tensor.hpp that allows this

libs/ml/include/ml/ops/max_pool_2d.hpp
```diff
@@ -119,39 +123,43 @@ class MaxPool2D : public BatchOps<T>
...
               }
             }
-          }
 
-          // Add error to max node
-          return_signal.Set(c, max_iterw, max_iterh,
-                            return_signal.At(c, max_iterw, max_iterh) + *erit);
-          ++erit;
+            // Add error to max node
+            return_signal.Set(c, max_iterw, max_iterh, n_i,
```
baykaner: return_signal(c, max_iterw, max_iterh, n_i) = return_signal(c, max_iterw, max_iterh, n_i) + *erit;

libs/ml/include/ml/ops/reshape.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+#include "math/matrix_operations.hpp"
+#include "ml/ops/ops.hpp"
+
+namespace fetch {
+namespace ml {
+namespace ops {
+
+template <class T>
+class Reshape : public fetch::ml::BatchOps<T>
```
baykaner: unify batchOps and elementwiseOps as just Ops

libs/ml/include/ml/ops/transpose.hpp
```diff
@@ -32,28 +32,67 @@ class Transpose : public fetch::ml::BatchOps<T>
...
+      return {inputs.front().get().shape().at(1), inputs.front().get().shape().at(0)};
+    }
+    // Transpose with given vector
+    else
+    {
+      std::vector<SizeType> input_shape = inputs.front().get().shape();
+      std::vector<SizeType> shape;
+
+      shape.reserve(shape.size());
+      for (SizeType i{0}; i < transpose_vector_.size(); ++i)
```
baykaner: use iterator for transpose_vector?

libs/ml/tests/ml/ops/leaky_relu_op.cpp
```diff
@@ -35,21 +35,22 @@ TYPED_TEST_CASE(LeakyReluOpTest, MyTypes);
...
-  std::vector<double> data_input({1, -2, 3, -4, 5, -6, 7, -8});
-  std::vector<double> alphaInput({0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8});
-  std::vector<double> gt_input({1, -0.4, 3, -1.6, 5, -3.6, 7, -6.4});
-  for (std::uint64_t i(0); i < 8; ++i)
-  {
-    data.Set(i, DataType(data_input[i]));
-    gt.Set(i, DataType(gt_input[i]));
-    alpha.Set(i, DataType(alphaInput[i]));
-  }
+  ArrayType data = ArrayType::FromString(R"(
```
baykaner, line 40: no need for raw string literals btw

libs/ml/tests/ml/ops/max_pool_1d.cpp
```diff
@@ -68,20 +71,30 @@ TYPED_TEST(MaxPool1DTest, backward_test)
...
+  std::vector<double> gt_input1({0, 0, 2, 0, 7, 0, 0, 0, 5, 0});
+  std::vector<double> gt_input2({0, 0, 3, 0, 9, 0, 0, 0, 6, 0});
+
+  for (SizeType i_b{0}; i_b < 2; i_b++)
   {
-    data.Set(0, i, static_cast<DataType>(data_input[i]));
-    gt.Set(0, i, static_cast<DataType>(gt_input[i]));
+    for (SizeType i{0}; i < 10; ++i)
+    {
+      data.Set(0, i, i_b, static_cast<DataType>(data_input[i]) + static_cast<DataType>(i_b));
```
baykaner: prefer not to use Set
data(0, i, i_b) = ...

baykaner (changes requested):

libs/ml/include/ml/ops/embeddings.hpp
MissingNO57: Because you cannot Slice a Slice. I will try to come with something better.

libs/ml/tests/ml/ops/leaky_relu_op.cpp
MissingNO57: But without that I can't have FromString nicely on multiple lines.

MissingNO57: But without that I cannot have FromString on multiple lines.

libs/ml/include/ml/ops/matrix_multiply.hpp
MissingNO57: Can I avoid that?
      // Copy data to original array
      output_slice.Assign(output_slice_tensor);

MissingNO57: Are you sure that it will work?
```
: error: conversion from ‘fetch::math::Tensor<double>::TensorSlice’ to non-scalar type ‘fetch::ml::ops::MatrixMultiply<fetch::math::Tensor<double> >::ArrayType {aka fetch::math::Tensor<double>}’ requested
       ArrayType      output_slice_tensor        = output.Slice(i, 2);
```

libs/ml/include/ml/ops/embeddings.hpp
MissingNO57: We will discuss it on Monday.

libs/math/include/math/fundamental_operators.hpp
```diff
@@ -331,6 +329,16 @@ meta::IfIsMathArray<ArrayType, void> Add(ArrayType const &array1, ArrayType cons
       ++rit;
     }
   }
+  else
+  {
+    ArrayType a = array1.Copy();
```
baykaner, line 336: I guess we could probably get rid of the tensor copies if we can figure out a way to get Broadcast to work with const & inputs here?
Probably best to do it in another PR thought - just pointing it out

libs/math/include/math/fundamental_operators.hpp
```diff
@@ -400,37 +408,57 @@ template <typename ArrayType>
 meta::IfIsMathArray<ArrayType, void> Subtract(ArrayType const &array1, ArrayType const &array2,
                                               ArrayType &ret)
 {
-  assert(array1.shape() == array2.shape());
-  assert(array1.shape() == ret.shape());
```
baykaner, line 404: I guess we can still have the array1.shape == ret.shape assertion, but inside the if statement?

libs/ml/include/ml/clustering/tsne.hpp
```diff
@@ -173,16 +175,28 @@ class TSNE
...
+    if (input_matrix.shape().size() != 2)
+    {
+      fetch::ml::ops::Flatten<ArrayType> flatten_op;
+      ArrayType                          flat_input(flatten_op.ComputeOutputShape({input_matrix}));
+      flatten_op.Forward({input_matrix}, flat_input);
+      input_matrix_ = flat_input.Transpose();
+    }
+    else
+    {
+      input_matrix_ = input_matrix.Transpose();
```
baykaner, line 189: does this line work?
here you are calling .Transpose() on a tensor that cannot be 2D; but Tensor::Transpose() insists on a 2D tensor (otherwise you should be calling Tensor::Transpose(std::vector<SizeType> &))

libs/ml/include/ml/clustering/tsne.hpp
```diff
@@ -368,8 +382,8 @@ class TSNE
                                 fetch::math::DotTranspose(output_matrix, output_matrix));
 
     // num = 1 / (1 + (num+sum_y).T+sum_y)
-    ArrayType tmp_val((num + sum_y).Transpose());
-    num = fetch::math::Divide(DataType(1), fetch::math::Add(DataType(1), (tmp_val + sum_y)));
+    ArrayType val((num + sum_y).Transpose());
+    num = fetch::math::Divide(DataType(1), fetch::math::Add(DataType(1), (val + sum_y)));
```
baykaner: pedantic point but:
static_cast<DataType>(1) is preferable here

libs/ml/include/ml/clustering/tsne.hpp
```diff
@@ -415,33 +429,39 @@ class TSNE
...
 
         // /(1+||yi-yj||^2)
-        fetch::math::Multiply(num.At(i, j), tmp_val, tmp_val);
+        fetch::math::Multiply(num.At(i, j), val, val);
+
+        // val*(yi-yj), where val=(Pij-Qij)/(1+||yi-yj||^2)
+
+        ArrayType diff = (output_matrix.Slice(j).Copy()) - (output_matrix.Slice(i).Copy());
+
+        auto shape = diff.shape();
```
baykaner: very pedantic, but it would be nicer to put SizeVector shape = diff.shape() here, to make it clear to other developers what type of object this is.

libs/ml/include/ml/clustering/tsne.hpp
```diff
@@ -415,33 +429,39 @@ class TSNE
...
-        // tmp_val*(yi-yj), where tmp_val=(Pij-Qij)/(1+||yi-yj||^2)
-        tmp_slice += Multiply(tmp_val, (output_matrix.Slice(j).Copy().Squeeze()) -
-                                           (output_matrix.Slice(i).Copy().Squeeze()));
+        output_slice += Multiply(val, diff);
       }
 
       for (SizeType k = 0; k < output_matrix.shape().at(1); k++)
       {
-        ret.Set(i, k, fetch::math::Multiply(DataType(-1), tmp_slice.At(k)));
+        ret.Set(i, k, fetch::math::Multiply(DataType(-1), output_slice.At(k)));
```
baykaner: can we refactor to:
ret(i, k, fetch::math::Multiply(static_cast<DataType>(-1)) = output_slice(k);

libs/ml/include/ml/clustering/tsne.hpp
```diff
@@ -415,33 +429,39 @@ class TSNE
...
+        ArrayType diff = (output_matrix.Slice(j).Copy()) - (output_matrix.Slice(i).Copy());
+
+        auto shape = diff.shape();
+        shape.erase(shape.begin());
+        diff.Reshape(shape);
 
-        // tmp_val*(yi-yj), where tmp_val=(Pij-Qij)/(1+||yi-yj||^2)
-        tmp_slice += Multiply(tmp_val, (output_matrix.Slice(j).Copy().Squeeze()) -
-                                           (output_matrix.Slice(i).Copy().Squeeze()));
+        output_slice += Multiply(val, diff);
```
baykaner: I believe that output_slice is a Tensor, and not a TensorSlice. name is maybe slightly confusing

libs/ml/include/ml/dataloaders/dataloader.hpp
```diff
@@ -28,23 +29,79 @@ namespace dataloaders {
...
+  virtual ReturnType    PrepareBatch(fetch::math::SizeType subset_size);
+  virtual std::uint64_t Size() const   = 0;
+  virtual bool          IsDone() const = 0;
+  virtual void          Reset()        = 0;
 
 protected:
   bool random_mode_ = false;
 };
 
+template <typename LabelType, typename DataType>
```
baykaner, line 65: doxygen style function comment

libs/ml/include/ml/dataloaders/dataloader.hpp
```diff
@@ -28,23 +29,79 @@ namespace dataloaders {
...
+  std::pair<LabelType, std::vector<DataType>> training_pair = GetNext();
+  std::vector<SizeType>                       labels_size   = training_pair.first.shape();
+  std::vector<DataType>                       data;
+
+  // Prepare output data tensors
+  for (SizeType i{0}; i < training_pair.second.size(); i++)
+  {
+    std::vector<SizeType> current_data_shape             = training_pair.second.at(i).shape();
+    current_data_shape.at(current_data_shape.size() - 1) = subset_size;
+    data.push_back(DataType{current_data_shape});
```
baykaner: emplace back?

libs/ml/include/ml/dataloaders/dataloader.hpp
```diff
@@ -28,23 +29,79 @@ namespace dataloaders {
...
+template <typename LabelType, typename DataType>
+std::pair<LabelType, std::vector<DataType>> DataLoader<LabelType, DataType>::PrepareBatch(
+    fetch::math::SizeType subset_size)
+{
+  if (IsDone())
+  {
+    Reset();
+  }
+  std::pair<LabelType, std::vector<DataType>> training_pair = GetNext();
+  std::vector<SizeType>                       labels_size   = training_pair.first.shape();
```
baykaner: the optimiser will be calling PrepareBatch in the training loop. Therefore it is important to make this method efficient.
instantiating two new vectors at the start of every call is likely to be expensive - so would it be possible to make these member variables instead?

libs/ml/include/ml/dataloaders/dataloader.hpp
baykaner: actually, if you make data a member variable this will be an assignment operation instead.

libs/ml/include/ml/dataloaders/dataloader.hpp
```diff
@@ -28,23 +29,79 @@ namespace dataloaders {
...
+  std::vector<DataType>                       data;
+
+  // Prepare output data tensors
+  for (SizeType i{0}; i < training_pair.second.size(); i++)
+  {
+    std::vector<SizeType> current_data_shape             = training_pair.second.at(i).shape();
```
baykaner, line 83: this vector creation is probably an unnecessary expense in the training loop. Presumably all of this work could be front loaded at the time that the data is added?

libs/ml/include/ml/dataloaders/mnist_loaders/mnist_loader.hpp
```diff
@@ -99,27 +102,35 @@ class MNISTLoader : public DataLoader<LabelType, T>
...
       }
+      ret_labels(labels_[cursor_], index) = static_cast<typename LabelType::Type>(1);
+
+      ++cursor_;
     }
 
     return std::make_pair(ret_labels, ret_images);
   }
 
   std::pair<LabelType, std::vector<T>> ToArray()
```
baykaner: I suspect this ToArray method is not used anywhere. Can you please double check and ditch it if that's true.

libs/ml/include/ml/ops/add.hpp
```diff
@@ -24,33 +24,52 @@ namespace ml {
...
     assert(inputs.size() == 2);
-    assert(inputs.at(0).get().size() == inputs.at(1).get().size());
-    assert(error_signal.size() == inputs.at(1).get().size());
+    assert(inputs.at(0).get().shape().size() == inputs.at(1).get().shape().size());
+    assert(inputs.at(0).get().shape() == error_signal.shape());
+    assert(error_signal.shape() == ComputeOutputShape(inputs));
 
-    return {error_signal, error_signal};
+    // Test if input is broadcastable by batch dimension
+    assert(inputs.at(1).get().shape().at(inputs.at(1).get().shape().size() - 1) == 1);
```
baykaner, line 53: oh wow - this is hard to read. We will have to get rid of these reference wrappers in ml library soon to make this easier to understand

libs/ml/include/ml/ops/embeddings.hpp
```diff
@@ -35,10 +35,10 @@ class Embeddings : public fetch::ml::ops::Weights<T>
   using SizeType      = typename ArrayType::SizeType;
   using VecTensorType = typename Weights<T>::VecTensorType;
 
-  Embeddings(SizeType dataPoints, SizeType dimensions)
+  Embeddings(SizeType dimensions, SizeType dataPoints)
```
baykaner: style: data_points not dataPoints

libs/ml/include/ml/ops/embeddings.hpp
```diff
@@ -53,44 +53,72 @@ class Embeddings : public fetch::ml::ops::Weights<T>
...
+    for (SizeType i{0}; i < inputs.front().get().shape().at(0); i++)
     {
-      auto tmp  = this->embeddings_output_->Slice(j);
-      auto tmp2 = this->output_->Slice(typename ArrayType::SizeType(i));
-      tmp.Assign(tmp2);
-      j++;
+      for (SizeType n{0}; n < batch_size; n++)
+      {
+
+        indices1.at(0)       = i;
```
baykaner: private member variables should have a trailing _
indices1_.at(0)

libs/ml/include/ml/ops/embeddings.hpp
```diff
@@ -53,44 +53,72 @@ class Embeddings : public fetch::ml::ops::Weights<T>
...
+        indices1.at(1)       = n;
+        auto embedding_slice = this->embeddings_output_->View(indices1);
```
baykaner: are you using indices1 as a member variable for efficiency reasons?
I think you can also do this at about the same cost:
auto embedding_slice = this->embeddings_output_->View({i, n});
because {i, n} will be an initializer_list

libs/ml/include/ml/ops/leaky_relu_op.hpp
```diff
@@ -53,49 +54,68 @@ class LeakyReluOp : public fetch::ml::ElementWiseOps<T>
...
-      if (*input1_it >= zero)
-      {
-        *rs1_it = one;
-        *rs2_it = zero;
-      }
-      else
+      a_size *= inputs.at(0).get().shape().at(i);
+    }
+    ArrayType return_signal2({a_size, 1});
+    return_signal2.Fill(static_cast<DataType>(0));
```
baykaner: I believe a tensor is filled with 0's on construction anyway?

libs/ml/include/ml/ops/leaky_relu_op.hpp
```diff
@@ -53,49 +54,68 @@ class LeakyReluOp : public fetch::ml::ElementWiseOps<T>
...
+    // Test if batch dimension for alpha is 1
+    assert(inputs.at(1).get().shape().at(inputs.at(1).get().shape().size() - 1) == 1);
 
-    auto     rs1_it    = return_signal1.begin();
-    auto     rs2_it    = return_signal2.begin();
-    auto     input1_it = inputs.at(0).get().begin();
-    auto     input2_it = inputs.at(1).get().begin();
-    DataType zero{0};
-    DataType one{1};
+    ArrayType return_signal1{inputs.at(0).get().shape()};
```
baykaner: style:
return_signal_1
return_signal_2

libs/ml/include/ml/ops/matrix_multiply.hpp
baykaner: ah, my example was wrong because then tensor looks to construct on a slice.
The right way to do it is to have output_slice_ be a private member variable of  known size, constructed earlier, then just:
output_slice_tensor = output.Slice(i, 2);

As an example i just tested and this works:

  Tensor<TypeParam> x({10, 10});
  Tensor<TypeParam> z({10});
  z = (x.Slice(0));

libs/ml/include/ml/ops/matrix_multiply.hpp
```diff
@@ -18,54 +18,235 @@
...
+    }
+    else
+    {
+      batch_size = inputs.at(1).get().shape().at(2);
+    }
+
+    // Iterate over batch
+    for (SizeType i{0}; i < batch_size; i++)
+    {
+      ArrayType err_sig_slice = error_signal.Slice(i, 2).Copy();
```
baykaner: maybe we can avoid this Copy() also?

libs/ml/include/ml/ops/reshape.hpp
```diff
@@ -0,0 +1,85 @@
...
+namespace ml {
+namespace ops {
+
+template <class T>
+class Reshape : public fetch::ml::Ops<T>
+{
+public:
+  using ArrayType     = T;
+  using SizeType      = typename ArrayType::SizeType;
+  using ArrayPtrType  = std::shared_ptr<ArrayType>;
```
baykaner: ptr type not used anywhere

libs/ml/include/ml/ops/reshape.hpp
```diff
@@ -0,0 +1,85 @@
...
+  using VecTensorType = typename Ops<T>::VecTensorType;
+
+  Reshape(std::vector<SizeType> new_shape)
+    : new_shape_(new_shape)
+  {}
+  ~Reshape() = default;
+
+  void Forward(VecTensorType const &inputs, ArrayType &output)
+  {
+    ASSERT(inputs.size() == 1);
```
baykaner: lower case asserts preferred

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -122,29 +121,59 @@ typename T::Type Optimiser<T, C>::Run(std::vector<ArrayType> const &data, ArrayT
   DataType loss_sum{0};
 
   SizeType step{0};
-  while (step < n_data)
+
+  // Prepare output data tensors
+  std::vector<ArrayType> batch_data;
```
baykaner: can batch_data and batch_labels be made private member variables? Under this current implementation we create a new std::vector for each of these once per epoch. Not the most expensive, but could be cheaper too

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -157,7 +186,7 @@ typename T::Type Optimiser<T, C>::Run(std::vector<ArrayType> const &data, ArrayT
   UpdateLearningRate();
   epoch_++;
 
-  return loss_sum / static_cast<DataType>(step);
+  return loss_sum;
```
baykaner, line 201: removed loss normalisation? is this by mistake?

baykaner: or are the loss functions now responsible for the normalisation?

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -223,7 +250,7 @@ typename T::Type Optimiser<T, C>::Run(
...
```
baykaner, line 265: loss normalisation removed?

baykaner (changes requested):
really nice work.
Mostly style changes, a couple of places where we can be slightly more efficient.

libs/ml/include/ml/ops/embeddings.hpp
baykaner: actually i'm not sure about this. leave it as member variable but add the missing trailing _

libs/math/include/math/fundamental_operators.hpp
MissingNO57: Raised Issue  #1257

libs/ml/include/ml/ops/leaky_relu_op.hpp
MissingNO57: It is, but I think that we should not fill it in the future if it's not necessary.

libs/ml/include/ml/optimisation/optimiser.hpp
MissingNO57: Because you call it on a batch, Criterion returns average of batch now. It will be rewritten anyway In next PR, where criterions are ops and will return Tensor of errors instead.

libs/ml/include/ml/optimisation/optimiser.hpp
MissingNO57: Criterion which get batch will do averaging. It will be changed in next PR, because criterion will be op and will return Tensor of losses.

baykaner (dismissed):
LGTM

baykaner (approved):

PR #1160, opened 2019-06-09
Optimise headers

wkaluza:
This is a subset of what I extracted out of the IWYU tool which I was able to clean up without upsetting the codebase too much. Overall headers got moved into source files a lot, plenty of transitive includes are now included directly (especially std headers), and quite a few forward declarations were added where appropriate. I was doing cleanup as I went, so there are some misc changes mixed in here and there. Hopefully it's uncontroversial.

* HttpConnectionManager was header-only for no discernible reason. I moved the implementation to a source file.
* Checked-in headers for Google Test, Mock and Benchmark now use double quotes instead of angle brackets throughout


n-hutton (approved):
looks great

PR #1162, opened 2019-06-10
xplatform backtrace on segfault

n-hutton:
Catches the following signals:

```
SIGABRT,  // Abort signal from abort(3)
SIGBUS,   // Bus error (bad memory access)
SIGFPE,   // Floating point exception
SIGILL,   // Illegal Instruction
SIGIOT,   // IOT trap. A synonym for SIGABRT
SIGQUIT,  // Quit from keyboard
SIGSEGV,  // Invalid memory reference
SIGSYS,   // Bad argument to routine (SVr4)
SIGTRAP,  // Trace/breakpoint trap
SIGXCPU,  // CPU time limit exceeded (4.2BSD)
SIGXFSZ,  // File size limit exceeded (4.2BSD)
#if defined(BACKWARD_SYSTEM_DARWIN)
SIGEMT,  // emulation instruction executed
```

Note: not Ctrl-C (SIGINT) nor SIGKILL.

When it does, it looks like this (everything is using fetch logger):

```
 2019-06-10 15:50:50.555, # 7 INFO  :                          BasicMiner ] Starting block packing. Pool Size: 0
[ 2019-06-10 15:50:50.556, # 7 INFO  :                          BasicMiner ] Finished block packing (packed: 0 remaining: 0)
[ 2019-06-10 15:50:50.591, # 6 WARN  :                       DocumentStore ] Attempted to commit an already committed hash
[ 2019-06-10 15:50:50.592, # 7 ERROR :                         SIG_HANDLER ] Stack trace (most recent call last) in thread 123145477066752:
#6    Object "libsystem_pthread.dylib", at 0x7fff5d148248, in _pthread_start + 65
#5    Object "libsystem_pthread.dylib", at 0x7fff5d1452ea, in _pthread_body + 125
#4    Object "constellation", at 0x100d5436d, in void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void (fetch::core::Reactor::*)(), fetch::core::Reactor*> >(void*) + 61
#3    Object "constellation", at 0x100d53c7b, in fetch::core::Reactor::Monitor() + 987
#2    Object "constellation", at 0x100a89c7d, in fetch::core::StateMachine<fetch::ledger::BlockCoordinator::State>::Execute() + 333
#1    Object "constellation", at 0x100a884ee, in fetch::ledger::BlockCoordinator::OnTransmitBlock() + 30
#0    Object "constellation", at 0x100a90c14, in fetch::ledger::MainChain::AddBlock(fetch::ledger::Block const&) + 148

Stack trace for #7
  0: In thread #7: (root) , 0

Active locks:

Segmentation fault: 11
```

It is possible to install and link a tool that prints out the line numbers in a pretty way. I need to experiment a bit with that. This addition also opens the door for a nicer backtraces than we have (probably in a future PR)


libs/core/include/core/fetch_backward.hpp
```diff
@@ -0,0 +1,43 @@
...
+#pragma GCC diagnostic ignored "-Wconversion"
+#pragma GCC diagnostic ignored "-Wpedantic"
+#pragma GCC diagnostic ignored "-Wsign-compare"
+#endif
+
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wconversion"
+#pragma clang diagnostic ignored "-Wpedantic"
+#pragma GCC diagnostic ignored "-Wsign-compare"
```
wkaluza: GCC pragma in clang block

wkaluza (changes requested):

libs/core/include/core/fetch_backward.hpp
n-hutton: It passed CI though! :) Good spot

ejfitzgerald:
Can you give an example of how this integrates?

libs/core/include/core/fetch_backward.hpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    backward::Printer printer;
+    printer.address = true;
+
+    std::ostringstream oss;
+
+    printer.print(st, oss);
+
+    FETCH_LOG_ERROR("SIG_HANDLER", oss.str());
```
n-hutton: Basically the same functionality as the normal sig handler from backward-cpp, except using the fetch logger instead of cerr

libs/core/include/core/fetch_backward.hpp
```diff
@@ -17,6 +17,8 @@
 //
 //------------------------------------------------------------------------------
 
+#include "core/logger.hpp"
```
wkaluza, line 20: Why include this here?

.gitmodules
```diff
@@ -18,4 +18,4 @@
 	url = https://github.com/msgpack/msgpack-c.git
 [submodule "vendor/backward-cpp"]
 	path = vendor/backward-cpp
-	url = https://github.com/bombela/backward-cpp.git
+	url = https://github.com/n-hutton/backward-cpp.git
```
wkaluza, line 21: If we're forking a 3rd party lib, then it should probably live under the fetchai org.

wkaluza (changes requested):

.gitmodules
ejfitzgerald: This is fine for the moment

ejfitzgerald (approved):
Looks good to me

PR #1163, opened 2019-06-10
Defensive check for the subtree syncing

ejfitzgerald:
Added initial check that `roots_to_sync_` was not empty. This should not happen however, we have at least one case of this on the test net. Further investigation is needed.

### DIFF
--- a/libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
+++ b/libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
@@ -206,22 +206,27 @@ TransactionStoreSyncService::State TransactionStoreSyncService::OnQuerySubtree()
   assert(!roots_to_sync_.empty());
 
   FETCH_LOCK(mutex_);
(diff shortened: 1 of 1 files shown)


libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
```diff
@@ -206,22 +206,27 @@ TransactionStoreSyncService::State TransactionStoreSyncService::OnQuerySubtree()
   assert(!roots_to_sync_.empty());
 
   FETCH_LOCK(mutex_);
-  for (auto const &connection : muddle_->AsEndpoint().GetDirectlyConnectedPeers())
+
+  // sanity check that this is not the case
+  if (!roots_to_sync_.empty())
```
ejfitzgerald, line 211: This is the only addition check added

n-hutton (approved):
Might be worth putting in a ticket to look over this once more - as you say, after inspecting this it is not clear how this could ever happen

ejfitzgerald:
Done #1164

Issue #1164, opened 2019-06-10
Investigate test net cause of test net seg fault


ejfitzgerald:
The following fix PR was added to work around this issue #1163. From the code it is not clear why this would need to be added and points to an alternative underlying issue.

This issue covers the investigation of this issue and associated fixes

PR #1165, opened 2019-06-10
Defensive check for the subtree syncing (#1163)

ejfitzgerald:
Port of previous bugfix #1163 to `master`

### DIFF
--- a/libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
+++ b/libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
@@ -207,22 +207,27 @@ TransactionStoreSyncService::State TransactionStoreSyncService::OnQuerySubtree()
   assert(!roots_to_sync_.empty());
 
   FETCH_LOCK(mutex_);
-  for (auto const &connection : muddle_->AsEndpoint().GetDirectlyConnectedPeers())
+
+  // sanity check that this is not the case
+  if (!roots_to_sync_.empty())
   {
-    auto root = roots_to_sync_.front();
-    roots_to_sync_.pop();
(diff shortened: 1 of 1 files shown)


libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
```diff
@@ -207,22 +207,27 @@ TransactionStoreSyncService::State TransactionStoreSyncService::OnQuerySubtree()
...
-        transactions_prefix, root_size_));
+      auto promise = PromiseOfTxList(client_->CallSpecificAddress(
+          connection, RPC_TX_STORE_SYNC, TransactionStoreSyncProtocol::PULL_SUBTREE,
+          transactions_prefix, root_size_));
 
-    promise_id_to_roots_[promise.id()] = root;
-    pending_subtree_.Add(root, promise);
+      promise_id_to_roots_[promise.id()] = root;
+      pending_subtree_.Add(root, promise);
+    }
```
wkaluza, line 230: You could combine this with the following if statement into one block. Also, the assertion above is redundant now.

wkaluza (changes requested):

libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
ejfitzgerald: I think additional refactoring is out of the scope of this PR. We can add the updates to the context of #1164.

As for the assertion it is very much still required. As discussed, it is not expect behaviour for the `roots_to_sync_` to be empty when entering this function

wkaluza (approved):

PR #1166, opened 2019-06-10
Commodity prediction example

evsmithx:


libs/ml/examples/commodity_prediction/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+#include "math/tensor.hpp"
+#include "ml/dataloaders/tensor_dataloader.hpp"
+#include "ml/graph.hpp"
+#include "ml/layers/fully_connected.hpp"
+#include "ml/ops/activation.hpp"
+#include "ml/state_dict.hpp"
+
+#include <fstream>
+#include <iostream>
+#include <ml/ops/transpose.hpp>
```
baykaner: if you're going to separate your includes like this then transpose should be with the top group in ""

libs/ml/examples/commodity_prediction/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+    layer_type = LayerType::DENSE;
+    match_counter++;
+  }
+  if (layer_name.find("softmax") != std::string::npos)
+  {
+    layer_type = LayerType::SOFTMAX;
+    match_counter++;
+  }
+
+  assert(match_counter == 1);
```
baykaner: better to throw than to assert. In release mode the assertion will be ignored

libs/ml/examples/commodity_prediction/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+  return layer_type;
+}
+
+/**
+ * Loads a csv file into a Tensor
```
baykaner, line 77: presumably you're assuming a certain layout of the data (i.e. are rows == features or data?).
Should specify your format

libs/ml/examples/commodity_prediction/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+ * @param filename  name of the file
+ * @param cols_to_skip  number of columns to skip
+ * @param rows_to_skip  number of rows to skip
+ * @param transpose  whether to transpose the resulting Tensor
+ * @return  Tensor with data
+ */
+ArrayType read_csv(std::string const &filename, SizeType cols_to_skip = 0,
```
baykaner: a bit pedantic, but rows_to_skip and cols_to_skip should be const

libs/ml/examples/commodity_prediction/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+  if (transpose)
+  {
+    weights = weights.Transpose();
+  }
+  return weights;
+}
+
+/**
+ * Loads a single model architecture from a csv file and adds the specified nodes to the graph
+ * Example csv line: keras_h7_aluminium_px_last_us,num_input,118,dropout_0,output_dense,54,softmax
```
baykaner: maybe instead of keras_h7..... you can just put {model_name} here

libs/ml/examples/commodity_prediction/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+  for (SizeType j = 0; j < output_data_size; j++)
+  {
+    auto current_input = test_x.Slice(j).Copy().Transpose();
+    g_ptr->SetInput("num_input", current_input);
+    auto slice_output = g_ptr->Evaluate(node_names.back());
+    output.Slice(j).Assign(slice_output);
+  }
+
+  assert(output.AllClose(test_y, 0.00001f));
```
baykaner: instead of asserting just print a statement indicating success or failure (otherwise running this example in release mode does nothing interesting)

baykaner (changes requested):
a few minor changes needed

baykaner (approved):

PR #1167, opened 2019-06-10
Remove dead code

wkaluza:
The `log2_num_lanes_` field is unused and fails Linux builds locally on Clang 6. Surprising that Jenkins did not pick it up.

### DIFF
--- a/libs/miner/include/miner/transaction_layout_queue.hpp
+++ b/libs/miner/include/miner/transaction_layout_queue.hpp
@@ -74,13 +74,11 @@ public:
   TransactionLayoutQueue &operator=(TransactionLayoutQueue &&) = delete;
 
 private:
-  uint32_t       log2_num_lanes_;
   DigestSet      digests_;  ///< Set of digests stored within the list
   UnderlyingList list_;     ///< The list of transaction layouts
 };
 
-inline TransactionLayoutQueue::TransactionLayoutQueue(uint32_t log2_num_lanes)
-  : log2_num_lanes_{log2_num_lanes}
+inline TransactionLayoutQueue::TransactionLayoutQueue(uint32_t /* log2_num_lanes */)
 {}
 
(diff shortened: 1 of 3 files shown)


libs/miner/include/miner/transaction_layout_queue.hpp
```diff
@@ -74,15 +74,11 @@ class TransactionLayoutQueue
...
-  static bool RemapAndAdd(UnderlyingList &list, TransactionLayout const &tx, uint32_t num_lanes);
-
-  uint32_t       log2_num_lanes_;
   DigestSet      digests_;  ///< Set of digests stored within the list
   UnderlyingList list_;     ///< The list of transaction layouts
 };
 
-inline TransactionLayoutQueue::TransactionLayoutQueue(uint32_t log2_num_lanes)
-  : log2_num_lanes_{log2_num_lanes}
+inline TransactionLayoutQueue::TransactionLayoutQueue(uint32_t /* log2_num_lanes */)
```
ejfitzgerald: Please just remove this constructor and add a `= default` instead

ejfitzgerald (changes requested):

libs/miner/src/basic_miner.cpp
```diff
@@ -151,7 +151,7 @@ void BasicMiner::GenerateBlock(Block &block, std::size_t num_lanes, std::size_t
     std::vector<QueuePtr> transaction_lists(num_threads);
```
ejfitzgerald: Since the constructor is no longer needed here we can make this a `std::vector<Queue>` and remove `num_threads` extra allocations

ejfitzgerald (changes requested):

libs/miner/src/basic_miner.cpp
wkaluza: Nice.

ejfitzgerald (approved):

PR #1173, opened 2019-06-11
[VM] Fix for segfault when serialising null-ref `Ptr<Object>` objects

### LINKED ISSUE #1168: VM segfault
The following script causes `vm-lang` and `VmTestToolkit` to segfault. Identified by @dHonerkamp.

```
function main()
  var default_array = Array<Array<UInt64>>(2);
  var bids = State<Array<Array<UInt64>>>("blub", default_array);
endfunction
```

pbukva:
Resolves: #1168

### DIFF
--- a/libs/vm/include/vm/state.hpp
+++ b/libs/vm/include/vm/state.hpp
@@ -166,6 +166,13 @@ public:
     try
     {
       FlushIO();
+      // TODO (issue 1171): The the code bellow is is the correct code, however it
+      // is failing on various occasions (e.g. in our uTests on `Status::PERMISSION_DENIED`
+      // in `StateAdapter::Write(...)` method.).
+      // if (FlushIO() != Status::OK)
+      //{
+      //  vm_->RuntimeError("Failure of writing state to the storage.");
+      //}
     }
     catch (std::exception const &ex)
     {
@@ -216,13 +223,15 @@ private:
     return v;
   }
 
-  void FlushIO()
+  Status FlushIO()
   {
     // if we have an IO observer then inform it of the changes
     if (!vm_->HasError() && vm_->HasIoObserver())
     {
-      WriteHelper(name_, value_, vm_->GetIOObserver());
+      return WriteHelper(name_, value_, vm_->GetIOObserver());
     }
+
+    return Status::OK;
   }
 
   std::string name_;
--- a/libs/vm/include/vm/variant.hpp
+++ b/libs/vm/include/vm/variant.hpp
@@ -511,44 +511,61 @@ inline void Serialize(ByteArrayBuffer &buffer, Variant const &variant)
 
   switch (variant.type_id)
   {
+  case TypeIds::Unknown:
+  case TypeIds::Null:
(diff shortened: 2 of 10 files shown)


libs/vm/include/vm/state.hpp
```diff
@@ -216,13 +223,15 @@ class State : public IState
...
+  Status FlushIO()
   {
     // if we have an IO observer then inform it of the changes
     if (!vm_->HasError() && vm_->HasIoObserver())
     {
-      WriteHelper(name_, value_, vm_->GetIOObserver());
+      return WriteHelper(name_, value_, vm_->GetIOObserver());
     }
+
+    return Status::OK;
```
troels-im, line 234: Is the status OK if it was not able to write?

libs/vm/include/vm/variant.hpp
```diff
@@ -511,44 +511,61 @@ inline void Serialize(ByteArrayBuffer &buffer, Variant const &variant)
 
   switch (variant.type_id)
   {
+  case TypeIds::Unknown:
+  case TypeIds::Null:
+  case TypeIds::Void:
+    // No need to serialise the value (conceptually there is no value).
```
troels-im, line 517: Wouldn't the correct behaviour be to do `RuntimeError("Cannot serialize void/null/unknown")`?

libs/vm/include/vm/state.hpp
pbukva: Have been thinking this through a few times before, and it felt like this is the right implementation if we are going by return value only (= no exception).
Reason being that this (`State::OK`) is returned only under conditions when:
 * VM is already in error state (`vm_->HasError() == true`), in which case it does not make much sense to go ahead with reporting another nested error 
 * and/or when VM does not have registered IO observer, which is responsible for actual write to storage (`vm_->HasIoObserver() == false`), what was, by original design intention, considered as valid state (whoever is setting up VM can decide not to register IO observer) - see original usage of the observer in the `State::State(...)` constructor.

However, I have to admit that original design intention somehow never felt completely kosher exactly for the same reason which caught you eye.

If we want to change this, there are 2 ways to go:
 1) either throw exception here in the case when `vm_->HasIoObserver() == false`, 
 2) or check IO observer validity at the `State<...>` construction time and fail at that point. Keep in mind that VM has `VM::SetIOObserver(IoObserverInterface &observer)` method, which allows to change IO observer instance - it can not be invalidated (null-ed), just changed to another object (what is however highly improbable scenario in real life)

libs/vm/include/vm/variant.hpp
pbukva: Left it in for the scenarios when someone instantiates `Variant` instance without specifying type and value, or as explicit `Null` or `Void` types, where all these cases are harmless - we can serialise & deserialise it with no effort or impact on functionality.

I can take it out.

ejfitzgerald (commented):
Seems good from my side. Over to you @troelsfr

troels-im (approved):

PR #1175, opened 2019-06-11
Introducing UInt256 to the VM and SHA256

### LINKED ISSUE #1074: Support required for error checking, i.e. asserts, require, revert, Rust-type panic, etc.
> Warning: This is currently not a well thought out issue and needs discussion, clarification, and agreement.

`etch` needs support for error checking such as `assert`, `require`, `revert`, etc. in Solidity, and/or `panic` in Rust.

troels-im:
Both of these are used in many popular smart contracts and hence something that would be useful to have. Specially you can with the new version do things like:
```
function main()
  var x : UInt256 = UInt256(2u64);
  var y = UInt256("hello world");  
  var hasher = SHA256();
  hasher.update("string literal");
  hasher.update(x);
  hasher.update(y);
  var ret : UInt256 = hasher.final();
  panic("forgot to implement the rest of the function");
endfunction
```

Closes #1074


libs/crypto/include/crypto/stream_hasher.hpp
```diff
@@ -59,6 +61,11 @@ class StreamHasher
     return Update(reinterpret_cast<uint8_t const *>(vect.data()), vect.size() * sizeof(T));
   }
 
+  UInt256 Update(UInt256 const &val)
```
ejfitzgerald: This is not a great fit for this generic interface. While it is true that we only support SHA256 currently, this will change in the future.

libs/vm_modules/include/vm_modules/math/bignumber.hpp
```diff
@@ -17,17 +17,16 @@
...
 //------------------------------------------------------------------------------
 
-#include "math/bignumber.hpp"
-
+#include "vectorise/uint/uint.hpp"
+#include "vm/module.hpp"
 #include "vm_modules/core/byte_array_wrapper.hpp"
 
-#include "vm/module.hpp"
+// This is does not really say much about what Fetch.AI is and how it empowers the community
```
ejfitzgerald: ?

libs/vm_modules/src/core/panic.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "vm/vm.hpp"
+
+#include <ostream>
+
+namespace fetch {
+namespace vm_modules {
+
+void Panic(fetch::vm::VM *vm, fetch::vm::Ptr<fetch::vm::String> const &s)
+{
+  vm->RuntimeError(s->str);
```
ejfitzgerald, line 32: 👍 👍

libs/crypto/include/crypto/stream_hasher.hpp
troels-im: I am happy to make this a templated function of the number of bits in `UInt`, Note this is an overloaded `Update` function, not the `Final` function and it is neither `virtual`. Hence, it is not tied to SHA256 specifically and is just a convenience function.

libs/vm_modules/include/vm_modules/math/bignumber.hpp
troels-im: An accurate description of the file? Obviously not supposed to be there.

libs/vm_modules/src/core/panic.cpp
wkaluza: Closes #1074

libs/crypto/include/crypto/stream_hasher.hpp
troels-im: Updated in latest to any `UInt<S>`.

libs/vm_modules/src/core/panic.cpp
troels-im: @wkaluza thanks for providing issue number.

PR #1176, opened 2019-06-12
Add regression test for null serialisation segfault

wkaluza:
Tests for #1168 bug


libs/vm_modules/tests/unit/state_tests.cpp
```diff
@@ -195,6 +195,19 @@ TEST_F(StateTests, querying_state_constructed_from_null_string_fails_gracefully)
   ASSERT_FALSE(toolkit.Run(&output));
 }
 
+TEST_F(StateTests, serialising_compound_object_with_null_values_does_not_segfault)
```
pbukva, line 198: 👍
Good spot. I actually added test for this while working on segfault fix, but then after messing with merges back and forth it did't make it in to the segfault fix PR for some reason.

pbukva (approved):
LGTM

PR #1178, opened 2019-06-12
Fix testnet crash issues

ejfitzgerald:
* Fix race condition causing fees to not be settle all the time (ultimately causing block invalidation)
* Catch corner case segmentation fault on corrupted disk files
* Correct issue when syncing transaction roots
* Fix race in muddle packet routing (by removing the ability to compute the transmitter address - only needed for enhanced trust system modelling)


n-hutton:
Closed due to inactivity

PR #1179, opened 2019-06-13
DataLoader Cleanup

baykaner:
- removed tensor dataloader (nothing was using it)
- removed GetDataPair & GetRandom methods from dataloaders
- reworked dataloaders to all use GetNext (which checks internally the random_mode_ flag set on construction


evsmithx (commented):
I'm not very familiar with most of this code yet, but the changes make sense to me.

PR #1181, opened 2019-06-13
Fix bug which causes constellation to lock when recovering

n-hutton:
I observed that constellation can occasionally lock when trying to recover from a bad merkle tree db file - this fixes that bug


ejfitzgerald (approved):

ejfitzgerald:
LGTM - Can you also back port it to v0.3.x and v0.4.x

Issue #1172, opened 2019-06-11
[VM] Support for general serialisation concept of `Object` & `Variant` types


pbukva:
Current (ledger native) serialisation principle firmly requires types to be `default constructable`, reason being that in order for types to be deserializable they need to have default constructor (= non-parametric constructor).
For VM purposes, it is necessary to extend/change current serialisation concept and enable default-like construction of VM `Object` based types (what in extend also covers `Variant` wrapper type).

The the full extent of the issue manifests in nested scenarios (Object based types are nested inside of Object based types, ... ).

PR #1193, opened 2019-06-14
Bugfix/enable string serialisation

troels-im:
Enabling string serialisation and fixing issue #1172 .

### DIFF
--- a/libs/vm_modules/tests/unit/state_tests.cpp
+++ b/libs/vm_modules/tests/unit/state_tests.cpp
@@ -259,8 +259,6 @@ TEST_F(StateTests, test_serialisation_of_complex_type)
 
   ASSERT_TRUE(toolkit.Compile(deser_src));
 
-  // TODO (issue 1172): This expectation is just temporary, it shall be replaced
-  // with commented-out code bellow once the issue is resolved.
   try
   {
     auto const retval = toolkit.Run();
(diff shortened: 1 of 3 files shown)


libs/vm_modules/tests/unit/state_tests.cpp
```diff
@@ -259,18 +259,14 @@ TEST_F(StateTests, test_serialisation_of_complex_type)
 
   ASSERT_TRUE(toolkit.Compile(deser_src));
 
-  // TODO (issue 1172): This expectation is just temporary, it shall be replaced
-  // with commented-out code bellow once the issue is resolved.
   try
```
pbukva, line 262: All of this try-catch block & commented-out code bellow shall be replaced by code bellow (which however is failing in runtime - there is still an outstanding issue with deserialisation of Array of objects):

```suggestion
  Variant output;
  ASSERT_TRUE(toolkit.Run(&output));
  ASSERT_FALSE(output.IsPrimitive());

  auto retval{output.Get<Ptr<IArray>>()};
  ASSERT_TRUE(static_cast<bool>(retval));
  ASSERT_EQ(int32_t{3}, retval->Count());
  EXPECT_EQ(std::string{"aaa"}, retval->PopFrontOne().Get<Ptr<String>>()->str);
  EXPECT_EQ(std::string{"bbb"}, retval->PopFrontOne().Get<Ptr<String>>()->str);
  EXPECT_EQ(std::string{"ccc"}, retval->PopFrontOne().Get<Ptr<String>>()->str);
}
```

PR #1194, opened 2019-06-14
Bugfix/top level object alloc for serialisation

troels-im:
Fixing top level constructor in serialise to state function.


pbukva (dismissed):
LGTM

pbukva (approved):
LGTM

PR #1200, opened 2019-06-14
Vm break continue fix

troels-im:


libs/ml/examples/commodity_prediction/main.cpp
```diff
@@ -0,0 +1,373 @@
...
+ */
+ArrayType read_csv(std::string const &filename, SizeType const cols_to_skip = 0,
+                   SizeType rows_to_skip = 0, bool transpose = false)
+{
+  std::ifstream file(filename);
+  std::string   buf;
+
+  // find number of rows and columns in the file
+  std::string           delimiter = ",";
+  ulong                 pos;
```
wkaluza, line 95: This `ulong` caused some problems on Mac, I think. It's changed to `SizeType` on master.

PR #1206, opened 2019-06-16
Fix possible main chain segfault

n-hutton:
I observed that it was possible for main chain to segfault after block coord calls remove block. This invalidated the heaviest hash and when the BC calls main chain->getHeaviest next time it can return a nullptr

### DIFF
--- a/libs/ledger/src/chain/main_chain.cpp
+++ b/libs/ledger/src/chain/main_chain.cpp
@@ -194,7 +194,9 @@ bool MainChain::LoadBlock(BlockHash const &hash, Block &block) const
 MainChain::BlockPtr MainChain::GetHeaviestBlock() const
 {
   FETCH_LOCK(lock_);
-  return GetBlock(heaviest_.hash);
+  auto block_ptr = GetBlock(heaviest_.hash);
+  assert(block_ptr);
+  return block_ptr;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


libs/ledger/src/chain/main_chain.cpp
```diff
@@ -263,6 +265,11 @@ bool MainChain::RemoveBlock(BlockHash hash)
 {
   FETCH_LOCK(lock_);
 
+  // Step 0. Manually set heaviest to a block we still know is valid
+  auto block_before_one_to_del = GetBlock(hash);
+  heaviest_.weight             = block_before_one_to_del->body.weight;
+  heaviest_.hash               = block_before_one_to_del->body.hash;
```
wkaluza: I think this needs
```
heaviest_.Update(*block_before_one_to_del);
```

n-hutton: I believe this will only update if the proposed block is heavier - since it is not in this case I have done it manually. Perhaps I should have reset the heaviest tip then called update would be cleaner

wkaluza (approved):

ejfitzgerald (approved):
Seems reasonable, certainly we have seen this failure case on the test net I am sure. I added another addition to the block coordinator to handle this case but this is a nicer solution

PR #1210, opened 2019-06-17
[VM]: Adding unit test for complex nested Array<...> type

pbukva:
* In addition fixing bug in the `FakeIoObserver::Read(...)` method

### DIFF
--- a/libs/vm_modules/tests/unit/state_tests.cpp
+++ b/libs/vm_modules/tests/unit/state_tests.cpp
@@ -42,46 +42,30 @@ TEST_F(StateTests, SanityCheck)
   ASSERT_TRUE(toolkit.Run());
 }
 
-TEST_F(StateTests, AddressSerializeTest)
+TEST_F(StateTests, AddressSerialisationTest)
 {
-  static char const *TEXT = R"(
+  static char const *ser_src = R"(
     function main()
-      var data = Address("MnrRHdvCkdZodEwM855vemS5V3p2hiWmcSQ8JEzD4ZjPdsYtB");
-      var state = State<Address>("addr");
-      state.set(data);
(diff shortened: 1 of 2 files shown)


libs/vm_modules/tests/unit/state_tests.cpp
```diff
@@ -252,6 +236,103 @@ TEST_F(StateTests, test_serialisation_of_complex_type)
   EXPECT_EQ(std::string{"ccc"}, retval->PopFrontOne().Get<Ptr<String>>()->str);
 }
 
+template <typename T>
+std::enable_if_t<!IsPtr<T>::value> ArrayFromVariant(Variant const &array, int32_t expected_size,
```
wkaluza, line 240: I don't think this overload would compile if it was ever instantiated - gtest ASSERT* macros contain a return statement which yields IIRC a testing::AssertionResult, but SFINAE tries to make this a void function. May be worth writing a test that exercises a nested array of non-pointers.

wkaluza (changes requested):

libs/vm_modules/tests/unit/state_tests.cpp
pbukva: `ASSERT_XYZ(...)` macros can be used in any function/method with return type `void` . The proof  that it would compile just fine is the 2nd specialisation of the `ArrayFromVariant<...>` funtion for `Ptr<IArray>` type right bellow (it also returns `void`), or multiple other test-helper methods/functions where ASSERT* marcore are used, like for example `SmartContractTests::CreateContract(...)`.

wkaluza (approved):

ejfitzgerald (approved):

PR #1214, opened 2019-06-18
abs fix + optimiser refactor

baykaner:
- bugfixed vm_modules abs example
- fixed math abs for unsinged ints (new math SFINAE)
- optimiser refactor (const& -> std::move())


markos (approved):
LGTM

PR #1222, opened 2019-06-19
ML commodity example with training mode

evsmithx:
Added an optimiser to the ml example so the the network can be trained as well as tested.


libs/ml/examples/commodity_prediction/main.cpp
```diff
@@ -303,70 +314,99 @@ int main(int argc, char **argv)
...
+    // Initialise Optimiser
+    OptimiserType optimiser(g_ptr, {node_names.front()}, node_names.back(), learning_rate);
+
+    // Training loop
+    DataType loss{0};
+    for (SizeType j{0}; j < epochs; j++)
+    {
+      loss = optimiser.Run(loader, batch_size, subset_size);
+      std::cout << "Loss: " << loss << std::endl;
+    }
```
baykaner, line 543: we still want to make some predictions after training, so we should load train data set and test data set and see how close the result are on prediction. As we know, they will not be identical to the previously trained models due to dropout randomisation, but hopefully they will be close. Therefore we will also need to measure the distance between our predictions and the loaded test data set predictions; since predictions are vectors you may wish to use cosine distance method in math library.

Also, the training schedule requires several bouts of training and testing - so we should load multiple train and test files (specify directory, train and validate in a loop)

libs/ml/include/ml/dataloaders/commodity_dataloader.hpp
```diff
@@ -0,0 +1,241 @@
...
+  }
+}
+
+/**
+ * Constructor for CommodityDataLoader
+ * @tparam LabelType Type for the labels, usually tensor<float>
+ * @tparam InputType Type for the data, usually tensor<float>
+ * @param random_mode whether to access the data randomly
+ */
+template <typename LabelType, typename InputType>
```
baykaner: when the implementation is just an empty brace you can just do that inside the class

baykaner (changes requested):
the implementation looks fine but there's a bit more work to do to match the original training schedule.

libs/ml/include/ml/ops/activations/dropout.hpp
```diff
@@ -65,6 +65,7 @@ class Dropout : public fetch::ml::ElementWiseOps<T>
       UpdateRandomValues();
 
       fetch::math::Multiply(inputs.front().get(), drop_values_, output);
+      fetch::math::Multiply(output, DataType{1.0} / probability_, output);
```
baykaner: this will work, but it's inefficient.
UpdateRandomValues will iterate over drop_values (size of output tensor). Multiply will iterate over input & output tensor once. Multiply is called twice, so we now have 3x iterations over a tensor of size output on every call to dropout during training.

I think we can do a lot better.

libs/ml/include/ml/ops/activations/dropout.hpp
```diff
@@ -65,6 +65,7 @@ class Dropout : public fetch::ml::ElementWiseOps<T>
...
     }
   }
```
baykaner, line 86: also - do we need to change the backpropagation? since dropout now scales at train time, is the gradient any different?

baykaner (approved):
LGTM

PR #1228, opened 2019-06-20
Make etch Array of Booleans printable

wkaluza:
I managed to add this functionality and enable a unit test I wrote speculatively some time ago. Thanks to @robertdickson for helping out with the gnarliest bits.


robertdickson (approved):
LGTM

PR #1231, opened 2019-06-21
Miscellaneous changes

wkaluza:


apps/constellation/constellation.cpp
```diff
@@ -97,15 +97,12 @@ std::shared_ptr<ledger::DAGInterface> GenerateDAG(bool generate, std::string con
                                                   bool                          load_on_start,
                                                   Constellation::CertificatePtr certificate)
 {
-  std::shared_ptr<ledger::DAGInterface> ret;
-
   if (generate)
   {
-    ledger::DAG *new_dag = new ledger::DAG{db_name, load_on_start, certificate};
-    ret.reset(new_dag);
+    return std::make_shared<ledger::DAG>(db_name, load_on_start, certificate);
```
n-hutton, line 109: 👍

libs/core/include/core/vector.hpp
```diff
@@ -1,55 +0,0 @@
-#pragma once
```
n-hutton: @baykaner

n-hutton: Is this unused?

n-hutton (changes requested):

libs/math/include/math/clustering/k_means.hpp
```diff
@@ -180,8 +180,8 @@ class KMeansImplementation
     }
 
     // initialise size of euclidean distance container
-    k_euclids_      = fetch::core::Vector<ArrayType>(n_clusters_);
-    empty_clusters_ = fetch::core::Vector<SizeType>(n_clusters_);
+    k_euclids_      = std::vector<ArrayType>(n_clusters_);
```
ejfitzgerald: There was a purposeful reason for using the `fetch::Vector` specifically the updated `operator[]` behaviour. What prompted this change?

ejfitzgerald (changes requested):

libs/core/include/core/vector.hpp
wkaluza: It's only used by k_means.hpp, which is itself unused. Khan asked me not to delete k_means, as it's in need of a redesign, but a lot of it can be reused, so I left it alone.

libs/math/include/math/clustering/k_means.hpp
wkaluza: It's dead code - it's only used by k_means.hpp, which is not used anywhere. I'm not deleting k_means.hpp itself at Khan's request.

n-hutton (dismissed):
lgtm

n-hutton (approved):

ejfitzgerald (approved):

PR #1232, opened 2019-06-21
[VM] Add erase member function to Array

### LINKED ISSUE #1230: Add array .erase(element) function
There are many cases where `erase(x)` would save a lot of time. For example, managing a list of orders or actions: when one is done, many loops are required in order to "pack and stack" so that a hole does not exist in the array. Far easier would be able erase an element by index and have that action performed automatically.

wkaluza:
Closes #1230


robertdickson (approved):
LGTM

PR #1236, opened 2019-06-22
Feature/genesisfile creator

n-hutton:


scripts/end_to_end_test/run_end_to_end_test.py
```diff
@@ -175,6 +175,22 @@ def start_node(self, index):
 
         time.sleep(0.5)
 
+    def restart_node(self, index):
+        print('Restarting Node {}...'.format(index))
+
+        self._nodes[index].stop()
+
+        # Optically remove db files when testing recovering from a genesis file
+        if False:
```
ejfitzgerald, line 184: Make an option?

libs/ledger/src/genesis_file_creator.cpp
```diff
@@ -0,0 +1,145 @@
...
+  ledger::Block genesis_block;
+
+  genesis_block.body.merkle_hash  = merkle_commit_hash;
+  genesis_block.body.block_number = 0;
+  genesis_block.body.miner        = ledger::Address(crypto::Hash<crypto::SHA256>(""));
+  genesis_block.UpdateDigest();
+
+  FETCH_LOG_INFO(LOGGING_NAME, "Created genesis block hash: ", genesis_block.body.hash.ToBase64());
+
+  ledger::GENESIS_MERKLE_ROOT = merkle_commit_hash;
```
ejfitzgerald, line 139: This feels potentially raced but fine for a v1

ejfitzgerald (approved):

scripts/end_to_end_test/run_end_to_end_test.py
n-hutton: it's going to be useful to keep that there for a short while as I will have to add an e2e test for the custom genesis block etc.

libs/ledger/src/genesis_file_creator.cpp
n-hutton: You're right, there is one potential race in main chain rpc service here:    ``` if (it->body.previous_hash == GENESIS_DIGEST)```

Addressed in another PR. I am now fairly confident there are no other races at this time

PR #1240, opened 2019-06-24
vm_modules standardising bindings

baykaner:
- bugfixing/re-enabling vm_modules abs
- normlising vm_modules bindings


markos (approved):
Agree with the changes, some attention might be needed with the FixedPoint test, perhaps merging the fixedpoint etch file used in the fixedpoint example into the test.

PR #1241, opened 2019-06-24
moved math tests into all_tests where they will be enabled

baykaner:


MissingNO57 (approved):
LGTM, just wait for Jenkins.

PR #1250, opened 2019-06-25
Upgrade Logging Subsystem

ejfitzgerald:
Logging subsystem now using `spdlog`. In addition HTTP module added for controlling runtime log configuration.


libs/core/include/core/logger.hpp
```diff
@@ -731,67 +667,3 @@ extern log::details::LogWrapper logger;
 #define LOG_PRINT_STACK_TRACE(name, custom_name)
 
 #endif
-
```
n-hutton, line 734: :)

libs/core/src/logging.cpp
```diff
@@ -0,0 +1,199 @@
...
+    break;
+  }
+
+  return new_level;
+}
+
+LogRegistry::LogRegistry()
+{
+  spdlog::set_level(
+      spdlog::level::trace);  // this should be kept in sync with the compilation level
```
n-hutton, line 126: presumably debug mode compiles all levels and release some subset?

scripts/log_tool.py
```diff
@@ -0,0 +1,75 @@
...
+
+def parse_commandline():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('logger', nargs='?', help='The target logger')
+    parser.add_argument('level', nargs='?', type=_level,
+                        help='The level to be set')
+    parser.add_argument('-H', '--host', default='127.0.0.1',
+                        help='The target host')
+    parser.add_argument('-P', '--port', type=int,
+                        default=8000, help='The tareget port')
```
n-hutton: spelling: target

scripts/log_tool.py
```diff
@@ -0,0 +1,75 @@
...
+        print('Unable to determine level')
+        sys.exit(1)
+
+    return candidates[0]
+
+
+def parse_commandline():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('logger', nargs='?', help='The target logger')
+    parser.add_argument('level', nargs='?', type=_level,
```
n-hutton, line 29: I hadn't seen this functionality before - type is a closure?

n-hutton (dismissed):
Looks great - minor bug bug you can fix in another PR

scripts/log_tool.py
ejfitzgerald: Yep I quite like this trick too. It basically works because in python calling `int(value)` as a constructor is the same syntax as `functor(value)`

scripts/log_tool.py
ejfitzgerald: Yep whoops will fix later

libs/core/src/logging.cpp
ejfitzgerald: Current the same behaviour is kept. i.e. info messages and "above" are compiled into the code for all modes.

Stage 2. Will be to alter this so that the default debug mode is configuration but will default to debug messages I think. There is a trade off here between slow down in debug mode and usefulness to be able to switch any mode on.

This particular line is the global runtime log filter level. When I change the compile time configuration I will make sure this is matched here. This sort of gives the user a runtime API to query what the compiled log level is. (that was the thinking anyway)

PR #1251, opened 2019-06-25
Feature/pos additions

n-hutton:


libs/ledger/src/chaincode/token_contract.cpp
```diff
@@ -62,8 +62,9 @@ bool IsOperationValid(WalletRecord const &record, Transaction const &tx,
   return true;
 }
 
-constexpr uint64_t MAX_TOKENS           = 0xFFFFFFFFFFFFFFFFull;
-constexpr uint64_t STAKE_WARM_UP_PERIOD = 5;
+constexpr uint64_t MAX_TOKENS             = 0xFFFFFFFFFFFFFFFFull;
+constexpr uint64_t STAKE_WARM_UP_PERIOD   = 5;
```
ejfitzgerald: We should actually adjust these. I made them small for testing but I think that we should have them as something in the order of at least 120 for the time being.

libs/ledger/src/chaincode/token_contract.cpp
```diff
@@ -288,6 +292,70 @@ Contract::Status TokenContract::AddStake(Transaction const &tx, BlockIndex block
...
+    if (Extract(data, AMOUNT_NAME, amount))
+    {
+      // look up the state record (to see if there is a deed associated with this address)
+      WalletRecord record{};
+      if (GetStateRecord(record, tx.from().display()))
+      {
+        // ensure the transaction has authority over this deed
+        if (IsOperationValid(record, tx, STAKE_NAME))
+        {
+          FETCH_LOG_WARN(LOGGING_NAME, "Destaking! : ", amount, " of ", record.stake);
```
ejfitzgerald: Debug message

libs/ledger/include/ledger/chaincode/wallet_record.hpp
```diff
@@ -153,16 +158,30 @@ struct WalletRecord
     return false;
   }
 
+  void CollectStake(uint64_t block_index)
+  {
+    // Point to stake equal to or greater than block index
+    auto stop_point = cooldown_stake.lower_bound(block_index);
```
ejfitzgerald, line 164: Feel like there is a bug here. I think you want `upper_bound()` and then decrement the iterator.

```c++
  std::set<int> data = {0, 1, 5, 10, 15, 20};
  auto it = data.lower_bound(12);
  std::cout << *it << std::endl;
```
Will give you:
```
15
```

ejfitzgerald (changes requested):

libs/ledger/src/chaincode/token_contract.cpp
n-hutton: done

libs/ledger/src/chaincode/token_contract.cpp
n-hutton: done

libs/ledger/include/ledger/chaincode/wallet_record.hpp
n-hutton: Don't think it is a bug - in your example you would want to remove 0, 1, 5 and 10 from the cooldown pool (12 would be block number). The current logic is that you iterate from begin until lower_bound removing these

ejfitzgerald (approved):

PR #1253, opened 2019-06-26
BLS Signature and signature verification for the library

troels-im:
This implements basic BLS identity with signer and verifier. It introduces two new dependencies, namely `bls` and `mcl`. Furthermore it contains the start of a refactor of the crypto library to ensure that we will be signature agnostic for transactions. A ticket will be opened for this as it is not within the scope of this PR.


.gitmodules
```diff
@@ -19,3 +19,9 @@
 [submodule "vendor/backward-cpp"]
 	path = vendor/backward-cpp
 	url = https://github.com/n-hutton/backward-cpp.git
+[submodule "vendor/mcl"]
```
n-hutton, line 25: we usually use https here

cmake/BuildTargets.cmake
```diff
@@ -220,6 +220,40 @@ function (configure_vendor_targets)
   # Google Test
   add_subdirectory(${FETCH_ROOT_VENDOR_DIR}/googletest)
 
+  # MCL
+  set(USE_GMP OFF CACHE BOOL "use gmp" FORCE)
+  set(USE_OPENSSL OFF CACHE BOOL "use openssl" FORCE)
+  # TODO: Work out how to get this to work with the already found version of OpenSSL
```
n-hutton, line 230: you may already have tried this but can you not link against vendor-openssl?

ejfitzgerald:
Also @troelsfr looks like there is still the dependency on `libgmp`. That is fine if we want to go ahead to use it, but we will need to update the CI docker image

libs/crypto/include/crypto/bls.hpp
```diff
@@ -0,0 +1,165 @@
...
+    bls::Signature s;
+    private_key_.sign(s, m.c_str());
+    std::stringstream signature;
+    signature << s;
+    return static_cast<ConstByteArray>(signature.str());
+  }
+
+  Identity identity() const final
+  {
+    // TODO: Fix this
```
n-hutton, line 141: create an issue for this?

PR #1263, opened 2019-06-26
Initial Telemetry Support

ejfitzgerald:
Initial OpenTSDB compatible metrics generated from `/api/telemetry` endpoint.

Example output:

```
# HELP ledger_log_messages The number of log messages printed
# TYPE ledger_log_messages counter
ledger_log_messages 521
# HELP ledger_miner_mining_pool_size 
# TYPE ledger_miner_mining_pool_size gauge
ledger_miner_mining_pool_size 434
# HELP ledger_miner_max_mining_pool_size 
# TYPE ledger_miner_max_mining_pool_size gauge
ledger_miner_max_mining_pool_size 434
# HELP ledger_miner_max_pending_pool_size 
# TYPE ledger_miner_max_pending_pool_size gauge
ledger_miner_max_pending_pool_size 434
# HELP ledger_miner_duplicate_count 
# TYPE ledger_miner_duplicate_count counter
ledger_miner_duplicate_count 0
# HELP ledger_miner_duplicate_filtered_count 
# TYPE ledger_miner_duplicate_filtered_count counter
ledger_miner_duplicate_filtered_count 0
# HELP ledger_block_coordinator_reload_state_count 
# TYPE ledger_block_coordinator_reload_state_count counter
ledger_block_coordinator_reload_state_count 1
# HELP ledger_block_coordinator_synchronising_state_count 
# TYPE ledger_block_coordinator_synchronising_state_count counter
ledger_block_coordinator_synchronising_state_count 28
# HELP ledger_block_coordinator_synchronised_state_count 
# TYPE ledger_block_coordinator_synchronised_state_count counter
ledger_block_coordinator_synchronised_state_count 2542
# HELP ledger_block_coordinator_pre_valid_state_count 
# TYPE ledger_block_coordinator_pre_valid_state_count counter
ledger_block_coordinator_pre_valid_state_count 1
```


libs/telemetry/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+# Test Targets
+# ------------------------------------------------------------------------------
+
+add_test_target()
+
+# ------------------------------------------------------------------------------
+# Benchmark Targets
+# ------------------------------------------------------------------------------
+
+# add_benchmark_target()
```
n-hutton: dead on lines 18 & 31

libs/telemetry/src/metric.cpp
```diff
@@ -0,0 +1,25 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "telemetry/metric.hpp"
```
n-hutton: not intentional is it?

libs/telemetry/tests/unit/gauge_tests.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+  this->gauge_->set(2);
+  this->gauge_->add(2);
+  EXPECT_EQ(this->gauge_->get(), 4);
+}
+
+TYPED_TEST(IntegerGaugeTests, CheckRemove)
+{
+  this->gauge_->set(4);
+  this->gauge_->remove(2);
```
n-hutton: I would have thought a gague would only allow updating the value, which would override old values  - interested in why it's this way

libs/telemetry/tests/unit/registry_tests.cpp
```diff
@@ -0,0 +1,54 @@
...
+    registry_.reset();
+  }
+
+  RegistryPtr registry_;
+};
+
+TEST_F(RegistryTests, SimpleCheck)
+{
+  auto counter = registry_->CreateCounter("foo_bar_baz", "Description");
+  counter->add(200);
```
n-hutton: don't think this actually tests anything

libs/telemetry/src/metric.cpp
ejfitzgerald: Well it is a metric, but I have renamed it. I think I will probably pull the previous metric library at some point. Think that the project has outgrown it.

ejfitzgerald: #1288

libs/telemetry/tests/unit/gauge_tests.cpp
ejfitzgerald: Mostly connivence methods

libs/telemetry/tests/unit/registry_tests.cpp
ejfitzgerald: Nope removed!

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -109,6 +110,38 @@ BlockCoordinator::BlockCoordinator(MainChain &chain, DAGPtr dag, StakeManagerPtr
   , exec_wait_periodic_{EXEC_NOTIFY_INTERVAL}
   , syncing_periodic_{NOTIFY_INTERVAL}
   , synergetic_exec_mgr_{CreateSynergeticExecutor(features, dag, storage_unit_)}
+  , reload_state_count_{telemetry::Registry::Instance().CreateCounter(
```
engineerkatie, line 113: Counter names should probably be dot-delimited; that's normal for tools like Graphite.

ejfitzgerald: Looks like that is not the case in prometheus. I want to get these changes in so that we can start experimenting with the client QL.

libs/telemetry/src/measurement.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+      // after the first element seperators should always be added
+      add_sep = true;
+    }
+
+    stream << '}';
+  }
+
+  // add the value spacer
+  stream << ' ';
```
pbukva, line 59: I'm wondering if it would not be better to use the `=` here instead.

libs/telemetry/include/telemetry/gauge.hpp
```diff
@@ -0,0 +1,217 @@
...
+  /// @{
+  void ToStream(std::ostream &stream) const override;
+  /// @}
+
+  // Operators
+  Gauge &operator=(Gauge const &) = delete;
+  Gauge &operator=(Gauge &&) = delete;
+
+private:
+  using Mutex     = std::mutex;
```
pbukva, line 67: Based on set of operations/functionality of current `Gauge` class, it feels like it can be lock-less (using `std::atomic`). Requirements for type to be usable with `std::atomic<T>` are:
    - std::is_trivially_copyable<T>::value,
    - std::is_copy_constructible<T>::value,
    - std::is_move_constructible<T>::value,
    - std::is_copy_assignable<T>::value,
    - std::is_move_assignable<T>::value,
what should be perfectly fine for current requirement (`std::is_arithmetic<ValueType>`) defined by static assert bellow.

libs/telemetry/src/registry.cpp
```diff
@@ -0,0 +1,102 @@
...
+ * @return The reference to the registry
+ */
+Registry &Registry::Instance()
+{
+  static Registry instance;
+  return instance;
+}
+
+/**
+ * Ensure the name consists of lowercase letters and underscores
```
pbukva, line 37: Digits not allowed?

libs/telemetry/include/telemetry/measurement.hpp
```diff
@@ -0,0 +1,73 @@
...
+#include <string>
+#include <unordered_map>
+
+namespace fetch {
+namespace telemetry {
+
+class Measurement
+{
+public:
+  using Labels = std::unordered_map<std::string, std::string>;
```
pbukva, line 30: Unordered map will make it complex (for client) for search for specific set of labels in the serialized stream using for example simple regexp, unless client does full JSON parsing.

pbukva (approved):
LGTM

PR #1264, opened 2019-06-26
Staking weight incorporation

n-hutton:


ejfitzgerald (commented):
Looks good to me is it worth creating that issue for moving around the weight value or are you planning on doing that next?

ejfitzgerald (approved):

PR #1268, opened 2019-06-27
Sgns fix

XiaodongWuFetch:
1. Fixed some issues in Skip gram negative sampling (Sgns) data loading process to make it align with the paper.
2. Incorporate the legacy w2v_dataloader with the SGNS graph implementation.


libs/ml/examples/word2vec_SGNS/main.cpp
```diff
@@ -99,18 +102,73 @@ std::string Model(fetch::ml::Graph<ArrayType> &g, SizeType embeddings_size, Size
   return ret_name;
 }
 
+void NormVector(ArrayType &vector)
```
baykaner: this should be a generalised method rather than specific to this example.
There exists a math function for calculating the L2Norm of a tensor already, but not for normalising a tensor; so please move your implementation to the math library and invoke it.
see math/ml/loss_functions/l2_norm.hpp and math/kernels/L2Loss.hpp
(i'll walk you through this process)

libs/ml/examples/word2vec_SGNS/main.cpp
```diff
@@ -99,18 +102,73 @@ std::string Model(fetch::ml::Graph<ArrayType> &g, SizeType embeddings_size, Size
...
+{
+  ArrayType arr = embeddings;
+
+  SizeType word1_idx = dl.VocabLookup(word1);
+  SizeType word2_idx = dl.VocabLookup(word2);
+  SizeType word3_idx = dl.VocabLookup(word3);
+
+  if (word1_idx > dl.VocabSize() || word2_idx > dl.VocabSize() || word3_idx > dl.VocabSize())
+  {
+    std::cout << "WARNING! not all to-be-tested words are in vocabulary" << std::endl;
```
baykaner: decide whether it might be better to throw here?

libs/ml/examples/word2vec_SGNS/main.cpp
```diff
@@ -122,8 +180,31 @@ void PrintKNN(SkipGramLoader<ArrayType> const &dl, ArrayType const &embeddings,
   }
 }
 
+void PrintEmbedding(Graph<ArrayType> const &g, std::string const &skip_gram_name,
```
baykaner: its not clear why you want to do this. is it for debugging purposes?

libs/ml/examples/word2vec_SGNS_new_dataloader/main.cpp
```diff
@@ -0,0 +1,286 @@
+//------------------------------------------------------------------------------
```
baykaner: if this will be the new word2vec_SGNS example we should delete the old one.

libs/ml/examples/word2vec_SGNS_new_dataloader/main.cpp
```diff
@@ -0,0 +1,286 @@
...
+#include "ml/graph.hpp"
+#include "ml/layers/skip_gram.hpp"
+#include "ml/ops/loss_functions/cross_entropy.hpp"
+#include "ml/optimisation/adam_optimiser.hpp"
+#include "ml/optimisation/sgd_optimiser.hpp"
+
+#include <iostream>
+#include <map>
+#include <memory>
+#include <set>
```
baykaner: it doesn't look like you actually use all of these includes.

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
+#pragma once
```
baykaner, line 1: if this is going to be the dataloader for SGNS you should delete the previous dataloaders that are now unused.
Also is it possible to unify this with the loader used by the w2v_simple example

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+  , current_sentence_(0)
+  , current_word_(0)
+  , window_size_(window_size)
+  , negative_samples_(negative_samples)
+  , mode_(mode)
+  , target_({window_size_ * 2, 1})
+  , label_({negative_samples_, 1})
+{
+  // setup temporary buffers for training purpose
+  label_one     = LabelType({1});
```
baykaner: should be private member variables and should have a trailing _
Also, initialise in the initialiser list above. just set the values here

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+  tmp_output    = DataType({1});
+}
+
+/**
+ * reports the total size of the outputs iterating through the dataloader
+ * @tparam T
+ * @return
+ */
+template <typename T>
+math::SizeType W2VLoader<T>::Size() const
```
baykaner: size methods should not be this expensive. Probably this value should be computed when the data is initially parsed, and then this method should just return the private member variable that stores the value;

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+{
+  SizeType size(0);
+  for (auto const &s : data_)
+  {
+    if ((SizeType)s.size() > (2 * window_size_))
```
baykaner, line 180: please avoid this C-style cast. in this case use:
static_cast<SizeType>(s.size())

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+
+/**
+ * checks if we've passed through all the data and need to reset
+ * @tparam T
+ * @return
+ */
+template <typename T>
+bool W2VLoader<T>::IsDone() const
+{
+  if (current_sentence_ == data_.size())
```
baykaner: since the vast majority of calls to IsDone will be false; maybe it would be marginally faster if you write it this way:

if (current_sentence_ < (data_.size() - 1))
{
  return false;
}
else if (current_sentence_ == data_.size() - 1)
{
  ...
}
else
{
  return true;
}

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+
+/**
+ * Remove words that appears less than MIN times. operation is destructive
+ * @tparam T
+ * @param min
+ */
+template <typename T>
+void W2VLoader<T>::RemoveInfrequent(SizeType min)
+{
+  W2VLoader new_loader(window_size_, negative_samples_, mode_);
```
baykaner: i believe you intend to rewrite this so that it doesn't create a secondary data loader

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+ */
+template <typename T>
+void W2VLoader<T>::InitUnigramTable()
+{
+  std::vector<SizeType> frequencies(vocab_size());
+  for (auto const &kvp : vocab_.data)
+  {
+    frequencies[kvp.second.first] = kvp.second.second;
+  }
+  unigram_table_.Reset(1e8, frequencies);
```
baykaner: 1e8 should probably be a static const expr or a class member variable, and not hardcoded

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+}
+
+template <typename T>
```
baykaner, line 317: doxygen style function comments please

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+  if (current_word_ < window_size_)
+  {
+    current_word_ = window_size_;
+  }
+
+  // select random window size
+  SizeType dynamic_size = rng_() % window_size_ + 1;
+
+  // for the interested one word
+  tmp_input(0) = T(data_[current_sentence_][current_word_]);
```
baykaner: use .at() accessors for bounds checking

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+
+  // set the context samples
+  for (SizeType i = 0; i < dynamic_size; ++i)
+  {
+    tmp_output(0) = T(data_[current_sentence_][current_word_ - i - 1]);
+    buffer.push_back(ReturnType(label_one, {tmp_input, tmp_output}));
```
baykaner: emplace back will be faster i think

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+
+    tmp_output(0) = T(data_[current_sentence_][current_word_ + i + 1]);
+    buffer.push_back(ReturnType(label_one, {tmp_input, tmp_output}));
```
baykaner: tmp variable name not ideal. current_input seems logical?

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+  }
+
+  // negative sampling for every context word
+  SizeType neg_sample;
+  for (SizeType i = 1; i < negative_samples_ * window_size_ * 2; ++i)
+  {
+    bool success = unigram_table_.SampleNegative(SizeType(tmp_input(0)), neg_sample);
```
baykaner: static_cast<SizeType>(tmp_input(0))

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+    if (success)
```
baykaner: you can wrap this up as:

if (!(unigram_table_.SampleNegative........))
{
throw
}
else
{
// do something useful
}

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+    {
+      tmp_output(0) = T(neg_sample);
+      buffer.push_back(ReturnType(label_zero, {tmp_input, tmp_output}));
```
baykaner: emplace

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+    frequencies[kvp.second.first] = kvp.second.second;
+  }
+  unigram_table_.Reset(1e8, frequencies);
+}
+
+template <typename T>
+void W2VLoader<T>::BufferNextSamples()
+{
+  // clear the buffer
+  buffer.clear();
```
baykaner: the problem with clearing the vector and then pushing back is that its a lot of memory allocation which is taking place inside the training loop. This will be slower than keeping a fixed array and reserving positions in that array dynamically.

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+ * Adds a dataset to the dataloader
+ * @tparam T
+ * @param s input string containing all the text
+ * @return bool indicates success
+ */
+template <typename T>
+bool W2VLoader<T>::BuildVocab(std::string const &s)
+{
+  std::vector<SizeType> indexes = StringsToIndices(PreprocessString(s));
+  if (indexes.size() >=
```
baykaner: typo - indices

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+/**
+ * converts string to indices and inserts into vocab as necessary
+ * @tparam T
+ * @param strings
+ * @return
+ */
+template <typename T>
+std::vector<math::SizeType> W2VLoader<T>::StringsToIndices(
+    std::vector<std::string> const
+        &strings)  // it is more like words to indices (each string is a word not a sentence)
```
baykaner: leftover comment - rename function if you think appropriate

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+{
+  std::vector<SizeType> indexes;
+  if (strings.size() >= 2 * window_size_ + 1)  // Don't bother processing too short inputs
+  {
+    indexes.reserve(strings.size());
+    for (std::string const &s : strings)
+    {
+      auto value = vocab_.data.insert(
+          std::make_pair(s, std::make_pair((SizeType)(vocab_.data.size() + 1),
```
baykaner: be consistent with reserved index for unknown word

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -0,0 +1,445 @@
...
+                                           0)));  // this line allows index is 0 right???
+      indexes.push_back((*value.first).second.first);
+      value.first->second.second++;
```
baykaner: pair of pairs is confusing. consider using a different data structure

libs/ml/include/ml/dataloaders/word2vec_loaders/w2v_dataloader.hpp
```diff
@@ -211,22 +211,29 @@ void W2VLoader<T>::InitUnigramTable()
...
+    ret.second.at(0).Set(i, 0, T(data_[current_sentence_][current_word_ - i - 1]));
+    ret.second.at(0).Set(i + dynamic_size, 0, T(data_[current_sentence_][current_word_ + i + 1]));
   }
 
-  // pad the unused part of the window
+  // denote the unused part of the window
   for (SizeType i = (dynamic_size * 2); i < ret.second.at(0).size(); ++i)
   {
-    ret.second.at(0)(i, 0) = -1;
+    ret.second.at(0)(i, 0) = -1;  // set the max of SizeType as a tag for un-updated positions
```
baykaner: you should do as your comment suggests with an enum for example

libs/ml/include/ml/graph.hpp
```diff
@@ -319,7 +319,7 @@ template <class OperationType>
 meta::IfIsTrainable<ArrayType, OperationType, void> Graph<ArrayType>::AddTrainable(
     std::string const &name, std::shared_ptr<Node<ArrayType, OperationType>> op)
 {
-  FETCH_LOG_INFO("ML_LIB", "Created trainable node [", name, "]");
+  //  FETCH_LOG_INFO("ML_LIB", "Created trainable node [", name, "]");
```
baykaner: just delete these

libs/ml/include/ml/layers/skip_gram.hpp
```diff
@@ -72,13 +72,18 @@ class SkipGram : public SubGraph<T>
     std::string in_ctx_matmul = this->template AddNode<fetch::ml::ops::MatrixMultiply<ArrayType>>(
         name + "_In_Ctx_MatMul", {transpose_ctx, embed_in_});
 
-    // dense layer
-    std::string fc_out = this->template AddNode<fetch::ml::layers::FullyConnected<ArrayType>>(
-        name + "_Dense", {in_ctx_matmul}, in_size, out);
-
+    //    // dense layer
+    //    std::string fc_out = this->template AddNode<fetch::ml::layers::FullyConnected<ArrayType>>(
```
baykaner: delete commented out code

libs/ml/include/ml/node.hpp
```diff
@@ -114,7 +114,7 @@ std::vector<std::reference_wrapper<const T>> Node<T, O>::GatherInputs() const
 template <typename T, class O>
 T &Node<T, O>::Evaluate()
 {
-  FETCH_LOG_INFO("ML_LIB", "Evaluating node [", name_, "]");
+  //  FETCH_LOG_INFO("ML_LIB", "Evaluating node [", name_, "]");
```
baykaner: delete

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -189,7 +191,7 @@ typename T::Type Optimiser<T, C>::Run(std::vector<ArrayType> const &data, ArrayT
     // Compute and apply gradient
     ApplyGradients(batch_size);
 
-    FETCH_LOG_INFO("ML_LIB", "Batch loss: ", loss);
+    // FETCH_LOG_INFO("ML_LIB", "Batch loss: ", loss);
```
baykaner: delete

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -229,8 +231,36 @@ typename T::Type Optimiser<T, C>::Run(
...
   SizeType                                     step{0};
   std::pair<ArrayType, std::vector<ArrayType>> input;
+
+  high_resolution_clock::time_point cur_time;
+  high_resolution_clock::time_point prev_time = high_resolution_clock::now();
+  SizeType                          prevStep{0};
+
   while (!loader.IsDone() && step < subset_size)
   {
+    // print the training stats every batch
```
baykaner: you should set a flag in optimiser that defaults to not printing these statistics

baykaner (changes requested):

libs/ml/examples/word2vec_SGNS/main.cpp
XiaodongWuFetch: This was created to test if the NN is actually learning
And it turned out to be. I am not sure if I shout remove this function?

libs/ml/examples/word2vec_SGNS/main.cpp
XiaodongWuFetch: done

libs/ml/examples/word2vec_SGNS_new_dataloader/main.cpp
XiaodongWuFetch: yeah done

XiaodongWuFetch:
I have resolved all the reviews, but couldnt figure out why the loss is increasing. Since there are other problems such as no linear weight updating avialable, I may as well leave this till next week. For the part I have implemented, I dont think there are any trivial bugs or logical issues.

libs/ml/examples/word2vec_SGNS_new_dataloader/main.cpp
```diff
@@ -26,17 +26,12 @@
 #include "ml/graph.hpp"
```
baykaner: please rename this example directory to word2vec_SGNS_dataloader

libs/ml/examples/word2vec_SGNS_new_dataloader/main.cpp
```diff
@@ -86,23 +67,23 @@ void PrintWordAnology(W2VLoader<DataType> const &dl, ArrayType const &embeddings
   SizeType word2_idx = dl.IndexFromWord(word2);
   SizeType word3_idx = dl.IndexFromWord(word3);
 
-  if (word1_idx == 0 || word2_idx == 0 || word3_idx == 0)
+  if (word1_idx == fetch::math::numeric_max<SizeType>() || word2_idx == fetch::math::numeric_max<SizeType>() || word3_idx == fetch::math::numeric_max<SizeType>())
```
baykaner: it would be nice if at the top of this example file you put something like:

static constexpr SizeType UNKNOWN_WORD = fetch::math::numeric_max<SizeType>();
then everywhere in the code you check against numeric max just put UNKNOWN_WORD and it's easier to understand what's going on

libs/ml/examples/word2vec_SGNS_new_dataloader/main.cpp
```diff
@@ -193,22 +150,22 @@ std::string ReadFile(std::string const &path)
...
+  SizeType negative_sample_size = 5;    // number of negative sample per word-context pair
   SizeType window_size          = 8;     // window size for context sampling
   bool     train_mode           = true;  // reserve for future compatibility with CBOW
-
+  DataType freq_thresh     = 1e-5;     // frequency threshold for subsampling
   SizeType min_count = 5;  // infrequent word removal threshold
 
-  SizeType    output_size     = 1;
-  SizeType    batch_size      = 128;      // training data batch size
+  SizeType    batch_size      = 100000;      // training data batch size
```
baykaner: this seems extremely large....

libs/ml/examples/word2vec_SGNS_new_dataloader/main.cpp
```diff
@@ -235,12 +192,14 @@ int main(int argc, char **argv)
 
   std::cout << "Setting up training data...: " << std::endl;
 
-  W2VLoader<DataType> data_loader(tp.window_size, tp.negative_sample_size, tp.train_mode);
+  W2VLoader<DataType> data_loader(tp.window_size, tp.negative_sample_size, tp.freq_thresh, tp.train_mode);
```
baykaner: your new dataloader has the same name as an existing dataloader - this must be changed

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -18,16 +18,14 @@
 //------------------------------------------------------------------------------
 
 #include "core/random/lcg.hpp"
+#include "core/random/lfg.hpp"
```
baykaner, line 20: do you need both lcg AND lfg?
also, theres a little bit of helper code in core/random.hpp - if you include that instead, you can just invoke the Random.generator  whenever you want to use it

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -37,21 +35,21 @@ template <typename T>
 class W2VLoader : public DataLoader<fetch::math::Tensor<T>, fetch::math::Tensor<T>>
```
baykaner: this is slightly confusing:
w2v_dataloader.hpp has the class W2VLoader
you've implemented a class with the same name W2VLoader in the file sgns_w2v_dataloader.
If you're not going to unify these two dataloaders, please give them different names

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -37,21 +35,21 @@ template <typename T>
...
 {
 public:
+    const T BufferPositionUnused = -1;
```
baykaner: is this an implicit cast? T is probably not integral right?
maybe cleaner to use either:
const T BufferPositionUnused = static_cast<T>(-1); or
const T BufferPositionUnused(-1);

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -128,16 +123,24 @@ W2VLoader<T>::W2VLoader(SizeType window_size, SizeType negative_samples, bool mo
...
 {
-  SizeType size(0);
-  for (auto const &s : data_)
-  {
-    if ((SizeType)s.size() > (2 * window_size_))
+    return size_;
+}
+
+/**
+ * update stats of data in dataloader, including: size_,
```
baykaner, line 171: rename this method UpdateSize (since thats all it does and Update is not very informative)

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -183,89 +181,147 @@ void W2VLoader<T>::Reset()
...
+    auto old2new = vocab_.Compactify();
+
+    // create reverse vocab
+    auto reverse_vocab = vocab_.GetReverseVocab();
+
+    // create a new data_ for storing text
+    std::vector<std::vector<SizeType>> new_data;
+
+    for(auto sent_it = data_.begin(); sent_it != data_.end(); sent_it++){
+        std::vector<SizeType> new_sent_buffer; // buffer for this sentence
```
baykaner: instantiating a new vector for each iteration of the loop is not very efficient. In practice you're only calling remove infrequent once though so it's not very important

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -183,89 +181,147 @@ void W2VLoader<T>::Reset()
...
+    buffer_pos_ = 0;
 
   // the current word should start from position that allows full context window
   if (current_word_ < window_size_)
   {
     current_word_ = window_size_;
   }
 
+  // subsample too frequent word
+  fetch::random::LaggedFibonacciGenerator<> lfg_(seed);
```
baykaner: this looks like you're instantiating lfg in the training loop. Shouldn't this be a private member variable, and you just invoke it?

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -183,89 +181,147 @@ void W2VLoader<T>::Reset()
...
+  while(true){
+      T word_freq = static_cast<T>(vocab_.reverse_data[data_.at(current_sentence_).at(current_word_)].second) / static_cast<T>(vocab_.total_count);
+      T random_var = static_cast<T>(lfg_.AsDouble()); // random variable between 0-1
+      if(random_var < 1 - fetch::math::Sqrt(freq_thresh_ / word_freq)){ // subsample for a cumulative prob of 1 - sqrt(thresh/freq) // N.B. if word_freq < freq_thresh, then subsampling would not happen
```
baykaner: are we 100% sure this should happen during training, and not be a data pre-processing step?

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -183,89 +181,147 @@ void W2VLoader<T>::Reset()
...
 
   // the current word should start from position that allows full context window
   if (current_word_ < window_size_)
   {
     current_word_ = window_size_;
   }
 
+  // subsample too frequent word
+  fetch::random::LaggedFibonacciGenerator<> lfg_(seed);
+  while(true){
```
baykaner: break only if IsDone. does this mean the BufferNextSamples buffers an entire epoch?

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -183,89 +181,147 @@ void W2VLoader<T>::Reset()
...
-    else
-    {
-      throw std::runtime_error(
-          "unigram table timed out looking for a negative sample. check window size for sentence "
-          "length and that data loaded correctly.");
+    if (!unigram_table_.SampleNegative(static_cast<SizeType>(cur_word_id), neg_sample)){
+        throw std::runtime_error(
+                "unigram table timed out looking for a negative sample. check window size for sentence "
+                "length and that data loaded correctly.");
+    }else{
```
baykaner: run apply_style

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -303,19 +374,23 @@ typename W2VLoader<T>::ReturnType W2VLoader<T>::GetNext()
...
-      2 * window_size_ + 1)  // each sentence stored in the data_ are guaranteed to have minimum
-                             // length to handle window_size context sampling
-  {
-    data_.push_back(std::move(indexes));
-    return true;
-  }
-  return false;
+    std::vector<std::string> preprocessed_string = PreprocessString(s);
+
+    if(preprocessed_string.size() <= 2*window_size_){ // dispose short sentences before we create vocabulary and count frequency: if we are not gonna train on it, the sentence does not exist
```
baykaner: this checks the size of one string? shouldn't you be checking the size of every sentence?

libs/ml/include/ml/dataloaders/word2vec_loaders/vocab.hpp
```diff
@@ -37,6 +48,65 @@ class Vocab
...
+ */
+void Vocab::Update() {
+    // update total count
+    total_count = 0;
+    for(auto const &w : data){
+        total_count += w.second.second;
+    }
+
+    // update reverse vocab
+    reverse_data = GetReverseVocab();
```
baykaner: private variables must have a trailing underscore

libs/ml/include/ml/dataloaders/word2vec_loaders/vocab.hpp
```diff
@@ -37,6 +48,65 @@ class Vocab
...
+/**
+ * reserve numerical max of SizeType for unknown word
+ *
+ */
+Vocab::Vocab(){}
+
+/**
+ * Get the total count of the vocabulary
+ */
+void Vocab::Update() {
```
baykaner: this is too generic a function name. it does two (apparently unrelated) things; maybe just make this one called UpdateCount, and separately just update the reverse_data_ outside of this function

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -16,89 +16,83 @@
 //   limitations under the License.
 //
 //------------------------------------------------------------------------------
-
 #include "math/base_types.hpp"
```
baykaner: optimiser.hpp has been given wrong style indentations - make it impossible to see quickly what changes have been made

baykaner (changes requested):

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -0,0 +1,284 @@
...
+#include <string>
+#include <vector>
+
+using namespace fetch::ml;
+using namespace fetch::ml::dataloaders;
+using namespace fetch::ml::ops;
+using namespace fetch::ml::layers;
+using DataType     = double;
+using ArrayType    = fetch::math::Tensor<DataType>;
+using ArrayPtrType = std::shared_ptr<ArrayType>;
```
baykaner: you dont use ArrayPtrType so remove it

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -0,0 +1,284 @@
...
+  return ret_name;
+}
+
+void PrintWordAnology(GraphW2VLoader<DataType> const &dl, ArrayType const &embeddings,
+                      std::string const &word1, std::string const &word2, std::string const &word3,
+                      SizeType k)
+{
+  ArrayType arr = embeddings;
+
+  if (!dl.WordKnown(word1) || !dl.WordKnown(word1) || !dl.WordKnown(word1))
```
baykaner: testing word1 all three times

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -0,0 +1,284 @@
...
+  return std::string((std::istreambuf_iterator<char>(t)), std::istreambuf_iterator<char>());
+}
+
+////////////////////////////////
+/// PARAMETERS AND CONSTANTS ///
+////////////////////////////////
+
+struct TrainingParams
+{
+  SizeType max_word_count       = 100000000000000;  // maximum number to be trained
```
baykaner: i suspect we might run out of memory on our laptops before this word count was ever triggered. maybe set it to something more reasonable like 100M?

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -0,0 +1,284 @@
...
+  SizeType negative_sample_size = 5;       // number of negative sample per word-context pair
+  SizeType window_size          = 8;       // window size for context sampling
+  bool     train_mode           = true;    // reserve for future compatibility with CBOW
+  DataType freq_thresh          = 1e-5;       // frequency threshold for subsampling
+  SizeType min_count            = 5;       // infrequent word removal threshold
+
+  SizeType batch_size      = 10000;  // training data batch size
+  SizeType embedding_size  = 32;     // dimension of embedding vec
```
baykaner: usually we set the embedding size to something like 200-300 ish

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -0,0 +1,284 @@
...
+  ///////////////////////////////////////
+
+  std::cout << "Setting up training data...: " << std::endl;
+
+  GraphW2VLoader<DataType> data_loader(tp.window_size, tp.negative_sample_size, tp.freq_thresh,
+                                       tp.max_word_count, tp.train_mode);
+  // set up dataloader
+  /// DATA LOADING ///
+  std::cout << "building vocab " << std::endl;
+  data_loader.BuildVocab({ReadFile(train_file)});
```
baykaner: I think you should invoke RemoveInfrequent and InitUnigramTable from within BuildVocab.

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -0,0 +1,284 @@
...
+  std::cout << "data_loader.EstimatedSampleNumber(): " << data_loader.EstimatedSampleNumber()
+            << std::endl;
+
+  ////////////////////////////////
+  /// SETUP MODEL ARCHITECTURE ///
+  ////////////////////////////////
+
+  // set up model architecture
+  std::cout << "building model architecture...: " << std::endl;
+  std::shared_ptr<fetch::ml::Graph<ArrayType>> g(std::make_shared<fetch::ml::Graph<ArrayType>>());
```
baykaner: neater if you rewrite this line to:
auto g = std::make_shared ......

libs/ml/include/ml/dataloaders/dataloader.hpp
```diff
@@ -122,7 +123,8 @@ std::pair<LabelType, std::vector<DataType>> DataLoader<LabelType, DataType>::Pre
     {
       if (IsDone())
       {
-        Reset();
+        break;  // we stop the prepare batch when the dataloader is drained, this might modify the
```
baykaner: I'm still not 100% sure about this. lets discuss it tomorrow in the office

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -154,31 +154,31 @@ std::string ReadFile(std::string const &path)
...
-  SizeType max_word_count       = 100000000000000;  // maximum number to be trained
-  SizeType negative_sample_size = 5;       // number of negative sample per word-context pair
-  SizeType window_size          = 8;       // window size for context sampling
-  bool     train_mode           = true;    // reserve for future compatibility with CBOW
-  DataType freq_thresh          = 1e-5;       // frequency threshold for subsampling
-  SizeType min_count            = 5;       // infrequent word removal threshold
-
-  SizeType batch_size      = 10000;  // training data batch size
-  SizeType embedding_size  = 32;     // dimension of embedding vec
+  SizeType max_word_count       = 1000000;  // maximum number to be trained
```
baykaner: using maximum of 1M words; so you don't train on the entire text8 dataset. This is deliberate?

libs/ml/examples/word2vec_simple/w2v_model.hpp
```diff
@@ -215,6 +215,134 @@ void W2VModel<ArrayType>::Train(SizeType iter, SizeType print_frequency, bool cb
   std::cout << "Done Training" << std::endl;
 }
 
+template <typename ArrayType>
+void W2VModel<ArrayType>::SGNSTrain(
+    ArrayType const &target,
+    ArrayType const &context)  // this function is simply a slower implementation of CBOW!!!!! it
```
baykaner: has this been fixed or not? if so remove leftover comment

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -117,7 +118,9 @@ GraphW2VLoader<T>::GraphW2VLoader(SizeType window_size, SizeType negative_sample
   // setup temporary buffers for training purpose
   input_words_  = DataType({negative_samples * window_size_ * 2 + window_size_ * 2});
   output_words_ = DataType({negative_samples * window_size_ * 2 + window_size_ * 2});
-  labels_       = DataType({negative_samples * window_size_ * 2 + window_size_ * 2 +
+  output_words_buffer_ =
+      fetch::math::Tensor<SizeType>({negative_samples * window_size_ * 2 + window_size_ * 2});
+  labels_ = DataType({negative_samples * window_size_ * 2 + window_size_ * 2 +
```
baykaner, line 123: can you explain what's going on with this + 1 please?

libs/ml/include/ml/dataloaders/word2vec_loaders/unigram_table.hpp
```diff
@@ -118,6 +119,41 @@ bool UnigramTable::SampleNegative(SizeType positive_index, SizeType &ret)
...
+ * @return
+ */
+bool UnigramTable::SampleNegative(fetch::math::Tensor<SizeType> positiveIndices, SizeType &ret)
+{
+  ret = data_[rng_() % data_.size()];
+
+  SizeType attempt_count = 0;
+  while (true)
+  {
+    bool inPositiveIndices = false;
```
baykaner: style, should be:
bool in_positive_indices = false;

libs/ml/include/ml/dataloaders/word2vec_loaders/unigram_table.hpp
```diff
@@ -118,6 +119,41 @@ bool UnigramTable::SampleNegative(SizeType positive_index, SizeType &ret)
   return true;
 }
 
+/**
+ *
+ * @param positive_index
+ * @param ret
+ * @return
+ */
+bool UnigramTable::SampleNegative(fetch::math::Tensor<SizeType> positiveIndices, SizeType &ret)
```
baykaner: style and const correctness; I think this should be
fetch::math::Tensor<SizeType> const & positive_indices

libs/ml/include/ml/optimisation/adam_optimiser.hpp
```diff
@@ -90,6 +96,29 @@ AdamOptimiser<T, C>::AdamOptimiser(std::shared_ptr<Graph<T>>
   ResetCache();
 }
 
+template <class T, class C>
+AdamOptimiser<T, C>::AdamOptimiser(
+    std::shared_ptr<Graph<T>> graph, std::vector<std::string> const &input_node_names,
```
baykaner, line 108: instead of std::vector<std::string> const & input_node_names,
can we use std::move in this context for efficiency?
same for output_node_name string

libs/ml/include/ml/optimisation/adam_optimiser.hpp
```diff
@@ -90,6 +96,29 @@ AdamOptimiser<T, C>::AdamOptimiser(std::shared_ptr<Graph<T>>
...
+    fetch::ml::optimisers::LearningRateParam<DataType> const &learning_rate_param,
+    DataType const &beta1, DataType const &beta2, DataType const &epsilon)
+  : Optimiser<T, C>(graph, input_node_names, output_node_name, learning_rate_param)
+  , beta1_(beta1)
+  , beta2_(beta2)
+  , beta1_t_(beta1)
+  , beta2_t_(beta2)
+  , epsilon_(epsilon)
+{
+  for (auto &train : this->graph_trainables_)
```
baykaner: since this is a duplicate implementation as for the other constructor, it would be preferable to write a private function like this:

void Init()
{
  for (auto &train : this->graph_trainables_)
  {
    this->cache_.emplace_back(ArrayType(train->get_weights().shape()));
    this->momentum_.emplace_back(ArrayType(train->get_weights().shape()));
    this->mt_.emplace_back(ArrayType(train->get_weights().shape()));
    this->vt_.emplace_back(ArrayType(train->get_weights().shape()));
  }
  ResetCache();
}

and then have both constructors just invoke Init()

libs/ml/include/ml/optimisation/momentum_optimiser.hpp
```diff
@@ -77,6 +85,24 @@ MomentumOptimiser<T, C>::MomentumOptimiser(std::shared_ptr<Graph<T>>
...
+    std::shared_ptr<Graph<T>>
+
+                                    graph,
+    std::vector<std::string> const &input_node_names, std::string const &output_node_name,
+    fetch::ml::optimisers::LearningRateParam<DataType> const &learning_rate_param,
+    DataType const &                                          momentum_update)
+  : Optimiser<T, C>(graph, input_node_names, output_node_name, learning_rate_param)
+  , momentum_update_(momentum_update)
+{
+  for (auto &train : this->graph_trainables_)
```
baykaner: DRY - don't repeat yourself
implement a function that carries out this initialisation and call it from both constructors

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -136,6 +154,28 @@ Optimiser<T, C>::Optimiser(std::shared_ptr<Graph<T>> graph,
...
+  : graph_(graph)
+  , input_node_names_(std::move(input_node_names))
+  , output_node_name_(std::move(output_node_name))
+  , epoch_(0)
+  , learning_rate_param_(learning_rate_param)
+{
+  // initialize learning rate
+  learning_rate_ = learning_rate_param_.starting_learning_rate;
+
+  graph_trainables_ = graph_->get_trainables();
```
baykaner: repeated implementation

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -269,20 +332,24 @@ typename T::Type Optimiser<T, C>::Run(
...
     loss_ = DataType{0};
-    // Do batch back-propagation
-    input_ = loader.PrepareBatch(batch_size);
+    // update learning rate
     UpdateLearningRate();
+
+    // print the training stats every batch
+    PrintStats(batch_size, subset_size);
+
+    //    FETCH_LOG_INFO("ML_LIB", "Batch loss: ", loss);
```
baykaner: remove commented out code

libs/ml/include/ml/optimisation/rmsprop_optimiser.hpp
```diff
@@ -84,6 +93,29 @@ RMSPropOptimiser<T, C>::RMSPropOptimiser(std::shared_ptr<Graph<T>>
...
+    std::vector<std::string> const &input_node_names, std::string const &output_node_name,
+    fetch::ml::optimisers::LearningRateParam<DataType> const &learning_rate_param,
+    DataType const &decay_rate, DataType const &epsilon)
+  : Optimiser<T, C>(graph, input_node_names, output_node_name, learning_rate_param)
+  ,
+
+  decay_rate_(decay_rate)
+  , epsilon_(epsilon)
+{
+  auto weights = this->graph_->get_weights();
```
baykaner: repeated implementation

libs/ml/tests/ml/dataloader/skipgram_loader.cpp
```diff
@@ -81,37 +72,20 @@ TYPED_TEST(SkipGramDataloaderTest, loader_test)
...
   {
     if (loader.IsDone())
     {
       loader.Reset();
     }
-    left_and_right            = loader.GetNext().second;
-    std::string input         = loader.VocabLookup(SizeType(double(left_and_right.at(0).At(0, 0))));
-    std::string context       = loader.VocabLookup(SizeType(double(left_and_right.at(1).At(0, 0))));
-    auto        input_context = std::make_pair(input, context);
+    auto        left_and_right = loader.GetNext().second;
```
baykaner: slightly pedantic point but:
use of auto here is not preferred because other developers can't immediately see what type this object is.

libs/python/include/python/ml/all.hpp
```diff
@@ -20,7 +20,7 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
-#include "python/ml/dataloaders/py_cbow_loader.hpp"
+//#include "python/ml/dataloaders/py_cbow_loader.hpp"
```
baykaner: delete don't comment

libs/python/include/python/ml/all.hpp
```diff
@@ -42,7 +42,7 @@ void BuildMLLibrary(pybind11::module &module)
   fetch::ml::ops::BuildFullyConnected<T>("FullyConnected", module);
   fetch::ml::ops::BuildMeanSquareError<T>("MeanSquareError", module);
   fetch::ml::dataloaders::BuildMNISTLoader<T>("MNISTLoader", module);
-  fetch::ml::dataloaders::BuildCBOWLoader<T>("CBOWLoader", module);
+  //  fetch::ml::dataloaders::BuildCBOWLoader<T>("CBOWLoader", module);
```
baykaner: delete

baykaner (changes requested):
a couple of repeated implementations need fixing but otherwise nice work. I think you're getting pretty close to merging this in.

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
XiaodongWuFetch: lmao....

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
XiaodongWuFetch: max word count is only used to cut off too long data
so it should be fine.

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
XiaodongWuFetch: I did that when running the simulation

libs/ml/examples/word2vec_simple/w2v_model.hpp
XiaodongWuFetch: oh I thought u are going to fix that

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
XiaodongWuFetch: this is only handy when we have dynamic_window_size = window_size. So when we access the next sample pair after the buffer is drained, we will access out of bound samples. but extending the length of label, we can still access the node and see if it is filled with BUFFERNOTUSED. It saves us the effort of writing a extra logic, and make the buffer-is-drained-check more consistent

libs/ml/include/ml/optimisation/optimiser.hpp
XiaodongWuFetch: I thought we will restore them afterwards

libs/ml/include/ml/optimisation/adam_optimiser.hpp
XiaodongWuFetch: This is the implementation of constructor for generic optimiser, where these two string variables are passed to:
template <class T, class C>
Optimiser<T, C>::Optimiser(std::shared_ptr<Graph<T>>      graph,
                           std::vector<std::string> const input_node_names,
                           std::string const output_node_name, DataType const learning_rate)
  : graph_(graph)
  , input_node_names_(std::move(input_node_names))
  , output_node_name_(std::move(output_node_name))
  , learning_rate_(learning_rate)
  , epoch_(0)

So if we add std::move to the adam constructor (and all similar constructors such as sgd, momentum etc.), we need to change the implementation of generic optimiser constructor to use std::move as well?

XiaodongWuFetch:
glad it actually passed all tests~~

.gitignore
```diff
@@ -10,6 +10,7 @@ CMakeCache.txt
 nodes/
 .ipynb_checkpoints/
 /docker-images/
+fetch-ledger-develop-image
```
baykaner: generally speaking this is non-ideal. presumably this is the name of your symbolic link to the docker-images repo - but not everyone necessarily names it the same thing, so this wouldn't work for others anyway.

What you really want to do us edit your excludes file:
.git/info/exclude

it does exactly the same thing as gitignore but will only affect your local git repo and wont get pushed to remote

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -0,0 +1,279 @@
...
+  for (SizeType i{0}; i < tp.training_epochs; i++)
+  {
+    std::cout << "start training for epoch no.: " << i << std::endl;
+    optimiser.Run(data_loader, tp.batch_size);
+    std::cout << std::endl;
+    // Test trained embeddings
+    if (i % tp.test_frequency == 0)
+    {
+      TestEmbeddings(*g, skipgram_layer, data_loader, tp.word0, tp.word1, tp.word2, tp.word3, tp.k);
+      TestEmbeddings(*g, skipgram_layer, data_loader, "france", "france", "paris", "italy", tp.k);
```
baykaner: making two calls to TestEmbeddings once with variables and once with hard-coded values is not great. If you really want to do this you should at least add the hard-coded values as variables to the training params struct and invoke them properly. Then another developer can look at the training params struct and know they exist

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -0,0 +1,279 @@
...
+    }
+  }
+
+  //////////////////////////////////////
+  /// EXTRACT THE TRAINED EMBEDDINGS ///
```
baykaner: comment not attached to any code? delete it

libs/ml/include/ml/dataloaders/word2vec_loaders/unigram_table.hpp
```diff
@@ -107,6 +120,42 @@ bool UnigramTable::SampleNegative(uint64_t positive_index, SizeType &ret)
   return true;
 }
 
+/**
+ *
```
baykaner: comment is empty - please give a brief description of what this method is doing

libs/ml/include/ml/dataloaders/word2vec_loaders/vocab.hpp
```diff
@@ -42,6 +57,101 @@ class Vocab
   SizeType    IndexFromWord(std::string const &word) const;
 };
 
+Vocab::Vocab()
```
baykaner: I would say that an empty constructor is probably the one exemption from the rule of thumb that you should define methods in the class but define them outside the class.
You can ditch this and just add an {} to the definition in the class (or a = default)

libs/ml/include/ml/dataloaders/word2vec_loaders/vocab.hpp
```diff
@@ -42,6 +57,101 @@ class Vocab
...
+{}
+
+/**
+ * Update the data in vocabulary every time vocabulary is changed
```
baykaner: i guess you mean the 'meta-data'? since that's whats being updated here

libs/ml/include/ml/dataloaders/word2vec_loaders/vocab.hpp
```diff
@@ -42,6 +57,101 @@ class Vocab
...
+{
+  if (IndexFromWord(word) == UNKNOWN_WORD)
+  {
+    return false;
+  }
+  return true;
+}
+
+/**
+ * check if a word is known
```
baykaner: ... based on the word index

libs/ml/include/ml/dataloaders/word2vec_loaders/vocab.hpp
```diff
@@ -42,6 +57,101 @@ class Vocab
...
+{
+  if (id == UNKNOWN_WORD)
+  {
+    return false;
+  }
+  return true;
+}
+
+/**
+ * remove word that have less counts then min
```
baykaner: fewer ;-)

libs/ml/include/ml/dataloaders/word2vec_loaders/w2v_dataloader.hpp
```diff
@@ -37,6 +37,8 @@ template <typename T>
 class W2VLoader : public DataLoader<fetch::math::Tensor<T>, fetch::math::Tensor<T>>
 {
 public:
+  static constexpr T WindowContextUnused = -1;
```
baykaner, line 42: can you just add a comment here stating what is the expected type T used here? We spoke about this before, and I forget, but if T were SizeType this would roll around to be numeric_max which might be unintended

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -17,17 +17,56 @@
 //
 //------------------------------------------------------------------------------
 
+//------------------------------------------------------------------------------
```
baykaner: you seem to have pasted the copyright info in twice. please delete

libs/ml/include/ml/optimisation/optimiser.hpp
baykaner: or uncomment

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -262,24 +356,63 @@ typename T::Type Optimiser<T>::Run(fetch::ml::dataloaders::DataLoader<ArrayType,
...
+        "%) -- " + "learning rate: " + std::to_string(static_cast<double>(learning_rate_)) +
+        " -- " +
+        std::to_string(static_cast<double>(step_) / static_cast<double>(time_span_.count())) +
+        " samples / sec ";
+  }
+  // print it in log
+  //    FETCH_LOG_INFO("ML_LIB", "Training speed", stat_string_);
+  //    FETCH_LOG_INFO("ML_LIB", "Batch loss: ", loss_sum / step);
+  //      // print it in cout
+  std::cout << "\r" << stat_string_
```
baykaner: ditch the cout.

uncomment the logging

libs/ml/include/ml/optimisation/rmsprop_optimiser.hpp
```diff
@@ -75,13 +91,25 @@ RMSPropOptimiser<T>::RMSPropOptimiser(std::shared_ptr<Graph<T>>
...
+RMSPropOptimiser<T>::RMSPropOptimiser(
+    std::shared_ptr<Graph<T>>
+
+                                    graph,
+    std::vector<std::string> const &input_node_names, std::string const &label_node_name,
+    std::string const &                                       output_node_name,
+    fetch::ml::optimisers::LearningRateParam<DataType> const &learning_rate_param,
+    DataType const &decay_rate, DataType const &epsilon)
+  : Optimiser<T>(graph, input_node_names, label_node_name, output_node_name, learning_rate_param)
+  ,
```
baykaner: fix weird formatting!

baykaner (changes requested):

libs/ml/include/ml/dataloaders/word2vec_loaders/vocab.hpp
XiaodongWuFetch: what do u mean by meta-data?

baykaner (dismissed):

baykaner (approved):

PR #1270, opened 2019-06-27
Change Criterions to be Ops

MissingNO57:
- Removed Criterion and made all Criterions Ops.


libs/math/include/math/ml/loss_functions/mean_square_error.hpp
```diff
@@ -28,11 +28,7 @@ typename ArrayType::Type MeanSquareError(ArrayType const &A, ArrayType const &B)
 {
   typename ArrayType::Type ret = distance::SquareDistance(A, B);
```
baykaner, line 30: can we please move SquareDistance into its own file - its not easy to find at the moment

libs/math/include/math/ml/loss_functions/mean_square_error.hpp
```diff
@@ -28,11 +28,7 @@ typename ArrayType::Type MeanSquareError(ArrayType const &A, ArrayType const &B)
...
 
-  ret = Divide(ret, typename ArrayType::Type(A.size()));
-
-  // TODO(private 343)
-  // division by 2 allows us to cancel out with a 2 in the derivative
-  ret = Divide(ret, typename ArrayType::Type(2));
+  ret = Divide(ret, typename ArrayType::Type(A.shape().at(0)));
```
baykaner: lets change this mse implementation to use iterators so that we can calculate square distance and divide by data_points all in one iteration of the tensor

libs/ml/examples/asynchronous_distributed_learning/main.cpp
```diff
@@ -87,11 +90,13 @@ class TrainingClient
         // training data is low
         input = dataloader_.GetNext();
         g_.SetInput("Input", input.second.at(0));
+        g_.SetInput("Label", input.first);
         {
           std::lock_guard<std::mutex> l(m_);
-          ArrayType const &           results = g_.Evaluate("Softmax").Copy();
-          loss += criterion.Forward({results, input.first});
-          g_.BackPropagate("Softmax", criterion.Backward({results, input.first}));
+          ArrayType                   loss_tensor = g_.Evaluate("Error").Copy();
```
baykaner: do you still need the Copy here?

libs/ml/examples/asynchronous_distributed_learning/main.cpp
```diff
@@ -87,11 +90,13 @@ class TrainingClient
...
+          loss += fetch::math::statistics::Mean(loss_tensor);
+          loss_tensor.Fill(static_cast<DataType>(1));
```
baykaner: slightly confused about this. I dont think the user should be required to fill a tensor with ones and pass them in to backprop. Backprop should handle that for the user.

libs/ml/examples/centralised_distributed_learning/main.cpp
```diff
@@ -78,10 +81,12 @@ class TrainingClient
         // training data is low
         input = dataloader_.GetNext();
         g_.SetInput("Input", input.second.at(0));
+        g_.SetInput("Label", input.first);
 
-        ArrayType results = g_.Evaluate("Softmax").Copy();
-        loss += criterion.Forward({results, input.first});
-        g_.BackPropagate("Softmax", criterion.Backward({results, input.first}));
+        ArrayType loss_tensor = g_.Evaluate("Error").Copy();
```
baykaner: copy needed?

libs/ml/examples/code2vec/main.cpp
```diff
@@ -219,9 +219,13 @@ int main(int ac, char **av)
...
       "PredictionSoftMax", {prediction_softmax_kernel}, SizeType{1});
 
+  std::string label = g->AddNode<fetch::ml::ops::PlaceHolder<ArrayType>>("Label", {});
+
+  std::string error = g->AddNode<fetch::ml::ops::CrossEntropy<ArrayType>>("Error", {result, label});
+
   // Initialise Optimiser
-  fetch::ml::optimisers::AdamOptimiser<ArrayType, fetch::ml::ops::CrossEntropy<ArrayType>>
-      optimiser(g, {input_source_words, input_paths, input_target_words}, result, LEARNING_RATE);
+  fetch::ml::optimisers::AdamOptimiser<ArrayType> optimiser(
```
baykaner, line 228: it's cool that we don't have to pass the criterion template into optimisers now :+1

libs/ml/examples/commodity_prediction/main.cpp
```diff
@@ -212,6 +211,11 @@ std::pair<std::string, std::vector<std::string>> ReadArchitecture(
 
   g->AddNode<PlaceHolder<ArrayType>>(layer_name, {});  // add node for input
 
+  // Add label node
+  std::string label_name = "num_label";
+  g->AddNode<PlaceHolder<ArrayType>>(label_name, {});  // add node for input
```
baykaner: wrong comment

libs/ml/examples/commodity_prediction/main.cpp
```diff
@@ -299,9 +307,10 @@ DataType get_loss(std::shared_ptr<GraphType> const &g_ptr, std::string const &te
   {
     auto input = loader.GetNext();
     g_ptr->SetInput(node_names.front(), input.second.at(0).Transpose());
+    g_ptr->SetInput(node_names.at(1), input.first);
```
baykaner, line 238: this is pretty confusing code to read. The function is called SetInput is node_names.at(1) - so it's pretty unclear that what's happening here is that the Label is being set. Please add a comment, and please check with @EVSmithFetchAI that this node_names.at(1) implementation is a design choice that makes sense to her.

libs/ml/examples/commodity_prediction/main.cpp
```diff
@@ -412,7 +421,7 @@ int main(int argc, char **argv)
       auto input = loader.GetNext();
       g_ptr->SetInput("num_input", input.second.at(0).Transpose());
 
-      auto slice_output = g_ptr->Evaluate(node_names.back(), false);
+      auto slice_output = g_ptr->Evaluate(node_names.at(node_names.size() - 2), false);
```
baykaner, line 354: is g_ptr->Evaluate(node_names.at(node_names.size() - 2) really safe? feels like this could easily break if node adding is accidentally rearranged

libs/ml/examples/decentralised_distributed_learning/main.cpp
```diff
@@ -82,9 +85,14 @@ class TrainingClient
         // Randomly sampling the dataset, should ensure everyone is training on different data
         input = dataloader_.GetNext();
         g_.SetInput("Input", input.second.at(0));
-        ArrayType results = g_.Evaluate("Softmax").Copy();
-        loss += criterion.Forward({results, input.first});
-        g_.BackPropagate("Softmax", criterion.Backward({results, input.first}));
+        g_.SetInput("Label", input.first);
+
+        ArrayType results     = g_.Evaluate("Softmax").Copy();
```
baykaner: Aren't these Copies unnecessary? You're instantiating a new Tensor - which constructor is being called?

libs/ml/include/ml/ops/loss_functions/cross_entropy.hpp
```diff
@@ -25,64 +25,63 @@
...
 
   CrossEntropy()          = default;
   virtual ~CrossEntropy() = default;
 
-  /**
-   * @param inputs vector of 2 Tensors
-   * @return CrossEntropy value of 2 inputs
-   */
-  virtual DataType Forward(std::vector<ArrayType> const &inputs)
+  virtual void Forward(VecTensorType const &inputs, ArrayType &output)
```
baykaner: ditch "virtual" add override

libs/ml/include/ml/ops/loss_functions/cross_entropy.hpp
```diff
@@ -25,64 +25,63 @@
...
+      ret = fetch::math::Sigmoid(inputs.at(0).get());
+      fetch::math::Subtract(ret, inputs.at(1).get(), ret);
+      fetch::math::Multiply(ret, inputs.at(0).get(), ret);
     }
-    else if (inputs[0].shape().size())  // one-hot
+    else if (inputs.at(0).get().shape().size())  // one-hot
     {
-      ret = fetch::math::Softmax(inputs[0], 1);
-      fetch::math::Divide(inputs[1], ret, ret);
+      ret = fetch::math::Softmax(inputs.at(0).get(), 1);
```
baykaner: I think we can make this all more optimised by performing all the math in one tensor iteration

libs/ml/include/ml/ops/loss_functions/cross_entropy.hpp
```diff
@@ -25,64 +25,63 @@
...
-    return ret;
+    // chain rule
+    fetch::math::Multiply(ret, error_signal, ret);
+
+    return {ret, ret};
+  }
+
+  std::vector<typename T::SizeType> ComputeOutputShape(VecTensorType const &inputs) const
+  {
+    (void)inputs;
```
baykaner: I think you need to actually implement this function correctly in terms of inputs. Then from graph.BackPropogate you can instantiate a Tensor of 1's by making a call to ComputeOutputShape. Then the user won't have to pass a tensor of 1's into the error function.

libs/ml/include/ml/ops/loss_functions/mean_square_error.hpp
```diff
@@ -22,37 +22,66 @@
...
-    assert(inputs[0].shape() == inputs[1].shape());
+    assert(inputs.at(0).get().shape() == inputs.at(1).get().shape());
 
-    typename ArrayType::Type ret = fetch::math::MeanSquareError(inputs[0], inputs[1]);
-    return ret;
+    SizeType batch_dim = inputs.at(0).get().shape().size() - 1;
+    for (SizeType i{0}; i < inputs.at(0).get().shape().at(batch_dim); i++)
+    {
+      typename ArrayType::Type ret = fetch::math::MeanSquareError(
+          inputs.at(0).get().Slice(i, batch_dim), inputs.at(1).get().Slice(i, batch_dim));
```
baykaner: Am I right in saying that this performs implicit Tensor construction and copying from Slice into input parameters?

maybe add a TODO referring to when we implement isiterable throughout math library and we wont need to copy

libs/ml/include/ml/ops/loss_functions/mean_square_error.hpp
```diff
@@ -22,37 +22,66 @@
...
+  // grad[0]=2*err*(in[0]-in[1])/mean_size, grad[1]=-2*err*(in[0]-in[1])/mean_size,
+  virtual std::vector<ArrayType> Backward(VecTensorType const &inputs,
+                                          ArrayType const &    error_signal)
   {
     assert(inputs.size() == 2);
-    assert(inputs[0].shape() == inputs[1].shape());
-    return fetch::math::Subtract(inputs[0], inputs[1]);
+    assert(inputs.at(0).get().shape() == inputs.at(1).get().shape());
+
+    ArrayType return_signal_1(inputs.front().get().shape());
```
baykaner: dont construct new tensors in the training loop if you don't have to. Make these private member variables; cache the input shape - if it hasn't changed, just fill the private member tensors - otherwise reshape them.

libs/ml/include/ml/ops/loss_functions/mean_square_error.hpp
```diff
@@ -22,37 +22,66 @@
...
+    ArrayType return_signal_2(inputs.front().get().shape());
+
+    // return_signal_1=in[0]-in[1]
+    fetch::math::Subtract(inputs.at(0).get(), inputs.at(1).get(), return_signal_1);
+
+    // return_signal_1=err*(in[0]-in[1])
+    fetch::math::Multiply(return_signal_1, error_signal, return_signal_1);
```
baykaner: each of these math calls iterates over tensor. Again I think it's probably more efficient to do this all in one loop. We may need to do some benchmarking soon and investigate what size of tensor results in faster implementation for separate calls to math library ( because of vectorisation); but I suspect for this it will be small tensors we compute MSE over so one loop would be better

libs/ml/include/ml/ops/loss_functions/softmax_cross_entropy.hpp
```diff
@@ -21,49 +21,63 @@
...
   {
     assert(inputs.size() == 2);
-    assert(inputs[0].size() == inputs[1].size());
+    assert(inputs.at(0).get().size() == inputs.at(1).get().size());
+
+    ArrayType ret =
+        fetch::math::Subtract(fetch::math::Softmax(inputs.at(0).get()), inputs.at(1).get());
+
+    // chain rule
+    fetch::math::Multiply(ret, error_signal, ret);
```
baykaner: subtraction and multiply can be combined into one loop

libs/ml/include/ml/ops/loss_functions/softmax_cross_entropy.hpp
```diff
@@ -21,49 +21,63 @@
...
 
-    return fetch::math::Subtract(fetch::math::Softmax(inputs[0]), inputs[1]);
+    return {ret, ret};
+  }
+
+  std::vector<typename T::SizeType> ComputeOutputShape(VecTensorType const &inputs) const
+  {
+    (void)inputs;
```
baykaner, line 79: see previous comment regarding reworking these functions for CostFunction Ops

libs/ml/tests/ml/optimisation/optimisers.cpp
```diff
@@ -118,29 +123,30 @@ TYPED_TEST(OptimisersTest, sgd_optimiser_training)
...
+  fetch::ml::optimisers::SGDOptimiser<TypeParam> optimiser(g, {input_name}, label_name, output_name,
+                                                           learning_rate);
 
   // Do 2 optimiser steps
   optimiser.Run({data}, gt);
   DataType loss = optimiser.Run({data}, gt);
 
   // Test loss
-  EXPECT_NEAR(static_cast<double>(loss), 0.459031165, 1e-5);
+  EXPECT_NEAR(static_cast<double>(loss), 0.9180623, 1e-5);
```
baykaner: why has this changed?

libs/ml/tests/ml/optimisation/optimisers.cpp
```diff
@@ -158,25 +164,26 @@ TYPED_TEST(OptimisersTest, sgd_optimiser_training_2D)
...
+  fetch::ml::optimisers::SGDOptimiser<TypeParam> optimiser(g, {input_name}, label_name, output_name,
+                                                           learning_rate);
 
   // Do 2 optimiser steps
   optimiser.Run({data}, gt);
   DataType loss = optimiser.Run({data}, gt);
 
   // Test loss
-  EXPECT_NEAR(static_cast<double>(loss), 39.3290558, 1e-4);
+  EXPECT_NEAR(static_cast<double>(loss), 78.658111, 1e-4);
```
baykaner: these should not change. the division by 2 in the loss function should still take place in forward, not backward

libs/ml/tests/ml/training/basic_training.cpp
```diff
@@ -113,11 +116,12 @@ void PlusOneTest()
     cur_input.At(0, 0) = data.At(step, 0);
     g.SetInput(input_name, cur_input);
     cur_gt.At(0, 0) = gt.At(step, 0);
+    g.SetInput(label_name, cur_gt);
 
-    auto results = g.Evaluate(output_name);
-    loss += criterion.Forward({results, cur_gt});
-
-    g.BackPropagate(output_name, criterion.Backward({results, cur_gt}));
+    auto error_tensor = g.Evaluate(error_name);
```
baykaner, line 121: can the loss ops not be made batch aware so that the user doesn't have to manually compute the mean of the loss?

baykaner (changes requested):

libs/ml/tests/ml/optimisation/optimisers.cpp
MissingNO57: Because MeanSquareError used to be just SquareError divided by 2 instead of SquareError divided by size.

libs/ml/benchmark/training/training.cpp
```diff
@@ -68,9 +69,12 @@ void BM_Setup_And_Train(benchmark::State &state)
         "FC2", {a_1}, hidden_size, output_size);
     std::string output_name = g->template AddNode<fetch::ml::ops::Relu<TensorType>>("", {h_2});
 
+    std::string error_name = g->template AddNode<fetch::ml::ops::MeanSquareErrorLoss<TensorType>>(
+        "", {output_name, label_name});
+
     // Initialize Optimiser
-    fetch::ml::optimisers::SGDOptimiser<TensorType, fetch::ml::ops::MeanSquareError<TensorType>>
-        optimiser(g, {input_name}, output_name, learning_rate);
+    fetch::ml::optimisers::SGDOptimiser<TensorType> optimiser(g, {input_name}, label_name,
```
baykaner, line 76: :+1 this looks much neater like this

libs/ml/examples/centralised_distributed_learning/main.cpp
```diff
@@ -83,10 +83,9 @@ class TrainingClient
         g_.SetInput("Input", input.second.at(0));
         g_.SetInput("Label", input.first);
 
-        ArrayType loss_tensor = g_.Evaluate("Error").Copy();
-        loss += fetch::math::statistics::Mean(loss_tensor);
-        loss_tensor.Fill(static_cast<DataType>(1));
-        g_.BackPropagate("Error", loss_tensor);
+        ArrayType loss_tensor = g_.Evaluate("Error");
```
baykaner, line 86: it would be nice if we didn't have to do this. It's inefficient and ugly to generate a tensor as output from loss function when we know it should only be a DataType.

I'm not 100% sure the best way to achieve this, and maybe it should go in the next PR but how about something like this:
1. all ops that are loss functions have a flag/enum indicating that they are a loss type or not (we wanted to do this anyway to protect against using the output from the loss function in forward layers, or back-propping through the ground truth input)
2. Evaluate calls one of two different functions (EvalImplementation) depending on whether "Error" refers to a loss function op or not. Return type could be templated?

Not sure if that will work or not, but maybe we can think of something like this that makes sense

libs/ml/include/ml/graph.hpp
```diff
@@ -56,7 +56,8 @@ class Graph
   virtual ~Graph() = default;
 
   ArrayType    Evaluate(std::string const &node_name, bool is_training = true);
-  void         BackPropagate(std::string const &node_name, ArrayType const &error_signal);
+  void         BackPropagateSignal(std::string const &node_name, ArrayType const &error_signal);
```
baykaner, line 59: these names are confusing. If i understand correctly wouldn't it be better to rename as follows:
BackPropagate -> BackPropagateLoss
BackPropagate -> BackPropagateSignal

baykaner (changes requested):
Please just rename the BackPropagate then it looks ready to go

baykaner (approved):

PR #1271, opened 2019-06-27
[CORE] Fixing FromBase58(...) when input array contains trailing space(s) beyond it's end

### LINKED ISSUE #1216: Flakey ledger chaincode test CheckBasicTokenContract
This failure was observed on Clang 6 Debug; on replaying the build, the test passed.
```
[ RUN      ] SmartContractTests.CheckBasicTokenContract

/root/workspace/ledger_PR-1108-head/libs/ledger/tests/chaincode/smart_contract_tests.cpp:78: Failure
Expected equality of these values:
  SmartContract::Status::OK
    Which is: 4-byte object <00-00 00-00>
  SendQuery(query_method_name, request, response)
    Which is: 4-byte object <01-00 00-00>

unknown file: Failure
C++ exception with description "Unable to access keys of non-object variant" thrown in the test body.

/root/workspace/ledger_PR-1108-head/libs/ledger/tests/chaincode/smart_contract_tests.cpp:351: Failure
Actual function call count doesn't match EXPECT_CALL(*storage_, Get(target_resource))...
         Expected: to be called once
           Actual: never called - unsatisfied and active
[  FAILED  ] SmartContractTests.CheckBasicTokenContract (10 ms)
```

pbukva:
Root cause of this issue is bug in the `fetch::byte_array::FromBase58(...)` function, which allows to iterate the input buffer beyond its end boundary, what occurs in following two scenarios:

- when memory immediately after the buffer end boundary is filled with continuous sequence of _space_ character(s) (one or more). Besides accessing memory beyond end boundary, this leads to returning empty byte array as result.

- when *whole* buffer is filled with  _1_ characters (ascii hex value `0x31`) **AND** memory immediately after the buffer end boundary continues to be filled with continuous sequence of _1_ character(s) (one or more). Besides accessing memory beyond end boundary, this ultimately leads to `std::bad_alloc` exception. 

Resolves #1216.

### DIFF
--- a/libs/core/src/byte_array/base58.cpp
+++ b/libs/core/src/byte_array/base58.cpp
@@ -79,7 +79,7 @@ ConstByteArray FromBase58(ConstByteArray const &str)
   auto const *raw_end   = raw_start + str.size();
 
   // Skip leading spaces.
-  while ((raw_start != raw_end) && IsSpace(*raw_start))
+  while ((raw_start < raw_end) && IsSpace(*raw_start))
   {
     raw_start++;
   }
@@ -87,7 +87,7 @@ ConstByteArray FromBase58(ConstByteArray const &str)
   // Skip and count leading '1's.
   int zeroes = 0;
   int length = 0;
-  while (*raw_start == '1') {
+  while ((raw_start < raw_end) && *raw_start == '1') {
     zeroes++;
     raw_start++;
   }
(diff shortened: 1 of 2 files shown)


pbukva:
This commit is just introducing unit test to reproduce the issue & demonstrate the root cause.

pbukva:
This commit is just introducing unit test to reproduce the issue & demonstrate the root cause.
Thus the unit test is intentionally **expected to fail** the CI build in order to demonstrate it's effectiveness.

The following commit will actually fix the issue, with obvious expectation of unit test to pass, ultimately verifying effectiveness of the fix.

ejfitzgerald (approved):

ejfitzgerald:
FYI @pbukva  I have merged your change

PR #1272, opened 2019-06-28
Mark tests longer than 1s as slow

wkaluza:


pbukva:
In general I agree with the intention here, but I would be really careful about this.

Reason being that most people run `ci-tool.py` with just `-T` parameter, this especially applies to guys making changes to components like `math`, `ml`, `vm`, etc. ... .

Thus if some tests will be taken out of unit test group controlled by `-T` option, it might create false impression that everything is ok, but in reality the test(s) which people thing have been executed & passed during the test run have *not* been actually executed.

On different, but not unrelated note - separating/grouping unit tests based on their speed of execution factor does not feel the best. I would suggest to create group of `essential` unit tests which would contain unit tests based on their importance to be executed factor in _basic_ test run rather than on their time-to-execute factor.

baykaner:
as @wkaluza suggested let's get @ejfitzgerald opinion on this as well. This looks fine to me, and I'm happy to approve if he's on board with adding more 'slow' tests

wkaluza:
@pbukva Your comment applies to any situation where it is possible to execute a subset of the tests. It is the responsibility of the developer to verify that they are running the test they are going to change. Conventionally one does this by temporarily introducing a failing assertion.

Dividing unit tests into essential and non-essential is a blind alley in my opinion. Apart from the fact that the classification would be completely arbitrary, it would be indefensible from an engineering standpoint - which features of ledger are we happy to be broken? All tests are either essential or else they deserve to be deleted. Categorisation based on execution time is objective and has practical benefits, namely letting devs easily select a subset of tests that offers the fastest feedback loop during development.

pbukva:
@wkaluza I tend to disagree. First, the arbitrarity argument you mentioned actually applies way more on selecting time execution boundary which separates _slow_ and _normal_ unit tests - as by now set to 1 sec. First obvious reason is that it is simply arbitrarily selected number, but second, much more important reason is that this pre-selected time boundary actually separates unit tests based on test run timing results executed on very specific hardware configuration (CPU core speed, number of CPU cores, RAM sped, disk speed, etc ...) at this moment, which obviously can and does differ significantly between machines (e.g. docker containers configuration for CI build on cloud, developer machines, ...) , and it will become obsolete with new hardware as time will go on.

On the other hand, creating group of essential tests gives developers much higher value, since that group should used by developers locally on daily bases to verify whether the essential/core functionality is still working, and they should be executed in time reasonable for local development.
I'm not saying we should drop separation to slow / fast groups, just that creation of another - the _essential_ group would be useful.

wkaluza:
I think it would be challenging to come up with a measure of core-ness or essential-ness which we could agree on. Execution time is an objective measure, even if we use it by selecting arbitrary thresholds like 1 second. In any case, we have left engineering and entered philosophy. The issue is that we have long-running tests which should be marked as slow, but aren't.

pbukva:
That objective measure for essential tests is quite simple - the code coverage as the first factor and time-to-execute as the secondary factor.Where long executing tests should be 

But anyway, My point The issue qe are having

pbukva:
That objective measure for essential tests is quite simple - the code coverage as the first factor and time-to-execute as the secondary factor. As developer, I would rather wait a minute to execute essential unit test tests and have reasonably high confidence that my change has been tested enough, than get results in 30secs but with holes in code coverage (holes that I might not be aware of because some tests have been relocated).

But anyway, at this point, the main slowness in CI is not caused by `-T` unit test execution, but time spent on C++ build. I fact there is staggering difference between these two times.
Just to put this to perspective, with our current CI kubernetes configuration it takes just cca. 18secs to execute `-T` tests vs. **17mins** to execute C++build alone (for release build), and for debug build less than 2mins for `-T` unit tests versus 10+ mins for C++ build alone. This means, that when looking at big picture, we can neglect execution time for `-T` tests when compared to C++ build time.

wkaluza:
The point of this change is not to speed builds up but to save machine time by helping builds fail fast. See discussion on slack this morning.

ejfitzgerald (approved):

PR #1273, opened 2019-06-28
Miscellaneous changes

wkaluza:


libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -128,8 +128,8 @@ class MainChain
   /// @name Block Management
   /// @{
   BlockStatus AddBlock(Block const &block);
-  BlockPtr    GetBlock(BlockHash hash) const;
-  bool        RemoveBlock(BlockHash hash);
+  BlockPtr    GetBlock(BlockHash const &hash) const;
```
pbukva, line 131: I would be careful here, this does not feel correct. The internal implementation of this method uses moving semantic on `hash` input parameter. Changing this to reference in method signature effectively results to invalidating content of the external variable passed in by caller of this method.

I'm not completely sure why it was originally implemented this way, but my understanding is that it was intentional to avoid risk that pointer to data which `ConstByteArray` carries to won't get changed, what is critical in multi-threaded scenario in which this method is actually used (e.g. mining thread pool, etc. ...).

Keep in mind that `ConstByteArray` is stores data  using `std::shared_ptr<...>` thus, passing it _by-value_ carries minimal performance penalty when compared to passing _by-reference_.

This comment applies to all methods bellow with the analogical change.

pbukva (changes requested):
check my inline comment

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
```diff
@@ -447,14 +447,7 @@ class FixedPoint : public BaseFixedpointType
...
   {
-    if (Type(n) <= MAX && Type(n) >= MIN)
-    {
-      return true;
-    }
-    else
-    {
-      return false;
-    }
+    return Type(n) <= MAX && Type(n) >= MIN;
```
pbukva: This is just suggestion since we are changing it here - following  should improve readability and should performance too (I think compiler optimization won't be able to detect, in general, that it does not need to create 2 instances of Type since this is templated method and we have multiple specialisations of `TypeFromSize<...>`).

```suggestion
    auto const Type x{n};
    return MIN <= x && x <= MAX;
```

libs/vm/src/sharded_state.cpp
```diff
@@ -71,7 +71,7 @@ class ShardedState : public IShardedState
   }
 
 private:
-  inline Ptr<String> ComposeFullKey(Ptr<String> const &key)
+  Ptr<String> ComposeFullKey(Ptr<String> const &key)
```
pbukva, line 74: That `inline` might be useful here (from performance perspective) to give compiler hint of our wish, since it this is _not_ templated class (= inlining is not implicit).

libs/vm_modules/tests/unit/state_tests.cpp
```diff
@@ -20,8 +20,7 @@
 
 #include "gmock/gmock.h"
 
-namespace fetch {
```
pbukva, line 23: Just comment: I would not drop the namespaces here. They were here intentionally. The idea behind this is to set unit test code in to the very same namespace as tested production code is implemented in, in order to have the very same setup & experience (access to types) as the code would be written in tested production code. This helps to identify types which are missing (code won't compile) what it might indicate that there is potential unexpected dependency.

The most inner-nested _anonymous_ namespace is here intentionally as well, in order to avoid potential/accidental leak of types declared & defined by unit test code in to production code by linking. This can happen **IF** there is miss-configuration in build setup with unintended consequences which make production code to link with unit test objects.

libs/vm/src/sharded_state.cpp
wkaluza: Modern compilers do not need hints about what functions should be inlined. The `inline` keyword these days is only required as a directive for the linker to avoid apparent ODR violations.

libs/vm_modules/tests/unit/state_tests.cpp
wkaluza: As for setting the namespaces to provide the same 'environment' in terms of the available symbols as the production code, the using namespace declaration I introduced achieves the same goal.

The anonymous namespace is still there. The need for internal linkage in this case is clear.

I think these namespaces should be removed, because all they achieve is making the classes defined under the hood by `TEST_F` to think that they belong in a `fetch` namespace. I do not think hey do, as these classes are code defined by gtest.

libs/ledger/include/ledger/chain/main_chain.hpp
wkaluza: As far as I can tell none of the methods in question are called in a multithreaded context - even `DetectDuplicateTransactions` is called by the miner on a single thread, not the thread pool. The `std::move`s you refer to are no longer in place.

pbukva: The guts (e.g. BlockCoordinator) have been reworked in to state machine (=single thread workflow) since I saw it the last time. Back then methods have been actually called from multiple threads, and that was probably the reason behind passing _by-value_ & using moving semantics (rather than _by-reference_).

pbukva (dismissed):
lgtm

libs/vm/src/sharded_state.cpp
pbukva: :+1:

libs/vm_modules/tests/unit/state_tests.cpp
pbukva: :+1:

pbukva (approved):
LGTM

PR #1276, opened 2019-06-28
[CORE]@v0.5.x Fixing FromBase58(...) when input array contains trailing space(s) beyond it's end

pbukva:
This is cherry-pick of fix from master(PR #1271).

### DIFF
--- a/libs/core/src/byte_array/base58.cpp
+++ b/libs/core/src/byte_array/base58.cpp
@@ -79,7 +79,7 @@ ConstByteArray FromBase58(ConstByteArray const &str)
   auto const *raw_end   = raw_start + str.size();
 
   // Skip leading spaces.
-  while ((raw_start != raw_end) && IsSpace(*raw_start))
+  while ((raw_start < raw_end) && IsSpace(*raw_start))
   {
     raw_start++;
   }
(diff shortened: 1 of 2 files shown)


pbukva:
Above commit just introduces unit tests to reproduce the issues & demonstrate the root cause.
Thus these new unit tests are intentionally **expected to fail** in CI build in order to demonstrate it's effectiveness.

The following commit will actually fix the issues in prod. code, with obvious expectation of new unit tests to pass, ultimately verifying effectiveness of the fix.

n-hutton (commented):
I don't understand what causes this bug. So this means the bitcoin base58 implementation is flawed does it?

pbukva:
The PR #1271 contains detailed description of the bug.
The bug is actually caused by us, since we had to adjust original BTC code (https://github.com/bitcoin/bitcoin/blob/bccb4d29a8080bf1ecda1fc235415a11d903a680/src/base58.cpp) to take input in form of `ConstByteArray` type rather than original C-string `const char*` type, where original BTC code assumed & relied on proper C style string ending by `'\0'` character, what is **not** the case for `ConstByteArray` type.

ejfitzgerald:
This has been merged as of 0e3e484c

PR #1277, opened 2019-06-28
Feature/bls dkg

n-hutton:


libs/ledger/src/consensus/naive_entropy_generator.cpp
```diff
@@ -24,14 +24,17 @@
...
 namespace ledger {
 
+using Status = NaiveEntropyGenerator::Status;
+
 /**
  * Generate Entropy for a specified block period
  *
  * @param block_digest The specified block digest for the period
  * @param block_number The block number
  * @return The generated entropy
```
troels-im, line 34: Description needs updating

libs/dkg/src/dkg_service.cpp
```diff
@@ -0,0 +1,577 @@
...
+}
+
+State DkgService::OnBuildAeonKeysState()
+{
+  State next_state{State::REQUEST_SECRET_KEY};
+
+  // if (is_dealer_)
+  //{
+  //  next_state = State::BROADCAST_SIGNATURE;
+  //}
```
troels-im: Redundant?

PR #1278, opened 2019-06-28
Added tests for math and ml

evsmithx:
Added tests for minkowski distance, pairwise distance and l2loss and fixed broken code.


libs/math/tests/math/all_tests/distance/minkowski.cpp
```diff
@@ -0,0 +1,61 @@
...
+                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<32, 32>>>;
+
+TYPED_TEST_CASE(MinkowskiTest, MyTypes);
+
+TYPED_TEST(MinkowskiTest, simple_test)
+{
+  using SizeType = typename TypeParam::SizeType;
+
+  Tensor<double> A = Tensor<double>(3);
+  A.Fill(0);
```
MissingNO57: Tensor is zero-initialized, so you don't need to fill.

libs/math/tests/math/all_tests/distance/minkowski.cpp
```diff
@@ -0,0 +1,61 @@
...
+  A.Set(SizeType{0}, 1);
+  EXPECT_NEAR(Minkowski(A, A, SizeType{2}), 1.41421, 0.00001);
+
+  Tensor<double> B = Tensor<double>(3);
+  B.Fill(0);
```
MissingNO57: Tensor is zero-initialized, so you don't need to fill.

libs/math/tests/math/all_tests/distance/minkowski.cpp
```diff
@@ -0,0 +1,61 @@
...
+  B.Set(SizeType{1}, 1);
+  EXPECT_EQ(Minkowski(A, B, SizeType{2}), 1);
+
+  B.Fill(0);
```
MissingNO57: Tensor is zero-initialized, so you don't need to fill.

libs/math/tests/math/all_tests/distance/minkowski.cpp
```diff
@@ -0,0 +1,61 @@
...
+  B.Set(SizeType{1}, 2);
+  EXPECT_EQ(Minkowski(A, B, SizeType{2}), 1);
+
+  B.Fill(0);
```
MissingNO57: Tensor is zero-initialized, so you don't need to fill.

libs/math/include/math/ml/loss_functions/l2_loss.hpp
```diff
@@ -24,12 +24,26 @@ namespace fetch {
...
   return tmp;
 }
 
+template <typename ArrayType>
+meta::IfIsMathNonFixedPointArray<ArrayType, typename ArrayType::Type> L2Loss(ArrayType const &at)
+{
+  auto a = at.data();
+
+  using VectorRegisterType = typename ArrayType::VectorRegisterType;
+  using type               = typename ArrayType::Type;
```
MissingNO57: Style - Should have first letter upper-case and please name it DataType, because it's named like that everywhere else.

libs/math/include/math/distance/pairwise_distance.hpp
```diff
@@ -18,34 +18,31 @@
...
-
-    for (SizeType j = i + 1; j < a.height(); ++j)
+    for (SizeType j = i + 1; j < a.shape(0); ++j)
     {
-      ArrayType slice2 = a.data().slice(offset2, a.width());
-      offset2 += a.padded_width();
-      r(SizeType(0), k++) = metric(slice1, slice2);
+      ArrayType slice1    = a.Slice(i).Copy();
+      ArrayType slice2    = a.Slice(j).Copy();
+      r(SizeType{0}, k++) = metric(slice1, slice2);
```
MissingNO57: I would suggest renaming r to ret, because it would make the code clearer.

MissingNO57 (dismissed):
Looks good, just a few style changes.

libs/math/tests/math/all_tests/distance/minkowski.cpp
```diff
@@ -0,0 +1,61 @@
...
+//
+//------------------------------------------------------------------------------
+
+#include "core/random/lcg.hpp"
+#include "math/distance/minkowski.hpp"
+#include "math/tensor.hpp"
+
+#include "gtest/gtest.h"
+
+using namespace fetch::math::distance;
```
wkaluza, line 29: Small point - try to remember to wrap test cpp files' contents in an anonymous namespace. This helps to keep things tidy for the linker, and prevents us from shooting ourselves in the foot accidentally. See `lib/vm_modules/tests/unit/vm_modules_tests.cpp` for an example.

libs/math/tests/math/all_tests/distance/pairwise_distance.cpp
```diff
@@ -0,0 +1,81 @@
...
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "math/distance/pairwise_distance.hpp"
+#include "math/tensor.hpp"
+
+#include "gtest/gtest.h"
+
+using namespace fetch::math::distance;
```
wkaluza, line 25: Please add anonymous namespace.

libs/ml/tests/ml/ops/divide.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+#include "math/tensor.hpp"
+#include "vectorise/fixed_point/fixed_point.hpp"
+
+#include "gtest/gtest.h"
+
+#include <cstdint>
+#include <vector>
+
+template <typename T>
```
wkaluza, line 30: anonymous namespace

wkaluza (dismissed):

libs/math/include/math/distance/minkowski.hpp
```diff
@@ -26,8 +29,8 @@ namespace math {
 namespace distance {
 
 template <typename ArrayType>
-inline typename ArrayType::Type Minkowski(ArrayType const &a, ArrayType const &b,
-                                          typename ArrayType::type n)
+inline meta::IfIsMathArray<ArrayType, typename ArrayType::Type> Minkowski(
+    ArrayType const &a, ArrayType const &b, typename ArrayType::Type n)
 {
```
baykaner, line 34: have you confirmed that this implementation works as intended? It seems like count is initialised at 0 and never incremented. Please compare with an external implementation to verify

libs/math/include/math/distance/pairwise_distance.hpp
```diff
@@ -18,34 +18,31 @@
...
 template <typename ArrayType, typename F>
-inline ArrayType &PairWiseDistance(ArrayType &r, ArrayType const &a, F &&metric)
+inline meta::IfIsMathArray<ArrayType, ArrayType> &PairWiseDistance(ArrayType const &a, F &&metric,
+                                                                   ArrayType &r)
 {
   using SizeType = typename ArrayType::SizeType;
 
-  detailed_assert(r.height() == 1);
-  detailed_assert(r.width() == (a.height() * (a.height() - 1) / 2));
+  detailed_assert(r.shape(0) == 1);
```
baykaner: you should probably also assert that all tensors are 2D

libs/math/include/math/distance/pairwise_distance.hpp
```diff
@@ -18,34 +18,31 @@
...
-    offset1 += a.padded_width();
-    SizeType offset2 = offset1;
-
-    for (SizeType j = i + 1; j < a.height(); ++j)
+    for (SizeType j = i + 1; j < a.shape(0); ++j)
     {
-      ArrayType slice2 = a.data().slice(offset2, a.width());
-      offset2 += a.padded_width();
-      r(SizeType(0), k++) = metric(slice1, slice2);
+      ArrayType slice1    = a.Slice(i).Copy();
```
baykaner: please add a TODO and raise an issue (if it's not already raised) to implement math functions with IsIterable. When that is complete we'll remove these copies.

libs/math/tests/math/all_tests/distance/minkowski.cpp
baykaner: or better yet replace lines 43 - 45 with:
Tensor<double> t = Tensor<double>::FromString("1, 0, 0");

libs/math/tests/math/all_tests/distance/minkowski.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  B.Fill(0);
+  B.Set(SizeType{1}, 2);
+  EXPECT_EQ(Minkowski(A, B, SizeType{2}), 1);
+
+  B.Fill(0);
+  B.Set(SizeType{0}, 1);
+  B.Set(SizeType{1}, 1);
+  EXPECT_NEAR(Minkowski(A, B, SizeType{2}), 1.41421, 0.00001);
+}
```
baykaner, line 78: it would be nice to have a test that checks minkowski 1 == manhattan; minkowski2 == euclidean

libs/math/tests/math/all_tests/distance/minkowski.cpp
```diff
@@ -0,0 +1,61 @@
...
+                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<16, 16>>,
+                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<32, 32>>>;
+
+TYPED_TEST_CASE(MinkowskiTest, MyTypes);
+
+TYPED_TEST(MinkowskiTest, simple_test)
+{
+  using SizeType = typename TypeParam::SizeType;
+
+  Tensor<double> A = Tensor<double>(3);
```
baykaner: this is a typed test - but you're explicitly instantiating Tensor<double>.
If the implementation does not work for FixedPoint yet please add the relevant non-vectorised implementation.

libs/ml/tests/ml/ops/divide.cpp
```diff
@@ -0,0 +1,69 @@
...
+TYPED_TEST(DivideTest, forward_test)
+{
+  TypeParam data = TypeParam::FromString("1, 1, 2; 3, 4, 5");
+
+  TypeParam gt = TypeParam::FromString("1, 1, 1; 1, 1, 1");
+
+  fetch::ml::ops::Divide<TypeParam> op;
+
+  TypeParam prediction(op.ComputeOutputShape({data}));
+  op.Forward({data, data}, prediction);
```
baykaner: this is not a great test. dividing values by themselves and checking the output is 1 does not really test that much. we should check:
1. do some reasonable values give function tolerance precision results.
2. what happens for division by 0?
3. negative values?

baykaner (changes requested):

libs/math/include/math/distance/minkowski.hpp
evsmithx: Good catch! The implementation was totally wrong.

libs/math/include/math/distance/manhattan.hpp
```diff
@@ -30,15 +30,14 @@ template <typename ArrayType>
...
 {
   assert(a.size() == b.size());
-  using Type     = typename ArrayType::Type;
-  using SizeType = typename ArrayType::SizeType;
+  using Type = typename ArrayType::Type;
 
-  Type     result = 0;
-  SizeType count  = 0;
+  Type result{0};
+  auto b_ptr = b.begin();
```
baykaner: b is an iterator - it's slightly confusing to name it pointer. b_it would be fine.
Also, would be better to use cbegin() if possible.

libs/math/include/math/distance/minkowski.hpp
```diff
@@ -34,15 +34,15 @@ inline meta::IfIsMathArray<ArrayType, typename ArrayType::Type> Minkowski(
...
   detailed_assert(a.size() == b.size());
   using DataType = typename ArrayType::Type;
-  using SizeType = typename ArrayType::SizeType;
 
-  SizeType count = 0;
-  DataType sum   = 0;
+  DataType sum{0};
+  auto     b_ptr = b.begin();
```
baykaner: same as previous

libs/math/tests/math/all_tests/distance/minkowski.cpp
```diff
@@ -38,24 +42,38 @@ TYPED_TEST_CASE(MinkowskiTest, MyTypes);
...
+
+  // minkowski with lambda=1 == manhattan
+  a         = TypeParam::FromString("1, 2, 3");
+  b         = TypeParam::FromString("10, 11, 12");
+  ret       = Minkowski(a, b, Type{1});
+  Type ret2 = Manhattan(a, b);
+  EXPECT_NEAR(static_cast<double>(ret), static_cast<double>(ret2),
+              static_cast<double>(function_tolerance<Type>()));
+
+  // minkowski with lambda=2 == euclidean
```
baykaner, line 71: nice - much better tests

baykaner (changes requested):

baykaner (approved):

PR #1282, opened 2019-06-28
optimised embeddings and matrix multiply implementations

baykaner:
- embeddings had unnecessary tensor traversals - this is now computed all in one loop
- matrix multiply had multiple tensor creations that can be avoided with private members
- these changes should hopefully speed up backprop for these ops


MissingNO57 (approved):
LGTM

PR #1289, opened 2019-06-30
Synchronised access wrappers

wkaluza:
A common pattern in our code is having a variable which may be susceptible to race conditions, and an associated mutex. Before accessing the protected variable, we need to remember to manually lock the mutex; this is a potential source of bugs which we could offload onto the compiler.

This defines a wrapper class template which would own the mutex and the protected quantity, and would allow threadsafe access via handles which lock the mutex until they are destroyed. Const-correctness is enforced via the type system. See `protect_tests.cpp` in this PR for intended usage.

I also implemented the wrapper in two places in our code as an example: `StateMachine` and `Reactor`. In the process I also found some locks which I don't think are necessary.

Is this something we would be interested in? If so, please feel free to suggest improvements and use cases.


ejfitzgerald:
@wkaluza have you seen the synchronised state class:
https://github.com/fetchai/ledger/blob/master/libs/core/include/core/threading/synchronised_state.hpp

wkaluza:
@ejfitzgerald I have, yes. It was `SynchronisedState` that inspired me to write this.

`Protect` is designed to be a lower level class: `SynchronisedState` can be easily reimplemented on top of `Protect`. I think `SynchronisedState` suffers from trying to be several things at once, which makes it more complicated than it needs to be. This could partly explain why it hasn't really caught on in the codebase, and people prefer to keep track of mutexes by hand.

To my mind, the true strength of `SynchronisedState` lies in it being a waitable, rather than access control, which is what `Protect` is meant to do. I think separating the two responsibilities would result in a cleaner design, as we would then have two synchronisation classes which Do One Thing Well. The fact that `SynchronisedState` is not unit tested does not help matters.

One bit of production code that you will be familiar with which could be immediately simplified by using `Protect` is `StakeUpdateQueue`: it has no use for the `WaitFor` functionality of `SynchronisedState`, as all it needs is a thread-safe box to hold the `updates_` member variable. Right now any manipulations of `updates_` have to be done by passing around lambdas, resulting in code that is needlessly complicated.

Let's sync up when we get in tomorrow.

n-hutton:
Although I think it could be an interesting idea, I don't really get what the point is of Protect<const int> for example. Your variable is const and so would not need a mutex. I am not sure of the advantage for either over an atomic

Also, imo the reason that SyncronisedState hasn't caught on is probably because you often want to lock more than one thing at once.

wkaluza:
> Although I think it could be an interesting idea, I don't really get what the point is of Protect for example. Your variable is const

You'd need to clarify by what you mean by 'my variable'.. The payload of Protect is not const, or doesn't have to be. See tests for example, specifically the one illustrating read-write access.

In any case, this PR is pending rework following review.

n-hutton:
I guess I was looking specifically at ```Protect<const int>                protected_number;``` and the test that ensures that you can only read protected_number

ejfitzgerald:
Closing to be reopened at a later time

wkaluza:
Closing temporarily to reduce load on build server.

libs/core/include/core/synchronisation/protected.hpp
```diff
@@ -0,0 +1,92 @@
...
+      -> std::enable_if_t<std::is_void<decltype(handler(payload_))>::value, void>
+  {
+    std::lock_guard<M> lock(mutex_);
+
+    handler(payload_);
+  }
+
+  template <typename Handler>
+  auto Apply(Handler &&handler)
+      -> std::enable_if_t<!std::is_void<decltype(handler(payload_))>::value,
```
ejfitzgerald: Worth an informative alias?

libs/core/include/core/synchronisation/waitable.hpp
```diff
@@ -0,0 +1,128 @@
...
+#include <functional>
+#include <mutex>
+#include <utility>
+
+namespace fetch {
+
+template <typename T, typename M = std::mutex>
+class Waitable
+{
+private:
```
ejfitzgerald, line 33: Can we switch this section around in the class definition?

libs/core/include/core/synchronisation/waitable.hpp
```diff
@@ -0,0 +1,128 @@
...
+  {
+    protected_payload_.Apply([this, handler](auto const &payload) -> void {
+      handler(payload);
+      condition_.notify_all();
+    });
+  }
+
+  template <typename Handler>
+  auto Apply(Handler &&handler)
+      -> std::enable_if_t<!std::is_void<decltype(protected_payload_.Apply(handler))>::value,
```
ejfitzgerald: These are getting a little hairy.

ejfitzgerald (approved):
Not sure that I would have called it `ApplyVoid` but we can rename that in the future

PR #1291, opened 2019-07-01
Bloom filter

wkaluza:
Still left to do:
- preventing filter degradation: rebuilding filter when false positive rate exceeds threshold, adaptive size and hash count
- benchmarking of hash function selection and order
- lazy hash evaluation
- building up filter from a non-empty store


libs/core/include/core/bloom_filter.hpp
```diff
@@ -0,0 +1,224 @@
...
+  std::size_t getHash(std::size_t index) const;
+
+  std::vector<std::size_t> data_;
+
+  friend class HashSourceFactory;
+};
+
+}  // namespace internal
+
+class BloomFilter
```
ejfitzgerald: Style 🚓 : `BloomFilterInterface`

ejfitzgerald: Also this should be moved to its own header file to reduce the coupling

libs/core/include/core/bloom_filter.hpp
```diff
@@ -0,0 +1,224 @@
...
+  internal::HashSourceFactory const hash_source_factory_;
+  std::size_t                       entry_count_{};
+  std::size_t                       positive_count_{};
+  std::size_t                       false_positive_count_{};
+};
+
+/*
+ * A fake Bloom filter which holds no data and treats any query as positive.
+ */
+class DummyBloomFilter : public BloomFilter
```
ejfitzgerald: As far as I can tell this is only used in tests therefore we should move it to the test folder.

libs/core/include/core/feature_flags.hpp
```diff
@@ -31,6 +31,8 @@ namespace core {
 class FeatureFlags
 {
 public:
+  constexpr static char const *const MAIN_CHAIN_BLOOM_FILTER = "main_chain_bloom_filter";
```
ejfitzgerald: Pedantic: I believe the final `const` is redundant because of the `constexpr` also as for consistency normally these are done: `static constexpr char const *`

libs/core/src/bloom_filter.cpp
```diff
@@ -0,0 +1,375 @@
...
+HashSource HashSourceFactory::operator()(Bytes const &element) const
+{
+  return HashSource(this->hash_functions_, element);
+}
+
+HashSource::HashSource(HashSourceFactory::Functions const &hash_functions,
+                       HashSourceFactory::Bytes const &    input)
+  : data_{}
+{
+  // TODO(WK): lazily evaluate hashes
```
ejfitzgerald: Please raise as an issue

libs/core/src/bloom_filter.cpp
```diff
@@ -0,0 +1,375 @@
...
+// No point in evaluating the filter's quality until this many positive queries
+// have been executed
+constexpr std::size_t const MEANINGFUL_STATS_THRESHOLD = 5u * INVERSE_TARGET_FALSE_POSITIVE_RATE;
+
+constexpr std::size_t const INITIAL_SIZE_IN_BITS = 8 * 10 * 1024 * 1024;
+
+namespace fetch {
+namespace internal {
+
+HashSourceFactory::HashSourceFactory(HashSourceFactory::Functions hash_functions)
```
ejfitzgerald, line 43: Can we add documentation to all of these methods please.

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -17,30 +17,53 @@
...
 
 using fetch::byte_array::ToBase64;
 using fetch::generics::MilliTimer;
 
 namespace fetch {
 namespace ledger {
 
+namespace {
+
+void addBlockToBloomFilter(BloomFilter &bf, Block const &block)
```
ejfitzgerald: Style 🚓 

```suggestion
void AddBlockToBloomFilter(BloomFilter &bf, Block const &block)
```

libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -118,7 +120,7 @@ class MainChain
   };
 
   // Construction / Destruction
-  explicit MainChain(Mode mode = Mode::IN_MEMORY_DB);
+  explicit MainChain(std::unique_ptr<BloomFilter> bloom_filter, Mode mode = Mode::IN_MEMORY_DB);
```
ejfitzgerald: I think that it would be cleaner for the moment to allow this bloom filter to be optional i.e. it can be a defaulted to a null parameter

libs/core/src/bloom_filter.cpp
```diff
@@ -0,0 +1,375 @@
...
+{
+  return HashSource::HashSourceIterator{this, data_.size()};
+}
+
+std::size_t HashSource::getHash(std::size_t index) const
+{
+  return data_[index];
+}
+
+HashSource::HashSourceIterator::~HashSourceIterator() = default;
```
ejfitzgerald: All these `= default`s should be the headers

libs/core/src/bloom_filter.cpp
```diff
@@ -0,0 +1,375 @@
...
+      openssl_type = EVP_sha512();
+      break;
+    case Type::SHA_1:
+      openssl_type = EVP_sha1();
+      break;
+    case Type::SHA_2_512:
+      openssl_type = EVP_sha512();
+      break;
+    case Type::SHA_3_512:
+      openssl_type = EVP_sha3_512();
```
ejfitzgerald: This has doesn't work on my machine

libs/core/src/bloom_filter.cpp
```diff
@@ -0,0 +1,375 @@
...
+    switch (type)
+    {
+    case Type::MD5:
+      openssl_type = EVP_sha512();
+      break;
+    case Type::SHA_1:
+      openssl_type = EVP_sha1();
+      break;
+    case Type::SHA_2_512:
+      openssl_type = EVP_sha512();
```
ejfitzgerald, line 171: Can we not add these hashes to our existing crypto library?

libs/core/src/bloom_filter.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+void DummyBloomFilter::Add(Bytes const &)
+{}
+
+bool DummyBloomFilter::ReportFalsePositives(std::size_t)
+{
+  return true;
+}
+
+std::unique_ptr<BloomFilter> BloomFilter::create(Type type)
```
ejfitzgerald: Style 🚓 

```suggestion
std::unique_ptr<BloomFilter> BloomFilter::Create(Type type)
```

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -17,30 +17,53 @@
...
 /**
  * Constructs the main chain
  *
  * @param mode Flag to signal which storage mode has been requested
  */
-MainChain::MainChain(Mode mode)
+MainChain::MainChain(std::unique_ptr<BloomFilter> bloom_filter, Mode mode)
+  : bloom_filter_(std::move(bloom_filter))
+  , bloom_filter_false_positive_count_(telemetry::Registry::Instance().CreateCounter(
+        "ledger_main_chain_bloom_filter_false_positive_total"))
```
ejfitzgerald: This will need a description to, otherwise that is a warning in prometheus. (I have a PR open to fix the existing ones)

libs/core/src/bloom_filter.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+}  // namespace
+
+}  // namespace internal
+
+BasicBloomFilter::Functions const default_hash_functions{internal::raw_data,  internal::md5,
+                                                         internal::sha_1,     internal::sha_2_512,
+                                                         internal::sha_3_512, internal::fnv};
+
+BasicBloomFilter::~BasicBloomFilter() = default;
```
ejfitzgerald: In header file

libs/core/include/core/bloom_filter.hpp
```diff
@@ -0,0 +1,224 @@
...
+  using Function  = internal::HashSourceFactory::Function;
+  using Functions = internal::HashSourceFactory::Functions;
+
+  enum class Type
+  {
+    DUMMY,
+    BASIC
+  };
+
+  static std::unique_ptr<BloomFilter> create(Type);
```
ejfitzgerald: Style 🚓 

```suggestion
  static std::unique_ptr<BloomFilter> Create(Type);
```

libs/core/include/core/bloom_filter.hpp
```diff
@@ -0,0 +1,224 @@
...
+  HashSourceIterator end() const;
+  HashSourceIterator cbegin() const;
+  HashSourceIterator cend() const;
+
+private:
+  HashSource(HashSourceFactory::Functions const &, HashSourceFactory::Bytes const &);
+
+  std::size_t getHash(std::size_t index) const;
+
+  std::vector<std::size_t> data_;
```
ejfitzgerald: This is repeatedly used, consider a type alias?

libs/core/include/core/bloom_filter.hpp
```diff
@@ -0,0 +1,224 @@
...
+
+    HashSourceIterator &operator=(HashSourceIterator const &);
+    HashSourceIterator &operator=(HashSourceIterator &&) noexcept;
+
+    HashSourceIterator() = delete;
+
+    bool operator!=(HashSourceIterator const &other) const;
+    bool operator==(HashSourceIterator const &other) const;
+
+    HashSourceIterator const operator++(int);
```
ejfitzgerald: Do we really need a post-increment operator? I tend not to define them if I can help it.

libs/core/include/core/bloom_filter.hpp
```diff
@@ -0,0 +1,224 @@
...
+
+    ~HashSourceIterator();
+
+    HashSourceIterator(HashSourceIterator const &);
+    HashSourceIterator(HashSourceIterator &&) noexcept;
+
+    HashSourceIterator &operator=(HashSourceIterator const &);
+    HashSourceIterator &operator=(HashSourceIterator &&) noexcept;
+
+    HashSourceIterator() = delete;
```
ejfitzgerald: Style: In general can we group constructors and destructors together and in a separate group the assignment operators

libs/core/include/core/bloom_filter.hpp
```diff
@@ -0,0 +1,224 @@
...
+ * to an input, invoke the factory's operator() and use the resulting
+ * HashSource.
+ *
+ * The factory must be kept alive while its HashSource instances remain
+ * in use.
+ */
+class HashSourceFactory
+{
+public:
+  using Bytes     = fetch::byte_array::ConstByteArray;
```
ejfitzgerald: I would prefer no to have this type alias because I will (at some point) remove byte array and its friends from the core namespace and leave it in the main fetch namespace. When this happens it will be an easy job to remove all the forwarding type aliases we have (provided they have the same name 😁 )

libs/core/include/core/bloom_filter.hpp
```diff
@@ -0,0 +1,224 @@
...
+  using Function  = std::function<std::vector<std::size_t>(Bytes const &)>;
```
ejfitzgerald: Move to private section

libs/core/src/bloom_filter.cpp
```diff
@@ -0,0 +1,375 @@
...
+}
+
+HashSource::HashSourceIterator::HashSourceIterator(HashSource const *source, std::size_t index)
+  : hash_index_{index}
+  , source_{source}
+{}
+
+namespace {
+
+class OpenSslHasher
```
ejfitzgerald, line 118: Do we not already have a similar concept for this in the crypto library?

ejfitzgerald (changes requested):
Looks very promising. Let me know what you think.

libs/core/src/bloom_filter.cpp
wkaluza: Why? I think the opposite is true. This is implementation just like an explicit function body.

libs/core/include/core/bloom_filter.hpp
wkaluza: :+1:

libs/core/include/core/bloom_filter.hpp
wkaluza: No: it's used by the main chain by default as long as the feature flag is not set, in which case the BasicBloomFilter is used instead

libs/core/include/core/feature_flags.hpp
wkaluza: But more const == more better...! Fine.

libs/core/src/bloom_filter.cpp
wkaluza: Done.

libs/ledger/src/chain/main_chain.cpp
wkaluza: Done

libs/ledger/include/ledger/chain/main_chain.hpp
wkaluza: I'll default it to the DummyBloomFilter, which is the null implementation

libs/core/src/bloom_filter.cpp
wkaluza: Regarding our conversation: this is now done.

libs/core/src/bloom_filter.cpp
wkaluza: I was thinking of doing it as a follow up, once we decide which ones we actually need.

libs/core/src/bloom_filter.cpp
wkaluza: :+1:

libs/ledger/src/chain/main_chain.cpp
wkaluza: Done

libs/core/src/bloom_filter.cpp
wkaluza: Done.

libs/core/include/core/bloom_filter.hpp
wkaluza: Fair.

libs/core/include/core/bloom_filter.hpp
wkaluza: Done

libs/core/include/core/bloom_filter.hpp
wkaluza: It's gone.

libs/core/src/bloom_filter.cpp
wkaluza: Not really - there's a FNV implementation that I chose to use here, but in general this is quite lightweight. I think there's room for adding some general purpose digest functions to our crypto lib

libs/core/include/core/bloom_filter.hpp
wkaluza: :+1:

libs/core/include/core/bloom_filter.hpp
wkaluza: Done

libs/core/include/core/bloom_filter.hpp
wkaluza: :+1:

libs/core/src/bloom_filter.cpp
wkaluza: Done

wkaluza:
Closing to force full CI rebuild

libs/core/src/bloom_filter.cpp
wkaluza: You're right - after clearing the CI workspace the mac build started failing. I removed the sha3 hash function.

ejfitzgerald (approved):

PR #1292, opened 2019-07-01
L1 and L2 regularization implemented

MissingNO57:


libs/ml/examples/mnist/main.cpp
```diff
@@ -66,11 +66,14 @@ int main(int ac, char **av)
 
   std::string input   = g->AddNode<PlaceHolder<ArrayType>>("Input", {});
   std::string layer_1 = g->AddNode<FullyConnected<ArrayType>>(
-      "FC1", {input}, 28u * 28u, 10u, fetch::ml::details::ActivationType::RELU);
+      "FC1", {input}, 28u * 28u, 10u, fetch::ml::details::ActivationType::RELU,
+      fetch::ml::details::RegularisationType::L1, DataType{0.01f});
```
baykaner: instead of hard-coding. define above:
DataType l1_reg = 0.01f;

libs/ml/include/ml/graph.hpp
```diff
@@ -96,6 +105,32 @@ class Graph
   std::vector<TrainablePtrType>                trainable_;
 };
 
+template <typename ArrayType>
```
baykaner, line 112: add doxygen style function comments and explanation of what each of the two set regularisation functions does differently

libs/ml/include/ml/graph.hpp
```diff
@@ -96,6 +105,32 @@ class Graph
...
+                                         fetch::ml::details::RegularisationType regularisation,
+                                         DataType                               regularisation_rate)
+{
+  if (nodes_[node_name])
+  {
+    trainable_[node_name]->SetRegularisation(regularisation, regularisation_rate);
+  }
+  else
+  {
+    throw std::runtime_error("Cannot change regularisation parameters for node [" + node_name +
```
baykaner: how about we change the design so that SetRegularisation returns a success bool instead of throwing? I think in general anything that get used in the VM will need to return a success bool so that the VM can handle the error in whatever way is most suitable.

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -57,6 +59,36 @@ class Trainable
...
+      break;
+    }
+  }
+
+protected:
+  fetch::ml::details::RegularisationType regularisation_ =
+      fetch::ml::details::RegularisationType::NONE;
+  DataType regularisation_rate_ = static_cast<DataType>(0);
+
+  virtual void L1Regularisation(DataType regularisation_rate) = 0;
```
baykaner: I think it would be better if you make an abstract class called Regulariser. then make child classes L1Regularisation and L2Regularisation. Then weights.hpp can have a regulariser_ private member variable; Then SetRegularisation constructs the variable, and you don't need the switch case in apply regularisation anymore - you just invoke the regulariser.

This way, the regularisation implementations can live in ml/regularisers rather than being implemented directly in weights.

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -107,6 +140,27 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<T>
     ResetGradients();
   }
 
+  virtual void L1Regularisation(DataType regularisation_rate)
+  {
+    ArrayType weight     = *this->output_;
+    ArrayType reg_output = ArrayType(weight.shape());
+
+    fetch::math::Multiply(weight, static_cast<DataType>(-2) * regularisation_rate, reg_output);
```
baykaner: static casting -2 to a DataType is dangerous if -2 is ever an unsigned integer. I think we should probably template guard graph from permitting that use case though

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -107,6 +140,27 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<T>
...
+    this->output_->InlineAdd(reg_output);
```
baykaner: multiply and Add with iterators in a loop to avoid iterating over tensor twice

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -107,6 +140,27 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<T>
...
+  }
+
+  virtual void L2Regularisation(DataType regularisation_rate)
+  {
+    ArrayType weight     = *this->output_;
+    ArrayType reg_output = ArrayType(weight.shape());
+
+    fetch::math::Abs(weight, reg_output);
+    fetch::math::Divide(weight, reg_output, reg_output);
```
baykaner: same here - this iterates over the tensor three times.

baykaner (changes requested):
would be good to implement a test specifically to numerically verify the weight decay

baykaner (approved):

PR #1294, opened 2019-07-01
Re-enable unreliable tests

wkaluza:


n-hutton:
I think we need to re-enable these flaky tests only (and as a priority) if there is a corresponding fix to the flakyness - I also just noticed that ```thread_pool_gtest``` timed out on another PR

wkaluza:
Yes.

n-hutton:
Looks like that makes it containers, network, muddle and thread_pool. Maybe we should divvy them up in the next sprint (?)

wkaluza:
Close temporarily to reduce load on CI

PR #1303, opened 2019-07-02
Fix issue with DAG work ordering

ejfitzgerald:


libs/ledger/tests/unit/work_queue_tests.cpp
```diff
@@ -0,0 +1,125 @@
...
+    work_queue_->pop();
+
+    EXPECT_EQ(item->score(), 500);
+  }
+}
+
+TEST_F(WorkQueueTests, CheckOrderingWhenSameScore)
+{
+  auto item1 = CreateWork(2000, 12);
+  auto item2 = CreateWork(2000, 12321321);
```
wkaluza: Could you add `item3` with a nonce of, say, `1234`, then add the items to the queue in order `item1`, then `item2`, then `item3`? The previous test is nice and robust because it has a few items to work with, but here the sorting is not obvious, and the queue looks like a simple FILO.

wkaluza (approved):

wkaluza (changes requested):
Misclicked, see inline comment.

wkaluza (approved):

PR #1307, opened 2019-07-03
Etch Language Tests

ejfitzgerald:
Adding in initial set of language tests for Etch. To avoid issues in the future. These are similar to the end 2 end tests but using the `etch` executable.

In order to start laying the foundation for Etch based unit tests I also added the `assert(condition)` and `assert(condition, message)` free functions.


scripts/run-language-tests.py
```diff
@@ -0,0 +1,82 @@
...
+    parser.add_argument('build_folder', type=os.path.abspath,
+                        help='The path to the build folder')
+    return parser.parse_args()
+
+
+def main():
+    args = parse_commandline()
+
+    etch_cli = os.path.join(args.build_folder, 'apps', 'etch', 'etch')
+    assert os.path.isfile(etch_cli)
```
wkaluza:
```suggestion
    assert os.path.isfile(etch_cli), 'etch executable not found at {}'.format(etch_cli)
```

Jenkinsfile
```diff
@@ -97,6 +97,10 @@ def fast_tests_stage(Platform platform, Configuration config)
     stage("Unit Tests ${stage_name_suffix(platform, config)}") {
       sh "./scripts/ci-tool.py -T ${config.label}"
     }
+
+    stage("Etch Lang Tests ${stage_name_suffix(platform, config)}") {
```
wkaluza, line 101: This will make the tests run on all platforms including Mac, and on all parallel PR merge builds. I'm not opposed; just checking this is what you were going for.

scripts/run-language-tests.py
```diff
@@ -0,0 +1,82 @@
...
+    print()
+    print('=== ETCH Language Tests ===')
+    print()
+
+    # loop through and execute each one of them
+    failures = []
+    for name, path in sorted(tests, key=lambda x: x[0]):
+
+        # create a temp file and call the executable inside it
+        output_file, output_filename = tempfile.mkstemp(
```
wkaluza, line 47: I'd use `TemporaryFile()` context manager to improve exception safety. https://docs.python.org/3.4/library/tempfile.html#tempfile.TemporaryFile . Then you would not need the call to `close()` and `remove()` below.

wkaluza (changes requested):
Nice one!

Jenkinsfile
ejfitzgerald: That was sort of what I was expected. I didn't really consider the Mac. They are very short set of tests to I tend to think let's keep them and then move them if they start to take a long time.

scripts/run-language-tests.py
ejfitzgerald: `TemporaryFile` is racey however (and I have been bitten in the past). This obv. would be fine in this case but I had planned for this code to be parallelised in the future so might be more useful.

wkaluza: The docs in the link state that it also uses mkstemp under the hood. Why is doing it this way less racey?

wkaluza (approved):

PR #1308, opened 2019-07-04
Miscellaneous changes

wkaluza:


apps/constellation/CMakeLists.txt
```diff
@@ -6,14 +6,6 @@ project(fetch-constellation)
 
 setup_compiler()
 
-add_executable(constellation
```
ejfitzgerald: Adding the header files are useful for CLion

libs/http/include/http/json_client.hpp
```diff
@@ -89,86 +90,5 @@ class JsonClient
   ClientPtr client_;
 };
 
-/**
```
ejfitzgerald: Haha isn't this the second time you have proposed this change. I would simply keep these in since they are effectively function aliases

libs/telemetry/include/telemetry/registry.hpp
```diff
@@ -74,6 +72,10 @@ class Registry
   using Mutex          = std::mutex;
   using LockGuard      = std::lock_guard<std::mutex>;
 
+  // Construction / Destruction
```
ejfitzgerald, line 75: As we just talked about - I would revert this particular change

libs/vm/include/vm/state.hpp
```diff
@@ -28,8 +27,8 @@ namespace vm {
 class IState : public Object
 {
 public:
-  IState()          = delete;
-  virtual ~IState() = default;
+  IState()           = delete;
+  ~IState() override = default;
```
ejfitzgerald, line 31: 👍

ejfitzgerald (changes requested):

apps/constellation/CMakeLists.txt
wkaluza: ok

libs/http/include/http/json_client.hpp
wkaluza: That's very possible. These things do bug me.

libs/telemetry/include/telemetry/registry.hpp
wkaluza: ok

ejfitzgerald (approved):

PR #1309, opened 2019-07-04
Correcting names for telemetry values

ejfitzgerald:


libs/core/include/core/string/ends_with.hpp
```diff
@@ -0,0 +1,38 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include <string>
```
wkaluza, line 20: Let's put the implementation in a cpp file. It worries me how much executable code we put into headers in the name of performance on no evidence, or just for convenience. Our rebuild times will thank us if we kick the habit.

libs/core/tests/unit/string_tests.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include "core/string/ends_with.hpp"
+
+#include "gtest/gtest.h"
+
+namespace {
+
+TEST(StringTests, CheckEndsWith)
+{
+  EXPECT_TRUE(fetch::core::EndsWith("Hello World", "World"));
```
wkaluza, line 28:
```suggestion
  EXPECT_TRUE(fetch::core::EndsWith("Hello World", "Hello World"));
  EXPECT_TRUE(fetch::core::EndsWith("Hello World", "World"));
```

libs/core/tests/unit/string_tests.cpp
```diff
@@ -0,0 +1,31 @@
...
+  EXPECT_FALSE(fetch::core::EndsWith("Hello World", "World2"));
```
wkaluza, line 29:
```suggestion
  EXPECT_FALSE(fetch::core::EndsWith("Hello World", "World2"));
  EXPECT_FALSE(fetch::core::EndsWith("Hello World", ""));
  EXPECT_FALSE(fetch::core::EndsWith("Hello World", "o"));
  EXPECT_FALSE(fetch::core::EndsWith("Hello World", "Hello"));
```

wkaluza (changes requested):

libs/core/include/core/string/ends_with.hpp
ejfitzgerald: In general, I would agree, however in this case it is only used in the one file and we want to promote the inlining.

wkaluza: Do we, though? How much time do we expect ledger to spend on constructing telemetry variables that an inlined function will make a difference?

wkaluza (approved):

PR #1313, opened 2019-07-06
Ensure naming consistency of Math/ML library

ejfitzgerald:
A small change to ensure that the Math / ML library is using camelCase for method names.

The rest of the updates are fallout from these changes. I have updated the examples but might have missed some bits, so please let me know.

In addition added some new language tests


libs/vm_modules/examples/commodity_prediction/script/commodity_prediction.etch
```diff
@@ -6,16 +6,16 @@ var first_dense_size = 216;
...
-graph.AddDropout("dropout_0", "num_input", dropout_prob);
-graph.AddFullyConnected("hidden_dense_1", "dropout_0", input_size, first_dense_size);
-graph.AddSoftmax("softmax_1", "hidden_dense_1");
-graph.AddDropout("dropout_1", "softmax_1", dropout_prob);
-graph.AddFullyConnected("hidden_dense_2", "dropout_1", first_dense_size, second_dense_size);
-graph.AddSoftmax("softmax_2", "hidden_dense_2");
-graph.AddDropout("dropout_2", "softmax_2", dropout_prob);
-graph.AddFullyConnected("output_dense", "dropout_2", second_dense_size, output_size);
-graph.AddSoftmax("softmax_3", "output_dense");
+graph.addPlaceholder("num_input");
```
baykaner, line 9: where is the etch style guide that indicates this is preferred?

scripts/etch-language-tests/maths/trig.etch
```diff
@@ -0,0 +1,112 @@
+function main()
```
baykaner, line 1: a variation on this already exists as an example. I recommend we delete the example since we don't need it if we have this test.

scripts/etch-language-tests/maths/abs.etch
```diff
@@ -0,0 +1,23 @@
+function main()
```
baykaner, line 1: there are already several math tests in vm_modules/tests/unit/math.cpp (though not as extensive as this one) - so what is the correct place for tests? are we moving them all to scripts/etch-language-tests?

libs/vm_modules/examples/commodity_prediction/script/commodity_prediction.etch
troels-im: The style guide does not exist atm.

baykaner (approved):

PR #1316, opened 2019-07-07
Refactor hashers

wkaluza:
This is slightly bulkier than usual. Highlights include:

* SHA256 hash no longer uses OpenSSL internals, but the recommended high level EVP digest functions.
* SHA256 and other OpenSSL digests brought under unified wrapper
* Code using OpenSSL digests no longer ends up indirectly including OpenSSL headers
* StreamHasher renamed to HasherInterface
* Crypto tests moved to `unit` directory
* Unit tests for digests added and rearranged - FNV hasher is now tested together with other digests, with a separate suite fnv_detail testing the algo.


libs/crypto/include/crypto/openssl_hasher.hpp
```diff
@@ -0,0 +1,67 @@
...
+public:
+  explicit OpenSslHasher(OpenSslDigestType);
+  ~OpenSslHasher();
+  OpenSslHasher(OpenSslHasher const &) = delete;
+  OpenSslHasher(OpenSslHasher &&)      = delete;
+
+  OpenSslHasher &operator=(OpenSslHasher const &) = delete;
+  OpenSslHasher &operator=(OpenSslHasher &&) = delete;
+
+  bool        reset();
```
ejfitzgerald: Style 🚓 - PascalCase for function names

libs/crypto/include/crypto/openssl_hasher.hpp
```diff
@@ -0,0 +1,67 @@
...
+  bool        update(uint8_t const *data_to_hash, std::size_t size);
+  bool        final(uint8_t *data_to_hash);
+  std::size_t hash_size() const;
+
+private:
+  internal::OpenSslHasherImpl *impl_;
```
ejfitzgerald: `std::shared_ptr` for this? (instead of unique pointer so that you can still maintain things being decoupled)

libs/crypto/src/openssl_hasher.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+OpenSslHasher::OpenSslHasher(OpenSslDigestType const type)
+  : impl_(new OpenSslHasherImpl(type))
+{
+  reset();
+}
+
+OpenSslHasher::~OpenSslHasher()
+{
+  delete impl_;
```
ejfitzgerald: See shared_pointer comment

libs/network/src/muddle/router.cpp
```diff
@@ -73,7 +73,11 @@ std::size_t GenerateEchoId(Packet const &packet)
   hash.Update(reinterpret_cast<uint8_t const *>(&channel), sizeof(channel));
   hash.Update(reinterpret_cast<uint8_t const *>(&counter), sizeof(counter));
 
-  return hash.Final<std::size_t>();
+  auto const res = hash.Final();
+
+  std::size_t const out = *reinterpret_cast<std::size_t const *>(res.pointer());
```
ejfitzgerald: This is an odditiy of FNV. Since we have the concrete implementation on hand can we refactor this into new method

ejfitzgerald (commented):
Looks really good. Just some minor niggles

libs/crypto/include/crypto/openssl_hasher.hpp
wkaluza: Done.

libs/network/src/muddle/router.cpp
wkaluza: :+1:

libs/crypto/include/crypto/openssl_hasher.hpp
wkaluza: can do. if the extra cost is a problem, we can always go back to raw pointers

libs/crypto/src/openssl_hasher.cpp
wkaluza: :+1:

libs/crypto/include/crypto/fnv.hpp
```diff
@@ -54,9 +55,11 @@ class FNV : public HasherInterface
   std::size_t HashSizeInBytes() const override;
 
 private:
-  internal::FnvHasherInternals *impl_;
+  std::shared_ptr<internal::FnvHasherInternals> impl_;
 };
 
+inline FNV::~FNV() = default;
```
ejfitzgerald: This can be moved to the main body of the header surely

libs/network/src/muddle/router.cpp
```diff
@@ -73,9 +73,11 @@ std::size_t GenerateEchoId(Packet const &packet)
   hash.Update(reinterpret_cast<uint8_t const *>(&channel), sizeof(channel));
   hash.Update(reinterpret_cast<uint8_t const *>(&counter), sizeof(counter));
 
-  auto const res = hash.Final();
+  std::size_t out = 0;
 
-  std::size_t const out = *reinterpret_cast<std::size_t const *>(res.pointer());
+  static_assert(sizeof(decltype(out)) == decltype(hash)::size_in_bytes,
```
ejfitzgerald: Pedantic: think it would read easier like: `static_assert(sizeof(out) == hash.size_in_bytes, "<message>");`

libs/crypto/include/crypto/openssl_hasher.hpp
```diff
@@ -53,15 +53,17 @@ class OpenSslHasher
...
+  bool        Update(uint8_t const *data_to_hash, std::size_t size);
+  bool        Final(uint8_t *data_to_hash);
+  std::size_t HashSize() const;
 
 private:
-  internal::OpenSslHasherImpl *impl_;
+  std::shared_ptr<internal::OpenSslHasherImpl> impl_;
 };
 
+inline OpenSslHasher::~OpenSslHasher() = default;
```
ejfitzgerald: Again think this can be moved into the declaration.

ejfitzgerald (changes requested):

libs/network/src/muddle/router.cpp
wkaluza: I think clang-tidy may object to accessing static members through an instance, but I'll give it a go

libs/crypto/include/crypto/openssl_hasher.hpp
wkaluza: :+1:

libs/crypto/include/crypto/fnv.hpp
wkaluza: :+1:

ejfitzgerald (approved):

PR #1317, opened 2019-07-07
Improved telemetry + performance fixes

ejfitzgerald:
* Updated the telemetry library with `Histogram`, `HistogramMap` and `CounterMap` primitives
* Sticky plaster over `MerkleTreeBlock` limitations
* Add extra checking on `TransactionBuilder` to catch common mistakes
* Fix issue with execution manager generating fatal errors in non-fatal occasions
* Add tool for generation of bulk transactions + reinstate that API
* Fix race condition in Muddle RPC Client (rework in the pipeline)
* HTTP Server Telemetry (counts and request durations)


apps/tx-generator/main.cpp
```diff
@@ -0,0 +1,165 @@
...
+    // compute the range of data to be populated
+    std::size_t const start = num_tx_per_thread * i;
+    std::size_t const end   = std::min(count, start + num_tx_per_thread);
+
+    pool.Dispatch([start, end, &encoded_txs]() {
+      static constexpr std::size_t LOG2_VECTOR_SIZE = 7u;
+      static constexpr std::size_t VECTOR_SIZE      = 1u << LOG2_VECTOR_SIZE;
+
+      // initialise the RNG
+      std::random_device rd{};
```
wkaluza: I think it would be more correct to move `rd` and `rng` out of the loop and have one (suitably locked) generator for all threads. Otherwise there's potential for collisions.

Is `std::minstd_rand` reproducible, or would we have to seed it?

libs/ledger/include/ledger/storage_unit/storage_unit_client.hpp
```diff
@@ -38,6 +38,7 @@
 #include "storage/document_store_protocol.hpp"
 #include "storage/object_store_protocol.hpp"
 
+#include <array>
```
wkaluza, line 41: This file doesn't seem to use `std:array`. You could include `<cassert>` and also `<cstring>` for `std::memcpy`.

libs/ledger/include/ledger/transaction_verifier.hpp
```diff
@@ -37,12 +38,8 @@ class TransactionVerifier
...
 
   // Construction / Destruction
-  explicit TransactionVerifier(TransactionSink &sink, std::size_t verifying_threads,
-                               std::string name)
-    : verifying_threads_(verifying_threads)
-    , name_(std::move(name))
-    , sink_(sink)
-  {}
+  TransactionVerifier(TransactionSink &sink, std::size_t verifying_threads,
+                      std::string const &name);
```
wkaluza, line 42: :+1::+1::+1::+1::+1:

libs/ledger/src/chain/transaction_builder.cpp
```diff
@@ -52,6 +52,12 @@ TransactionBuilder::Sealer::Sealer(TransactionPtr tx)
     }
   }
 
+  // ensure there is a from field
+  if (partial_transaction_->from_.empty())
+  {
+    throw std::runtime_error("Malformed transcation, missing transaction");
```
wkaluza:
```suggestion
    throw std::runtime_error("Malformed transaction, missing 'from' field");
```

libs/ledger/src/transaction_verifier.cpp
```diff
@@ -31,6 +35,77 @@ static const std::chrono::milliseconds POP_TIMEOUT{300};
...
+
+  // ensure there are no invalid characterss
+  std::transform(metric_name.begin(), metric_name.end(), metric_name.begin(), [](char c) -> char {
+    if (c == '-')
+    {
+      return '_';
+    }
+    else
+    {
+      return static_cast<char>(std::tolower(c));
```
wkaluza, line 54: Include a check using `std::isalnum` to discard non-alphanumerics?

libs/ledger/src/transaction_verifier.cpp
```diff
@@ -31,6 +35,77 @@ static const std::chrono::milliseconds POP_TIMEOUT{300};
...
+  });
+
+  return metric_name;
+}
+
+telemetry::GaugePtr<uint64_t> CreateGauge(std::string const &prefix, std::string const &name,
+                                          std::string const &description)
+{
+  std::string const metric_name = CreateMetricName(prefix, name);
+  return Registry::Instance().CreateGauge<uint64_t>(metric_name, description);
```
wkaluza: Remove `const` and `std::move(metric_name)`?

libs/ledger/src/transaction_verifier.cpp
```diff
@@ -31,6 +35,77 @@ static const std::chrono::milliseconds POP_TIMEOUT{300};
...
+}
+
+telemetry::CounterPtr CreateCounter(std::string const &prefix, std::string const &name,
+                                    std::string const &description)
+{
+  std::string const metric_name = CreateMetricName(prefix, name);
+  return Registry::Instance().CreateCounter(metric_name, description);
```
wkaluza: Ditto

libs/network/include/network/muddle/rpc/client.hpp
```diff
@@ -91,8 +90,10 @@ class Client : public service::ServiceClientInterface
 
   SharedHandler handler_;
 
+  std::mutex call_mutex_;  // priority 1
```
wkaluza, line 93: What do the priorities mean in practice?

libs/telemetry/src/histogram.cpp
```diff
@@ -0,0 +1,119 @@
...
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "telemetry/histogram.hpp"
+
+#include <iostream>
```
wkaluza: I think `<ostream>` will do here.

wkaluza (changes requested):

libs/network/include/network/muddle/rpc/client.hpp
ejfitzgerald: The order in which the mutexs are locked in order not to deadlock

libs/ledger/src/transaction_verifier.cpp
ejfitzgerald: Discarding of the characters would require a little more effort here. Ultimately this is more of a connivence thing than a necessity so I am inclined to leave it as is at the moment.

apps/tx-generator/main.cpp
ejfitzgerald: `std::minstd_rand` is an LCG. 

This RNG is redundant anyway so have removed.

libs/ledger/src/transaction_verifier.cpp
wkaluza: sure

libs/ledger/include/ledger/storage_unit/storage_unit_client.hpp
ejfitzgerald: `std::array` is used in the `MerkleTreeBlock` (this will hopefully be updated soon). I have added those header files

wkaluza (approved):

PR #1319, opened 2019-07-08
Embeddings bugfix

baykaner:
@XiaodongWuFetch found bug in not reshaping according to updated batch size for embeddings


evsmithx (approved):
LGTM

PR #1322, opened 2019-07-09
Miscellaneous changes

wkaluza:
Early drop this week - they're stacking up.


apps/etch/main.cpp
```diff
@@ -23,6 +23,7 @@
 #include "core/commandline/parameter_parser.hpp"
 #include "core/json/document.hpp"
 #include "core/serializers/byte_array.hpp"
+#include "fetch_version.hpp"
```
ejfitzgerald: I was trying to keep this separate to try and indicate that it is a little special because it be generated everytime by cmake. What are your thoughts on this convention?

wkaluza: I don't think it matters where a header comes from. If this one is absent or malformed because something in the generation bugged out, the compilation would fail; in that sense it's like every other header under the sun, not special at all. Leaving a blank line between it and the other includes in client code would not help in that situation, so why do it?

What actually helps with generated headers like this one is some extra information readily accessible to someone who encountered a problem related to them being generated (as opposed to checked in). One way to achieve this would be to have two headers: a checked in header `fetch_version.hpp` used by client code, and a generated `fetch_version_detail.hpp`. The contents of `fetch_version.hpp` could read as follows:

```
#pragma once, copyright, etc.

// fetch_version_detail.hpp is a generated header. To generate it,
// run Foo, for extra information see Bar. A valid fetch_version_detail.hpp
// file should define the following:
//
// namespace fetch {
// namespace version {
// 
// constexpr char const *FULL        = "v0.6.0-alpha13-7-gc19889bb";
// constexpr uint32_t    MAJOR       = 0;
// constexpr uint32_t    MINOR       = 6;
// constexpr uint32_t    PATCH       = 0;
// constexpr char const *COMMIT_HASH = "";
// constexpr bool        VALID       = true;
// 
// } // namespace version
// } // namespace fetch

#include "path/to/fetch_version_detail.hpp"

```

This indirection would allow anyone who encounters a problem to find out what to do, because `fetch_version.hpp` would be checked in (implicitly: always present) and would contain guidance. Users could even hack out their own stub `fetch_version_detail.hpp` file based on this info if they can't generate it or can't be bothered to try.

ejfitzgerald:
Is GitHub being weird but the generated diff seems to include contents I have added. Can you double check this.

wkaluza:
> Is GitHub being weird but the generated diff seems to include contents I have added. Can you double check this.

I see nothing that should not be there; could you point me to a specific file to show what you mean?

ejfitzgerald (approved):
Think it must have been a temporary github issue

PR #1325, opened 2019-07-09
Tweak test infrastructure

wkaluza:
* Move installation of test dependencies out of Jenkinsfile and into ci-tool in -E mode.
* Install pyyaml and wheel as part of dependencies installation (make it easier to set up a new venv)
* Move etchlang tests to earlier in the build pipeline
* Replace call to `yaml.load_all` with `safe_load_all` to silence deprecation warning

### DIFF
--- a/scripts/ci/install-test-dependencies.sh
+++ b/scripts/ci/install-test-dependencies.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
 set -e
 
-# install the latest version of the ledger api
-pip3 install -e git+https://github.com/fetchai/ledger-api-py.git@master#egg=fetchai-ledger-api
+pip3 install wheel
 
-# install the latest version of the net utils
+pip3 install fetchai-ledger-api==0.5.1
 pip3 install -i https://test.pypi.org/simple/ fetchai-netutils==0.0.5a1
+
+pip3 install pyyaml
--- a/scripts/end_to_end_test/run_end_to_end_test.py
+++ b/scripts/end_to_end_test/run_end_to_end_test.py
@@ -520,7 +520,7 @@ def run_test(build_directory, yaml_file, constellation_exe):
     # Read YAML file
     with open(yaml_file, 'r') as stream:
         try:
-            all_yaml = yaml.load_all(stream)
+            all_yaml = yaml.safe_load_all(stream)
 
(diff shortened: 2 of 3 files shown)


scripts/ci/install-test-dependencies.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
 set -e
 
-# install the latest version of the ledger api
-pip3 install -e git+https://github.com/fetchai/ledger-api-py.git@master#egg=fetchai-ledger-api
+pip3 install wheel
 
-# install the latest version of the net utils
+pip3 install -e git+https://github.com/fetchai/ledger-api-py.git@master#egg=fetchai-ledger-api
```
ejfitzgerald: Can we update this to the pypi package please

scripts/end_to_end_test/run_end_to_end_test.py
```diff
@@ -520,7 +520,7 @@ def run_test(build_directory, yaml_file, constellation_exe):
     # Read YAML file
     with open(yaml_file, 'r') as stream:
         try:
-            all_yaml = yaml.load_all(stream)
+            all_yaml = yaml.safe_load_all(stream)
```
ejfitzgerald, line 523: 👍 👍 👍

ejfitzgerald (commented):
Not 100% comfortable adding the dependency injection into the ci-tool script. I think it would be nicer if the script attempted to import those packages and maybe post a warning. What do you think?

wkaluza:
> Not 100% comfortable adding the dependency injection into the ci-tool script. I think it would be nicer if the script attempted to import those packages and maybe post a warning. What do you think?

It's simplifying the setup of a fresh Python virtual environment that I was after, as currently it requires multiple steps. With the injection, there is no setup at all (simply running the e2e tests ensures that dependencies are installed). Without the injection, it's still only one step, which is a big improvement. I care about the changes to `install-test-dependencies.sh` much more than about running it from `ci-tool.py`, which was more of an afterthought. I can remove the installation from `ci-tool` if you'd rather the two to stay separate.

I'm not keen on having a check for the dependencies because it's duplication: the checks in `ci-tool` would need to be kept in sync with `install-test-dependencies.sh` to be useful.

ejfitzgerald (approved):

PR #1323, opened 2019-07-09
Added tests and benchmarks

MissingNO57:

PR #1328, opened 2019-07-09
Benchmarks for ops

MissingNO57:
Can be merged in after #1323


libs/ml/benchmark/ops/ops.cpp
```diff
@@ -153,4 +160,319 @@ BENCHMARK_TEMPLATE(BM_SqrtForward, double, 1024)->Unit(benchmark::kMicrosecond);
...
+  inputs.push_back(input);
+  fetch::ml::ops::Sqrt<fetch::math::Tensor<T>> sqrt1;
+
+  for (auto _ : state)
+  {
+    sqrt1.Backward(inputs, error_signal);
+  }
+}
+
+BENCHMARK_TEMPLATE(BM_SqrtBackward, double, 2)->Unit(benchmark::kNanosecond);
```
baykaner, line 240: Please benchmark for double, fp32_t, and fp_64_t

libs/ml/benchmark/ops/ops.cpp
```diff
@@ -153,4 +160,319 @@ BENCHMARK_TEMPLATE(BM_SqrtForward, double, 1024)->Unit(benchmark::kMicrosecond);
...
+BENCHMARK_TEMPLATE(BM_ExpBackward, double, 256)->Unit(benchmark::kMicrosecond);
+BENCHMARK_TEMPLATE(BM_ExpBackward, double, 512)->Unit(benchmark::kMicrosecond);
+BENCHMARK_TEMPLATE(BM_ExpBackward, double, 1024)->Unit(benchmark::kMicrosecond);
+BENCHMARK_TEMPLATE(BM_ExpBackward, double, 2048)->Unit(benchmark::kMicrosecond);
+BENCHMARK_TEMPLATE(BM_ExpBackward, double, 4096)->Unit(benchmark::kMicrosecond);
+
+template <class T, int N>
+void BM_DivideForward(benchmark::State &state)
+{
+  fetch::math::Tensor<T> input_1({1, N});
```
baykaner, line 462: by default tensors are filled with 0s; so this benchmark will compute 0/0
I wonder if that will be faster than performing a real division?

I suggest for all of these ops that you create or fill all the tensors with Tensor<T>::UniformRandom

baykaner (changes requested):

baykaner (approved):

PR #1330, opened 2019-07-10
updated ML docs inheritence tree to match recent changes

baykaner:


evsmithx (changes requested):
There's no longer an arrow pointing from Subgraph to Graph, but apart from that it look good.

evsmithx (approved):
LGTM

PR #1333, opened 2019-07-10
Quick hack to add support for signees in etch interpreter

pbukva:
The address can be provided as command-line parameter  in display format  (Base58 encoded), and it can be **prefixed** with the `signee:` **non**-mandatory prefix **IF** it is supposed to represent signee (= `Address.signedTx()` will return `true`), see etch example bellow.

Essential: In order to get instance of `Address` type in the Etch code from such input parameter, it is necessary to use the `System.ArgvAsAddr(...)` method **instead** of ordinary `System.Argv(...)` method.

Here is Etch code demonstrating usage (save it as `etch_signee_address_test.etch` file):
```
function main()
    var addr = System.ArgvAsAddr(1);
    printLn("addr: " + addr.display() + ", is signee: " + toString(addr.signedTx()));
endfunction
```

Here is how to pass the address as command-line parameter to etch interpreter:
 * As  **NOT** signee: `./apps/etch/etch etch_signee_address_test.etch -- yJtFTWgacYmR1HKqBTfWAY3rKSSRXfYCkuk8GZR6aicLLr6UR`

 * As **SIGNEE** (please note presence of the **signee:** prefix in front of address value): `./apps/etch/etch etch_signee_address_test.etch -- signee:yJtFTWgacYmR1HKqBTfWAY3rKSSRXfYCkuk8GZR6aicLLr6UR`


libs/vm/src/module.cpp
```diff
@@ -242,7 +242,8 @@ Module::Module()
       .CreateSerializeDefaultConstuctor<>()
       //      .CreateConstuctor<>()
       .CreateConstuctor<Ptr<String>>()
-      .CreateMemberFunction("signedTx", &Address::HasSignedTx);
+      .CreateMemberFunction("signedTx", &Address::HasSignedTx)
+      .CreateMemberFunction("display", &Address::AsString);
```
pbukva, line 246: This change is not necessary, but useful, I was thinking about to add new standalone  `toString(...)` function for address. but this was way simpler and it aligns with what we have in native code `ledger::Address` type.

ejfitzgerald (changes requested):
I am not sure this is the right approach. Lets catchup with @n-hutton and we can work out a more complete solution for the simulation of the etch contract lifecycle

ejfitzgerald:
Closing because now stale

PR #1335, opened 2019-07-10
Create fetch-version library

wkaluza:


apps/etch/main.cpp
```diff
@@ -303,7 +302,7 @@ int main(int argc, char **argv)
   }
 
   // print the header
-  fetch::commandline::DisplayCLIHeader("vm-lang");
+  fetch::version::DisplayCLIHeader("vm-lang");
```
ejfitzgerald:
```suggestion
  fetch::version::DisplayCLIHeader("etch");
```

ejfitzgerald (changes requested):
Just a minor request

apps/etch/main.cpp
wkaluza: Good spot!

ejfitzgerald (approved):
LGTM

PR #1337, opened 2019-07-11
remove obsolete docs

n-hutton:


baykaner (changes requested):
these are not obselete!

PR #1338, opened 2019-07-11
scale back upper bound for operations to main chain

### LINKED ISSUE #976: Clients are able to request the full block chain (crashed alpha net)
```MAX_SUB_CHAIN_SIZE```(main_cahin_rpc_service) - there should be guards on the server to disallow asking for the whole block chain/too many blocks.

### PR DESCRIPTION
**Pull request by n-hutton**
Closes #976 

What I found was that there are limits on what a client can call on a node. It is currently quite high but will not break if this limit is low enough to come into force. I have stopped here because @bipll is going to make some changes to the main chain and main chain rpc that allow syncing forwards instead of backwords on startup which would conflict with any changes I might make.

### DIFF
--- a/libs/ledger/include/ledger/protocols/main_chain_rpc_protocol.hpp
+++ b/libs/ledger/include/ledger/protocols/main_chain_rpc_protocol.hpp
@@ -33,7 +33,6 @@ public:
   enum
   {
     HEAVIEST_CHAIN   = 1,
-    CHAIN_PRECEDING  = 2,
     COMMON_SUB_CHAIN = 3
   };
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: libs/ledger/include/ledger/protocols/main_chain_rpc_protocol.hpp [hunk 1]
```diff
@@ -33,15 +33,13 @@ class MainChainProtocol : public service::Protocol
   enum
   {
     HEAVIEST_CHAIN   = 1,
-    CHAIN_PRECEDING  = 2,
```
n-hutton, line 36: dead code

scripts/end_to_end_test/run_end_to_end_test.py
```diff
@@ -153,6 +154,9 @@ def append_node(self, index, load_directory=None):
             clear_path=False
         )
 
+        # Possibly soon to be depreciated functionality - set the block interval
+        instance._block_interval = self._block_interval
```
ejfitzgerald, line 158: Does this mean all test instances are now mining?

n-hutton: no - instance._mining = True would set that

ejfitzgerald (approved):

PR #1339, opened 2019-07-11
Configuration Overhaul

ejfitzgerald:
The main focus of this PR is to clean up the `main.cpp` of Constellation which has grown to be a bit out of control.

I have introduced a new `Setting` class which is a new abstraction of a command line parameter and a potential environment variable. In the future, this can also be mapped to a simple configuration structure.


apps/constellation/key_generator.cpp
```diff
@@ -0,0 +1,108 @@
...
+  bool      certificate_loaded = false;
+
+  // Step 1. Attempt to load the existing key
+  {
+    std::ifstream input_file(key_path, std::ios::in | std::ios::binary);
+
+    if (input_file.is_open())
+    {
+      fetch::byte_array::ByteArray private_key_data;
+      private_key_data.Resize(ECDSASigner::PrivateKey::ecdsa_curve_type::privateKeySize);
```
wkaluza, line 71: I think ByteArray has a ctor which takes a size_t argument and calls resize for you

apps/constellation/key_generator.cpp
```diff
@@ -0,0 +1,108 @@
...
+namespace fetch {
+
+using fetch::crypto::Prover;
+using fetch::crypto::ECDSASigner;
+
+using SignerPtr = std::shared_ptr<ECDSASigner>;
+using ProverPtr = std::shared_ptr<Prover>;
+
+char const *LOGGING_NAME         = "KeyGenerator";
+char const *DEFAULT_KEY_FILENAME = "p2p.key";
```
wkaluza, line 33: Probably meaningless, but these may be constexpr

apps/constellation/manifest_builder.cpp
```diff
@@ -0,0 +1,141 @@
...
+ * Create the manifest from an environment variable based configuration
+ *
+ * @param manifest The manifest to be populated
+ * @return true if successful, otherwise false
+ */
+bool LoadManifestFromEnvironment(Manifest &manifest)
+{
+  bool success{false};
+
+  char const *manifest_data = std::getenv("CONSTELLATION_MANIFEST");
```
wkaluza, line 102: rather than calling std::getenv directly, I'd like to see a thin wrapper around it which we own and can mock out. Aids testability. Rather than a string, the wrapper could take an enum class argument like ConstellationEnvVar, and would look up the right env var string name based on the arg value. This would also have the benefit of putting all env vars constellation depends on in one place - we only have a couple for now, but it's easy to imagine their number growing over time, and the code getting peppered with magic strings like this one.

apps/constellation/settings.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+  /// @name Proof of Stake
+  /// @{
+  settings::Setting<bool>        proof_of_stake;
+  settings::Setting<std::string> beacon_address;
+  /// @}
+
+  // Operators
+  Settings &operator=(Settings const &) = delete;
+  Settings &operator=(Settings &&) = delete;
```
wkaluza, line 112: Hissss....

apps/constellation/settings.hpp
```diff
@@ -0,0 +1,120 @@
...
+#include <cstdint>
+#include <string>
+#include <vector>
+
+namespace fetch {
+
+/**
+ * Command line / Environment variable settings
+ */
+class Settings : protected settings::SettingCollection
```
wkaluza: inheritance may be private?

libs/settings/include/settings/setting_collection.hpp
```diff
@@ -0,0 +1,57 @@
...
+class SettingBase;
+
+/**
+ * A very simple collection of pointers to settings object. The use must make sure that the
+ * lifetime of the settings is longer than this container
+ */
+class SettingCollection
+{
+public:
+  using Settings = std::vector<SettingBase *>;
```
wkaluza, line 36: may this be a vector of unique_ptrs? Lifetime concerns would go away if so.

libs/settings/include/settings/detail/setting_list.hpp
```diff
@@ -0,0 +1,111 @@
+#pragma once
```
wkaluza: this file is usable more generally; namespacing it as a detail of settings may need a rethink. Could it find a home together with the EndsWith string helper that you recently wrote?

Unrelatedly, setting_list.hpp as a name may be confusing, especially given that setting_collection.hpp exists and contains very different functionality.

libs/settings/tests/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+#
+# F E T C H   M I N E R   T E S T S
```
wkaluza: Copypasta error.

libs/settings/tests/unit/setting_collection_tests.cpp
```diff
@@ -0,0 +1,74 @@
...
+  {
+    return argv_.data();
+  }
+
+private:
+  std::vector<std::string> args_;
+  std::vector<char *>      argv_;
+};
+
+TEST(SettingCollectionTests, SimpleCheck)
```
wkaluza: This is nice, but while we're here, can we test stuff typical edge cases? Passing in no settings, passing in a misspelled option, passing in `xyz` for the number of lanes?

libs/settings/tests/unit/setting_collection_tests.cpp
```diff
@@ -0,0 +1,74 @@
...
+  Setting<uint32_t>    lanes{collection, "lanes", 0, ""};
+  Setting<std::string> name{collection, "name", "default", ""};
+
+  SystemArgAdapter arguments({"-lanes", "256", "-name", "foo-bar-baz"});
+
+  collection.UpdateFromArgs(arguments.argc(), arguments.argv());
+
+  EXPECT_EQ(lanes.value(), 256);
+  EXPECT_EQ(name.value(), "foo-bar-baz");
+}
```
wkaluza, line 178: I think we must test the interaction etween env variables and their corresponding config/cli options, and which take precedence when. The wrapper I suggested above may be a good first step.

wkaluza (changes requested):
Looks good, some niggles and suggested tests.

apps/constellation/key_generator.cpp
ejfitzgerald: Nope that is not the case (unfortunately)

apps/constellation/key_generator.cpp
ejfitzgerald: `DEFAULT_KEY_FILENAME` certainly can't because we need to be able to make a pointer to it! 😁

apps/constellation/settings.hpp
ejfitzgerald: ?

apps/constellation/settings.hpp
ejfitzgerald: Yep it could be in this case. `protected` seems a good idea in case we wanted to have derived configs in the future.

libs/settings/include/settings/setting_collection.hpp
ejfitzgerald: Specifically the design doesn't want to control the lifetime of these objects. The `unique_ptr` would change this quite drastically. The closest version compatible version would be to use the `shared_ptr` or to use `std::reference_wrapper`. I did initially look at the reference wrapper and decided that it was more effort than was required.

The basic design here is that everything is designed to be very static so having this pointer vector is fine IMHO.

libs/settings/include/settings/detail/setting_list.hpp
ejfitzgerald: Yes it might be useful. I would suggest that we find that use case first and then generalise it. I am happy for filename suggestions but this seems fine to me

libs/settings/tests/CMakeLists.txt
ejfitzgerald: done

apps/constellation/settings.hpp
wkaluza: What we talked about - assignment operators far away from constructors. :) I know that's the style, but I refuse to like it.

apps/constellation/settings.hpp
wkaluza: Let's make it private for now, until something actually derives from it.

libs/settings/include/settings/setting_collection.hpp
wkaluza: ok

apps/constellation/key_generator.cpp
wkaluza: Seems like it would be useful. Consider adding it as part of this change.

wkaluza: Seems like it would be useful. Consider adding it as part of this change?

libs/settings/include/settings/detail/setting_list.hpp
wkaluza: csv_string_helpers.hpp seems truthful enough

apps/constellation/key_generator.cpp
ejfitzgerald: Not for this change. It would be a change in of itself since this pattern of create instance and resize is prevalent in the code base

libs/settings/tests/unit/setting_collection_tests.cpp
ejfitzgerald: Done

apps/constellation/manifest_builder.cpp
ejfitzgerald: This is sort of a chicken and egg problem a little bit because it is hard to test the value of an environment variable. 

The enum approach would not work because the whole point of this setting change is to allow dynamically query based on the collection settings.

I will raise an issue to track this and I think that we can tackle this as another PR.

libs/settings/tests/unit/setting_collection_tests.cpp
ejfitzgerald: See comment above

apps/constellation/key_generator.cpp
wkaluza: Fair. Please create a Jira issue - if the pattern is this prevalent, it'll be a nice bit of refactoring to do at some point.

apps/constellation/manifest_builder.cpp
wkaluza: I think we misunderstand each other - I didn't mean testing values of env variables, but injecting envvariables in a unit test context to check that they override defaults and are overridden by cli arguments. Raise the issue - it's a change we can make on top of this after it's merged.

wkaluza (dismissed):

wkaluza (approved):

PR #1340, opened 2019-07-11
Distributed Random Beacon Base Functionality

troels-im:
This update creates the base functionality needed to generate and manage keys for the distributed random beacon.


libs/dkg/src/beacon_member.cpp
```diff
@@ -0,0 +1,160 @@
+//------------------------------------------------------------------------------
```
n-hutton: This file should be called beacon_manager.cpp

libs/dkg/src/beacon_member.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+  verification_vectors_.resize(cabinet_size);
+  received_shares_.resize(cabinet_size);
+
+  // Creating member id
+  // TODO(tfr): Make loadable
+  certificate_.reset(new ECDSASigner());
+  certificate_->GenerateKeys();
+  auto id = crypto::bls::PrivateKeyByCSPRNG();
+  id_.v   = id.v;
```
n-hutton: Is it possible to avoid single character variable names?

n-hutton: Ah I see this is actually from the BLS library

libs/dkg/include/dkg/beacon_manager.hpp
```diff
@@ -0,0 +1,189 @@
...
+    return id_;
+  }
+  /// }
+private:
+  uint64_t cabinet_size_{0};
+  uint32_t threshold_{0};
+
+  /// Member details
+  /// @{
+  std::unordered_map<Identity, uint64_t> identity_to_id_;
```
n-hutton: The usage of this is a bit confusing - the code has ```identity_to_id_[identity] = participants_.size();``` which seems odd/incorrect

libs/crypto/examples/dkg/main.cpp
```diff
@@ -57,16 +57,18 @@ int main()
   bls::Init();
 
   // Creating members from predefined seeds
-  std::vector<ConstByteArray> member_seeds = {"12122",    "454323", "547456", "54",
-                                              "23423423", "68565",  "56465"};
-  std::vector<Member>         members;
-  uint32_t                    threshold = 4;
+  std::vector<ConstByteArray> member_seeds = {
```
n-hutton, line 47: Test() is actually unused in this file

libs/dkg/examples/beacon_manager/main.cpp
```diff
@@ -0,0 +1,149 @@
...
+  {
+    auto &n = *m.second;
+    participants.push_back({n.identity(), n.id()});
+  }
+
+  // Propagating identities
+  for (auto &n : nodes)
+  {
+    // Shuffling to similate random arrival order
+    std::random_shuffle(participants.begin(), participants.end());
```
n-hutton: There is an issue using random_shuffle  without that breaks mac builds - it happens to @baykaner compiler. Do this instead:

```
    std::random_device rd;
    std::mt19937       g(rd());
    std::shuffle(participants.begin(), participants.end(), g);

```

n-hutton (changes requested):
Looks great! I wish you'd write code like this for me sometimes :)

Might be worth as a next step me making a unit test for bls/beacon_manager so the framework is there for Jon and Jenny

libs/dkg/src/beacon_member.cpp
troels-im: No, unfortunately I did not invent this.

n-hutton (approved):

PR #1343, opened 2019-07-11
Change Bloom filter feature flag semantics; add telemetry

wkaluza:


libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1482,39 +1493,52 @@ DigestSet MainChain::DetectDuplicateTransactions(BlockHash const &starting_hash,
...
-  }
+
+    return duplicates;
+  };
+
+  DigestSet const duplicates =
+      search_chain_for_duplicates(enable_bloom_filter_ ? potential_duplicates : transactions);
 
   auto const false_positives =
       static_cast<std::size_t>(potential_duplicates.size() - duplicates.size());
```
ejfitzgerald: I suspect this static_cast is redundant

ejfitzgerald (commented):
Looks good one minor niggle

libs/ledger/src/chain/main_chain.cpp
wkaluza: Quite right.

ejfitzgerald (approved):
LGTM

PR #1347, opened 2019-07-12
Feature/improve uint class

### LINKED ISSUE #1260: [VECTORISE] Potential loss of data in `UInt::operator<ArrayType> = (...)` operator
The `ArrayType` template type can represent any array, where return value type `T` of its it's indexing operator `T& ArrayType::operator[](T_IDX index)` can differ (byte-size wise) from return type of indexing operator of destination array `decltype(UInt<S>::data_[0])` what is currently `uint8_t`.

Thus solution to this issue should handle cases where `sizeof(T) > sizeof(decltype(UInt::data_[0]))`.

### LINKED ISSUE #1290: Improve Uint256 class
Improve the class for performance, add tests

markos:
Improvement to the UInt class. Now operations are done at the 64-bit level -though 128-bit native type in gcc/clang is used for 256-bit multiplication.

Most operators +,-,*,/,%,<<,>> are provided, incl. methods to calculate MSB/LSB of the object.

Tests are provided for most operators.

Closes: #1290. Closes: #1260 .


libs/vectorise/include/vectorise/uint/uint.hpp
```diff
@@ -39,338 +40,1012 @@ namespace vectorise {
...
+  constexpr std::size_t msb() const;
+  constexpr std::size_t lsb() const;
+
+  /////////////////////////
+  /// element accessors ///
+  /////////////////////////
+
+  constexpr uint8_t   operator[](std::size_t const &n) const;
+  constexpr uint8_t & operator[](std::size_t const &n);
+  constexpr WideType  elementAt(std::size_t const &n) const;
```
troels-im: Please our standard style `ElementAt` rather than `elementAt`. When export to Etch, we `elementAt` in the Etch module definition, but not in the C++ version.

libs/vectorise/include/vectorise/uint/uint.hpp
```diff
@@ -39,338 +40,1012 @@ namespace vectorise {
...
+constexpr bool UInt<S>::operator!=(UInt const &other) const
+{
+  return !(*this == other);
+}
+
+template <uint16_t S>
+constexpr bool UInt<S>::operator<(UInt const &other) const
+{
+  // Simplified version, as we're dealing with wider elements
+  for (std::size_t i = WIDE_ELEMENTS - 1; i >= 0; --i)
```
troels-im: `i` is unsigned. Will it not always be greater than or equal to zero?

libs/vectorise/include/vectorise/uint/uint.hpp
markos: fixed

libs/vectorise/include/vectorise/uint/uint.hpp
```diff
@@ -39,338 +40,1012 @@ namespace vectorise {
...
-    {
-      uint64_t value;
-      uint8_t  bytes[sizeof(uint64_t) / sizeof(BaseType)];
-    } data;
-    data.value = number;
-
-    // FIXME: Assumes little endian
-    std::size_t i = 0;
-    for (; i < sizeof(uint64_t); ++i)
+    if (data_.wide[i] == other.elementAt(i))
```
troels-im: What happens if you compare two `UInt` which are equal?  It seems to me that the loop will run forever, but maybe I am wrong?

markos: hm, equality is at line 324 and it doesn't seem like an infinite loop.

libs/vectorise/include/vectorise/uint/uint.hpp
markos: indeed, fixed.

troels-im (approved):
LGTM

PR #1321, opened 2019-07-08
Ml bitcoin price predictor

baykaner:

PR #1349, opened 2019-07-12
Ml crypto etch fixed point

baykaner:
Extending #1321 
- sets fixed_point_64 as the default DataType in the vm_modules math/ml libraries
- implements normalisation for crypto prediction example
- implements sgd optimiser for etch binding
- various tidying/refactoring


libs/ml/include/ml/dataloaders/crypto_prediction_loaders/crypto_dataloader.hpp
```diff
@@ -17,93 +17,110 @@
...
-void TensorDataLoader<LabelType, InputType>::Reset()
-{
-  data_cursor_  = 0;
-  label_cursor_ = 0;
-}
-
-}  // namespace dataloaders
-}  // namespace ml
-}  // namespace fetch
+//#pragma once
```
wkaluza: Please don't.

wkaluza (changes requested):
Looks ok, but there's a commented out file that should not go in as is; also, the scaler and optimiser modules deserve to have some tests to keep them safe.

scripts/etch-language-tests/ml/dataloader.etch
```diff
@@ -0,0 +1,3 @@
+function main()
+    var data_loader = DataLoader();
+endfunction
```
wkaluza, line 3: I don't think this test is pulling its weight, as it's only exercising the constructor, especially given that the other tests in this PR construct an instance of a DataLoader. If there's nothing interesting one can do with a dataloader in isolation, I would get rid of this one.

scripts/etch-language-tests/ml/graph.etch
```diff
@@ -0,0 +1,24 @@
+function main()
```
wkaluza, line 1: This test is missing assertions that behaviour is correct. Etch now contains an `assert` function you can use, with an optional message.

scripts/etch-language-tests/ml/optimiser.etch
```diff
@@ -0,0 +1,41 @@
+function main()
```
wkaluza, line 1: This checks the code compiles and does not crash. It would be nice to also know if the optimisers do the right thing, if for no other reason than to protect the behaviour from being accidentally changed in the future. Add assertions, please.

wkaluza: This only checks that code involving optimisers compiles and does not crash. It would be nice to know that they also do the right thing. Add assertions, please.

scripts/etch-language-tests/ml/scaler.etch
```diff
@@ -0,0 +1,20 @@
+function main()
+
```
wkaluza, line 2: As above.

wkaluza (changes requested):

wkaluza (dismissed):

wkaluza (approved):

PR #1353, opened 2019-07-15
LDGR-321 init lists

bipll:
This commit introduces initializer lists, and notably demonstrates the very early commencement of combinatorial explosion that is imminent unless we reform types handling in etch.


libs/vm/include/vm/analyser.hpp
```diff
@@ -164,53 +164,58 @@ class Analyser
   TypePtr        map_type_;
   TypePtr        sharded_state_type_;
   TypePtr        state_type_;
+  TypePtr        initializer_tree_type_;
```
robertdickson: initializer_list_type_ here

libs/vm/include/vm/common.hpp
```diff
@@ -43,26 +43,27 @@ using TypeIndex      = std::type_index;
...
+static TypeId const UInt64          = 11;
+static TypeId const Float32         = 12;
+static TypeId const Float64         = 13;
+static TypeId const Fixed32         = 14;
+static TypeId const Fixed64         = 15;
+static TypeId const PrimitiveMaxId  = 15;
+static TypeId const String          = 16;
+static TypeId const Address         = 17;
+static TypeId const InitializerList = 18;
+static TypeId const NumReserved     = 19;
```
robertdickson: InitializerList is not an Object type, it is a primitive, it should go after Null and before Void probably

libs/vm/src/analyser.cpp
```diff
@@ -78,6 +80,9 @@ void Analyser::Initialise()
                       fixed32_type_);
   CreatePrimitiveType("Fixed64", TypeIndex(typeid(fixed_point::fp64_t)), true, TypeIds::Fixed64,
                       fixed64_type_);
+  CreatePrimitiveType("InitializerList", TypeIndex(typeid(InitializerListPlaceholder)), false,
```
robertdickson: Move this after Null type creation, and before Void type creation

libs/vm/src/analyser.cpp
```diff
@@ -1182,21 +1274,38 @@ bool Analyser::AnnotateEqualityOp(ExpressionNodePtr const &node)
...
+      }
+      else if (lhs->node_kind == NodeKind::InitializerList)
+      {
+        if (!ConvertInitializerList(lhs, rhs->type))
+        {
+          AddError(node->line, "incompatible types");
+          return false;
+        }
+      }
+
```
robertdickson, line 1492: Just curious what happens here if LHS and RHS are both init lists?

libs/vm/src/analyser.cpp
```diff
@@ -1237,20 +1346,35 @@ bool Analyser::AnnotateRelationalOp(ExpressionNodePtr const &node)
...
+    }
+  }
+  else if (lhs->node_kind == NodeKind::InitializerList)
+  {
+    if (!ConvertInitializerList(lhs, rhs->type))
+    {
+      AddError(node->line, "incompatible types");
+      return false;
+    }
+  }
```
robertdickson, line 1570: Same drill as previous note: what happens if both sides are initializer lists?

libs/vm/src/analyser.cpp
bipll: Compilation error, as current type handling is too narrow to contain proper matching of untyped lists.

libs/vm/src/analyser.cpp
```diff
@@ -1389,12 +1513,19 @@ bool Analyser::AnnotateIndexOp(ExpressionNodePtr const &node)
...
     }
   }
 
+  if (lhs->node_kind == NodeKind::InitializerList)
+  {
+    // TODO(bipll): while indexing into initialization trees makes perfect easily defined sense
+    // a correct implementation would take too much for now
+    AddError(lhs->line, "indexing into initializer lists not supported yet");
+  }
+
```
robertdickson, line 1726: {1, 2, 3}[index] will never be an allowed expression because the type that is being indexed cannot be inferred. This if block can be removed.

libs/vm/src/analyser.cpp
```diff
@@ -1954,6 +2112,14 @@ FunctionPtr Analyser::FindFunction(TypePtr const &type, FunctionGroupPtr const &
...
   {
     actual_types = array[0];
+    for (std::size_t i{}; i < actual_types.size(); ++i)
+    {
+      auto const &supplied_node = supplied_nodes[i];
+      if (supplied_node->node_kind == NodeKind::InitializerList)
+      {
+        ConvertInitializerList(supplied_node, actual_types[i]);
+      }
+    }
```
robertdickson, line 2327: +1 nice solution to this issue, convert the nodes as soon as we know the winning signature.

libs/vm/src/parser.cpp
```diff
@@ -1024,304 +1016,172 @@ ExpressionNodePtr Parser::ParseExpression(bool is_conditional_expression)
...
+    bool parses{true};
     switch (token_->kind)
     {
     case Token::Kind::Identifier:
-    {
-      if (HandleIdentifier() == false)
-      {
-        return nullptr;
-      }
+      parses = HandleIdentifier();
```
robertdickson, line 1190: I like. This style is much nicer.

libs/vm/src/parser.cpp
```diff
@@ -1024,304 +1016,172 @@ ExpressionNodePtr Parser::ParseExpression(bool is_conditional_expression)
...
-      {
-        return nullptr;
-      }
+      parses =
+          HandleOpener(NodeKind::Unknown, NodeKind::Index, Token::Kind::RightSquareBracket, "]");
       break;
-    }
+
+    case Token::Kind::LeftBrace:
+      parses = HandleOpener(NodeKind::Unknown, NodeKind::Index, Token::Kind::RightBrace, "}");
```
robertdickson: This is wrong. It should be HandleOpener(NodeKind::InitializerList, NodeKind::Unknown....

libs/vm/src/analyser.cpp
robertdickson: A compiler error is fine, because {1, 2, 3} op {1, 2, 3} (say) doesn't really make any sense when we can't infer the types.

libs/vm/src/analyser.cpp
bipll: Is there any order imposed on types creation? I thought it did not matter at all.

libs/vm/src/analyser.cpp
bipll: Indexing can be initlist's own operation, without need to infer any type at all, can't it? In this example it would be quite idiomatic to allow `{1, 2, 3}[1] == 2`. 'Never' is too long a term, IMO.

libs/vm/src/analyser.cpp
bipll: Sure. :)

libs/vm/src/parser.cpp
bipll: Thanks.

libs/vm/include/vm/common.hpp
bipll: Does its position among primitives really matter?

robertdickson: Consistency, init lists, like null and void, are special in that they are primitives that are not added to the symbol table.

libs/vm/src/analyser.cpp
robertdickson: Again, just for consistency.

libs/vm/src/analyser.cpp
robertdickson: No, init lists, like null, do not have a type at all, they must always be inferred to a real type before being put to use.

bipll: > No, init lists, like null, do not have a type at all, they must always be inferred to a real type before being put to use.

This is not an explanation or reasoning, but only a confirmation of status quo.

robertdickson: OK, fair play. Never say never -- maybe in the future we can do something with this. Can we leave this if check in then, but simply remove the "yet" from the error message? That way we don't raise an expectation that it will be done soon, but don't close the possibility of doing so.

bipll: Agreed. :-)

libs/vm/include/vm/common.hpp
```diff
@@ -43,26 +43,27 @@ using TypeIndex      = std::type_index;
...
-static TypeId const Float64        = 13;
-static TypeId const Fixed32        = 14;
-static TypeId const Fixed64        = 15;
-static TypeId const PrimitiveMaxId = 15;
-static TypeId const String         = 16;
-static TypeId const Address        = 17;
-static TypeId const NumReserved    = 18;
+static constexpr TypeId Unknown         = 0;
+static constexpr TypeId Null            = 1;
+static constexpr TypeId InitializerList = 2;
```
robertdickson: Given the chat on Slack re British spelling, can we take the opportunity now to change InitializerList, initializer_list_type_ etc to the UK versions?

libs/vm/include/vm/module.hpp
```diff
@@ -98,7 +98,7 @@ class Module
     {}
 
     template <typename... Ts>
-    ClassInterface &CreateConstuctor()
+    ClassInterface &CreateConstructor()
```
robertdickson: What on earth is a "Constuctor"? :-) Nice find!

libs/vm/src/analyser.cpp
```diff
@@ -555,6 +564,10 @@ void Analyser::AnnotateBlock(BlockNodePtr const &block_node)
       AnnotateInplaceModuloOp(ConvertToExpressionNodePtr(child));
       break;
     }
+    case NodeKind::InitializerList:
+      AddError(child->line, "cannot infer list type");
+      break;
```
robertdickson, line 658: Are the braces missing here?

libs/vm/src/parser.cpp
```diff
@@ -1191,304 +1183,173 @@ ExpressionNodePtr Parser::ParseExpression(bool is_conditional_expression)
...
-        return nullptr;
-      }
+      parses =
+          HandleOpener(NodeKind::Unknown, NodeKind::Index, Token::Kind::RightSquareBracket, "]");
       break;
-    }
+
+    case Token::Kind::LeftBrace:
+      parses =
+          HandleOpener(NodeKind::InitialiserList, NodeKind::Index, Token::Kind::RightBrace, "}");
```
robertdickson: Second argument needs to be NodeKind::Unknown not NodeKind::Index

robertdickson (commented):
Apart from one tiny little bug in parser.cpp, this LGTM. Let's get that fixed, and we can land this feature.

robertdickson (approved):
LGTM

PR #1354, opened 2019-07-15
gtest example

n-hutton:
This PR shows the flow for adding unit tests to the codebase.

./libs/crypto/tests/CMakeLists.txt contains:

```
...
fetch_add_test(fetch-crypto-tests fetch-crypto gtests)
```

This creates a target fetch-crypto-tests where everything in the folder gtests will be added to the target fetch-crypto-tests.

after building this, you can run ```./libs/crypto/tests/fetch-crypto-tests  --gtest_filter=*Bls* ``` to only run certain tests (with Bls in the name in this instance).

NOTE: It looks like ed has already added a test ./libs/crypto/tests/gtests/bls_tests.cpp which you might want to look at.

This should fail CI here since I have written a unit test that fails.

### DIFF
--- a/libs/crypto/tests/gtests/bls_test_example.cpp
+++ b/libs/crypto/tests/gtests/bls_test_example.cpp
@@ -1,0 +1,25 @@
+//------------------------------------------------------------------------------
+//
+//   Copyright 2018-2019 Fetch.AI Limited
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "gtest/gtest.h"
+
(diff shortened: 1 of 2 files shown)


libs/crypto/tests/gtests/bls_test_example.cpp
```diff
@@ -0,0 +1,25 @@
...
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "gtest/gtest.h"
+
+TEST(BlsTestsExample, Example1)
```
n-hutton, line 21: The most basic way of creating a test is to include the gtest header and use the TEST macro with assertions such as EXPECT_EQ.

libs/crypto/tests/gtests/bls_test_example_fixture.cpp
```diff
@@ -0,0 +1,39 @@
...
+  void SetUp() override
+  {
+    fixture_state_ = 9;
+  }
+
+  void TearDown() override
+  {}
+
+  // Classes and other state here etc.
+  int fixture_state_ = 0;
```
n-hutton, line 33: A useful pattern can be to use a so called fixture class that can manage/set up some state for your test.

Other useful patterns are:  faking, where your code is dependent on some behaviour you don't feel like testing, like writing to a file. You can 'fake' a file writer. And mocking, when you create a 'mock' of your class that you can verify will make the correct function calls on certain inputs.

PR #1355, opened 2019-07-15
[Ledger] Propagate result from transaction execution to Tx status cache

### LINKED ISSUE #900: Generalise smart contract return types to handle any type (OR restrict smart contract returns to int)
Possibly an open question about the return types from executing a smart contract. At one point Ed suggested effectively an int error code

pbukva:
Resolves #900


libs/ledger/include/ledger/execution_result.hpp
```diff
@@ -0,0 +1,121 @@
...
+  RESOURCE_FAILURE,
+  INEXPLICABLE_FAILURE,
+};
+
+struct ContractExecutionResult
+{
+  ContractExecutionStatus status{
+      ContractExecutionStatus::NOT_RUN};  ///< The status of the transaction
+  TokenAmount charge{0};                  ///< The number of units of charge that
+  TokenAmount charge_rate{0};             ///< The cost of each unit of charge
```
ejfitzgerald, line 53: Would also be good to include the charge limit here also

libs/ledger/include/ledger/transaction_status_cache.hpp
```diff
@@ -20,14 +20,15 @@
...
+#include "ledger/execution_result.hpp"
 
 #include <chrono>
 #include <unordered_map>
 
 namespace fetch {
 namespace ledger {
 
-enum class TransactionStatus
+enum class eTransactionStatus : uint8_t
```
ejfitzgerald: Style 🚓 - Please revert to original name

libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -48,22 +48,28 @@ class Address;
 class Contract
 {
 public:
-  enum class Status
+  enum class eStatus
```
ejfitzgerald: Style  🚓 - Please revert to original name

libs/ledger/src/tx_status_http_interface.cpp
```diff
@@ -35,6 +35,25 @@ using fetch::variant::Variant;
 namespace fetch {
 namespace ledger {
 
+namespace {
+Variant toVariant(Digest const &digest, TransactionStatusCache::TxStatus const &tx_status)
```
ejfitzgerald: Style 🚓 - `ToVariant`

libs/ledger/src/tx_status_http_interface.cpp
```diff
@@ -35,6 +35,25 @@ using fetch::variant::Variant;
...
+{
+  auto retval{Variant::Object()};
+  retval["tx"]     = ToBase64(digest);
+  retval["status"] = ToString(tx_status.status);
+  auto contr_ex_res_v{Variant::Object()};
```
ejfitzgerald: Not sure if we need this extra object level here. I think it makes more sense to have a single level as a return value

libs/ledger/src/transaction_status_cache.cpp
```diff
@@ -27,54 +27,55 @@ using fetch::generics::MilliTimer;
...
+      status = it->second;
     }
   }
 
   return status;
 }
 
-void TransactionStatusCache::Update(Digest digest, TransactionStatus status, Timepoint const &now)
+void TransactionStatusCache::Update(Digest digest, eTransactionStatus status,
+                                    ContractExecutionResult const execRes, Timepoint const &now)
```
ejfitzgerald: Style 🚓  - `exec_result`?

Also I think that we want to remove the `const` in order to make the move useful

ejfitzgerald: I also think I would probably separate the two types of update call (high level and execution level) into two unique methods and sort of the details internally.

ejfitzgerald (changes requested):
This is looking good. I would not have bothered with all the renaming of the enums.

As discussed I think it is important to create a new enum that combines that high level status and execution status which can be returned from the HTTP queries. It might make sense to simple do this mapping in at the point of the query at the moment.

pbukva:
End-to-End test for `getBlockNumber()` is failing since the python api (`ledger-api-py` repo) has not been updated.

ejfitzgerald (dismissed):
LGTM

pbukva:
I'm merging this in, disregarding the failed `thread_pool_gtest` (on 677817d, PR-head @ GCC8-Release), since the test is empirically well known to be flaky, and it passed in PR-merge @ GCC8-Release.

PR #1356, opened 2019-07-15
Ml overview diagram update

baykaner:
- modified to represent true position of Node + added optimisers and estimators


evsmithx (approved):
LGTM

PR #1363, opened 2019-07-16
Reliable broadcast channel for DKG

jinmannwong:
Implements reliable broadcast protocol for broadcasting in P2P network to ensure that all honest parties act on the same message in the presence of some threshold number of Byzantine peers. Protocol involves exchanging various messages in different parts of the protocol where peers broadcast the hash of the message they received.


libs/dkg/include/dkg/rbc.hpp
```diff
@@ -0,0 +1,136 @@
...
+#include "network/muddle/muddle_endpoint.hpp"
+#include "network/muddle/rpc/server.hpp"
+#include "rbc_envelop.hpp"
+
+#include <bitset>
+
+namespace fetch {
+namespace dkg {
+class DkgService;
+namespace rbc {
```
n-hutton, line 31: It might be nice to not have too many nested namespaces if possible - could everything go in fetch::dkg or fetch::crypto?

libs/dkg/include/dkg/rbc.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Reliable broadcast channel (RBC) is a protocol which ensures all honest
+ * parties receive the same message in the presence of a threshold number of
+ * Byzantine adversaries
```
n-hutton, line 36: 👍

libs/dkg/include/dkg/rbc.hpp
```diff
@@ -0,0 +1,136 @@
...
+ */
+class RBC
```
n-hutton, line 38: Picky: I like full names if possible : ReliableBroadcastChannel

libs/dkg/include/dkg/rbc.hpp
```diff
@@ -0,0 +1,136 @@
...
+public:
+  using Endpoint        = muddle::MuddleEndpoint;
+  using ConstByteArray  = byte_array::ConstByteArray;
+  using MuddleAddress   = ConstByteArray;
+  using CabinetMembers  = std::set<MuddleAddress>;
+  using Subscription    = muddle::Subscription;
+  using SubscriptionPtr = std::shared_ptr<muddle::Subscription>;
+
+  // Constructor
+  explicit RBC(Endpoint &endpoint, MuddleAddress address, CabinetMembers const &cabinet,
```
n-hutton, line 49: No need for explicit when there is more than one argument

libs/dkg/include/dkg/rbc.hpp
```diff
@@ -0,0 +1,136 @@
...
+    R_SEND,
+    R_ECHO,
+    R_READY,
+    R_REQUEST,
+    R_ANSWER
+  };
+
+  struct MsgCount
+  {
+    size_t echo_count, ready_count;  ///< Count of RReady and RRecho messages
```
n-hutton, line 68: Style: usually have functions, then variables

n-hutton: there are a few of these so I'll just comment the once

libs/dkg/include/dkg/rbc.hpp
```diff
@@ -0,0 +1,136 @@
...
+    {
+      deliver_s = 1;  // initialize sequence counter by 1
+    }
+  };
+
+  uint32_t           id_;  ///< Rank used in DKG (derived from position in current_cabinet_)
+  uint8_t            msg_counter_;  ///< Counter for messages we have broadcasted
+  std::vector<Party> parties_;      ///< Keeps track of messages from cabinet members
+  std::unordered_map<uint64_t, Broadcast> broadcasts_;  ///< map from tag to broadcasts
+  std::mutex                              mutex_flags_;
```
n-hutton, line 99: We usually define a mutex, then the following variables are the ones that would be protected by that mutex, like so:

```
mutex_flags_
int a
int b

mutex_deliver_
int c
```

libs/dkg/include/dkg/rbc.hpp
```diff
@@ -0,0 +1,136 @@
...
+  std::mutex                              mutex_deliver_;
+  std::mutex                              mutex_broadcast_;
+
+  // For broadcast
+  static constexpr uint16_t SERVICE_DKG       = 5001;
```
n-hutton, line 104: There is a header we usually put these IDs in, I can do this in another PR

libs/dkg/include/dkg/rbc.hpp
```diff
@@ -0,0 +1,136 @@
...
+  static constexpr uint8_t  CHANNEL_BROADCAST = 2;  ///< Channel for reliable broadcast
+
+  MuddleAddress const address_;   ///< Our muddle address
+  Endpoint &          endpoint_;  ///< The muddle endpoint to communicate on
+  CabinetMembers const
+      &current_cabinet_;       ///< The set of muddle addresses of the cabinet (including our own)
+  std::size_t &   threshold_;  ///< Number of byzantine nodes
```
n-hutton, line 111: Is this comment correct? Isn't it the minimum number of nodes required for the protocol

libs/dkg/include/dkg/rbc.hpp
n-hutton: Also, the threshold here is a size_t - we are trying to avoid using these in favour of uint64_t because size_t could be a different size for example on 32 vs 64 bit systems

libs/dkg/include/dkg/rbc_envelop.hpp
```diff
@@ -0,0 +1,75 @@
...
+  // Constructor
+  RBCEnvelop() = default;
+  explicit RBCEnvelop(const RBCMessage &msg)
+    : type_{msg.type()}
+    , payload_{msg.Serialize().data()} {};
+
+  template <typename T>
+  void Serialize(T &serialiser) const
+  {
+    serialiser << (uint8_t)type_ << payload_;
```
n-hutton, line 43: We usually don't use C-style casting like this, rather ```uint8_t(type_)```

libs/dkg/include/dkg/rbc_envelop.hpp
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  template <typename T>
+  void Deserialize(T &serialiser)
+  {
+    uint8_t val;
```
n-hutton, line 49: 👍

libs/dkg/include/dkg/rbc_envelop.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+private:
+  RBCMessage::MessageType type_;     ///< Type of message contained in envelop
+  Payload                 payload_;  ///< Serialised RBCMessage
+};
+
+template <typename T>
+inline void Serialize(T &serializer, RBCEnvelop const &env)
+{
+  env.Serialize(serializer);
```
n-hutton, line 65: Looks like this might be unneccessary - I think I recall you saying the serializer lookup failed and you had a work around. I can fix this in another PR

libs/dkg/include/dkg/rbc_messages.hpp
```diff
@@ -0,0 +1,253 @@
...
+#include "core/serializers/byte_array_buffer.hpp"
+#include "core/serializers/stl_types.hpp"
+
+#include <cstdint>
+
+namespace fetch {
+namespace dkg {
+namespace rbc {
+
+using TruncatedHash        = byte_array::ByteArray;
```
n-hutton, line 30: We usually put these aliases in a class since any file including rpc_messages.hpp will see these aliases which can cause conflicts sometimes. It's ok in cpp files since these are never included.

libs/dkg/include/dkg/rbc_messages.hpp
```diff
@@ -0,0 +1,253 @@
...
+    RECHO,
+    RREADY,
+    RREQUEST,
+    RANSWER
+  };
+
+  // Destruction
+  virtual ~RBCMessage() = default;
+
+  /// @name Getter functions
```
n-hutton, line 60: unneccessary comment

libs/dkg/include/dkg/rbc_messages.hpp
```diff
@@ -0,0 +1,253 @@
...
+   *
+   * @return Tag of message
+   */
+  TagType tag() const
+  {
+    TagType msg_tag = channel_;
+    msg_tag <<= 48;
+    msg_tag |= id_;
+    msg_tag <<= 32;
+    return (msg_tag | uint64_t(counter_));
```
n-hutton, line 74: Not entirely sure about what the tag on a message entails

libs/dkg/include/dkg/rbc_messages.hpp
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  RBCMessage::MessageType type() const
+  {
+    return type_;
+  }
+
+  uint8_t channel() const
+  {
+    return static_cast<uint8_t>(channel_);
```
n-hutton, line 84: what's going on with channel_ being a uint 16 vs 8?

libs/dkg/include/dkg/rbc_messages.hpp
```diff
@@ -0,0 +1,253 @@
...
+  const MessageType type_;  ///< Type of the message
+  uint16_t
+           channel_;  ///< Channel Id of the broadcast channel (is this safe to truncate to uint8_t?)
+  uint32_t id_;       ///< Unique Id of the node
+  uint8_t  counter_;  ///< Counter for messages sent on RBC
+
+  // Construction
+  explicit RBCMessage(MessageType type)
+    : type_{type} {};
+  explicit RBCMessage(uint16_t channel, uint32_t id, uint8_t counter, MessageType type)
```
n-hutton, line 110: No need for explicit on this second constructor

libs/dkg/include/dkg/rbc_messages.hpp
```diff
@@ -0,0 +1,253 @@
...
+    , channel_{channel}
+    , id_{id}
+    , counter_{counter} {};
+};
+
+class RMessage : public RBCMessage
+{
+public:
+  // Destruction
+  virtual ~RMessage() = default;
```
n-hutton, line 121: Optional : make this class final if it's not inherited from

n-hutton: ah, I see it actually is

libs/dkg/include/dkg/rbc_messages.hpp
```diff
@@ -0,0 +1,253 @@
...
+    : RBCMessage{type} {};
+  RHash(uint16_t channel, uint32_t id, uint8_t counter, TruncatedHash msg_hash, MessageType type)
+    : RBCMessage{channel, id, counter, type}
+    , hash_{std::move(msg_hash)} {};
+};
+
+class RBroadcast : public RMessage
+{
+public:
+  // Construction/Destruction
```
n-hutton, line 178: can probably do away with comment on construction/dest

libs/dkg/src/dkg_service.cpp
```diff
@@ -114,6 +114,7 @@ DkgService::DkgService(Endpoint &endpoint, ConstByteArray address, ConstByteArra
   , rpc_server_{endpoint_, SERVICE_DKG, CHANNEL_RPC}
   , rpc_client_{"dkg", endpoint_, SERVICE_DKG, CHANNEL_RPC}
   , state_machine_{std::make_shared<StateMachine>("dkg", State::BUILD_AEON_KEYS, ToString)}
+  , rbc_{endpoint_, address_, current_cabinet_, current_threshold_, *this}
```
n-hutton, line 117: Just a note: as wojciech pointed out, need to be careful when refrencing this in the constructor of something - I think that if rbc took a ```DkgService dkg_service``` (it doesn't) it would copy construct DkgService partway through constructing it, which would likely cause a segfault.

I checked and this usage is safe, though.

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+ */
+TruncatedHash MessageHash(const SerialisedMessage &msg)
+{
+  byte_array::ByteArray msg_hash_256{crypto::Hash<crypto::SHA256>(msg)};
+  return msg_hash_256.SubArray(24);
+}
+
+std::string RBC::MsgTypeToString(MsgType msg_type)
+{
+  switch (msg_type)
```
n-hutton, line 42: todo : make enums have no size so default isn't needed (I can do this)

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+#include "dkg/rbc.hpp"
+
+namespace fetch {
+namespace dkg {
+namespace rbc {
+
+constexpr char const *LOGGING_NAME = "RBC";
+
+/**
+ * Helper function to compute truncation of message hash to 8 bytes. Truncation from left side.
```
n-hutton, line 32: 👍

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+  rbc_subscription_->SetMessageHandler([this](MuddleAddress const &from, uint16_t, uint16_t,
+                                              uint16_t, muddle::Packet::Payload const &payload,
+                                              MuddleAddress) {
+    RBCSerializer serialiser(payload);
+
+    // Deserialize the RBCEnvelop
+    RBCEnvelop env;
+    serialiser >> env;
+
+    // Dispatch the event
```
n-hutton, line 88: 👍

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+}
+
+/**
+ * Resets the RBC for a new cabinet
+ */
+void RBC::ResetCabinet()
+{
+  assert(!current_cabinet_.empty());
+  assert(current_cabinet_.find(address_) != current_cabinet_.end());
+  assert(current_cabinet_.size() > 3 * threshold_);
```
n-hutton, line 100: Where does the 3 come from here?

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+  std::lock_guard<std::mutex> lock{mutex_broadcast_};
+  id_ = static_cast<uint32_t>(
```
n-hutton, line 102: can id_ ever change ? current_cab is const

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+ * @param address Destination muddle address
+ */
+void RBC::Send(const RBCEnvelop &env, const MuddleAddress &address)
+{
+  // Serialise the RBCEnvelop
+  RBCSerializerCounter env_counter;
+  env_counter << env;
+
+  RBCSerializer env_serializer;
+  env_serializer.Reserve(env_counter.size());
```
n-hutton, line 122: 👍

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+  env_serializer << env;
+
+  std::lock_guard<std::mutex> lock{mutex_broadcast_};
```
n-hutton, line 143: don't think this lock is neccessary?

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+ * @param tag Unique tag of a message sent via RBC
+ * @param msg_ptr Shared pointer to REcho message
+ * @return Bool for whether we have reached the message count for REcho messages
+ */
+bool RBC::ReceivedEcho(TagType tag, std::shared_ptr<REcho> msg_ptr)
+{
+  std::lock_guard<std::mutex> lock(mutex_broadcast_);
+  auto &                      msg_count = broadcasts_[tag].msgs_count[msg_ptr->hash()];
+  msg_count.echo_count++;
+  return (msg_count.echo_count == current_cabinet_.size() - threshold_ and
```
n-hutton, line 214: style : we use ```&&```

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+ * Handler for a new RBCEnvelop message
+ *
+ * @param from Muddle address of sender
+ * @param envelop RBCEnvelop message
+ * @param transmitter Muddle address of transmitter
+ */
+void RBC::OnRBC(MuddleAddress const &from, RBCEnvelop const &envelop)
+{
+  auto msg_ptr = envelop.Message();
+  if (msg_ptr == nullptr)
```
n-hutton, line 243: todo (self) : switch enum adjust

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+
+/**
+ * Handler for RBroadcast messages. If valid, broadcasts REcho message for this tag
+ *
+ * @param msg_ptr Shared pointer to RBroadcast message
+ * @param sender_index Index of sender in current_cabinet_
+ */
+void RBC::OnRBroadcast(const std::shared_ptr<RBroadcast> msg_ptr, uint32_t sender_index)
+{
+  uint64_t tag{msg_ptr->tag()};
```
n-hutton, line 284: could use auto here?

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+  {
+    broadcasts_[tag].dbar = msg_ptr->hash();
+
+    TruncatedHash msg_hash;
+    if (!SetDbar(tag, msg_ptr))
+    {
+      RBCEnvelop env{RRequest{msg_ptr->channel(), msg_ptr->id(), msg_ptr->counter()}};
+
+      uint32_t counter{0};
+      auto     im{current_cabinet_.begin()};
```
n-hutton, line 378: our style for auto is ```auto xx = current_cabinet_.begin();```

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+  }
+}
+
+/**
+ * Computes index of an address in current_cabinet_
+ *
+ * @param other_address Muddle address of cabinet member
+ * @return Index of other_address
+ */
+uint32_t RBC::CabinetIndex(const MuddleAddress &other_address) const
```
n-hutton, line 508: this could be used in ResetCabinet right?

libs/dkg/src/rbc.cpp
```diff
@@ -0,0 +1,571 @@
...
+ *
+ * @param l Index of party for which we received a new RBCMessage
+ * @param tag Unique tag of a message sent via RBC
+ * @param m Message type of new RBCMessage received
+ * @return Bool for whether the message flag for this message type has been set
+ */
+bool RBC::SetPartyFlag(uint32_t sender_index, TagType tag, MsgType msg_type)
+{
+  std::lock_guard<std::mutex> lock(mutex_flags_);
+  auto &                      iter  = parties_[sender_index].flags[tag];
```
n-hutton, line 557: bit confused about this iterator(?)

n-hutton (changes requested):
mostly style changes

libs/dkg/include/dkg/rbc.hpp
jinmannwong: No, the threshold is the number of Byzantine nodes. There is then a condition that number of nodes must be greater than 3 * threshold_.

libs/dkg/src/rbc.cpp
jinmannwong: This is a RBC related constraint. If there are over 1/3 of Byzantine nodes then the guarantees of RBC do not hold.

libs/dkg/src/rbc.cpp
jinmannwong: Yes id_ can change if the cabinet is reset inside DkgService. current_cabinet should be a const reference in RBC.

libs/dkg/src/rbc.cpp
jinmannwong: Ok, I inserted this yesterday when I was having problems with double messages and through it might be some multithread problem.

n-hutton: ah sure. Well it's not causing much harm there. Btw, you can detect multithreaded/race conditions using a tool called thread sanitizer which effectively tries to find places in your code that are unprotected. Might be a bit heavyweight for now but worth having in the back of your mind.

libs/dkg/src/rbc.cpp
n-hutton: Oh I see - it looks like id_ is uninitialized until reset is called at which point it will be set to a value that can never change (since current_cabinet can't change) - would it make sense to set id_ in the constructor instead?

n-hutton: Ignore me - I see now that the current cabinet could be changed in the DKG service and the RBC just holds a reference to that, so it can change (I misread it as just const and not const &). Note this could be the source of race conditions depending on when the cabinet gets reset.

n-hutton (approved):

PR #1365, opened 2019-07-17
HTTP Open API implementation

troels-im:
This PR creates open API implementation that allows the HTTP interface to be dumped into a JSON file.,


apps/constellation/logging_http_module.hpp
```diff
@@ -30,59 +30,61 @@ class LoggingHttpModule : public http::HTTPModule
...
+          Variant response{Variant::Object()};
+          for (auto const &element : GetLogLevelMap())
+          {
+            response[element.first] = ToString(element.second);
+          }
+
+          return http::CreateJsonResponse(response);
+        });
+
+    Patch("/api/logging/", "TODO(tfr): yet to be written",
```
ejfitzgerald:
```suggestion
    Patch("/api/logging/", "Update the logging level for specified subsystem(s)",
```

apps/constellation/open_api_http_module.hpp
```diff
@@ -0,0 +1,106 @@
...
+          response["paths"] = paths;
+
+          return http::CreateJsonResponse(response);
+        });
+  }
+
+  void Reset(http::HTTPServer *srv)
+  {
+    std::lock_guard<std::mutex> lock(server_lock_);
+    server_ = std::move(srv);
```
ejfitzgerald: I guess move doesn't really make much sense here?

apps/constellation/open_api_http_module.hpp
```diff
@@ -0,0 +1,106 @@
...
+            path[method] = details;
+          }
+
+          response["paths"] = paths;
+
+          return http::CreateJsonResponse(response);
+        });
+  }
+
+  void Reset(http::HTTPServer *srv)
```
ejfitzgerald:
```suggestion
  void Reset(http::HTTPServer *srv = nullptr)
```

Since you sort of use this as a shared pointer kind of wrapper

libs/http/include/http/route.hpp
```diff
@@ -162,10 +202,14 @@ class Route
...
+    return var;
   }
 
-  byte_array::ByteArray            original_;
-  std::vector<match_function_type> match_;
+  byte_array::ByteArray                                                 original_;
+  byte_array::ByteArray                                                 path_;
+  std::vector<match_function_type>                                      match_;
+  std::vector<byte_array::ConstByteArray>                               path_parameters_;
+  std::unordered_map<byte_array::ConstByteArray, validators::Validator> validators_;
```
ejfitzgerald: I think it makes sense to create type aliases for these

libs/http/include/http/server.hpp
```diff
@@ -226,20 +227,42 @@ class HTTPServer : public AbstractHTTPServer
...
+               view_type const &view)
   {
-    views_.push_back({method, Route::FromString(path), view});
+    auto route = Route::FromString(path);
+
+    for (auto const &param : parameters)
+    {
+      validators::Validator v = param.validator;
+      v.description           = param.description;
+      route.AddValidator(param.name, v);
```
ejfitzgerald:
```suggestion
      route.AddValidator(param.name, std::move(v));
```

libs/http/include/http/validators.hpp
```diff
@@ -0,0 +1,49 @@
...
+namespace validators {
+
+struct Validator
+{
+  byte_array::ConstByteArray                      description;
+  std::function<bool(byte_array::ConstByteArray)> validator;
+  variant::Variant                                schema;
+};
+
+inline Validator StringValue(uint16_t min_length = 0, uint16_t max_length = uint16_t(-1))
```
ejfitzgerald: Can we move to a cpp file

libs/ledger/src/tx_status_http_interface.cpp
```diff
@@ -38,7 +38,10 @@ namespace ledger {
 TxStatusHttpInterface::TxStatusHttpInterface(TransactionStatusCache &status_cache)
   : status_cache_{status_cache}
 {
-  Get("/api/status/tx/(digest=[a-fA-F0-9]{64})",
+  Get("/api/status/tx/(digest=[a-fA-F0-9]{64})", "Retrieves a transaction status.",
+      {
+          {"digest", "The transaction hash.", http::validators::StringValue()},
```
ejfitzgerald, line 43: These usages of the validators are kind of curious because I guess the values are a little redundant because of the regex matching. Will be useful in the future though.

apps/constellation/open_api_http_module.hpp
```diff
@@ -0,0 +1,106 @@
...
+#include "http/server.hpp"
+
+#include <string>
+
+namespace fetch {
+
+class OpenAPIHttpModule : public http::HTTPModule
+{
+public:
+  OpenAPIHttpModule()
```
ejfitzgerald: Can we move this definition to a CPP file. The build times are starting to hurt now 😁

libs/http/include/http/route.hpp
```diff
@@ -49,6 +50,7 @@ class Route
       {
         return false;
       }
+      // TODO(tfr): Add validators
```
ejfitzgerald: Can we create a tracking issue for this if still valid

libs/http/include/http/route.hpp
```diff
@@ -79,15 +81,18 @@ class Route
 
         if (count != 0)
         {
-          TODO_FAIL("unclosed parameter.");
+          throw std::runtime_error("unclosed parameter.");
```
ejfitzgerald, line 89: 👍

ejfitzgerald (changes requested):

libs/ledger/src/tx_status_http_interface.cpp
troels-im: I agree with you and you will see that they are not actively validating at the moment if you dig deeper into the code. There are two reasons for having them:
1) OpenAPI provides schemas for all parameters 
2) They can provide more stringent validation than regex. For instance, an Address can be validated to check if the address checksum is actually correct.

We could rename them to Schema, rather than validators?

apps/constellation/open_api_http_module.hpp
troels-im: 👍

libs/http/include/http/route.hpp
troels-im: It is and yes, can do. It boils down to whether we want the interface to do additional validation beyond the regex. If so, the code should be added here. Curious to hear your thoughts.

apps/constellation/open_api_http_module.hpp
troels-im: No, you are right.

libs/http/include/http/validators.hpp
troels-im: Sure

ejfitzgerald (dismissed):

ejfitzgerald (dismissed):

libs/http/include/http/request.hpp
```diff
@@ -98,8 +98,8 @@ class HTTPRequest
   json::JSONDocument JSON() const
   {
     LOG_STACK_TRACE_POINT;
-
-    return json::JSONDocument(body());
+    json::JSONDocument x{body()};
+    return x;
```
ejfitzgerald: Presumably not needed any more

libs/variant/src/variant.cpp
```diff
@@ -62,23 +60,17 @@ Variant &Variant::operator=(Variant const &value)
...
     for (auto const &element : value.object_)
     {
-      // create an element
-      Variant *obj = pool().Allocate();
-      obj->parent_ = parent();
-
-      // make a deep copy
-      *obj = *element.second;
-
+      std::unique_ptr<Variant> variant{new Variant(*element.second)};
```
ejfitzgerald:
```suggestion
auto variant = std::make_unique<Variant>(*element.second);
```

libs/variant/include/variant/variant.hpp
```diff
@@ -754,14 +736,13 @@ inline Variant &Variant::operator[](ConstByteArray const &key)
   }
 
   // allocate an element
-  Variant *variant = pool().Allocate();
-  variant->parent_ = parent();
+  std::unique_ptr<Variant> variant{new Variant()};
```
ejfitzgerald:
```suggestion
auto variant = std::make_unique<Variant>();
```

libs/variant/include/variant/variant.hpp
```diff
@@ -301,9 +290,11 @@ inline Variant Variant::Object()
  *
  * @param pool_reserve The number of elements to preallocate
  */
+/*
```
ejfitzgerald: Lets just delete this entirely

libs/variant/include/variant/variant.hpp
```diff
@@ -77,7 +77,7 @@ class Variant
 
   // Construction / Destruction
   Variant() = default;
-  explicit Variant(std::size_t pool_reserve);
+  //  explicit Variant(std::size_t pool_reserve);
```
ejfitzgerald: Lets just delete entirely

ejfitzgerald (changes requested):
Think it is worth doing a little bit of clean up but looks okay to me

libs/http/include/http/request.hpp
troelsfr:
```suggestion
    return json::JSONDocument(body());
```

ejfitzgerald (approved):

PR #1366, opened 2019-07-17
Tensor slice/view improvements

baykaner:


libs/math/include/math/tensor.hpp
```diff
@@ -823,6 +843,13 @@ typename Tensor<T, C>::ViewType const Tensor<T, C>::View() const
   return TensorView<Type, ContainerType>(data_, height(), width);
 }
 
+/**
+ * returns a
```
evsmithx: Incomplete docs

libs/math/include/math/tensor.hpp
```diff
@@ -803,6 +810,13 @@ typename Tensor<T, C>::ConstIteratorType Tensor<T, C>::cend() const
 ///////////////////////
 /// View Extraction ///
 ///////////////////////
+
+/**
+ * returns a view of the entire tensor
+ * @tparam T
+ * @tparam C
```
evsmithx: I think T and C should either be documented or removed from this docstring?

evsmithx (approved):
LGTM

PR #1368, opened 2019-07-18
Feature/dkg test script

n-hutton:
script to drive dkg as a standalone app. Usage:

```
$ ./scripts/dkg/test_dkg.py -h
usage: test_dkg.py [-h] dkg_exe output_directory number_nodes threshold

High level DKG test

positional arguments:
  dkg_exe           Location of the dkg binary
  output_directory  directory to put output log files in
  number_nodes      Number of nodes to create
  threshold         Threshold for number of nodes to be honest

optional arguments:
  -h, --help        show this help message and exit
```


libs/dkg/src/dkg_service.cpp
```diff
@@ -433,6 +450,7 @@ State DkgService::OnCollectSignaturesState()
     }
   }
 
+  // TODO(HUT): looks like a bug here
```
n-hutton, line 473: @ejfitzgerald chat in the office? Looks like if round->HasSignature() gets set it can never get out of this state

jinmannwong (approved):

PR #1369, opened 2019-07-18
softmax that works with batch dimension, and  corresponding tests

XiaodongWuFetch:


libs/math/include/math/activation_functions/softmax.hpp
```diff
@@ -99,9 +99,20 @@ void Softmax(ArrayType const &array, ArrayType &ret, typename ArrayType::SizeTyp
   {
     details::Softmax2DImplementation(array, ret, axis);
   }
+  else if ((array.shape().size() == 3) && (ret.shape().size() == 3) && (axis == 0 || axis == 1))
+  {
+    ArrayType tmp_ret = ret.Slice(0, 2).Copy().Squeeze();
```
baykaner: use a view, like this:
auto tmp_ret = ret.View(0).Copy();

libs/math/include/math/activation_functions/softmax.hpp
```diff
@@ -99,9 +99,20 @@ void Softmax(ArrayType const &array, ArrayType &ret, typename ArrayType::SizeTyp
...
+    for (size_t i = 0; i < array.shape()[2]; i++)
+    {
+      details::Softmax2DImplementation(array.Slice(i, 2).Copy().Squeeze(), tmp_ret, axis);
```
baykaner: use a view:
array.View(i).Copy() // no need to squeeze - guaranteed to be 2D

libs/math/include/math/activation_functions/softmax.hpp
```diff
@@ -99,9 +99,20 @@ void Softmax(ArrayType const &array, ArrayType &ret, typename ArrayType::SizeTyp
...
+      ret.Slice(i, 2).Assign(tmp_ret);
```
baykaner: use view assign:
ret.View(i).Assign(tmp_ret);

libs/math/include/math/activation_functions/softmax.hpp
```diff
@@ -99,9 +99,20 @@ void Softmax(ArrayType const &array, ArrayType &ret, typename ArrayType::SizeTyp
...
+    }
+  }
   else
```
baykaner, line 117: add another else if to catch the case for 3D with axis == 2;
in this case you should use Slices

libs/math/tests/math/all_tests/activation_functions/softmax.cpp
```diff
@@ -76,6 +76,26 @@ TYPED_TEST(SoftmaxTest, equal_proportion_test)
...
+
+  TypeParam gt_array({4, 3, 1});
+  gt_array.Fill(typename TypeParam::Type(double(1) / double(3)));
+
+  typename TypeParam::SizeType axis = 0;
+
+  fetch::math::Softmax(test_array, test_array, axis);
+
+  // test correct values
+  for (std::size_t j = 0; j < gt_array.size(); ++j)
```
baykaner: please replace this for loop with something like this:

EXPECT_TRUE(test_array.AllClose(gt_array, fetch::math::function_tolerance<DataType>(), fetch::math::function_tolerance<DataType>()))

this will ensure that different types are tested at a fair precision for each type; and the AllClose will handle the loop and equivalence testing for you

libs/ml/include/ml/ops/activations/logsoftmax.hpp
```diff
@@ -64,17 +64,24 @@ class LogSoftmax : public fetch::ml::Ops<T>
...
+    // 2D softmax with 1 batch dimension
+    else if (inputs.front().get().shape().size() == 3)
     {
-      ArrayType sum;
-      sum = ReduceSum(return_signal, 1 - axis_);
+      assert((axis_ == 1) || (axis_ == 0));
+      ArrayType sum = return_signal.Slice(0, 1 - axis_).Copy();
+      for (size_t i = 0; i < return_signal.shape()[2]; i++)
+      {
+        sum.View(i).Assign(ReduceSum(return_signal.Slice(i, 2).Copy().Squeeze(), 1 - axis_).View());
```
baykaner: this is a pretty confusing 1-liner. Please rewrite as:

auto cur_sum = ReduceSum(return_signal.View(i).Copy(), 1 - axis);
sum.View(i).Assign(cur_sum);

libs/ml/include/ml/ops/activations/softmax.hpp
```diff
@@ -61,17 +61,23 @@ class Softmax : public fetch::ml::Ops<T>
...
     }
-    // 2D softmax
-    else if (inputs.front().get().shape().size() == 2)
+    // 2D softmax with 1 batch dimension
+    else if (inputs.front().get().shape().size() == 3)
     {
-      ArrayType sum;
-      sum = ReduceSum(return_signal, 1 - axis_);
+      assert((axis_ == 1) || (axis_ == 0));
+      ArrayType sum = return_signal.Slice(0, 1 - axis_).Copy();
```
baykaner: please rewrite the implementation as described above for logsoftmax

baykaner (changes requested):

libs/math/include/math/activation_functions/softmax.hpp
XiaodongWuFetch: I am not sure for now if we should implement the activation on batch dimension. It is better that we leave it till someone implemented multi dimension softmax.

libs/math/tests/math/all_tests/activation_functions/softmax.cpp
```diff
@@ -76,6 +76,24 @@ TYPED_TEST(SoftmaxTest, equal_proportion_test)
   }
 }
 
+TYPED_TEST(SoftmaxTest, multi_dimension_test)
```
baykaner, line 79: you should test this for both axis == 0 and axis == 1
also you should try a couple of other values (rather than 1/3 everywhere)

libs/math/include/math/activation_functions/softmax.hpp
```diff
@@ -99,9 +99,24 @@ void Softmax(ArrayType const &array, ArrayType &ret, typename ArrayType::SizeTyp
   {
     details::Softmax2DImplementation(array, ret, axis);
   }
+  else if ((array.shape().size() == 3) && (ret.shape().size() == 3) && (axis == 0 || axis == 1))
+  {
+    auto tmp_ret = ret.View(0).Copy();
```
baykaner, line 106: add a TODO(1320) - Copy()s can be removed when softmax2dimplementation handles IsIterable types

libs/math/include/math/activation_functions/softmax.hpp
```diff
@@ -99,9 +99,24 @@ void Softmax(ArrayType const &array, ArrayType &ret, typename ArrayType::SizeTyp
...
+    for (size_t i = 0; i < array.shape()[2]; i++)
+    {
+      details::Softmax2DImplementation(array.View(i).Copy(), tmp_ret, axis);
+      ret.View(i).Assign(tmp_ret);
+    }
+  }
+  else if ((array.shape().size() == 3) && (ret.shape().size() == 3) && (axis == 2))
```
baykaner, line 113: we can catch this with an assertion

libs/ml/include/ml/ops/activations/logsoftmax.hpp
```diff
@@ -64,17 +64,25 @@ class LogSoftmax : public fetch::ml::Ops<T>
...
     }
-    // 2D softmax
-    else if (inputs.front()->shape().size() == 2)
+    // 2D softmax with 1 batch dimension
+    else if (inputs.front()->shape().size() == 3)
     {
-      ArrayType sum;
-      sum = ReduceSum(return_signal, 1 - axis_);
+      assert((axis_ == 1) || (axis_ == 0));
+      ArrayType sum = return_signal.Slice(0, 1 - axis_).Copy();
```
baykaner: don't slice, construct an empty tensor:
ArrayType sum(shape);

libs/ml/include/ml/ops/activations/logsoftmax.hpp
```diff
@@ -64,17 +64,25 @@ class LogSoftmax : public fetch::ml::Ops<T>
...
+      for (size_t i = 0; i < return_signal.shape()[2]; i++)
+      {
+        auto cur_sum = ReduceSum(return_signal.Slice(i, 2).Copy().Squeeze(), 1 - axis_).View();
```
baykaner: i might be mistaken, but isn't this part:
return_signal.Slice(i, 2).Copy().Squeeze()
effectively a more expensive version of this:
return_signal.View(i).Copy()

libs/ml/include/ml/ops/activations/softmax.hpp
```diff
@@ -61,17 +61,24 @@ class Softmax : public fetch::ml::Ops<T>
...
     }
-    // 2D softmax
-    else if (inputs.front()->shape().size() == 2)
+    // 2D softmax with 1 batch dimension
+    else if (inputs.front()->shape().size() == 3)
     {
-      ArrayType sum;
-      sum = ReduceSum(return_signal, 1 - axis_);
+      assert((axis_ == 1) || (axis_ == 0));
+      ArrayType sum = return_signal.Slice(0, 1 - axis_).Copy();
```
baykaner: as above for logsoftmax

baykaner (dismissed):

MissingNO57 (approved):
LGTM, it will be rewritten with my and-D softmax anyway.

PR #1370, opened 2019-07-18
division fix

XiaodongWuFetch:
allow array / scalar division; fix incorrect backprop; change tests correspondingly


libs/ml/include/ml/ops/divide.hpp
```diff
@@ -44,39 +45,81 @@ class Divide : public fetch::ml::Ops<T>
   virtual void Forward(VecTensorType const &inputs, ArrayType &output)
   {
     assert(inputs.size() == 2);
-    assert(inputs.at(0).get().shape() == inputs.at(1).get().shape());
     assert(inputs.at(0).get().shape() == output.shape());
 
-    fetch::math::Divide(inputs.at(0).get(), inputs.at(1).get(), output);
+    if ((inputs.at(0).get().shape() == inputs.at(1).get().shape()) ||
+        (inputs.at(1).get().size() > 1))
+    {  // array / array
```
baykaner, line 51: I think you actually only need to check that inputs.at(1).get().size() > 1
This forward Divide will just invoke Broadcast if it needs to and work just fine right?

libs/ml/include/ml/ops/divide.hpp
```diff
@@ -44,39 +45,81 @@ class Divide : public fetch::ml::Ops<T>
...
+      fetch::math::Divide(inputs.at(0).get(), inputs.at(1).get(), output);
+    }
+    else
+    {  // array / scalar
```
baykaner, line 55: as it stands if you call this Forward with two arrays with different shapes (both with sizes > 1) it will trigger this else statement, and will divide the input(0) by the first element of input(1). That's a silent error so it's pretty dangerous.
Changing the if condition to simply (inputs(1).size() > 1) should fix it though

libs/ml/include/ml/ops/divide.hpp
```diff
@@ -44,39 +45,81 @@ class Divide : public fetch::ml::Ops<T>
...
+   * f'(a)=(1/b)*err
+   * f'(b)=-(a/(b^2))*err
    */
   virtual std::vector<ArrayType> Backward(VecTensorType const &inputs,
                                           ArrayType const &    error_signal)
   {
     ArrayType return_signal_1(inputs.at(0).get().shape());
-    ArrayType return_signal_2(return_signal_1.shape());
+    ArrayType return_signal_2(inputs.at(1).get().shape());
+    return_signal_2.Fill(static_cast<DataType>(0));
```
baykaner: this is unnecessary - When you create a Tensor it should be filled with 0 by default

libs/ml/include/ml/ops/divide.hpp
```diff
@@ -44,39 +45,81 @@ class Divide : public fetch::ml::Ops<T>
...
+        *r_2_it += -((*err_it) * (*a_it)) / ((*b_it) * (*b_it));
 
+        ++a_it;
+        ++err_it;
+        ++r_1_it;
+      }
+    }
+    else
+    {  // array / array different shape
+      // TODO () This implementation only works in limited scenarios, as it does not resolve the
```
baykaner: please raise an issue that Divide backwards broadcast needs to be implemented. You can paste this code snippet there if you think it would be helpful, and you must delete it from the codebase here.

baykaner (changes requested):

baykaner (dismissed):

baykaner (approved):

PR #1372, opened 2019-07-18
Ml slices to views

baykaner:
extends work begun in #1366 
more view implementations/specialisations


libs/math/include/math/tensor.hpp
```diff
@@ -823,6 +843,13 @@ typename Tensor<T, C>::ViewType const Tensor<T, C>::View() const
   return TensorView<Type, ContainerType>(data_, height(), width);
 }
 
+/**
+ * returns a
```
evsmithx: Unfinished comment

evsmithx (dismissed):
Looks good, apart from one unfinished comment!

evsmithx (approved):
Looks good!

PR #1374, opened 2019-07-18
Added System module to vm_modules

evsmithx:
Added a System vm module to do command-line parsing. Also fixed a few bugs in the etch examples. The System module is largely copied from ./apps/etch/main.cpp because etch scripts should run in the same way in both.


baykaner (dismissed):
LGTM

wkaluza (dismissed):
This PR is conspicuously missing an automated test or five. If we're elevating System to the actual modules library, we should give it some protection, as it's only going to be used more in the future. There is a a helper class for testing etch snippets Google Test - we can pair up to hack something out quickly, shouldn't take 20 minutes.

libs/vm_modules/include/vm_modules/core/system.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+class Parameters
+{
+public:
+  using ParamParser = fetch::commandline::ParamsParser;
+  using ArgList     = std::vector<char *>;
+  using StringList  = std::vector<std::string>;
+
+  void Parse(int argc, char **argv)
+  {
```
wkaluza: Additionally, I think we could move all the function definitions in this file into a cpp file.

evsmithx: Done!

wkaluza (dismissed):

wkaluza (approved):

PR #1375, opened 2019-07-18
Bugfix/fixedpoint overflow errors

markos:
Fix overflow checks
Also, removed Multiply and Divide methods and cleaned up some of the operators.

### DIFF
--- a/libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
+++ b/libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
@@ -27,8 +27,8 @@
 #include <cstdint>
 #include <functional>
 #include <iomanip>
-#include <iostream>
 #include <limits>
+#include <ostream>
 
 namespace fetch {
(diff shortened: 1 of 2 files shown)


libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
```diff
@@ -27,8 +27,8 @@
 #include <cstdint>
 #include <functional>
 #include <iomanip>
+#include <iostream>
```
wkaluza: I think we only need `<ostream>` here

wkaluza (changes requested):
Looks good, just a minor niggle.

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
markos: resolved.

wkaluza (approved):

PR #1376, opened 2019-07-19
Distributed Key Generation 

jinmannwong:
Distributed Key Generation protocol allows parties to collaboratively compute individual private and public keys which can be used to generate group signatures with a threshold number of individual signatures. This feature is used for the random beacon evolution, which requires a threshold signatures.


libs/dkg/include/dkg/dkg_messages.hpp
```diff
@@ -0,0 +1,234 @@
...
+  /// @}
+
+  virtual DKGSerializer Serialize() const = 0;
+
+protected:
+  const MessageType type_;       ///< Type of message of the three listed above
+  MsgSignature      signature_;  ///< ECDSA signature of message
+
+  explicit DKGMessage(MessageType type)
+    : type_{type} {};
```
wkaluza: Note: a semicolon is unnecessary after a function body (including a ctor). It's just an empty statement.

jinmannwong: Thanks. I have fixed this the latest commit.

n-hutton (approved):
LGTM - a bit of refactoring needs done but let's get this in then break it up

PR #1378, opened 2019-07-19
bugfix view copy in tensor dataloader

baykaner:
- implemented view copy method that allows specifying output tensor shape
- fixed tensor_dataloader getnext shapes


evsmithx (dismissed):
LGTM

evsmithx (approved):
LGTM

PR #1382, opened 2019-07-19
[Vectorise] `UInt<T>` - fix for buffer overrun

pbukva:
* Primarily, this fixes buffer overrun issue in constructors of the `UInt<T>`
 * As secondary change, this also removes constructors from `std::string` type
 for `UInt<T>` type, and so ultimately also constructors from `vm::String`
 for `vm::UInt256` type.


markos (dismissed):
Good catch, LGTM.

libs/vectorise/include/vectorise/uint/uint.hpp
```diff
@@ -52,40 +65,52 @@ class UInt
...
-  constexpr UInt(T const &other, meta::IfIsAByteArray<T> * = nullptr);
-  template <typename T>
-  constexpr UInt(T const &other, meta::IfIsStdString<T> * = nullptr);
-  template <typename T>
-  constexpr UInt(T const &number, meta::IfIsInteger<T> * = nullptr);
+  constexpr UInt()                      = default;
+  constexpr UInt(UInt const &other)     = default;
+  constexpr UInt(UInt &&other) noexcept = default;
+
+  template <typename... T, std::enable_if_t<meta::Is<WideType>::SameAsEvery<T...>::value &&
```
ejfitzgerald: I think it make sense to move all this SFINAE to a alias

libs/vectorise/include/vectorise/uint/uint.hpp
```diff
@@ -52,40 +65,52 @@ class UInt
...
+                                            (sizeof...(T) <= WIDE_ELEMENTS)> * = nullptr>
+  constexpr UInt(T &&... data)
+    : wide_{{std::forward<T>(data)...}}
+  {}
+
+  template <typename... T, std::enable_if_t<meta::Is<BaseType>::SameAsEvery<T...>::value &&
+                                            (sizeof...(T) <= ELEMENTS)> * = nullptr>
```
ejfitzgerald: Same here

libs/vectorise/include/vectorise/uint/uint.hpp
```diff
@@ -213,76 +239,73 @@ class UInt
   static const UInt max;
 
 private:
-  union
+  // TODO (issue 1393): Replace this with `std::array<...>` once C++17 feature set will be enabled.
```
ejfitzgerald: I think that we only really need one todo reference for this issue

libs/vectorise/include/vectorise/uint/uint.hpp
```diff
@@ -213,76 +239,73 @@ class UInt
...
+  WideContainerType wide_;
+
+  static constexpr WideType RESIDUAL_BITS_MASK{~WideType{0} >> RESIDUAL_BITS};
+
+  constexpr ContainerType const &base() const;
+  constexpr ContainerType &      base();
+
+  struct MaxConstr
```
ejfitzgerald: This naming is not 100% obvious. I would be more explicit with `MaxUIntValue` or something similar

ejfitzgerald (changes requested):

ejfitzgerald (approved):

PR #1385, opened 2019-07-21
Time travel across chains

bipll:
This commit introduces methods `MainChain::TimeTravel` and `MainChainProtocol::TimeTravel`. Unlike `Get\i*Chain`, they allow to travel in both directions, up- and downchain.


libs/ledger/include/ledger/protocols/main_chain_rpc_protocol.hpp
```diff
@@ -33,38 +33,44 @@ class MainChainProtocol : public service::Protocol
   enum
   {
     HEAVIEST_CHAIN   = 1,
-    COMMON_SUB_CHAIN = 3
+    COMMON_SUB_CHAIN = 2,
```
ejfitzgerald: I would keep the enums the same for the moment, shouldn't really be an issue but we should get in the habit of trying to keep backward compat.

libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -198,8 +200,8 @@ class MainChain
 
   struct HeaviestTip
   {
-    uint64_t  weight{0};
-    BlockHash hash{GENESIS_DIGEST};
+    std::uint64_t weight{0};
```
ejfitzgerald: Is this `std::` prefix necessary. We haven't done it in the rest of the code base. Happy to change it if the team feels the same way but this is something that we should do in one operation

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -402,8 +405,12 @@ MainChain::Blocks MainChain::GetHeaviestChain(uint64_t limit) const
  * @return The array of blocks
  * @throws std::runtime_error if a block lookup occurs
  */
-MainChain::Blocks MainChain::GetChainPreceding(BlockHash start, uint64_t limit) const
+MainChain::Blocks MainChain::GetChainPreceding(BlockHash start, std::uint64_t limit) const
 {
+  if (limit == 0)
```
ejfitzgerald, line 410: :+1:

scripts/apply_style.py
```diff
@@ -346,6 +350,22 @@ def print_diff_and_fail():
         sys.exit(1)
 
 
+def find_names(root, names):
```
ejfitzgerald: @wkaluza FYI

libs/meta/include/meta/type_traits.hpp
```diff
@@ -147,5 +148,45 @@ struct IsNotImplementedImpl
...
+
+template <class L, class R>
+struct IsSimilar : std::is_same<L, R>
+{
+};
+template <class L, class R>
+static constexpr auto IsSimilarV = IsSimilar<L, R>::value;
+
+template <template <class...> class Ctor, class... LArgs, class... RArgs>
+struct IsSimilar<Ctor<LArgs...>, Ctor<RArgs...>> : std::true_type
```
ejfitzgerald: These don't appear to be used in conjunction with this PR. Can we separate them into there own PR if that is the case

libs/meta/include/meta/value_util.hpp
```diff
@@ -103,5 +103,224 @@ constexpr auto Product(H &&h, Ts &&... ts) noexcept(
...
+    , recover_{values...}
+  {}
+
+  ~KeptValues()
+  {
+    recover_ = std::move(storage_);
+  }
+};
+template <class... StoredTypes>
+inline constexpr KeptValues<StoredTypes...> KeepValues(StoredTypes &... values) noexcept(
```
ejfitzgerald: Same for this

ejfitzgerald (changes requested):

libs/ledger/include/ledger/protocols/main_chain_rpc_protocol.hpp
bipll: It's just the numbering felt a big confusing to me. If subchain is 3 then what should time travel be? Is it 2? Or 5? And where did the original gap come from?

ejfitzgerald: Oh yep it is confusing. There used to be another RPC method that was removed can't remember which one it was but it was a long time ago I believe. 

It's not a big thing, but in the future these will need to stabilise. So I think you should use `2` or `4` as a value. I am happy with either. 😁

libs/ledger/include/ledger/chain/main_chain.hpp
bipll: Afaik, cstdint defines them in std. I wasn't able to find anything about them being injected in top-level so, as I understand, while everyone does it's still non-standard. And a separate PR would really be huge and cover almost the whole codebase. >_< No idea how it should be handled the most graceful.

bipll: I have added `using`s to main_chain.hpp as a preliminary solution, IMO this needs to be done at the rootest include of this whole codebase but not in this PR.

scripts/apply_style.py
wkaluza: I'd rather these changes were reverted, or at least tied more neatly into existing code. If the intent was to restrict the reformatting to a filesystem subtree, it's as simple as plumbing in the cwd into the `walk_source_directories` generator function instead of the `PROJECT_ROOT` constant. The rest of these changes are unnecessary. How many files with forbidden read or write access do we expect in a local working copy? I'm also not happy with introducing an execution path where `commit` is hijacked to be a list of files instead of a string specifying a Git ref. Apart from the fact that this is not what `commit` is for, this kind of type ambiguities and reliance on `isinstance` makes the code hard to follow and maintain. Just because Python allows it doesn't mean it's a good idea.

wkaluza (changes requested):
See comment under apply_style.py

scripts/apply_style.py
bipll: @wkaluza As I remember, these changes turned out necessary as otherwise, at some moment of time, apply_style.py wasn't able to find any file at all to process, unsure how it happened. Btw, I've added (slightly incomplete) path support in the command line but then it somehow disappeared from the script again few months ago.

wkaluza: Seeing that no one else reported issues, this appears to be a problem local to your machine, and that's where I would seek to resolve it. In any case, these changes have nothing to do with this PR, so rather than clutter up this discussion, please revert.

wkaluza (changes requested):

libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -44,6 +44,16 @@
 namespace fetch {
 namespace ledger {
 
+using std::int8_t;
```
ejfitzgerald: It might be non-standard but the rest of the code base doesn't using the std prefix. Lets ditch this and then we can cross the bridge when it doesn't work on a new platform in the future

ejfitzgerald (changes requested):
Looks like a good starting position. I would remove the type_utils stuff for the moment because they don't look like they are used in this change

wkaluza (dismissed):
@ejfitzgerald now that `apply_style.py` is fixed, I'm withdrawing my request for changes and leave the rest in your hands.

ejfitzgerald:
@bipll not sure if you are waiting on my review. From my side I think there are the following changes for this PR:

- [ ] Revert meta template changes (this should be a new PR)
- [ ] Remove the `using std::uint32_t` in the main chain header

Other than that this looks good. So lets make these changes and we can land this feature

ejfitzgerald (approved):

PR #1387, opened 2019-07-22
Add default threshold for DKG

jinmannwong:
Set default threshold for DKG to (n / 2 - 1), where n is the size of the new cabinet.

### DIFF
--- a/libs/dkg/include/dkg/dkg_service.hpp
+++ b/libs/dkg/include/dkg/dkg_service.hpp
@@ -144,16 +144,23 @@ public:
     return state_machine_;
   }
 
-  void ResetCabinet(CabinetMembers cabinet, uint32_t threshold)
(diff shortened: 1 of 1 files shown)


libs/dkg/include/dkg/dkg_service.hpp
```diff
@@ -144,16 +144,23 @@ class DkgService : public ledger::EntropyGeneratorInterface
     return state_machine_;
   }
 
-  void ResetCabinet(CabinetMembers cabinet, uint32_t threshold)
+  void ResetCabinet(CabinetMembers cabinet, uint32_t threshold = UINT32_MAX)
```
n-hutton: Style: the uint_32_max is C-style. We prefer C++-style which is ```std::numeric_limits<uint32_t>::max()```

n-hutton (changes requested):
Minor style

n-hutton (approved):
LGTM

PR #1392, opened 2019-07-23
Refactoring and fixing bug in cabinet reset in DKG

jinmannwong:
Added missing variables that need to be reset when the DKG is rerun with a different committee. Refactored compute qual complaints to remove computations out of broadcasting the complaints for testing.


libs/dkg/src/dkg.cpp
```diff
@@ -754,7 +719,53 @@ bool DistributedKeyGeneration::BuildQual()
 }
 
 /**
- * If in qual a member Computes individual share of the secret key and further computes and
+ * Checks coefficients sent by qual members and puts their address and the secret shares we received
+ * from them into a complaints maps if the coefficients are not valid
+ *
+ * @return Map of address and pair of secret shares for each qual member we wish to complain against
+ */
+std::unordered_map<
```
n-hutton: Stylistic: having a ```using``` for these types would avoid it getting forced onto multiple lines. Also, does iq mean anything in this context?

libs/dkg/src/dkg.cpp
```diff
@@ -754,7 +719,53 @@ bool DistributedKeyGeneration::BuildQual()
...
+
+  uint32_t i  = 0;
+  auto     iq = cabinet_.begin();
+  for (auto const &miner : qual_)
+  {
+    while (*iq != miner)
+    {
+      ++iq;
+      ++i;
+    }
```
n-hutton, line 739: Minor: This code seems like you could use std::distance

n-hutton (dismissed):

libs/dkg/src/dkg.cpp
jinmannwong: Yes, but using std::distance would mean having to start from the beginning of cabinet_ every time to find the index i for the miner in qual.

libs/dkg/src/dkg.cpp
jinmannwong: I have replaces the unordered_map with an alias. iq should not be there and has been removed in the tests branch I am working on. I will remove it here as well.

libs/dkg/src/dkg.cpp
jinmannwong: Yes, but then you would have to go through the same distance in cabinet_ twice: once to find the miner inside cabinet_ (which is what iq does) and then a second time to compute the index of the miner in cabinet. In the while loop both iq and i are iterated at the same time so we only have to traverse cabinet_ once. There might be a nicer way to do it though.

libs/dkg/src/dkg.cpp
jinmannwong: I have replaces the unordered_map with an alias.

n-hutton (approved):

PR #1394, opened 2019-07-23
temporarily disable flaky threadpool - timing out

n-hutton:


ejfitzgerald (dismissed):
I am happy with this can you create an issue to track this. We should look into why this is the case

wkaluza (changes requested):
Please disable the individual gtests - disabling the tests at the build level leads to code rot when dependencies are refactored.

wkaluza (approved):

PR #1396, opened 2019-07-24
Weight sharing

XiaodongWuFetch:


libs/math/include/math/activation_functions/softmax.hpp
```diff
@@ -99,9 +99,20 @@ void Softmax(ArrayType const &array, ArrayType &ret, typename ArrayType::SizeTyp
   {
     details::Softmax2DImplementation(array, ret, axis);
   }
+  else if ((array.shape().size() == 3) && (ret.shape().size() == 3) && (axis == 0 || axis == 1))
+  {
+    ArrayType tmp_ret = ret.Slice(0, 2).Copy().Squeeze();
+    for (size_t i = 0; i < array.shape()[2]; i++)
```
baykaner: SizeType not size_t

also array.shape(2), not shape()[2]

libs/math/include/math/activation_functions/softmax.hpp
```diff
@@ -99,9 +99,20 @@ void Softmax(ArrayType const &array, ArrayType &ret, typename ArrayType::SizeTyp
   {
     details::Softmax2DImplementation(array, ret, axis);
   }
+  else if ((array.shape().size() == 3) && (ret.shape().size() == 3) && (axis == 0 || axis == 1))
+  {
+    ArrayType tmp_ret = ret.Slice(0, 2).Copy().Squeeze();
```
baykaner: this is sort of redundant:
what you really want to do here is something like this:
ArrayType buffer({array.shape(0}, array.shape(1)});

libs/math/tests/math/all_tests/activation_functions/softmax.cpp
```diff
@@ -76,6 +76,26 @@ TYPED_TEST(SoftmaxTest, equal_proportion_test)
...
+  gt_array.Fill(typename TypeParam::Type(double(1) / double(3)));
+
+  typename TypeParam::SizeType axis = 0;
+
+  fetch::math::Softmax(test_array, test_array, axis);
+
+  // test correct values
+  for (std::size_t j = 0; j < gt_array.size(); ++j)
+  {
+    ASSERT_NEAR(double(test_array[j]), double(gt_array[j]), double(1e-7));
```
baykaner: do this with AllClose() and function tolerance

libs/ml/include/ml/graph.hpp
```diff
@@ -218,28 +223,69 @@ void Graph<ArrayType>::ApplyRegularisation()
...
+    std::string const &node_name, std::vector<std::string> const &inputs, Params... params)
+{
+  // guarantee unique op name
+  auto naming_result = UpdateVariableName<OperationType>(node_name);
+
+  std::string updated_name = naming_result.second;
+
+  auto op = std::make_shared<Node<ArrayType, OperationType>>(updated_name, params...);
+
+  nodes_[updated_name] = op;
```
baykaner: lines 275 - 285 are identical for both IfIsShareable and IfIsNotShareable;
Would be nice to make a private member function that does this and just call it from both so as not to have repeated code

libs/ml/include/ml/graph.hpp
```diff
@@ -218,28 +223,69 @@ void Graph<ArrayType>::ApplyRegularisation()
...
 template <typename ArrayType>
 template <class OperationType, typename... Params>
-std::string Graph<ArrayType>::AddNode(std::string const &             node_name,
-                                      std::vector<std::string> const &inputs, Params... params)
+meta::IfIsShareable<ArrayType, OperationType, std::string> Graph<ArrayType>::AddNode(
+    std::string const &node_name, std::vector<std::string> const &inputs, Params... params)
 {
   // guarantee unique op name
-  std::string name = UpdateVariableName<OperationType>(node_name);
+  auto        naming_result = UpdateVariableName<OperationType>(node_name);
```
baykaner: don't create and return a pair with UpdateVariableName; instead return only a bool, and change the input string const & to a string & then update the name internally.

Then, in AddNode you can just do:
bool is_duplicate = UpdateVariableName<OperationType>(node_name);

(and then re-use node_name later instead of "updated_name")

libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -46,16 +46,101 @@ class FullyConnected : public SubGraph<T>
...
+                 fetch::ml::details::RegularisationType regulariser =
+                     fetch::ml::details::RegularisationType::NONE,
+                 DataType    regularisation_rate = static_cast<DataType>(0),
+                 WeightsInit init_mode           = WeightsInit::XAVIER_GLOROT)
+    : in_size_(in)
+    , out_size_(out)
+  {
+    FETCH_UNUSED(init_mode);
+
+    // setup overall architecture of the model
```
baykaner: layer - not model

libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -46,16 +46,101 @@ class FullyConnected : public SubGraph<T>
...
+    WeightsPtrType target_bias_ptr  = std::dynamic_pointer_cast<WeightsType>(target_bias_node_ptr);
+    NodePtrType    weights_node_ptr = this->GetNode(name + "_Weights");
+    WeightsPtrType weights_ptr      = std::dynamic_pointer_cast<WeightsType>(weights_node_ptr);
+    NodePtrType    bias_node_ptr    = this->GetNode(name + "_Bias");
+    WeightsPtrType bias_ptr         = std::dynamic_pointer_cast<WeightsType>(bias_node_ptr);
+
+    // check if the shared weight is compatible with input shape
+    auto w_s = target_weights_ptr->get_weights().shape();
+    auto b_s = target_bias_ptr->get_weights().shape();
+    if ((w_s[0] != out_size_) || (w_s[1] != in_size_) || (b_s[0] != out_size_))
```
baykaner: make these assertions rather than a throw

libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -46,16 +46,101 @@ class FullyConnected : public SubGraph<T>
...
+    {
+      throw std::runtime_error("the shared weight from node: " + target_node_ptr->GetNodeName() +
+                               " is incompatible with the specified shape of this layer");
+    }
+
+    // Set sudo weights to handle initilization issues
```
baykaner: - Isn't there a way around this? This is wasteful to create irrelevant tensors
- Also you mean "pseudo" not sudo (which mean superuser do)

libs/ml/include/ml/meta/ml_type_traits.hpp
```diff
@@ -42,6 +47,13 @@ constexpr bool IsNotTrainable = !IsTrainable<T, OperationType>;
 template <typename T, typename OperationType>
 constexpr bool IsGraph = std::is_base_of<fetch::ml::Graph<T>, OperationType>::value;
 
+template <typename T, typename OperationType>
+constexpr bool IsShareable =
```
baykaner, line 51: please raise an issue that we should implement shareable weights in any other relevant ops/layers

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -95,6 +96,16 @@ class Weights : public fetch::ml::ops::PlaceHolder<T>, public Trainable<T>
   Weights()           = default;
   ~Weights() override = default;
 
+  ArrayPtrType GetSharableWeights()
```
baykaner: I think its spelt shareable ?

libs/ml/tests/ml/layers/fully_connected.cpp
```diff
@@ -28,8 +30,7 @@ class FullyConnectedTest : public ::testing::Test
 {
 };
 
-using MyTypes = ::testing::Types<fetch::math::Tensor<int>, fetch::math::Tensor<float>,
-                                 fetch::math::Tensor<double>,
+using MyTypes = ::testing::Types<fetch::math::Tensor<float>, fetch::math::Tensor<double>,
```
baykaner: actually I think we can handle Tensor<int> here as well.
The only change that needs to be made is to add to math/base_types a method for function tolerance:
fetch::math::meta::IfIsInteger<T, T> function_tolerance()
{
  return T(0);
}

libs/ml/tests/ml/layers/fully_connected.cpp
```diff
@@ -80,6 +81,169 @@ TYPED_TEST(FullyConnectedTest, ops_backward_test)  // Use the class as an Ops
...
+                       g.template AddNode<fetch::ml::layers::FullyConnected<ArrayType>>(
+                           "FC1", {input}, 11u, 20u));
+}
+
+TYPED_TEST(FullyConnectedTest, share_weight_backward_test)
+{
+  using ArrayType = TypeParam;
+  using DataType  = typename ArrayType::Type;
+  using GraphType = typename fetch::ml::Graph<ArrayType>;
+
```
baykaner, line 105: if you add a few using statements this will become more readable, e.g.:
using FCType = typename FullyConnected
using RegType = details::RegularisationType
using WeightsInitType = ops::WeightsInitialisation

libs/ml/tests/ml/layers/fully_connected.cpp
```diff
@@ -80,6 +81,169 @@ TYPED_TEST(FullyConnectedTest, ops_backward_test)  // Use the class as an Ops
...
+  for (size_t i = 0; i < 4; i++)
+  {
+    ASSERT_EQ(g_shared_weights_before[i], g_not_shared_weights_before[i]);
+  }
+
+  // start training
+  // set data
+  ArrayType data;
+  data.Resize({10, 1});
+  for (int i = 0; i < 10; i++)
```
baykaner: SizeType

libs/ml/tests/ml/layers/fully_connected.cpp
```diff
@@ -80,6 +81,169 @@ TYPED_TEST(FullyConnectedTest, ops_backward_test)  // Use the class as an Ops
...
+  {
+    data.Set(i, 0, DataType(i));
```
baykaner: static_cast<DataType>(i)

libs/ml/tests/ml/layers/fully_connected.cpp
```diff
@@ -80,6 +81,169 @@ TYPED_TEST(FullyConnectedTest, ops_backward_test)  // Use the class as an Ops
...
+  fetch::ml::optimisers::SGDOptimiser<ArrayType> g_shared_optimiser(
+      g_shared, {g_shared_input}, g_shared_label, g_shared_error, lr);
+  g_shared_optimiser.Run({data}, data, 1);
+  // Run 1 iteration of training on g not shared
+  fetch::ml::optimisers::SGDOptimiser<ArrayType> g_not_shared_optimiser(
+      g_not_shared, {g_not_shared_input}, g_not_shared_label, g_not_shared_error, lr);
+  g_not_shared_optimiser.Run({data}, data, 1);
+
+  // check that all weights are equal
+  std::cout << "**************************AFTER TRAINING**************************" << std::endl;
```
baykaner: remove print statments

libs/ml/tests/ml/layers/fully_connected.cpp
```diff
@@ -80,6 +81,169 @@ TYPED_TEST(FullyConnectedTest, ops_backward_test)  // Use the class as an Ops
...
+      g_not_shared_statedict_after.dict_["FC5_FullyConnected_Weights"].weights_->Copy());
+  g_not_shared_weights_after.emplace_back(
+      g_not_shared_statedict_after.dict_["FC5_FullyConnected_Bias"].weights_->Copy());
+
+  // check the shared weight graph has tied weights, and the not shared weight graph has untied
+  // weights
+  for (size_t i = 0; i < 2; i++)
+  {
+    ASSERT_TRUE(g_shared_weights_before[i] == g_shared_weights_before[i + 2]);
+    ASSERT_TRUE(g_not_shared_weights_before[i] == g_not_shared_weights_before[i + 2]);
```
baykaner, line 221: this looks like it checks that the untied weights are actually the same as each other (which is the opposite of what the comment claims"

libs/ml/tests/ml/layers/fully_connected.cpp
```diff
@@ -80,6 +81,169 @@ TYPED_TEST(FullyConnectedTest, ops_backward_test)  // Use the class as an Ops
...
+  }
+
+  // check the gradient of the shared weight matrices are the sum of individual weight matrice
```
baykaner, line 236: technically speaking, you do not examine the gradient - you check the change in the weights values which, with SGD optimiser, should be the same thing. I think its fine but update the comment to reflect this.

baykaner (changes requested):

libs/ml/include/ml/layers/fully_connected.hpp
XiaodongWuFetch: haha u caught me. tried to be lazy lol as this wont really affect efficiency given it happens only at initialization.

libs/ml/tests/ml/layers/fully_connected.cpp
XiaodongWuFetch: well I was thinking the other way around. I wanted to inspect the gradient, and that is the reason I chose SGD as optimizer.

libs/ml/include/ml/graph.hpp
```diff
@@ -266,20 +264,13 @@ meta::IfIsNotShareable<ArrayType, OperationType, std::string> Graph<ArrayType>::
...
-  nodes_[updated_name] = op;
-
-  // assign inputs and outputs
-  for (auto const &i : inputs)
-  {
-    nodes_[updated_name]->AddInput(nodes_[i]);
-    nodes_[i]->AddOutput(nodes_[updated_name]);
-  }
+  // assign inputs and outputs to the new node
+  LinkNodesInGraph(updated_name, inputs, op);
```
baykaner, line 273: nice bit of refactoring 👍

libs/ml/include/ml/graph.hpp
```diff
@@ -424,6 +415,23 @@ void Graph<ArrayType>::ApplyGradients(std::vector<ArrayType> &grad)
   }
 }
 
+template <typename ArrayType>
+template <class OperationType, typename... Params>
+void Graph<ArrayType>::LinkNodesInGraph(
```
baykaner, line 430: function comment needed

libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -65,7 +66,7 @@ class FullyConnected : public SubGraph<T>
   {
```
baykaner, line 77: add function comments to both constructors that highlight why we now need two constructors, and why they need to have the same list of parameters (leading to the FECH_UNUSED)

libs/ml/include/ml/node.hpp
```diff
@@ -77,16 +77,15 @@ class Node : public NodeInterface<T>, public O
   std::vector<std::pair<NodeInterface<T> *, ArrayType>> BackPropagateSignal(
       ArrayType const &error_signal) override;
 
-  std::string const &GetNodeName() override
+  void                                    AddInput(NodePtrType const &i) override;
+  void                                    AddOutput(NodePtrType const &o) override;
+  virtual std::vector<NodePtrType> const &GetOutputs() const override;
+  virtual void                            ResetCache(bool input_size_changed) override;
```
baykaner: i believe the virtual is unnecessary here

libs/ml/include/ml/node.hpp
```diff
@@ -77,16 +77,15 @@ class Node : public NodeInterface<T>, public O
   std::vector<std::pair<NodeInterface<T> *, ArrayType>> BackPropagateSignal(
       ArrayType const &error_signal) override;
 
-  std::string const &GetNodeName() override
+  void                                    AddInput(NodePtrType const &i) override;
+  void                                    AddOutput(NodePtrType const &o) override;
+  virtual std::vector<NodePtrType> const &GetOutputs() const override;
```
baykaner: i believe the virtual is unnecessary here

libs/ml/tests/ml/layers/fully_connected.cpp
```diff
@@ -82,45 +92,30 @@ TYPED_TEST(FullyConnectedTest, ops_backward_test)  // Use the class as an Ops
   // No way to test actual values for now as weights are randomly initialised.
 }
 
-TYPED_TEST(FullyConnectedTest, share_weight_with_incompatible_layer)
+TYPED_TEST(FullyConnectedNoIntTest, share_weight_backward_test)
```
baykaner, line 95: great 👍

baykaner (dismissed):
nice work. only a few small changes needed

libs/ml/include/ml/graph.hpp
XiaodongWuFetch: thx~

MissingNO57 (dismissed):
LGTM

MissingNO57 (approved):
LGTM, just wait for green ticks

PR #1404, opened 2019-07-24
[VM] Enhance type deduction for bindings

wkaluza:


libs/vm/include/vm/module.hpp
```diff
@@ -231,6 +205,43 @@ class Module
...
+    template <typename GetterReturnType, typename... GetterArgs, typename SetterReturnType,
+              typename... SetterArgs>
+    ClassInterface &InternalEnableIndexOperator(GetterReturnType (Type::*getter)(GetterArgs...),
+                                                SetterReturnType (Type::*setter)(SetterArgs...))
+    {
+      static_assert(sizeof...(GetterArgs) >= 1, "Getter should take at least one parameter");
+      static_assert(sizeof...(SetterArgs) >= 2, "Setter should take at least two parameters");
+      static_assert(
+          std::is_same<GetterReturnType,
+                       std::remove_cv_t<std::remove_reference_t<
```
ejfitzgerald: does `std::decay` make sense here?

wkaluza: Just to record our conversation just now: I don't think it does.

libs/vm/include/vm/module/base.hpp
```diff
@@ -245,5 +245,13 @@ struct FunctorTraits<ReturnType (Class_::*)(Args...)>
   using args_tuple_type = std::tuple<Args...>;
 };
 
+template <typename T>
+struct TypeFromPtr;
+template <typename T>
+struct TypeFromPtr<Ptr<T>>
```
wkaluza: use GetManagedType

robertdickson (approved):
looks great, let's get this in

PR #1406, opened 2019-07-25
Type dispatch refactored in vm (and in smart contracts)

bipll:
This commit simplifies and deduplicates vm code by getting _absolutely_ rid of numerous `case TypeIds::FieldN: lhsv.primary.fieldN = DoTheSameForEachN(rhsv.fieldN);` switches found all over the vm libs. Every such switch was replaced by a single few-lines call to a very small library function. A similar reduction was applied to smart_contract.cpp

To achieve this, a `Switch` template is introduced, similar to (though more generic than) `ApplyFunctor` found earlier in a now-closed PR (and `ApplyFunctor` is redefined to use `Switch` and be a more convenient tool), and a simple pack mechanics, contained in meta/pack.hpp, to back it up. The latter implements all the same variadic operations found earlier in meta/type_util.hpp, but does encapsulate variadic parameter packs in a special dedicated type Pack so they can be composed, combined and reused.


wkaluza (dismissed):
There's plenty going on here. This PR is both overkill and incomplete.

* A lot of complex code is introduced. It's commendable that it's under test, but otherwise it appears mostly unused.
* In the PR description you claim the Switch has improved asymptotic behaviour after this change. However, on master Switch is unused (other than by CopyReferenceKind, which is itself unused). There's more dead code that's been introduced in a similar manner - Sum and Product immediately come to mind.
* If this was supposed to be a change that paves the way to some optimisations, now would be the time to make the intended changes to production code and establish what speed improvement (if any) is observable using end-to-end benchmarks. Asymptotic behaviour alone is not a meaningful measure of performance.

I would not be comfortable with merging this code on faith and shouldering the associated maintenance burden. If it improves actual end-to-end performance, that's great. As is, I think this PR does more harm than good.

There are a few changes which I found surprising in review, but they pale in comparison to the general concerns listed above.

PR #1410, opened 2019-07-25
implemented fisher yates shuffle algorithm

baykaner:
- this is useful in places where we want to shuffle values in vector deterministically with our lfg/lcg

### DIFF
--- a/libs/core/include/core/random.hpp
+++ b/libs/core/include/core/random.hpp
@@ -19,6 +19,8 @@
 
 #include "core/random/lfg.hpp"
 
+#include <vector>
+
 namespace fetch {
 namespace random {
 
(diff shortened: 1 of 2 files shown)


libs/core/include/core/random.hpp
```diff
@@ -27,5 +29,45 @@ struct Random
...
+ * @tparam T
+ */
+template <typename T>
+static void Shuffle(LaggedFibonacciGenerator<> & gen, std::vector<T> const &in_vec, std::vector<T> &out_vec)
+{
+  out_vec = in_vec;
+
+  for (std::size_t i = out_vec.size() - 1; i > 0; --i)
+  {
+    std::size_t j = gen() % (i + 1);
```
troels-im: I would shift this by 19. LCG is known to be alternating on first bit and subsequent bits to be strongly correlated.
```suggestion
    std::size_t j = (gen() >> 19) % (i + 1);
```

libs/core/include/core/random.hpp
```diff
@@ -27,5 +29,45 @@ struct Random
...
+ * @tparam T
+ */
+template <typename T>
+static void Shuffle(LinearCongruentialGenerator & gen, std::vector<T> const &in_vec, std::vector<T> &out_vec)
+{
+  out_vec = in_vec;
+
+  for (std::size_t i = out_vec.size() - 1; i > 0; --i)
+  {
+    std::size_t j = gen() % (i + 1);
```
troels-im: Same here:
```suggestion
    std::size_t j = (gen()>>19) % (i + 1);
```

troels-im (changes requested):
Some minor changes suggested, but otherwise it looks good.

libs/core/tests/random/shuffle_tests.cpp
```diff
@@ -0,0 +1,78 @@
...
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "core/random.hpp"
+#include "vectorise/fixed_point/fixed_point.hpp"
+
+#include "gtest/gtest.h"
+
+#include <iostream>
```
wkaluza: Please do not include `<iostream>` unless it's actually used. It's a relatively large, expensive header to include, and it does our build times no favours.

troels-im (approved):
LGTM

PR #1414, opened 2019-07-25
Replace library shuffles

baykaner:
- replaces various std::random devices and shuffles with in library lfg to ensure determinisim
- must be merged in after #1410


MissingNO57 (approved):
LGTM

PR #1417, opened 2019-07-26
Tests for DKG and bug fixes

jinmannwong:
DKG is tested for various faults and checked that the correct number of nodes are able to succeed in the DKG. Bug fixes and refactoring as a result of tests are also included.


wkaluza (changes requested):
The DKG tests now take about three minutes - please categorise them as slow tests in the CMake file.

PR #1418, opened 2019-07-26
Sgd optimiser serialiser

baykaner:


libs/ml/include/ml/optimisation/learning_rate_params.hpp
```diff
@@ -0,0 +1,99 @@
...
+  DataType          ending_learning_rate   = static_cast<DataType>(0.0001);
+  DataType          linear_decay_rate      = static_cast<DataType>(0.0000000000001);
+  DataType          exponential_decay_rate = static_cast<DataType>(0.999);
+};
+}  // namespace optimisers
+}  // namespace ml
+
+namespace serializers {
+/**
+ * serializer for conv1d saveable params
```
MissingNO57: Why is serializer (serialise? for British English?)  for conv1D in file learning_rate_params? Shouldn't it be serialise for optimiser params?

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -493,4 +479,110 @@ void Optimiser<T>::ResetGradients()
 
 }  // namespace optimisers
 }  // namespace ml
+
+namespace serializers {
+/**
+ * serializer for conv1d saveable params
```
MissingNO57: Same as above, what it has to do with convolution?

libs/ml/include/ml/optimisation/sgd_optimiser.hpp
```diff
@@ -92,4 +107,36 @@ void SGDOptimiser<T>::ApplyGradients(SizeType batch_size)
 
 }  // namespace optimisers
 }  // namespace ml
+
+namespace serializers {
+/**
+ * serializer for conv1d saveable params
```
MissingNO57: Convolution?

MissingNO57 (changes requested):

libs/ml/include/ml/optimisation/learning_rate_params.hpp
baykaner: I would agree, but unfortunately it is not my choice since it is already serialize everywhere in the repo and it needs to be that namespace. Perhaps someone can put in a PR to replace it everywhere.

MissingNO57 (approved):
LGTM

PR #1419, opened 2019-07-26
Improve build times

ejfitzgerald:


baykaner (dismissed):

bipll:
Interesting. Once again it inlines with negative sizes.

PR #1420, opened 2019-07-27
Move vm_modules implementation to cpp files

wkaluza:


ejfitzgerald (dismissed):
Quite a difficult to review because there are a combination of moves and then namespace changes. May I recommend next time separating these steps into 2 different PRs

ejfitzgerald (approved):

PR #1421, opened 2019-07-27
Add message when unrecognised parameter is passed to constellation

n-hutton:
This pr adds the following behaviour to constellation:

```
$ ./constellation -foobar
[3/3] Linking CXX executable apps/constellation/constellation
 F E ╱     Constellation v0.7.0-alpha3-4-ga7634adf
   T C     Copyright 2018-2019 (c) Fetch AI Ltd.
     H

[E] 2019/07/27 13:12:38 | SettingCollection              : Unrecognised parameter(s) passed.
[E] 2019/07/27 13:12:38 | SettingCollection              : Unrecognised parameter: -foobar
[I] 2019/07/27 13:12:38 | SettingCollection              :
Valid parameters:
-lanes                        "The number of lanes to be used"
-slices                       "The number of slices to be used"
-block-interval               "The block interval is milliseconds"
-standalone                   "Signal the network should run in standalone mode"
-private-network              "Signal the network should run as part of a private network"
-db-prefix                    "The block interval is milliseconds"
-port                         "The starting port for ledger services"
-peers                        "The comma separated list of addresses to initially connect to"
-external                     "This node's global IP address or hostname"
-config                       "The path to the manifest configuration"
-max-peers                    "The max number of peers to connect to"
-transient-peers              "The number of the peers which will be random in answer sent to peer requests"
-peers-update-cycle-ms        "How fast to do peering updates"
-disable-signing              "Disable the signing of all network messages"
-bootstrap                    "Signal that we should connect to the bootstrap server"
-discoverable                 "Signal that this node can be advertised on the bootstrap server"
-host-name                    "The hostname or identifier for this node"
-network                      "The name of the bootstrap network to connect to"
-token                        "The authentication token when talking to bootstrap"
-processor-threads            "The number of processor threads"
-verifier-threads             "The number of verifier threads"
-executors                    "The number of transaction executors"
-experimental                 "The comma separated set of experimental features to enable"
```

Note: exit will be called on bad parameter passing.


libs/settings/src/setting_collection.cpp
```diff
@@ -103,8 +106,36 @@ void SettingCollection::UpdateFromArgs(int argc, char **argv)
...
+
+    for (auto const &setting : settings_)
+    {
+      oss << "\n-" << setting->name() << std::setw(int(30 - setting->name().size())) << " \""
+          << setting->description() << "\"";
+    }
+
+    FETCH_LOG_INFO(LOGGING_NAME, oss.str());
+
+    exit(1);
```
wkaluza: To improve testability, `std::exit` (or a wrapper) should be injected so it can be mocked out.

wkaluza (changes requested):

libs/settings/src/setting_collection.cpp
```diff
@@ -134,7 +134,7 @@ void SettingCollection::UpdateFromArgs(int argc, char **argv)
 
     FETCH_LOG_INFO(LOGGING_NAME, oss.str());
 
-    exit(1);
+    on_fail();
```
wkaluza: This needs a null check, I think.

wkaluza (changes requested):

libs/settings/src/setting_collection.cpp
n-hutton: Thanks, should be done now

wkaluza (dismissed):

apps/constellation/settings.cpp
```diff
@@ -89,7 +87,7 @@ Settings::Settings()
 bool Settings::Update(int argc, char **argv)
 {
   UpdateFromEnv("CONSTELLATION_");
-  UpdateFromArgs(argc, argv);
+  UpdateFromArgs(argc, argv, []() { std::exit(1); });
```
ejfitzgerald: My preference would be to get update from args to return a `bool` and handle this exit logic from here

libs/core/include/core/commandline/parameter_parser.hpp
```diff
@@ -52,7 +52,10 @@ class ParamsParser
 
   std::string GetParam(std::string const &key, std::string const &default_value) const;
 
+  std::map<std::string, std::string> &params();
```
ejfitzgerald:
```suggestion
  std::map<std::string, std::string> const &params() const;
```

ejfitzgerald (changes requested):

ejfitzgerald (approved):

PR #1428, opened 2019-07-29
Improved HTTP server security functions

troels-im:
This PR introduces functions to allow securing the HTTP interface. Specifically, it allows to secure the interface with a header token. This method can easily be extended to other procedures such as key signing etc. Examples of functions:

```
$ curl -i -H "AUTH_TOKEN: " http://localhost:8080/pages
HTTP/1.1 200 OK
content-length: 2
content-type: application/json

{}

$ curl -i -H "AUTH_TOKEN: " http://localhost:8080/pages/1
HTTP/1.1 511 Network Authentication Required
content-length: 23
content-type: text/html

authentication required

$ curl -i -H "AUTH_TOKEN: hello" http://localhost:8080/pages/1
HTTP/1.1 200 OK
content-length: 27
content-type: application/json

{"error": "It's all good!"}

$ curl -i -H "AUTH_TOKEN: hello" http://localhost:8080/throw
HTTP/1.1 500 Internal Server Error
content-length: 31
content-type: text/html

internal error: some exception!
```


libs/http/include/http/authentication_level.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+namespace fetch {
+namespace http {
+
+struct AuthenticationLevel
+{
+  enum
+  {
+    NO_ACCESS     = 0,
+    DEFUALT_LEVEL = 512,
```
ejfitzgerald: typo

libs/http/src/middleware/token_auth.cpp
```diff
@@ -0,0 +1,52 @@
...
+  // return the handler
+  for (auto &c : token_name)
+  {
+    c = static_cast<char>(std::tolower(c));
+  }
+
+  return [token, token_name, level](HTTPRequest &req) {
+    auto header = req.header();
+
+    if (header.Has(token_name))
```
ejfitzgerald: Does it make sense you simple match against the standardised token authentication method here i.e:

`Authorization: Token afafafafafafafafafafafaf`

libs/http/src/middleware/token_auth.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "core/logging.hpp"
+#include "http/middleware/token_auth.hpp"
+
+#include <memory>
+
+namespace fetch {
+namespace http {
+namespace middleware {
+
+HTTPServer::RequestMiddleware TokenAuth(byte_array::ConstByteArray const &token, uint32_t level,
```
ejfitzgerald: I think that is makes the most sense to pass in a reference to a class interface here like:

```
class AuthenticationInterface
{
public:
  virtual uint32_t AuthenticateToken(ConstByteArray const &token) = 0;
};
```
We can provide very simple concrete versions of this initially, but allows this to be easily extended in the future.

ejfitzgerald (changes requested):
I like the tidy up, I think that the authentication can be made more generic though relatively easily and I think that it is worth making the authentication follow the standards practises.

ejfitzgerald (approved):

PR #1430, opened 2019-07-29
Feature/vectorise refactor

markos:
Major vectorise refactor.
AVX2 only, both 128/256-bit supported
int32/int64/float/double and fixed32, fixed64 types supported
Basic operators/comparisons should work.

TODO:
* math functions


libs/meta/include/meta/type_traits.hpp
```diff
@@ -115,27 +108,12 @@ using IfIsUnsignedInteger = EnableIf<IsUnsignedInteger<T>, R>;
...
-using IfIsFixedPoint = EnableIf<IsFixedPoint<T>, R>;
-
-template <typename T, typename R = void>
-using IfIsNotFixedPoint = EnableIf<IsNotFixedPoint<T>, R>;
-
 template <typename T, typename R = void>
 using IfIsNullPtr = EnableIf<IsNullPtr<T>, R>;
 
 template <typename T, typename R = void>
 using IfIsPod = EnableIf<std::is_pod<T>::value, R>;
```
bipll: TBD: `std::is_pod` is deprecated in C++20. There's a fallback `IsPOD` in meta/type_traits.hpp.

libs/math/include/math/distance/minkowski.hpp
```diff
@@ -38,7 +38,7 @@ meta::IfIsMathArray<ArrayType, typename ArrayType::Type> Minkowski(ArrayType con
   auto     b_it = b.begin();
   for (auto &val : a)
   {
-    sum += Pow(Max(val, *b_it) - Min(val, *b_it), n);
+    sum += Pow(fetch::vectorise::Max(val, *b_it) - fetch::vectorise::Min(val, *b_it), n);
```
baykaner, line 42: shouldn't you also be including vectorise/math/min.hpp?

libs/math/include/math/tensor.hpp
```diff
@@ -828,7 +827,7 @@ typename Tensor<T, C>::ViewType Tensor<T, C>::View()
   assert(shape_.size() >= 1);
 
   SizeType N     = shape_.size() - 1;
-  SizeType width = shape_[N] * stride_[N] / padded_height_;
+  SizeType width = std::max(shape_[N] * stride_[N] / padded_height_, static_cast<SizeType>(1));
```
baykaner, line 831: as we discussed, i believe this may cause a problematic edge case: if we construct an empty tensor of size 0 and then call t.View(), what will be the behaviour? it will have a width of 1. is that okay?

libs/ml/include/ml/ops/tanh.hpp
```diff
@@ -73,9 +74,9 @@ class TanH : public fetch::ml::ops::Ops<T>
     for (auto &val : output)
     {
       // Minimum value of tanh is restricted to -1+epsilon
-      fetch::math::Max(val, fetch::math::Add(DataType(-1), epsilon_), val);
+      val = fetch::vectorise::Max(val, fetch::math::Add(DataType(-1), epsilon_));
```
baykaner, line 77: why do you not have an in-place implementation of fetch::vectorise::Max?

libs/vectorise/examples/01_elementwise_manipulation_fetch/main.cpp
```diff
@@ -24,16 +24,16 @@
 #include <cstdlib>
 #include <iostream>
 
-using type        = float;
+using type        = float;  // fetch::fixed_point::fp64_t;
```
baykaner: remove leftover comment

libs/vectorise/examples/02_reduction_fetch/main.cpp
```diff
@@ -24,15 +25,16 @@
 #include <cstdlib>
 #include <iostream>
 
-using type        = float;
+using type        = fetch::fixed_point::fp64_t;
```
baykaner, line 28: did you mean to leave this type changed?

libs/vectorise/include/vectorise/arch/avx2/math/approx_exp.hpp
```diff
@@ -0,0 +1,145 @@
...
+{
+  return VectorRegister<float, 32>(std::exp(x.data()));
+}
+
+inline VectorRegister<double, 64> approx_exp(VectorRegister<double, 64> const &x)
+{
+  return VectorRegister<double, 64>(std::exp(x.data()));
+}
+
+// inline VectorRegister<int32_t, 32> approx_exp(VectorRegister<int32_t, 32> const &x)
```
baykaner: leftover comments

libs/vectorise/include/vectorise/arch/avx2/register_fixed32.hpp
```diff
@@ -0,0 +1,467 @@
...
+  // set each of the 4 values in the vector register to either the solution of the division or 0
+  for (size_t i = 0; i < 8; i++)
+  {
+    ret[i] = d1[i] / d2[i];
+  }
+
+  return VectorRegister<fixed_point::fp32_t, 256>(ret);
+}
+
+/*
```
baykaner: lots of commented out code.
In such cases we usually prefer to raise an Issue and place the commented out code in the issue description.
Then here remove the commented out code and leave a TODO(#issue number)

libs/vectorise/include/vectorise/memory/array.hpp
```diff
@@ -40,10 +41,10 @@ class Array : public VectorSlice<T, type_size>
 {
 public:
   static_assert(sizeof(T) >= type_size, "Invalid object size");
+
   // TODO(issue 1424): check IfIsPodOrFixedPoint memory safe and reinstante appropriate static
   // asserts
-  //  static_assert(std::is_pod<T>::value, "Can only be used with POD types");
-  //  static_assert(meta::IfIsPodOrFixedPoint<T>::value, "can only be used with POD or FixedPoint");
+  static_assert(fetch::math::meta::IsPodOrFixedPoint<T>, "can only be used with POD or FixedPoint");
```
baykaner, line 44: does this fix the TODO/issue on the line above? if so remove that line and consider closing the issue

libs/vectorise/include/vectorise/memory/range.hpp
```diff
@@ -20,14 +20,15 @@
...
 namespace memory {
 
-class TrivialRange
+class Range
 {
 public:
   using SizeType = std::size_t;
 
-  TrivialRange(SizeType const &from, SizeType const &to)
+  Range(SizeType const &from = 0, SizeType const &to = SizeType(-1), SizeType const &step = 1)
```
baykaner: it feels like SizeType(-1) is a slightly strange way to set a default. would it be clearer if were set as numeric_max<SizeType>()?

baykaner (changes requested):

libs/math/include/math/distance/minkowski.hpp
markos: indeed.

libs/ml/include/ml/ops/tanh.hpp
markos: did some tests and I think I forgot to implement it back, will do. same for Min.

libs/vectorise/examples/01_elementwise_manipulation_fetch/main.cpp
markos: fixed

libs/vectorise/examples/02_reduction_fetch/main.cpp
markos: in this case actually yes, I wanted to test SumReduce for fixed point as well. In any case, this is just an example.

libs/vectorise/include/vectorise/arch/avx2/math/approx_exp.hpp
markos: fixed

libs/vectorise/include/vectorise/arch/avx2/register_fixed32.hpp
markos: that was one version that I would like to revisit because of performance gains, but you're right I'll do that in a separate issue

markos: actually not, because it's not bit-perfect to the integer division. I remember now. I'll just remove it.

libs/math/include/math/tensor.hpp
markos: I don't think I can answer that easily. in my mind a zero-width tensor is a special case, so maybe we should add such a fix in View instead so that it doesn't crash. this fix was only for the test (unrelated to vectorisation) to pass.

libs/vectorise/include/vectorise/memory/range.hpp
markos: since it's just a normal C++ type, I've set it to std::numeric_limits<SizeType>::max()

libs/vectorise/include/vectorise/memory/array.hpp
markos: done.

libs/math/include/math/tensor.hpp
markos: If we can't decide a fix now, I would suggest we allow it for now for the test to pass, but create an issue and revisit it with a proper fix.

libs/math/include/math/matrix_operations.hpp
```diff
@@ -98,7 +98,7 @@ void Product(ArrayType const &obj1, typename ArrayType::Type &ret)
 template <typename ArrayType>
 void Sum(ArrayType const &obj1, typename ArrayType::Type &ret)
 {
-  ret = obj1.data().in_parallel().Reduce(memory::TrivialRange(0, obj1.size()),
+  ret = obj1.data().in_parallel().Reduce(memory::Range(0, obj1.size()),
                                          [](typename ArrayType::VectorRegisterType const &a,
                                             typename ArrayType::VectorRegisterType const &b) ->
                                          typename ArrayType::VectorRegisterType { return a + b; });
```
troels-im: Is it not possible to use `auto` here? Would make the code more readable.

libs/math/src/math/linalg/blas/gemm_nn_vector.cpp
```diff
@@ -82,46 +78,38 @@ void Blas<S, Signature(_C <= _alpha, _A, _B, _beta, _C),
...
       auto ret_slice = c.data().slice(c.padded_height() * j, c.height());
       auto slice_c_j = c.data().slice(c.padded_height() * std::size_t(j), c.padded_height());
-      memory::TrivialRange range(std::size_t(0), std::size_t(c.height()));
-      ret_slice.in_parallel().Apply(
+      memory::Range range(std::size_t(0), std::size_t(c.height()));
+      ret_slice.in_parallel().RangedApplyMultiple(
           range,
-          [fetch_vec_beta](VectorRegisterType const &vr_c_j, VectorRegisterType &vw_c_j) {
-            vw_c_j = fetch_vec_beta * vr_c_j;
+          [beta](auto const &vr_c_j, auto &vw_c_j) {
```
troels-im, line 95: 👍

troels-im (dismissed):
Overall this looks good to me. 

Did you write any benchmarks we could try?

markos:
> Overall this looks good to me.
> 
> Did you write any benchmarks we could try?

Not separate examples per se, but here is a short indicative 
For floats:
$ ./example-ledger-03-sum-reduce 100000
1.34162 s to get 0.0999979
$ ./example-ledger-03-sum-reduce-fetch 100000
0.380341 s to get 0.1

doubles:
$ ./example-ledger-03-sum-reduce 100000
1.57286 s to get 0.100001
$ ./example-ledger-03-sum-reduce-fetch 100000
0.739113 s to get 0.100001

fp32_t:
$ ./example-ledger-03-sum-reduce 10000
0.436724 s to get 10005.2804 (0x271547c8)
$ ./example-ledger-03-sum-reduce-fetch 10000
0.0656059 s to get 10005.2804 (0x271547c8)

fp64_t:
$ ./example-ledger-03-sum-reduce 10000
0.68671 s to get 10006.34950306 (0x271659790843)
$ ./example-ledger-03-sum-reduce-fetch 10000
0.435022 s to get 10006.34950306 (0x271659790843)

fp64 is slower because multiplication currently is done by widening the type to 128bit and then taking the low 64bits. However, AVX512 has native 64-bit multiplication. You can expect similar results all around. We can definitely finetune some of the operations a bit more, but I would not expect dramatic differences.

libs/math/include/math/matrix_operations.hpp
markos: done

libs/math/include/math/tensor.hpp
markos: I've tried various tests and I cannot find a better solution than max with 1, or a case where setting it to 1 will cause a problem, so I will resolve this for now and revisit if there are doubts or problems with this fix.

troels-im (dismissed):

baykaner (approved):

PR #1435, opened 2019-07-30
[VM] UTF-8 support (the first take)

### LINKED ISSUE #1009: etch VM playground cannot deal with Chinese characters
Etch playground hangs on running the following code:

> please note that playground code version is behind the current code base `v0.2.0-alpha-35-5-g5dfdb758`

```
function main()

    var myString : String = "hello";
    var myInferredString = "hello again";
    var myChineseUTFString : String = "提供"; // prints on console, not in playground
    var x: String = null;

    Print(myString);
    Print(myInferredString);
    Print(myInferredString + " " + myString);
    Print(myChineseUTFString);

endfunction
```

pbukva:
This enables UTF-8 encoding in our VM Etch code, where the `String`
type seems to be the only place where we need to make effort to
handle UTF-8 encoding explicitly, since `String` is the only place
where we do encoding dependent operations, like using indexes to character position in string,
length counted in characters rather than bytes, etc. ...

 * The missing bit is to extend String::Trim(...) functionality, since for now
we are using just ASCII whitespace characters, however UTF-8 has wider set of
white-space characters.

Resolves #1009


libs/vm_modules/tests/unit/string_tests.cpp
```diff
@@ -551,4 +551,86 @@ TEST_F(
   ASSERT_EQ(stdout.str(), "2 | aaa | ");
 }
 
+TEST_F(StringTests, utf8_split_returns_an_array_of_string_segments_with_the_separators_taken_out)
+{
+  static char const *TEXT = R"(
+    function main()
+      var text = '他身旁放着一支磨尖的花岗岩长矛备用，脚边卧着一头犬族猛兽，它发出的喘hōu声表明它虽已入睡却睡不安稳。';
+      var output = text.split('，');
```
wkaluza: It may be worth using a non-ASCII character as the separator, like `着`. Another test above already checks that split works fine with an ASCII character.

wkaluza (dismissed):

libs/vm_modules/tests/unit/string_tests.cpp
pbukva: That is actually non-asci character (it just looks like normal ASCII coma & space,but in reality it is single utf-8 multi-byte character)

pbukva: But I agree, it would be better to use separator with more than single utf-8 multi-byte character.

wkaluza: Maybe let's split this example string on `着一` so that someone who sees it in the future doesn't get confused by the no-quite-comma.

ejfitzgerald:
Looks good as an initial step. I think it makes sense to add a couple of language tests too for completeness sake

scripts/etch-language-tests/strings/etch-lang-utf8-identifiers.etch_disabled
```diff
@@ -0,0 +1,15 @@
+function 他身旁放着() : String
```
pbukva: This new test file is temporarily disabled, since matcher for etch identifier **IDENT** in our lexer file `libs/vm/scripts/tokeniser.l` is defined using regexp matcher for classic ASCII C lang identifier.

ejfitzgerald: Okay - I think lets remove this test and defer this until having a catch up with @robertdickson about if we want to open this can of worms at this stage 😁

ejfitzgerald (changes requested):
Looks good  - just one minor niggle and lets get it merged

scripts/etch-language-tests/strings/etch-lang-utf8-identifiers.etch_disabled
robertdickson: I'd be very happy not to open the Unicode identifier can o' worms. Otherwise, LGTM.

ejfitzgerald (approved):

PR #1440, opened 2019-07-31
Disable tests but still build them

jinmannwong:
Change the way tests are disabled so that they are still built.


wkaluza (approved):
Cheers for that!

PR #1441, opened 2019-07-31
Timedistributed fc

XiaodongWuFetch:
implement timedistributed option in fully connected layer


libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -99,9 +100,10 @@ class FullyConnected : public SubGraph<T>
                  fetch::ml::details::RegularisationType regulariser =
                      fetch::ml::details::RegularisationType::NONE,
                  DataType    regularisation_rate = static_cast<DataType>(0),
-                 WeightsInit init_mode           = WeightsInit::XAVIER_GLOROT)
+                 WeightsInit init_mode = WeightsInit::XAVIER_GLOROT, bool time_distributed = false)
```
MissingNO57, line 103: I think we should consider some time in the future to pass parameters in a different way because there are too many of them and we would like to preserve default values. It's much easier in Python...

libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -112,19 +114,41 @@ class FullyConnected : public SubGraph<T>
...
     this->SetInput(name + "_Bias", bias_data);
   }
 
   void ShareWeights(NodePtrType target_node_ptr)
   {
+    // get correct name for the layer
+    std::string name = DESCRIPTOR;
+    if (time_distributed_)
+    {
+      name = "TimeDistributed_" + name;
```
MissingNO57: This code is used twice, so please move that logic to private member variable.

libs/ml/include/ml/node.hpp
```diff
@@ -141,6 +141,11 @@ std::shared_ptr<T> Node<T, O>::Evaluate(bool is_training)
     }
     this->Forward(inputs, cached_output_);
     cached_output_status_ = CachedOutputState::VALID_CACHE;
+
+    // debugging purpose
```
MissingNO57: Please remove that temporary code.

MissingNO57 (changes requested):

libs/ml/include/ml/layers/fully_connected.hpp
XiaodongWuFetch: well yeah khan brought that up to me several times about implementing a struct to store all the parameters. That might be a good idea, despite the fact that the user still need to take the effort to instantiate a para struct, and things are less explicit.

MissingNO57 (approved):
LGTM

PR #1442, opened 2019-07-31
Miscellaneous changes

wkaluza:


libs/core/include/core/serializers/binary_interface.hpp
```diff
@@ -76,34 +82,34 @@ class BinaryConstructorInterface
...
 
-      uint8_t size = static_cast<uint8_t>(count);
-      size         = platform::ToBigEndian(size);
+      auto size = static_cast<uint8_t>(count);
+      size      = platform::ToBigEndian(size);
       serializer_.Allocate(sizeof(size));
       serializer_.WriteBytes(&size, sizeof(size));
     }
-    else if (count < (1 << 16))
+    else if (count < (1u << 16u))
```
bipll: The three boundaries in this `if` statement are, in succession, `1 << 8`, `1u << 16u` and `1ull << 32`. Pick a single format perhaps?

libs/core/include/core/serializers/binary_interface.hpp
```diff
@@ -76,34 +82,34 @@ class BinaryConstructorInterface
...
     {
-      uint8_t opcode = static_cast<uint8_t>(CODE16);
+      auto opcode = static_cast<uint8_t>(CODE16);
       serializer_.Allocate(sizeof(opcode));
       serializer_.WriteBytes(&opcode, sizeof(opcode));
 
-      uint16_t size = static_cast<uint16_t>(count);
-      size          = platform::ToBigEndian(size);
+      auto size = static_cast<uint16_t>(count);
+      size      = platform::ToBigEndian(size);
```
bipll, line 101: Can this call be made in the previous line? Still clearly readable yet not introducing (even trivial and ephemeral) recursive dependencies.

libs/core/include/core/state_machine.hpp
```diff
@@ -47,13 +47,14 @@ namespace core {
 template <typename State>
 class StateMachine : public StateMachineInterface, public Runnable
 {
-public:
+private:
```
bipll: Is explicit `private` required here?

bipll (changes requested):

libs/core/include/core/serializers/binary_interface.hpp
wkaluza: done

libs/core/include/core/serializers/binary_interface.hpp
wkaluza: I think it's more readable the way it is. Room for personal taste.

There is no recursion here.

libs/core/include/core/state_machine.hpp
wkaluza: Fair, it's not.

ejfitzgerald:
Reviewed after the merge

PR #1450, opened 2019-08-02
Self attention encoder

XiaodongWuFetch:


libs/ml/include/ml/layers/self_attention_encoder.hpp
```diff
@@ -0,0 +1,154 @@
...
+#include "ml/ops/activations/dropout.hpp"
+#include "ml/ops/add.hpp"
+#include "ml/ops/placeholder.hpp"
+#include "multihead_attention.hpp"
+
+#include <cmath>
+#include <cstdint>
+#include <memory>
+#include <ml/ops/concatenate.hpp>
+#include <random>
```
MissingNO57, line 33: Do you need all of these headers?

MissingNO57 (changes requested):

MissingNO57 (approved):
LGTM

PR #1451, opened 2019-08-02
Pre-DKG sync

jinmannwong:
Pre-DKG sync implemented using RBC and tested on example-dkg-dkg and dkg_standalone. Also rpc for submitting shares during DKG change to direct sends to allow scaling to more than 20 nodes without timeouts. Note: timeouts still occur for rpc calls during signature share submission but nodes are no inhibited by this as long as not too many timeouts occur.

### DIFF
--- a/libs/dkg/include/dkg/pre_dkg_sync.hpp
+++ b/libs/dkg/include/dkg/pre_dkg_sync.hpp
@@ -1,0 +1,60 @@
+#pragma once
+//------------------------------------------------------------------------------
+//
+//   Copyright 2018-2019 Fetch.AI Limited
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "core/byte_array/const_byte_array.hpp"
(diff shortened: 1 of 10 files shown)


libs/dkg/src/pre_dkg_sync.cpp
```diff
@@ -0,0 +1,137 @@
...
+  else
+  {
+    FETCH_LOG_WARN(LOGGING_NAME, "Message from ", from, " discard as different connections");
+  }
+}
+
+bool PreDkgSync::CheckConnections(std::set<MuddleAddress> const &connections)
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+  for (auto const &id : connections)
```
n-hutton, line 70: minor note: probs slightly easier way to do '==' here else debug since you're checking two sets are equal

libs/dkg/src/pre_dkg_sync.cpp
```diff
@@ -0,0 +1,137 @@
...
+  }
+  else
+  {
+    FETCH_LOG_WARN(LOGGING_NAME, "Message from ", from, " discard as different connections");
+  }
+}
+
+bool PreDkgSync::CheckConnections(std::set<MuddleAddress> const &connections)
+{
+  std::lock_guard<std::mutex> lock(mutex_);
```
n-hutton, line 69: either this lock doesn't need to be here protecting peers_, or peers_ in onrbcmessage is unprotected

libs/dkg/include/dkg/pre_dkg_sync.hpp
```diff
@@ -0,0 +1,60 @@
...
+private:
+  using Cabinet = std::set<MuddleAddress>;
+
+  Muddle &                          muddle_;
+  PeersList                         peers_;
+  Cabinet                           cabinet_;
+  RBC                               rbc_;
+  std::mutex                        mutex_;
+  std::unordered_set<MuddleAddress> joined_;
+  uint32_t                          joined_counter_{0};
```
n-hutton, line 51: looks like you could just use joined_.size() here

n-hutton (approved):

PR #1452, opened 2019-08-05
Train/test splitting for dataloaders

MissingNO57:


libs/ml/include/ml/dataloaders/dataloader.hpp
```diff
@@ -50,13 +50,19 @@ class DataLoader
 
   virtual ~DataLoader() = default;
 
-  virtual ReturnType GetNext() = 0;
+  virtual ReturnType GetNext(bool is_validation = false) = 0;
```
baykaner: rather than pass a bool into every function, it seems more logical to have a mode flag for the dataloader, i.e.:

enum class DataLoaderMode
{
    TRAIN,
    VALIDATE,
    TEST,
    INVALID
} 

class DataLoader
{

private:
  DataLoaderMode mode_ = DataLoaderMode::INVALID;
}

baykaner (changes requested):

libs/ml/include/ml/dataloaders/dataloader.hpp
```diff
@@ -44,8 +52,9 @@ class DataLoader
    * @param label_shape
    * @param data_shapes
    */
-  explicit DataLoader(bool random_mode)
+  explicit DataLoader(bool random_mode, DataLoaderMode mode)
```
baykaner: I think DataLoader should have defaults for both modes:
random = false,
DataLoaderMode = TRAIN

baykaner: This is because it's completely valid to use a dataloader without being aware of either of these features

libs/ml/include/ml/dataloaders/dataloader.hpp
```diff
@@ -54,15 +63,18 @@ class DataLoader
 
   virtual ReturnType PrepareBatch(fetch::math::SizeType subset_size, bool &is_done_set);
 
-  virtual std::uint64_t Size() const   = 0;
-  virtual bool          IsDone() const = 0;
-  virtual void          Reset()        = 0;
+  virtual std::uint64_t Size() const        = 0;
```
baykaner: SizeType

libs/ml/include/ml/dataloaders/dataloader.hpp
baykaner: Oh I see that there is a default constructor and default initialisations so I guess this is fine if you just get rid of DataLoaderMode::INVALID, and set default to TRAIN

libs/ml/include/ml/dataloaders/commodity_dataloader.hpp
```diff
@@ -44,15 +44,16 @@ class CommodityDataLoader : public DataLoader<LabelType, InputType>
...
 
-  ReturnType GetNext() override;
-  SizeType   Size() const override;
-  bool       IsDone() const override;
-  void       Reset() override;
+  ReturnType  GetNext() override;
+  SizeType    Size() const override;
+  bool        IsDone() const override;
+  void        Reset() override;
+  inline bool IsValidable() const override;
```
baykaner: i'm not convinced this method should exist.

It seems like the only place it is used is by the estimators to check whether to compute validation loss. However, it seems to me that anything you do with an estimator you MUST be able to validate (or else the estimator should throw when you try to pass it a dataloader that doesn't handle that behaviour).

libs/ml/include/ml/dataloaders/dataloader.hpp
```diff
@@ -54,15 +63,18 @@ class DataLoader
...
+  virtual bool          IsValidable() const = 0;
+  void                  SetMode(DataLoaderMode new_mode);
 
   template <typename X, typename D>
   friend struct fetch::serializers::MapSerializer;
 
 protected:
-  bool random_mode_ = false;
+  bool           random_mode_ = false;
+  DataLoaderMode mode_        = DataLoaderMode::INVALID;
```
baykaner: default to train

(dataloader behaviour & interface should not change - or only minimally - from current implementation if you choose to ignore the new train/val/test split feature)

libs/ml/include/ml/dataloaders/tensor_dataloader.hpp
```diff
@@ -99,12 +107,19 @@ TensorDataLoader<LabelType, InputType>::GetNext()
   {
     throw std::runtime_error("random mode not implemented for tensor dataloader");
   }
+
+  if (this->mode_ == DataLoaderMode::VALIDATE)
```
baykaner: a single if/else is not very expensive, but it is happening inside of the GetNext() method, which is very time critical. We can leave this optimisation for another PR, but I'll point it out here:
Instead of SetMode(new_mode) switching the enum, and then having this if/else based on the enum, we could instead have a third cursor which is a pointer to either the validation or training cursor - Set mode would update the enum and make this third cursor point to either the validation or training data as necessary. Then inside GetNext() we have shorter code and no if/else statement


One way we could make these even faster is if when SetMode(new_mode) is called,

libs/ml/include/ml/dataloaders/tensor_dataloader.hpp
```diff
@@ -113,38 +128,76 @@ template <typename LabelType, typename InputType>
...
+  {
+    return n_train_samples_;
+  }
 }
 
 template <typename LabelType, typename InputType>
 bool TensorDataLoader<LabelType, InputType>::IsDone() const
 {
-  return (data_cursor_ >= data_.shape(batch_data_dim_));
+  if (this->mode_ == DataLoaderMode::TRAIN)
```
baykaner: similarly as with GetNext() we would not need the if/else statement here either.

If you do not implement the suggested change, please raise an issue and leave a TODO in the GetNext

libs/ml/include/ml/dataloaders/tensor_dataloader.hpp
```diff
@@ -113,38 +128,76 @@ template <typename LabelType, typename InputType>
...
+    throw std::runtime_error("Other modes than TRAIN and VALIDATE not supported.");
+  }
 }
 
 template <typename LabelType, typename InputType>
 void TensorDataLoader<LabelType, InputType>::Reset()
 {
-  data_cursor_  = 0;
-  label_cursor_ = 0;
+  if (this->mode_ == DataLoaderMode::TRAIN)
```
baykaner: same again

libs/ml/include/ml/dataloaders/tensor_dataloader.hpp
```diff
@@ -42,24 +42,29 @@ class TensorDataLoader : public DataLoader<LabelType, InputType>
 public:
   TensorDataLoader() = default;
   TensorDataLoader(SizeVector const &label_shape, std::vector<SizeVector> const &data_shapes,
-                   bool random_mode = false);
+                   bool random_mode = false, float VALIDATION_TO_TRAIN_RATIO = 0.0);
```
baykaner: now I think about it, maybe both of these parameters (random & validation ratio) should instead be part of AddData?

what do you think?

baykaner (changes requested):

libs/ml/include/ml/dataloaders/tensor_dataloader.hpp
MissingNO57: I am not convinced, maybe we can add a setter for that.

libs/ml/include/ml/dataloaders/word2vec_loaders/w2v_dataloader.hpp
```diff
@@ -434,6 +451,15 @@ std::vector<std::string> W2VLoader<T>::PreprocessString(std::string const &s)
   return words;
 }
 
+template <typename T>
+void W2VLoader<T>::UpdateCursor()
+{
+  if (this->mode_ != DataLoaderMode::TRAIN)
+  {
+    throw std::runtime_error("Other mode than training not supported yet.");
```
baykaner: It's not a case of 'yet' - it will never be supported for this dataloader.

libs/ml/include/ml/dataloaders/tensor_dataloader.hpp
```diff
@@ -127,20 +133,92 @@ template <typename LabelType, typename InputType>
...
+  *test_cursor_       = test_offset_;
+  *validation_cursor_ = validation_offset_;
+
+  UpdateCursor();
+}
+
+template <typename LabelType, typename InputType>
+void TensorDataLoader<LabelType, InputType>::UpdateCursor()
+{
+  if (this->mode_ == DataLoaderMode::TRAIN)
```
baykaner: switch case is more appropriate than if/else here

libs/ml/examples/mnist_estimator/main.cpp
```diff
@@ -57,9 +57,12 @@ int main(int ac, char **av)
   estimator_config.print_stats    = true;
 
   // setup dataloader
-  auto       data_loader_ptr = std::make_shared<DataLoaderType>(av[1], av[2]);
-  TensorType test_label      = (data_loader_ptr->GetNext()).first;
-  TensorType test_input      = (data_loader_ptr->GetNext()).second.at(0);
+  auto data_loader_ptr = std::make_shared<DataLoaderType>(av[1], av[2]);
+  data_loader_ptr->SetTestRatio(0.2f);
+
+  auto       test_data  = data_loader_ptr->GetNext();
```
baykaner: i think this gets a piece of training data right?
In any case, wouldn't it be better to use your new functionality instead of doing this and call Validate?

libs/ml/include/ml/estimators/dnn_classifier.hpp
```diff
@@ -139,8 +140,10 @@ bool DNNClassifier<TensorType>::Train(SizeType n_steps)
 template <typename TensorType>
 bool DNNClassifier<TensorType>::Train(SizeType n_steps, DataType &loss)
 {
+  data_loader_ptr_->SetMode(dataloaders::DataLoaderMode::TRAIN);
+
   loss = DataType{0};
-  DataType min_loss;
+  DataType min_loss, val_loss = fetch::math::numeric_max<DataType>();
```
baykaner: style, prefer:
DataType min_loss = fetch::math::numeric_max<>();
DataType val_loss = fetch::math::numeric_max<>();

libs/ml/include/ml/estimators/dnn_classifier.hpp
```diff
@@ -187,9 +192,20 @@ bool DNNClassifier<TensorType>::Train(SizeType n_steps, DataType &loss)
 }
 
 template <typename TensorType>
-bool DNNClassifier<TensorType>::Validate()
+bool DNNClassifier<TensorType>::Validate(DataType &validation_loss)
 {
-  throw std::runtime_error("validate not implemented");
+  data_loader_ptr_->SetMode(dataloaders::DataLoaderMode::TEST);
```
baykaner, line 200: slightly confusing - the Validate call has mode TEST?

libs/ml/include/ml/estimators/dnn_classifier.hpp
```diff
@@ -203,9 +219,17 @@ bool DNNClassifier<TensorType>::Predict(TensorType &input, TensorType &output)
 }
 
 template <typename TensorType>
-void DNNClassifier<TensorType>::PrintStats(SizeType epoch, DataType loss)
+void DNNClassifier<TensorType>::PrintStats(SizeType epoch, DataType loss, DataType val_loss)
 {
-  std::cout << "epoch: " << epoch << ", loss: " << loss << std::endl;
+  if (val_loss == fetch::math::numeric_max<DataType>())
```
baykaner: wouldn't it be better for this if/else to be a switch/case based on current mode_ enum?

libs/ml/include/ml/dataloaders/mnist_loaders/mnist_loader.hpp
```diff
@@ -147,8 +183,8 @@ class MNISTLoader : public DataLoader<LabelType, InputType>
   void SetupWithDataFiles(std::string const &images_file, std::string const &labels_file)
   {
     std::uint32_t record_length(0);
-    data_   = ReadMnistImages(images_file, size_, record_length);
-    labels_ = read_mnist_labels(labels_file, size_);
+    data_   = ReadMnistImages(images_file, total_size_, record_length);
+    labels_ = read_mnist_labels(labels_file, total_size_);
```
baykaner: style ReadMNistLabels

libs/ml/tests/ml/dataloader/tensor_dataloader.cpp
```diff
@@ -60,6 +60,8 @@ TYPED_TEST(TensorDataloaderTest, serialize_tensor_dataloader)
   // initialise a new tensor dataloader with the wrong shape parameters
   // these will get updated by deserialisation
   fetch::ml::dataloaders::TensorDataLoader<TypeParam, TypeParam> tdl_2({1, 1}, {{1, 1}});
+  tdl_2.SetTestRatio(0.5);
```
baykaner, line 63: what happens if the dataset is a single example?
Does it have a validation and test size of 0 as intended?

maybe we should cover this with a test?

libs/ml/include/ml/dataloaders/commodity_dataloader.hpp
```diff
@@ -152,6 +170,15 @@ void CommodityDataLoader<LabelType, InputType>::GetAtIndex(CommodityDataLoader::
   buffer_.second = std::vector<InputType>({data_.View(index).Copy()});
 }
 
+template <typename LabelType, typename InputType>
+void CommodityDataLoader<LabelType, InputType>::UpdateCursor()
+{
+  if (this->mode_ != DataLoaderMode::TRAIN)
```
baykaner: might want to double check this with emma - seems like it would be trivial to have a validation support for commodity dataloader

baykaner (changes requested):

libs/ml/include/ml/dataloaders/tensor_dataloader.hpp
```diff
@@ -186,15 +210,16 @@ void TensorDataLoader<LabelType, InputType>::UpdateRanges()
 template <typename LabelType, typename InputType>
 void TensorDataLoader<LabelType, InputType>::UpdateCursor()
 {
-  if (this->mode_ == DataLoaderMode::TRAIN)
+  switch (this->mode_)
   {
+  case DataLoaderMode::TRAIN:
     this->current_cursor_ = train_cursor_;
```
baykaner, line 217: style: put cases inside { } also

baykaner (changes requested):
just one style change

baykaner (dismissed):

baykaner (dismissed):

baykaner (approved):

PR #1454, opened 2019-08-05
Switch operation

XiaodongWuFetch:
1 implement a tensor switch operation


bipll:
`Switch` usually allows for more alternatives. Wouldn't `If` (or `IfThenElse`) make a better name?

MissingNO57 (approved):
LGTM

PR #1455, opened 2019-08-06
Remove path from telemetry update

wkaluza:


wkaluza:
@ejfitzgerald FYI

chr15murray (commented):
This looks good to me

But are you able to see what then displays it as the stats page? Would that just ignore the missing path? Or potentially could it still be trying to find that field to render the stats page?

wkaluza:
As far as I can see, block tracker doesn't care about that path. Don't know about other apps we may have scraping telemetry - I've had no visibility of how telemetry is being consumed.

chr15murray (approved):
Discussed - the page is constructed from the data so there is no need for other changes.

Block explorer performance sync is the only thing that I think will try and consume these stats and query on path but I'd be surprised. 

We don't have that component deployed currently so will deal with that if we need to (When Ed is back). Potentially there are some dashboards that look for 'hot' blocks that are seeing activity but we will need to rethink this approach as it's not scalable in my view to output a metric (or more ) per block.

Thanks Wojciech!

PR #1465, opened 2019-08-07
Refactored string primitives, plus tests for them.

bipll:
Functions in core/strings were refactored (we don't need to extract and compare substrings, or even less to run an in-string search to simply compare ranges of known length at the edges) and de-inlined, following the recent trends in code organization. Unit tests were added for `Replace()`, `ToLower()` and `Trim.*()`.

### DIFF
--- a/libs/core/tests/unit/string_tests.cpp
+++ b/libs/core/tests/unit/string_tests.cpp
@@ -33,7 +33,7 @@ namespace {
 std::string Escape(std::string const &s)
 {
   std::string::size_type i0{};
-  std::string            ret_val(1, '"');
+  std::string            ret_val{'"'};
   for (auto i = s.find_first_of("\n\t"); i != std::string::npos;
        i0 = i + 1, i = s.find_first_of("\n\t", i + 1))
   {
--- a/libs/core/tests/unit/string_tests.cpp
+++ b/libs/core/tests/unit/string_tests.cpp
@@ -24,17 +24,24 @@
 
 #include "gtest/gtest.h"
 
+#include <string>
+
 using namespace fetch::core;
 
-template <class T, class U>
-inline bool Equal(T &&t, U &&u)
-{
-  return t == u;
-}
-
-#define EXPECT_EQUAL(...) EXPECT_TRUE(Equal(__VA_ARGS__))
-
 namespace {
+
+std::string Escape(std::string const &s)
+{
+  std::string::size_type i0{};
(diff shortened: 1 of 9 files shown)


libs/core/tests/unit/string_tests.cpp
```diff
@@ -51,4 +62,85 @@ TEST(StringTests, check_StartsWith)
...
+
+  s = "1234";
+  TrimFromRight(s);
+  EXPECT_EQ(s, "1234");
+
+  s = "    ";
+  TrimFromRight(s);
+  EXPECT_EQ(s, "");
+
+  // TrimFromLeft
```
wkaluza: Rather than add these comments, I think it would be neater to split this test case into three, one for left, right and two-sided trim each.

libs/core/tests/unit/string_tests.cpp
```diff
@@ -17,12 +17,23 @@
...
 
 using namespace fetch::core;
 
+template <class T, class U>
+inline bool Equal(T &&t, U &&u)
+{
+  return t == u;
+}
+
+#define EXPECT_EQUAL(...) EXPECT_TRUE(Equal(__VA_ARGS__))
```
wkaluza: Let's not reinvent the wheel. This code is longer than what the check_replace test could be if you simply used gtest macros. Gtest not handling spaces in assertion macros is limitation that's more easily circumvented by extracting macro arguments to a suitably named local constant. In general I'd be careful with trying to shoehorn as much code as possible into every line (I'm referring to `check_Replace` here).

libs/core/tests/unit/string_tests.cpp
```diff
@@ -51,4 +62,85 @@ TEST(StringTests, check_StartsWith)
...
+  // Trim
+  s = "1234    ";
+  Trim(s);
+  EXPECT_EQ(s, "1234");
+
+  s = "    1234";
+  Trim(s);
+  EXPECT_EQ(s, "1234");
+
+  s = "    1234     ";
```
wkaluza: Please add a test or two which check that tabs and newlines are also trimmed.

wkaluza (changes requested):
Just a few niggles.

libs/core/tests/unit/string_tests.cpp
bipll: > Gtest not handling spaces
It's commas you mean.

wkaluza (approved):

PR #1384, opened 2019-07-21
VM opcode charge

wkaluza:

PR #1473, opened 2019-08-11
Redesign VM bindings

wkaluza:
This PR redesigns the way in which Etch functions (constructors, index operators, etc.) are bound to C++ handlers. There is some coupling and duplication which makes that area of code difficult to change, as experienced in #1384.

Carrying out an Etch function call has multiple aspects to it, each with its own challenges and room for useful abstractions:

1. Associating a C++ handler with the Etch function. Invoking  a C++ handler varies between functions, functors and non-static member functions
2. (since #1384) Estimating the charges involved. If the charge limit is exceeded, the operation is aborted with a runtime error
3. Retrieving argument values passed to the Etch function from the VM stack. This logic in particular could be completely generic, but it is tightly coupled to the type of C++ handler and the type of binding, which resulted in five nearly identical `Invoker` classes, one for functors, function pointers, member functions, constructors and static member functions.
4. Conveying Etch arguments to the C++ handler along with any additional arguments such as `VM*` and `TypeId`. The extra arguments vary depending on the binding.
5. In the case of member functions, retrieving a pointer to the object which is the context of the call
6. Passing the return value (if any) from the call onto the VM stack

This change has the following consequences, in no particular order:
* Binding Etch free functions to functors is no longer separate from binding to function pointers; the functor-specific `*_invoke.hpp` header and `CreateFreeFunction` methods in Module are gone
* Getting Etch argument values from the stack is now uniform across all bindings and handled by the `EtchArgumentTupleComposer`
* Invoking the C++ handler (currently encapsulated by five `*InvokerHelper` templates) is now handled by one of two `Invoker`s: `VmMemberFunctionInvoker` for non-static member functions and `VmFreeFunctionInvoker` for every type of other handler. These invokers also handle putting return values on the stack
* Charges are estimated in two places rather than six: separately for default serialisation constructors, and in the new `EtchInvocation` for all other bindings
* The `EtchInvocation` also handles any extra arguments that a binding may require, such as `VM*` (free functions), `VM*` and `TypeId` (constructors and static member functions) or no extra args at all (non-static member functions), keeping lower-level logic clean.

The code relies quite heavily on tuples so a fair amount of utility code is introduced. The `meta` library also has tests now. Two additions are notable:

* The `FunctorTraits` class template is replaced with the more general `meta::CallableTraits`. A `CallableTraits` instantiation lets clients query the most commonly needed properties of any callable: the return type, parameter types (as a tuple) and the owning type of member functions. Up to now we have relied on function template type deduction to glean callable traits for us, but this means additional boilerplate and accounting for the static typing demands of various callables. An additional benefit is the ability to query and use callable traits where they are needed, rather than holding on to them and laboriously (and often redundantly) passing them around. This helps metaprogramming code look more like normal C++, making it slightly more suitable for human consumption
* `meta::Invoke` and `meta::Apply` are introduced to roughly mimic the `std::invoke` and `std::apply` functionality of C++17 standard library. `Invoke` in particular is shouldering a lot of weight, with `Apply` serving more as a convenience function to avoid unnecessary tuple unpacking.


libs/vm/include/vm/module/member_function_invoke.hpp
```diff
@@ -97,34 +103,34 @@ struct MemberFunctionInvoker
...
+
+template <typename Callable>
+void InvokeMemberFunction(VM *vm, TypeId return_type_id, Callable const &callable)
 {
-  constexpr int num_parameters         = int(sizeof...(Ts));
-  constexpr int first_parameter_offset = num_parameters - 1;
-  constexpr int sp_offset              = num_parameters;
-  using MemberFunction                 = ReturnType (Type::*)(Ts...) const;
-  using MemberFunctionInvoker          = typename MemberFunctionInvoker<
-      Type, ReturnType, MemberFunction>::template Invoker<first_parameter_offset, Ts...>;
```
wkaluza: An example of what I meant in the PR description about passing around inferred arguments: it is redundant to pass around `Type`, `ReturnType` and `Callable`, as the former two are a property of the latter's type. `Type` is not even used until it makes it all the way from these `InvokeMemberFunction` overloads to the `InvokerHelper` at the top of this file.

libs/core/include/core/macros.hpp
```diff
@@ -19,6 +19,14 @@
 
 #define FETCH_UNUSED(x) (void)(x)
 
+template <typename>
+struct UNUSED
+{
+  constexpr static void NOOP()
+  {}
+};
+#define FETCH_UNUSED_ALIAS(x) (void)UNUSED<x>::NOOP();
```
bipll: It is generally not a good idea to include trailing semicolon in macro expansion: a) terminates expressions; b) confuses automatic indentation, e.g. in vim with default plugins.

libs/core/include/core/macros.hpp
```diff
@@ -19,6 +19,14 @@
 
 #define FETCH_UNUSED(x) (void)(x)
 
+template <typename>
+struct UNUSED
```
bipll: Why are struct and method names capitalized?

libs/meta/include/meta/callable_traits.hpp
```diff
@@ -0,0 +1,193 @@
...
+class member_fn_tag;
+class const_member_fn_tag;
+class functor_tag;
+
+template <typename R, typename... A>
+static free_or_static_member_fn_tag ClassifyCallable(R (*&&)(A...));
+template <typename R, typename T, typename... A>
+static member_fn_tag ClassifyCallable(R (T::*&&)(A...));
+template <typename R, typename T, typename... A>
+static const_member_fn_tag ClassifyCallable(R (T::*&&)(A...) const);
```
bipll: Nitpicking, but shouldn't `&`- and `&&`-qualified methods be accounted for?

libs/meta/include/meta/callable_traits.hpp
```diff
@@ -0,0 +1,193 @@
...
+  using OwningType    = void;
+  using ReturnType    = R;
+  using ArgsTupleType = std::tuple<Args...>;
+};
+
+template <typename F>
+struct FreeOrStaticMemberFunctionTraits
+{
+  template <typename ReturnType, typename... Args>
+  static ReturnType get_return(ReturnType (*)(Args...));
```
bipll: This template is defined anew for each instantiation of `FreeOrStaticMemberFunctionTraits`, so could better be defined not as member; next, it does not appear to be used anywhere else, so if defined in-class could be made `private`. Finally, it duplicates `type_util::InvokeResult`.

libs/meta/include/meta/param_pack.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <cstddef>
+#include <tuple>
+#include <utility>
+
+namespace fetch {
+namespace meta {
+
+namespace internal {
+
+template <template <typename...> class Source, typename SourceInstantiation,
```
bipll: Speaking of boilerplate, `Source` is doubled in parameter list: first, apparently excess, as a separate parameter, then as a constructor for `SourceInstantiation`.

libs/meta/include/meta/tuple.hpp
```diff
@@ -17,38 +17,79 @@
...
 #include <utility>
 
 namespace fetch {
 namespace meta {
 
-template <typename T, typename... Ts>
+template <std::size_t, typename...>
+struct SliceTuple;
+template <std::size_t N, typename... Ts>
+struct SliceTuple<N, std::tuple<Ts...>>
```
bipll: The type parameter needs not to be a `std::tuple`, any type with `std::tuple_size` and `std::tuple_element` defined can do.
`Slice` has a particular connotation in various languages, this looks more like a `TupleSplit`.

libs/meta/include/meta/tuple.hpp
```diff
@@ -17,38 +17,79 @@
...
+
+  template <std::size_t... Is>
+  static std::tuple<std::tuple_element_t<Is, std::tuple<Ts...>>...> get_start(
+      std::index_sequence<Is...>);
+  template <std::size_t... Is>
+  static std::tuple<std::tuple_element_t<N + Is, std::tuple<Ts...>>...> get_end(
+      std::index_sequence<Is...>);
+
+public:
+  using Head = decltype(get_start(std::make_index_sequence<start_length>()));
```
bipll: Nitpicking: traditionally, `Head` is simply the very first element of a list. How about `Init`?

libs/meta/include/meta/tuple.hpp
```diff
@@ -17,38 +17,79 @@
...
 };
 
 template <typename... Ts>
 struct GetLastType;
 template <typename... Ts>
 struct GetLastType<std::tuple<Ts...>>
 {
-  using type = typename std::tuple_element_t<sizeof...(Ts) - 1u, std::tuple<Ts...>>;
+  using type =
+      std::tuple_element_t<0, typename SliceTuple<sizeof...(Ts) - 1, std::tuple<Ts...>>::Tail>;
```
bipll: One problem is, how `make_index_sequence` is implemented internally. This can easily run in O(N) instantiations, while the last element can actually be retrieved logarithmically.

libs/meta/include/meta/tuple.hpp
```diff
@@ -17,38 +17,79 @@
...
 
 template <typename... Ts>
 struct RemoveLastType;
 template <typename... Ts>
 struct RemoveLastType<std::tuple<Ts...>>
 {
-  template <std::size_t... Is>
-  static std::tuple<std::tuple_element_t<Is, std::tuple<Ts...>>...> f(std::index_sequence<Is...>);
-  using type = decltype(f(std::make_index_sequence<sizeof...(Ts) - 1u>()));
+  using type = typename SliceTuple<sizeof...(Ts) - 1, std::tuple<Ts...>>::Head;
```
bipll: Do we really need to instantiate `Tail` here?

libs/meta/tests/unit/callable_traits_tests.cpp
```diff
@@ -0,0 +1,253 @@
...
+using NonVoid      = std::string;
+using NoOwningType = void;
+
+#define ASSERTIONS(callable, return_type, args_tuple_type, owning_type)      \
+  {                                                                          \
+    using ClassType = owning_type;                                           \
+    using Callable  = decltype(callable);                                    \
+                                                                             \
+    using Traits = CallableTraits<Callable>;                                 \
+    static_assert(Same<return_type, Traits::ReturnType>(), "");              \
```
bipll: Very subtle nitpicking: AFAIK, the Standard only mandates that the error message is included in compiler output. Would do no harm to double asserted expressions inside parentheses.

libs/meta/tests/unit/callable_traits_tests.cpp
```diff
@@ -0,0 +1,253 @@
...
+class TestFunctions
+{
+public:
+  // static member functions
+  static Void static_member_function_void_no_args()
+  {}
+
+  static NonVoid static_member_function_nonvoid_no_args()
+  {
+    return "";
```
bipll: `return NonVoid()`? Since it is typedefed anyway?

libs/vm/include/vm/module.hpp
```diff
@@ -179,31 +198,36 @@ class Module
       return *this;
     }
 
-    template <typename GetterReturnType, typename... GetterArgs, typename SetterReturnType,
-              typename... SetterArgs>
-    ClassInterface &InternalEnableIndexOperator(GetterReturnType (Type::*getter)(GetterArgs...),
-                                                SetterReturnType (Type::*setter)(SetterArgs...))
+    template <typename Getter, typename Setter>
+    ClassInterface &InternalEnableIndexOperator(Getter const &getter, Setter const &setter)
```
bipll: Here and below, is there any special sense in exactly `const &` templated arguments? It automatically decays template parameters but automatically rejects rvalues.

libs/vm/include/vm/module/constructor_invoke.hpp
```diff
@@ -75,25 +80,38 @@ struct ConstructorInvoker
...
-template <typename ReturnType, typename... Args>
-void InvokeConstructor(VM *vm, TypeId type_id, ReturnType (*constructor)(VM *, TypeId, Args...))
+template <typename... Args>
+struct UnpackConstructorArgs
 {
-  static_assert(IsPtr<ReturnType>::value, "Constructors must return a fetch::vm::Ptr");
+  template <typename Callable>
+  static void Invoke(VM *vm, TypeId type_id, Callable const &callable)
+  {
+    constexpr int num_parameters         = int(sizeof...(Args));
```
bipll: `static`?

libs/vm/include/vm/module/free_function_invoke.hpp
```diff
@@ -20,32 +20,34 @@
 namespace fetch {
 namespace vm {
 
-template <typename ReturnType, typename FreeFunction, typename... Ts>
+template <typename ReturnType, typename Callable, typename... Ts>
 struct FreeFunctionInvokerHelper
 {
-  static void Invoke(VM *vm, int sp_offset, TypeId return_type_id, FreeFunction f,
+  static void Invoke(VM *vm, int sp_offset, TypeId return_type_id, Callable const &callable,
```
bipll: Same as above, `Callable` is not even used on its own in `Invoke`'s body yet the argument is mandated to be `const &`.

libs/vm/include/vm/module/member_function_invoke.hpp
```diff
@@ -20,12 +20,14 @@
...
 namespace vm {
 
-template <typename Type, typename ReturnType, typename MemberFunction, typename... Ts>
+template <typename ReturnType, typename MemberFunction, typename... Ts>
 struct MemberFunctionInvokerHelper
 {
   static void Invoke(VM *vm, int sp_offset, TypeId return_type_id, MemberFunction f,
                      Ts const &... parameters)
   {
+    using Type = typename meta::CallableTraits<MemberFunction>::OwningType;
```
bipll: Isn't `OwningType` more descriptive?

bipll (changes requested):

libs/core/include/core/macros.hpp
wkaluza: ok, it's gone

libs/core/include/core/macros.hpp
wkaluza: No reason. I can switch it to conventional capitalisation.

libs/meta/include/meta/callable_traits.hpp
wkaluza: They could be, but I didn't include them for the same reason I didn't bother accounting for `volatile` non-static member functions or differentiating handlers by calling convention - we don't have a single case that would use those. I could make this class completely general until this file read like this site https://en.cppreference.com/w/cpp/types/is_function but I don't think it's good practice to introduce lots of dead code like that.

libs/meta/include/meta/callable_traits.hpp
wkaluza: Actually, I realised now that `get_return` and `get_args` are unnecessary and can be replaced via deduction. They're both gone now

libs/meta/include/meta/param_pack.hpp
wkaluza: I don't know what you mean here. The intent was to allow clients to easily create typesafe aliases for whichever template they happen to be working with - see for example `UnpackTuple` in this changeset. As it is, its definition is one line and it will only accept instantiations of `std::tuple`, but without putting the burden of implicitly specialising on `std::tuple` on the client. Not to mention that they would have to go through the usual `struct-struct-using` dance to do that, as you cannot specialise type alias templates. How would you have done this?

libs/meta/include/meta/tuple.hpp
wkaluza: Constraining it to `std::tuple` is intentional. Completely generic code is harder to use and read (the main objections people have to adopting metaprogramming more widely). On a separate note, I don't think we should encourage the proliferation of custom tuple types until we really need to. For now `std::tuple` is good enough.

I hesitated between split and slice for a long time. I still don't have a strong opinion, so I'm happy to switch.

libs/meta/include/meta/tuple.hpp
wkaluza: hmm, fair. I think I'll go with `Initial` and `Terminal`, though.

libs/meta/include/meta/tuple.hpp
wkaluza: You're not wrong, but given that there is no runtime overhead (`make_index_sequence` is only called in an unevaluated context here), I'm prepared to live with it and prefer keeping the source code consistent. If we ever deal with tuples thousands of elements long, it might make sense to worry about meta-algorithm complexity like this. Right now our build times face far worse bottlenecks in the form of poor modularisation and bloated headers.

libs/meta/include/meta/tuple.hpp
wkaluza: We do not, but I covered this in the previous comment.

libs/meta/tests/unit/callable_traits_tests.cpp
wkaluza: Nice

libs/meta/tests/unit/callable_traits_tests.cpp
wkaluza: Right, no point to reveal that it's a string by typing `""` if I had hidden `std::string` behind an alias to begin with. That said, I'll go with `{}` to keep things DRY.

libs/vm/include/vm/module.hpp
wkaluza: Right now the way we're using the callables is by capturing them in lambdas, which happens by value anyway. I don't think it makes a difference until we overcome this limitation.

libs/vm/include/vm/module/constructor_invoke.hpp
wkaluza: Yes

libs/vm/include/vm/module/free_function_invoke.hpp
wkaluza: Like I said above, I don't think it makes a difference. This code needs some love badly anyway. I have an idea how to rearchitecture it the `*_invoke.hpp` headers where we don't have to repeat everything five (now - four) times and we don't pass around so many arguments without using them (`sp_offset` in the `Invoker` classes gives me nightmares). Let's have this conversation after I cut the code down to size.

libs/vm/include/vm/module/member_function_invoke.hpp
wkaluza: It is, I was being lazy.

libs/meta/include/meta/callable_traits.hpp
wkaluza: To be clear, this does not mean we should not use any ref or noexcept-qualified functions in our APIs, only that we should write supporting code for them once it is actually needed.

libs/meta/include/meta/param_pack.hpp
bipll: What I mean is as it's written it is invoked as `ConveyTypeParameterPack<Source, Source<int, char>, Dest>`. Note the `Source` typed twice. The first template parameter is excessive as it can be deduced from the second one.

libs/vm/include/vm/module.hpp
bipll: So, for instance, by-value can be facilitated if we have it as rvalue; lvalues mandate copying.
I'm not saying it is exactly required but if we have two options that work approx. the same on current codebase but one of them is in theory better than the other, why go with the other?

wkaluza: If you put it like that, I suppose it makes sense to pass by value.

libs/meta/include/meta/param_pack.hpp
wkaluza: I explained above why I chose to write it the way I did. `ConveyTypeParameterPack` is a low-level abstraction that will likely be (or currently is, anyway) used via semantic aliasing to improve code readability - `UnpackTuple` being an example.

Let's compare the current API in action and what it could look like if it relied on deduction.

* Current definition (3 template arguments):
```
template <template <typename...> class, typename, template <typename...> class>
struct ConveyTypeParameterPack;
template <template <typename...> class Source, template <typename...> class Destination,
          typename... Args>
struct ConveyTypeParameterPack<Source, Source<Args...>, Destination>
{
  using type = Destination<Args...>;
};
```

Current usage (direct instantiation):
```
ConveyTypeParameterPack<Source, Source<T1, T2>, Destination>;
```

Current usage (type-safe alias):
```
template <typename Tuple, template <typename...> class Destination>
using UnpackTuple = ConveyTypeParameterPack<std::tuple, Tuple, Destination>;
```

* Remove redundant argument and rely on deduction. Definition:
```
template <typename, template <typename...> class>
struct ConveyTypeParameterPack;
template <template <typename...> class Destination, template <typename...> class Source,
          typename... Args>
struct ConveyTypeParameterPack<Source<Args...>, Destination>
{
  using type = Destination<Args...>;
};
```

Direct usage indeed becomes simpler:
```
ConveyTypeParameterPack<Source<T1, T2>, Destination>;
```

Definition of `UnpackTuple` with type-safety equivalent to previous one:
```
template <typename Tuple, template <typename...> class Destination>
struct UnpackTupleImpl;
template <template <typename...> class Destination, typename... Args>
struct UnpackTupleImpl<std::tuple<Args...>, Destination>
{
  using type = ConveyTypeParameterPack<std::tuple<Args...>, Destination>;
};
template <typename Tuple, template <typename...> class Destination>
using UnpackTuple = typename UnpackTupleImpl<Tuple, Destination>::type;
```

Can you think of a more parsimonious way to define `UnpackTuple` while maintaining specificity for `std::tuple` as the source template? If not, then I think my point is clear - at this early stage, the usage patterns of `ConveyTypeParameterPack` do not justify bloating one usage scenario to four times the size, just so the other can save a single word. Apart from the obvious maintenance considerations, this kind of unnecessary boilerplate is precisely what gives metaprogramming a bad name with so many developers.

bipll: If you want it to be `std::tuple`-specific why not keep a single partial instantiation?

libs/vm/include/vm/module.hpp
bipll: Quite reasonable.

libs/meta/include/meta/param_pack.hpp
wkaluza: I'm not going to guess what you men this time. What partial instantiation would that be? I posed a question in my previous comment and you ignored it. Kindly post an example like I did and we can compare the pros and cons of your solution. Otherwise there is no discussion to be had here.

bipll: Actually, you can keep both:
```
template <typename, template <typename...> class>
struct ConveyTypeParameterPack;

template <template <typename...> class Destination,
        template <typename...> class Source,
        typename... Args>
struct ConveyTypeParameterPack<Source<Args...>, Destination>
{
  using type = Destination<Args...>;
};

template <template<class...> class,
          typename,
          template <typename...> class>
struct CheckedConveyTypeParameterPack;

template <template<class...> class Source,
          template<class...> class Dest,
          class... Args>
struct CheckedConveyTypeParameterPack<Source, Source<Args...>, Dest>
  : ConveyTypeParameterPack<Source<Args...>, Dest> {};
```
But whatever.

libs/meta/tests/unit/callable_traits_tests.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+using namespace ::testing;
+
+namespace fetch {
+namespace meta {
+
+namespace {
+
+using EmptyArgs = std::tuple<>;
+#define ARGS std::size_t const &, int, std::vector<std::string> &
```
bipll, line 39: `std::size_t const &`? :)

libs/meta/include/meta/apply.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <type_traits>
+#include <utility>
+
+namespace fetch {
+namespace meta {
+
+namespace internal {
+
+template <class F, typename Tuple, std::size_t... I>
+constexpr decltype(auto) Apply(F &&f, Tuple &&t, IndexSequence<I...> &&)
```
bipll: Small types (which empty types definitely are) are better passed by value.

libs/meta/include/meta/callable_traits.hpp
```diff
@@ -0,0 +1,175 @@
...
+ *
+ *     // this works
+ *     using desired_overload = void (OverloadedFunctor::*)(int) const;
+ *     using OverloadedFunctorTraits = CallableTraits<desired_overload>
+ *
+ * @tparam F callable type
+ */
+template <typename F>
+struct CallableTraits
+  : internal::CallableTraitsImpl<typename std::decay_t<F>, internal::ClassifyCallable<F>>
```
bipll: `typename` is not needed before type aliases.

libs/meta/include/meta/tuple.hpp
```diff
@@ -26,34 +26,105 @@
...
+  static std::tuple<std::tuple_element_t<Is, std::tuple<Ts...>>...> get_start(IndexSequence<Is...>);
+  template <std::size_t... Is>
+  static std::tuple<std::tuple_element_t<N + Is, std::tuple<Ts...>>...> get_end(
+      IndexSequence<Is...>);
+
+public:
+  using Initial  = decltype(get_start(std::make_index_sequence<start_length>()));
+  using Terminal = decltype(get_end(std::make_index_sequence<end_length>()));
+
+  static_assert(std::tuple_size<Initial>::value == start_length, "");
```
bipll, line 55: These asserts probably better be found in tests, not the classdef itself.

libs/meta/include/meta/tuple.hpp
```diff
@@ -26,34 +26,105 @@
...
 {
-  template <std::size_t... Is>
-  static std::tuple<std::tuple_element_t<Is, std::tuple<Ts...>>...> f(std::index_sequence<Is...>);
-  using type = decltype(f(std::make_index_sequence<sizeof...(Ts) - 1u>()));
+  using type                        = std::tuple<Ts...>;
+  static constexpr std::size_t size = std::tuple_size<type>::value;
+
+private:
+  template <std::size_t N>
+  struct TakeInitialImpl : TupleOperations<typename SplitTuple<N, type>::Initial>
```
bipll: Empty struct with a single parent could simply be `using`.

libs/meta/include/meta/tuple.hpp
```diff
@@ -26,34 +26,105 @@
...
+struct ModifyTuple<std::tuple<>> : TupleOperations<std::tuple<>>
 {
-  using type = typename std::tuple_element_t<sizeof...(Ts) - 1u, std::tuple<Ts...>>;
 };
+template <typename T, typename... Ts>
+struct ModifyTuple<std::tuple<T, Ts...>> : TupleOperations<std::tuple<T, Ts...>>
+{
+  using FirstType = T;
+  using LastType  = std::tuple_element_t<
+      0, typename TupleOperations<std::tuple<T, Ts...>>::template TakeTerminal<1>::type>;
```
bipll, line 90: `std::tuple_element_t<sizeof...(Ts), type>`?

libs/meta/tests/unit/same.hpp
```diff
@@ -0,0 +1,32 @@
...
+//
+//------------------------------------------------------------------------------
+
+#include <type_traits>
+
+namespace fetch {
+namespace meta {
+
+template <typename T1, typename T2>
+constexpr bool Same()
```
bipll: Aside from an empty pair of parentheses, this function is no different in invocation from `std::is_same`. Perhaps it makes sense to add arguments, to retrieve types from, or change it into simply a template alias to `std::is_same`.

libs/meta/include/meta/param_pack.hpp
wkaluza: Thank you for posting the example. I think it makes the point very well: removing that first argument leads to pointless code bloat. It even fails on its own terms: you object to typing `Source` twice, but typing an extra `Checked` prefix every time is somehow ok? I'm glad we reached an agreement in the end.

bipll (changes requested):

libs/meta/include/meta/tuple.hpp
wkaluza: I disagree - they're in place as sanity checks

libs/meta/include/meta/tuple.hpp
wkaluza: ok

libs/meta/include/meta/callable_traits.hpp
wkaluza: ok

libs/meta/include/meta/tuple.hpp
wkaluza: no

wkaluza: no

libs/meta/tests/unit/same.hpp
wkaluza: ok

libs/meta/include/meta/apply.hpp
wkaluza: On what basis? And again, I'm not about to start guessing which type you're referring to. If you really have to review work in progress (as indicated by the label), then please be more explicit in your comments. This applies to all future reviews.

ejfitzgerald (approved):
Looks good to be but if @robertdickson is happy then so am I

PR #1474, opened 2019-08-11
Simplified variant

bipll:
This commit slightly reduces amount of templates and reduces complexity in basic vm headers, namely variant.hpp and object.hpp.


libs/vm/include/vm/variant.hpp
```diff
@@ -36,208 +36,211 @@ union Primitive
...
   {
     Construct(other, other_type_id);
   }
 
   template <typename T, typename std::enable_if_t<IsPtr<T>::value> * = nullptr>
-  Variant(T &&other, TypeId other_type_id)
+  Variant(T &&other, TypeId other_type_id) noexcept
   {
-    Construct(std::forward<T>(other), other_type_id);
+    Construct(std::move(other), other_type_id);
```
ejfitzgerald: Why not `std::forward`?

libs/vm/include/vm/object.hpp
```diff
@@ -328,7 +257,7 @@ class Ptr
     return *this;
   }
 
-  Ptr &operator=(Ptr &&other)
+  constexpr Ptr &operator=(Ptr &&other) noexcept(noexcept(std::declval<Ptr>().Release()))
```
ejfitzgerald: Honestly not a fan of all this conditional `noexpect` ness.

ejfitzgerald (changes requested):
I like a lot of the clean ups I think that I would not have gone down the `constexpr` `noexcept` route to the same degree. What was your thinking?

If we are not planning on using it in the short term it would be my preference to strip it especially considering that readability cost of the conditional `noexcept`ness

libs/vm/include/vm/variant.hpp
bipll: Currently, `IsPtr` is only true for `Ptr<T>` so this overload is only considered when `T` is `Ptr<U>` for some `U`, so `other` is an rvalue reference, and this wording is shorter.
I'm not sure, if `IsPtr` is ever to be true on references, then this line should be left intact. Should it?

bipll:
> If we are not planning on using it in the short term it would be my preference to strip it especially considering that readability cost of the conditional `noexcept`ness

My idea is, if these guarantees are ever needed, then it would be nice to have them (even if they're not adding any performance gain now). And conditional `noexcept`ness is added to merely require as little as needed while delivering as much as possible. I can strip them here, but frankly, function bodies like `i16 = value;` look like they're begging to be wrapped in `constexpr noexcept`.

libs/vm/include/vm/variant.hpp
ejfitzgerald: 👍 - nope it shouldn't be - do you think that it is worth some kind of static assert here to catch that?

ejfitzgerald (commented):
Looks good to me @wkaluza maybe just have a look over this too?

libs/vm/include/vm/variant.hpp
wkaluza: @ejfitzgerald  has a point - as things stand now `move` is appropriate here, but only under assumptions which are implicit in the code and which would be easy to violate by making reasonable changes to client code. Given that `forward` works, has no performance penalty, and requires no assertions, it is preferable to using `move` here.

wkaluza (dismissed):
While it was a good spot, I see dropping `forward` in favour of `move` in `variant.hpp` as an unnecessary risk. Other than that, I'd be ok ok with these changes. FYI @ejfitzgerald

ejfitzgerald (approved):
Looking good, let’s get this in

wkaluza (approved):

PR #1483, opened 2019-08-12
Broadcast without echo

jinmannwong:
Send broadcast messages using direct send function of endpoint, which does not echo. Significantly reduces the completion time of DKG.


n-hutton (approved):
lgtm

PR #1485, opened 2019-08-12
Committee renewal

troels-im:
This PR contains an initial version of committee renewal.


ejfitzgerald (commented):
Seems like a good starting point. Would be worth checking with @n-hutton and @jinmannwong I have not spun up on consensus at the moment since my head is still very much in the network stack

apps/beacon/main.cpp
```diff
@@ -0,0 +1,177 @@
...
+};
+
+int main()
+{
+  constexpr uint16_t number_of_nodes    = 16;
+  constexpr uint16_t cabinet_size       = 4;
+  constexpr uint16_t number_of_cabinets = number_of_nodes / cabinet_size;
+
+  // Initialising the BLS library
+  crypto::bls::Init();
```
n-hutton, line 106: Seems like the codebase indicates to use ```BLSInitialiser``` here

libs/vm/src/generator.cpp
```diff
@@ -416,10 +416,10 @@ void Generator::HandleWhileStatement(IRBlockNodePtr const &block_node)
   function_->AddInstruction(jf_instruction);
   AddLineNumber(condition_node->line, jf_pc);
 
-  uint16_t const while_block_start_pc = uint16_t(function_->instructions.size());
+  uint16_t const while_round_start_pc = uint16_t(function_->instructions.size());
```
n-hutton: is this supposed to be here?

libs/core/include/core/serializers/binary_interface.hpp
```diff
@@ -114,6 +114,9 @@ class BinaryConstructorInterface
           std::string("Cannot create container type with more than 1 << 32 elements"));
     }
 
+    // Allocating needed mempry
+    serializer_.Allocate(count);
```
n-hutton, line 118: unrelated?

troels-im: It's related in that only BLS uses the binary serialiser and this solves a problem with it.

libs/vm/src/generator.cpp
troels-im: Good catch.

libs/beacon/include/beacon/beacon_setup_service.hpp
```diff
@@ -0,0 +1,147 @@
...
+  struct DeliveryDetails
+  {
+    bool             was_delivered{false};
+    service::Promise response{nullptr};
+  };
+
+  BeaconSetupService()                           = delete;
+  BeaconSetupService(BeaconSetupService const &) = delete;
+  BeaconSetupService(BeaconSetupService &&)      = delete;
+  explicit BeaconSetupService(Endpoint &endpoint, Identity identity);
```
n-hutton: no need for explicit here

PR #1487, opened 2019-08-13
DKG related improvements and bug fixes

jinmannwong:
1. Refactored API for processing messages in DKG and put in catch for messages which cast to null
2. Bug fix in RBC for delivering old messages and data race
3. Bug fix in DKG related to receiving messages before qualified set is built


libs/dkg/examples/dkg/main.cpp
```diff
@@ -148,7 +145,7 @@ int main()
     }
 
     // Need to increase this depending on number of nodes to complete 3 rounds of DRB
-    std::this_thread::sleep_for(std::chrono::seconds(25));
+    std::this_thread::sleep_for(std::chrono::seconds(600));
```
n-hutton: Does your dkg_service have is_synced_ ? you could use that

libs/dkg/include/dkg/dkg_complaints_manager.hpp
```diff
@@ -59,8 +59,8 @@ class ComplaintsManager
 
   void ResetCabinet(uint32_t cabinet_size);
   void Count(MuddleAddress const &address);
-  void Add(std::shared_ptr<ComplaintsMessage> const &msg_ptr, MuddleAddress const &from_id,
-           uint32_t from_index, MuddleAddress node_address);
+  void Add(ComplaintsMessage const &msg, MuddleAddress const &from_id, uint32_t from_index,
```
n-hutton, line 62: This is different now, but in the previous code it was a ref to a shared pointer - was this done so it could be null?

libs/dkg/src/dkg.cpp
```diff
@@ -198,6 +198,9 @@ void DistributedKeyGeneration::BroadcastQualCoefficients()
       static_cast<uint8_t>(State::WAITING_FOR_QUAL_SHARES), coefficients, "signature"}});
   complaints_answer_manager_.Clear();
   state_ = State::WAITING_FOR_QUAL_SHARES;
+  std::unique_lock<std::mutex> lock{mutex_};
```
n-hutton: I would suggest to write it like this:

```
{
  unique_lock<> {mutex_}
  thing...
}
```

Since doing a manual unlock opens yourself to the chance of forgetting to unlock it at some point

libs/dkg/src/dkg.cpp
```diff
@@ -522,7 +547,9 @@ void DistributedKeyGeneration::OnNewCoefficients(
         return;
       }
     }
-    ++A_ik_received_;
+    std::unique_lock<std::mutex> lock{mutex_};
```
n-hutton: same here

n-hutton (changes requested):

libs/dkg/include/dkg/dkg_complaints_manager.hpp
jinmannwong: It should not be null and that is why it has been changed. The message should not get to this stage if it is null but I have changed to reference to be safe.

libs/dkg/src/dkg.cpp
jinmannwong: Thanks, I will make the change.

libs/dkg/examples/dkg/main.cpp
jinmannwong: Yes, I didn't realise that had been added. Thanks.

n-hutton (approved):

PR #1490, opened 2019-08-14
Feature/consensus telemetry add

n-hutton:


apps/constellation/constellation.hpp
```diff
@@ -217,6 +219,8 @@ class Constellation : public ledger::BlockSinkInterface
   HttpServer  http_;                  ///< The HTTP server
   HttpModules http_modules_;          ///< The set of modules currently configured
   /// @}
+
+  telemetry::GaugePtr<uint64_t> connected_peers_gauge_;
```
n-hutton, line 223: create here with headers

apps/constellation/constellation.cpp
```diff
@@ -486,24 +488,17 @@ void Constellation::Run(UriList const &initial_peers, core::WeakRunnable bootstr
...
   {
-    // wait until we have connected to as many peers as are required for DKG
-    if (dkg_ && muddle_.AsEndpoint().GetDirectlyConnectedPeers().size() + 1 == committee_size)
-    {
-      // Note: the DKG will already have its cabinet reset by this point
-      if (!dkg_attached)
-      {
-        // Required until we can guarantee the DRB isn't vulnerable to races
-        std::this_thread::sleep_for(std::chrono::milliseconds(5000));
+    connected_peers_gauge_->set(muddle_.AsEndpoint().GetDirectlyConnectedPeers().size());
```
n-hutton, line 491: set here

libs/dkg/include/dkg/pre_dkg_sync.hpp
```diff
@@ -29,32 +32,43 @@ namespace dkg {
...
+
+  uint64_t start_time_ = std::numeric_limits<uint64_t>::max();
 
   void OnRbcMessage(MuddleAddress const &from, std::set<MuddleAddress> const &connections);
-  bool CheckConnections(std::set<MuddleAddress> const &connections);
-  void ReceivedDkgReady();
+  bool CheckConnections(MuddleAddress const &from, std::set<MuddleAddress> const &connections);
+  void SetStartTime();
+
+  telemetry::GaugePtr<uint64_t> sync_time_gauge_;
```
n-hutton, line 69: gague + counter here (remember _total in counter descriptor)

libs/dkg/examples/dkg/main.cpp
```diff
@@ -144,22 +135,9 @@ int main()
       member->reactor.Start();
     }
 
-    // Sleep until everyone has finished
-    uint32_t qq = 0;
-    while (qq != cabinet_size)
+    while (!committee[0]->dkg_service.IsSynced())
```
jinmannwong: Should we not check everyone reaches IsSynced()? Otherwise they might be still stuck at some late stage in the DKG.

libs/dkg/include/dkg/pre_dkg_sync.hpp
```diff
@@ -29,32 +32,43 @@ namespace dkg {
...
-  Cabinet                           cabinet_;
-  RBC                               rbc_;
-  std::mutex                        mutex_;
-  std::unordered_set<MuddleAddress> joined_;
-  uint32_t                          joined_counter_{0};
-  bool                              committee_sent_{false};
-  bool                              ready_{false};
+  Endpoint &     endpoint_;
+  CabinetMembers members_;
+  Cabinet        cabinet_;
```
jinmannwong: We don't need cabinet_ and members_. Since you have removed peers_ we can just keep cabinet_.

libs/dkg/src/pre_dkg_sync.cpp
```diff
@@ -18,120 +18,155 @@
...
+    std::unordered_set<MuddleAddress> &peer_connections = other_peer_connections[from];
+
+    for (auto const &connection : connections)
     {
-      FETCH_LOG_WARN(LOGGING_NAME, "Message contains node ", id, " not in peers list");
-      return false;
+      peer_connections.insert(connection);
     }
+
+    ready_peers_++;
```
jinmannwong, line 86: There should be a check before this that you are not receiving a connections message from the same peer more than once. Otherwise the same person can spam connection messages to you and make ready_peers_ reach the threshold.

jinmannwong (changes requested):

libs/dkg/src/pre_dkg_sync.cpp
jinmannwong: I realise this was not the case in the original code, but changed it locally to just check the set joined_ (as you wrote in your previous review comment) but did not put in a PR for it.

libs/dkg/examples/dkg/main.cpp
n-hutton: You're right

libs/dkg/src/pre_dkg_sync.cpp
n-hutton: yup, fixed

PR #1494, opened 2019-08-14
Refactor messages in RBC

troels-im:
This PR refactors a few things related to the messages in the RBC.


libs/dkg/include/dkg/rbc_messages.hpp
```diff
@@ -94,163 +102,82 @@ class RBCMessage
...
-  {}
-  explicit RAnswer(RBCSerializer &serialiser)
-    : RMessage{MessageType::RANSWER}
-  {
-    serialiser >> channel_ >> id_ >> counter_ >> message_;
-  }
-  ~RAnswer() override = default;
-};
+using RBroadcast = RMessageImpl<RBCMessage::MessageType::R_BROADCAST>;
+using RRequest   = RMessageImpl<RBCMessage::MessageType::R_REQUEST>;
```
jinmannwong, line 176: A request does not contain any message or hash. It is only needs to contain the channel, id and counter.

libs/dkg/include/dkg/rbc_messages.hpp
```diff
@@ -46,14 +46,27 @@ class RBCMessage
...
+    : channel_{channel}
+    , id_{id}
+    , counter_{counter}
+    , payload_(std::move(msg))
+  {}
+  explicit RBCMessage(RBCSerializer &serialiser)
+  {
+    serialiser >> channel_ >> id_ >> counter_ >> payload_;
+  }
+
```
jinmannwong, line 68: I don't think the RBCMessage constructors should be public. One should not be able to create a message without specifying a type for it.

libs/dkg/include/dkg/rbc_messages.hpp
```diff
@@ -46,14 +46,27 @@ class RBCMessage
...
-    RECHO,
-    RREADY,
-    RREQUEST,
-    RANSWER
+    R_BROADCAST,
+    R_ECHO,
+    R_READY,
+    R_REQUEST,
+    R_ANSWER,
+    R_SEND
```
jinmannwong: R_SEND is not needed. This was in the consensus repo and is what is called R_BROADCAST here.

troels-im: OK, will change accordingly.

libs/dkg/include/dkg/rbc_messages.hpp
troels-im: Yes, I agree, but this will first be changed in subsequent PRs as this is just the first in many minor changes.

libs/dkg/include/dkg/rbc_messages.hpp
troels-im: It is worth adding an empty payload to request as this streamlines all message forms and will simplify the design significantly in coming PRs. The overhead of doing this is 4 bytes.

PR #1499, opened 2019-08-14
Deprecation of RBCEnvelope and fixing attack vector

troels-im:
This PR deprecates the `RBCEnvelope` and fixes an attack vector where a invalid payload causes an exception.


libs/dkg/src/rbc.cpp
```diff
@@ -73,11 +73,21 @@ RBC::RBC(Endpoint &endpoint, MuddleAddress address, CabinetMembers const &cabine
...
-    RBCEnvelope env;
-    serialiser >> env;
+    RBCMessage msg;
+    try
+    {
+      serialiser >> msg;
+    }
+    catch (...)
+    {
+      std::cout << "PANIC! PANIC! PANIC! PANIC! PANIC! PANIC! PANIC! PANIC! PANIC! PANIC! "
```
n-hutton: use logger_error here. also best to catch specific exceptions right

troels-im: Hmm... that was supposed to be gone. Also, error message is not all that good!

troels-im:
```suggestion
```

libs/dkg/src/rbc.cpp
```diff
@@ -73,11 +73,21 @@ RBC::RBC(Endpoint &endpoint, MuddleAddress address, CabinetMembers const &cabine
...
+                << std::endl;
```
troels-im:
```suggestion
```

jinmannwong (commented):
I agree with the streamlining of the messages however I am worried that having only a single message class everywhere will making debugging more difficult later, especially if things are changed a lot. Construction of the wrong message type, or passing to the wrong processing function will now only be evident at run time. I would be in favor of holding back on these changes if the rest of the code will change a lot.

PR #1502, opened 2019-08-15
RBC lock refactor

troels-im:
This PR refactors the locks in the RBC:
- The number of used mutices is reduced to 1
- It now has explicit testing for deadlocks in debug mode
- A number of potential race conditions have been eliminated


jinmannwong (approved):
Looks good.

PR #1504, opened 2019-08-15
Telemetry added for DKG stages

jinmannwong:
Telemetry for DKG stages added and renaming of threshold in DKG to polynomial degree to fix confusion with threshold.


libs/dkg/src/dkg.cpp
```diff
@@ -32,15 +33,14 @@ bn::G2                DistributedKeyGeneration::group_g_;
...
+    MuddleAddress address, std::function<void(DKGEnvelope const &)> broadcast_function,
     std::function<void(MuddleAddress const &, std::pair<std::string, std::string> const &)>
         rpc_function)
-  : cabinet_{cabinet}
-  , threshold_{threshold}
-  , address_{std::move(address)}
+  : address_{std::move(address)}
   , broadcast_function_{std::move(broadcast_function)}
   , rpc_function_{std::move(rpc_function)}
+  , dkg_state_gauge_{telemetry::Registry::Instance().CreateGauge<uint8_t>("dkg_state_gauge",
```
ejfitzgerald: Not sure if still relevant but can we have all metrics start with "ledger_" this is just makes Chris' life easier when coordinating these metrics at a network level

jinmannwong: Sure thing.

PR #1506, opened 2019-08-15
Semi-synchronous distributed learning by passing gradients

MissingNO57:


libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+#define NUMBER_OF_ITERATIONS 20
+#define BATCH_SIZE 128
+#define SYNCHRONIZATION_INTERVAL 3
+#define LEARNING_RATE .001f
+#define TEST_SET_RATIO 0.03f
+
+using namespace fetch::ml::ops;
+using namespace fetch::ml::layers;
+
+using DataType         = float;
```
baykaner: DataType should be fp64 (it should work in both cases, but we should leave default to fixed point)

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+using TensorType       = fetch::math::Tensor<DataType>;
+using TensorVectorType = std::vector<TensorType>;
```
baykaner: slightly confusing name - should be VectorTensorType

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+  STOP,
+};
+
+class Coordinator
+{
+public:
+  CoordinatorState state = CoordinatorState::RUN;
+};
+
+class TrainingClient
```
baykaner: 1. move this class out of main and into /examples/distributed_learning/distributed_learning_client.hpp
2. Then also make a source file and move the implementations into there (/examples/distributed_learning/distributed_learning_client.cpp)

This will set us up for a better more generic/reusable implementation as well; we might want to refactor training client to be more task agnostic - i.e. re-use it fore mnist/word2vec etc.

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+    g_.AddNode<Softmax<TensorType>>("Softmax", {"FC3"});
+    g_.AddNode<PlaceHolder<TensorType>>("Label", {});
+    g_.AddNode<CrossEntropyLoss<TensorType>>("Error", {"Softmax", "Label"});
+
+    // Clear loss file
+    std::ofstream lossfile("losses_" + id_ + ".csv", std::ofstream::out | std::ofstream::trunc);
+    lossfile.close();
+  }
+
+  void SetCoordinator(std::shared_ptr<Coordinator> &coordinator_ptr)
```
baykaner: why do you pass a reference to a shared_ptr? just pass the ptr

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+      BroadcastGradients();
+
+      // Load own gradient
+      TensorVectorType gradients = g_.GetGradients();
+      TensorVectorType new_gradients;
+
+      // Sum all gradient in queue
+      while (!gradient_queue.empty())
+      {
+        {
```
baykaner: move this anonymous scope into a function call, e.g.:
void GetNewGradients(TensorVectorType & new_gradients)
{
          std::lock_guard<std::mutex> l(queue_mutex_);
          new_gradients = gradient_queue.front();
          gradient_queue.pop();
}

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+          fetch::math::Add(gradients.at(j), new_gradients.at(j), gradients.at(j));
+        }
+      }
+
+      // Apply sum of all gradients from queue along with own gradient
+      ApplyGradient(gradients);
+
+      // Validate loss for logging purpose
+      Test(loss);
```
baykaner: try to be clear about your use of the words validation and testing; since they're not the same thing

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+  {
+    dataloader_.SetMode(fetch::ml::dataloaders::DataLoaderMode::TEST);
+
+    // Disable random to run model on whole test set
+    dataloader_.SetRandomMode(false);
+
+    SizeType test_set_size = dataloader_.Size();
+
+    dataloader_.Reset();
+    bool is_done_set;
```
baykaner: presumably you should give a default initialisation to is_done_set (i.e. false?)

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+    }
+  }
+
+  /**
+   * Adds own gradient to peers queues
+   */
+  void BroadcastGradients()
+  {
+    // Load own gradient
+    TensorVectorType current_gradient = g_.GetGradients();
```
baykaner: shouldn't you call GetGradients() so that you lock them with a mutex? or is that unnecessary?

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+    }
+  }
+
+  /**
+   * Adds gradient to own gradient queue
+   * @param gradient
+   */
+  void AddGradient(TensorVectorType gradient)
+  {
+    {
```
baykaner: is this pair of braces necessary? presumably the mutex goes out of scope immediately after the function call anyway?

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+  // Gradient queue access and mutex to protect it
+  std::queue<TensorVectorType> gradient_queue;
+  std::mutex                   queue_mutex_;
+
+  // random number generator for shuffling peers
+  fetch::random::LaggedFibonacciGenerator<> gen_;
+
+  // Learning hyperparameters
+  SizeType batch_size_      = BATCH_SIZE;
```
baykaner: pass in these hyperparameters on TrainingClient construction.

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+  SizeType test_set_ratio_  = TEST_SET_RATIO;
+  SizeType number_of_peers_ = NUMBER_OF_PEERS;
+  SizeType learning_rate_   = LEARNING_RATE
+
+      // Client id (identification name)
+      std::string id_;
```
baykaner: formatting?

baykaner: also, why is the ID a string and not a SizeType?

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+  // Timestamp for logging
+  std::string GetTimeStamp()
+  {
+    // TODO: Implement timestamp
```
baykaner: leftover TODO - raise an issue

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+  if (ac < 3)
+  {
+    std::cout << "Usage : " << av[0]
+              << " PATH/TO/train-images-idx3-ubyte PATH/TO/train-labels-idx1-ubyte" << std::endl;
+    return 1;
+  }
+
+  std::shared_ptr<Coordinator> coordinator = std::make_shared<Coordinator>();
+
+  std::cout << "FETCH Distributed MNIST Demo -- Synchronised" << std::endl;
```
baykaner: it's semi-synchronised

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+  std::srand((unsigned int)std::time(nullptr));
```
baykaner: why are you doing this? we should be using our core library randomisers lfg/lcg

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+  std::vector<std::shared_ptr<TrainingClient>> clients(NUMBER_OF_CLIENTS);
+  for (unsigned int i(0); i < NUMBER_OF_CLIENTS; ++i)
```
baykaner: please don't use "unsigned int".
Use either SizeType or a specific uint if you need that (e.g. uint32_t/uint64_t)

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+  {
+    // Instanciate NUMBER_OF_CLIENTS clients
```
baykaner: typo

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+    clients[i] = std::make_shared<TrainingClient>(av[1], av[2], std::to_string(i));
+  }
+
+  for (unsigned int i(0); i < NUMBER_OF_CLIENTS; ++i)
```
baykaner: unsigned int throughout

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+    // Start all clients
+    coordinator->state = CoordinatorState::RUN;
+    std::cout << "================= ITERATION : " << it << " =================" << std::endl;
+    std::list<std::thread> threads;
+    for (auto &c : clients)
+    {
+      threads.emplace_back([&c] { c->MainLoop(); });
+    }
+
+    std::this_thread::sleep_for(std::chrono::seconds(SYNCHRONIZATION_INTERVAL));
```
baykaner: add a boolean flag for synchronisation On/Off
At the moment if you want to try out fully asynchronous mode the only way to do it is to make this sleep forever.

libs/ml/include/ml/core/graph.hpp
```diff
@@ -490,6 +492,36 @@ std::vector<TensorType> Graph<TensorType>::get_weights() const
   return std::move(ret);
 }
 
+template <typename TensorType>
+void Graph<TensorType>::set_weights(std::vector<TensorType> &new_weights)
```
baykaner: function comments

libs/ml/include/ml/core/graph.hpp
```diff
@@ -490,6 +492,36 @@ std::vector<TensorType> Graph<TensorType>::get_weights() const
...
+{
+  SizeType index = 0;
+  for (auto &t : trainable_nodes_)
```
baykaner: it's a bit pedantic but in these cases where you want to iterate over two objects I think it's preferable to use
auto it1 = obj1.begin();
auto it2 = obj2.begin();
while (it1.is_valid()) // or equivalent code for the relevant object type
{
    // do work
    ++ it1;
    ++it2;
}

rather than mixing the use of one iterator and one counter/incrementer.

libs/ml/include/ml/core/graph.hpp
```diff
@@ -82,6 +82,8 @@ class Graph
   virtual void LoadStateDict(struct fetch::ml::StateDict<T> const &dict);
 
   std::vector<TensorType> get_weights() const;
+  void                    set_weights(std::vector<TensorType> &new_weights);
+  std::vector<TensorType> get_gradients_references() const;
```
baykaner: GetGradientsReferences

libs/ml/include/ml/core/graph.hpp
```diff
@@ -82,6 +82,8 @@ class Graph
   virtual void LoadStateDict(struct fetch::ml::StateDict<T> const &dict);
 
   std::vector<TensorType> get_weights() const;
+  void                    set_weights(std::vector<TensorType> &new_weights);
```
baykaner: SetWeights

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -95,7 +95,9 @@ class Trainable
   virtual struct fetch::ml::StateDict<T> StateDict() const                            = 0;
   virtual void              LoadStateDict(struct fetch::ml::StateDict<T> const &dict) = 0;
   virtual TensorType const &get_weights() const                                       = 0;
-  virtual TensorType const &get_gradients() const                                     = 0;
+  virtual void              set_weights(TensorType &new_value)                        = 0;
```
baykaner: SetWeights

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -95,7 +95,9 @@ class Trainable
...
+  virtual TensorType const &get_gradients_references() const                          = 0;
```
baykaner: GetGradientsReferences

libs/ml/include/ml/ops/ops.hpp
baykaner: weights and gradients are NOT common to all ops. Only to trainable ops. Presumably these virtual methods should exist in trainable and not Ops?

Also, by setting this pure virtuals here it presumably breaks every Op for which you haven't implemented these methods? (because they're now abstract classes)

baykaner (changes requested):

libs/ml/examples/distributed_learning/main.cpp
MissingNO57: I see

libs/ml/examples/distributed_learning/distributed_learning_client.hpp
```diff
@@ -0,0 +1,480 @@
...
+#include <utility>
+#include <vector>
+
+using namespace fetch::ml::ops;
+using namespace fetch::ml::layers;
+
+enum class CoordinatorMode
+{
+  SYNCHRONOUS,
+  SEMISYNCHRONOUS,
```
baykaner: SEMI_SYNCHRONOUS

libs/ml/examples/distributed_learning/distributed_learning_client.hpp
```diff
@@ -0,0 +1,480 @@
...
+#include <mutex>
+#include <queue>
+#include <string>
+#include <utility>
+#include <vector>
+
+using namespace fetch::ml::ops;
+using namespace fetch::ml::layers;
+
+enum class CoordinatorMode
```
baykaner: - Move the coordinator, coordinatormode and state into coordinator.hpp file

baykaner (changes requested):

libs/ml/examples/distributed_learning/coordinator.hpp
```diff
@@ -0,0 +1,96 @@
+#pragma once
+//------------------------------------------------------------------------------
+//
+//   Copyright 2018-2019 Fetch.AI Limited
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
```
baykaner, line 7: copyright appears twice

libs/ml/examples/distributed_learning/coordinator.hpp
```diff
@@ -0,0 +1,96 @@
...
+  using SizeType = fetch::math::SizeType;
+
+  Coordinator(CoordinatorMode mode, SizeType iterations_count = 0)
+    : mode_(mode)
+    , iterations_count_(iterations_count)
+  {}
+
+  void IncrementIterationsCounter()
+  {
+    std::lock_guard<std::mutex> l(iterations_mutex_);
```
baykaner: could also move implementations outside of class

libs/ml/examples/distributed_learning/distributed_learning_client.hpp
```diff
@@ -17,6 +17,7 @@
 //
 //------------------------------------------------------------------------------
 
+#include "../../examples/distributed_learning/coordinator.hpp"
```
baykaner: don't use relative paths in an include - you just need:
#include "coordinator.hpp"

libs/ml/examples/distributed_learning/main.cpp
```diff
@@ -16,6 +16,7 @@
 //
 //------------------------------------------------------------------------------
 
+#include "../../examples/distributed_learning/coordinator.hpp"
```
baykaner: don't use relative paths

baykaner (changes requested):

baykaner (dismissed):

baykaner (approved):

PR #1512, opened 2019-08-16
YAML parser

troels-im:
This PR implements a YAML parser based on the existing variant.


libs/core/include/core/byte_array/consumers.hpp
```diff
@@ -192,6 +192,52 @@ int StringConsumer(byte_array::ConstByteArray const &str, uint64_t &pos)
   return STRING;
 }
 
+template <int STRING>
```
n-hutton, line 195: This pattern looks a bit oddly inefficient - could you not pass as a parameter an int to return on success

libs/core/tests/yaml/gtest/document_tests.cpp
```diff
@@ -0,0 +1,686 @@
...
+  EXPECT_EQ(name.As<string>(), "Mark McGwire");
+
+  auto const &acc = root["accomplishment"];
+  EXPECT_EQ(acc.As<string>(), "Mark set a major league home run record in 1998.");
+
+  auto const &stats = root["stats"];
+  EXPECT_EQ(stats.As<string>(), "65 Home Runs\n0.278 Batting Average");
+}
+
+TEST(YamlTests, Example2_17Test)
```
n-hutton, line 366: This test seems more comprenensive than yaml_tests and it is testing the same thing?

libs/core/include/core/yaml/document.hpp
```diff
@@ -0,0 +1,177 @@
...
+  {
+    uint64_t first  = 0;
+    uint64_t second = 0;
+    uint8_t  type   = 0;
+    uint     ident  = 0;
+    uint     line   = 0;
+  };
+
+  void        Tokenise(ConstByteArray const &document);
+  static void ExtractPrimitive(Variant &variant, YamlToken const &token,
```
n-hutton: Some of this is defined here, whereas other is defined in the cpp file. Suggest all goes in the cpp file.

n-hutton (changes requested):
Seems like mostly a copy and paste job from the json parser right? I never thought the json parser was particularly easy to read but don't object so much to there being 2x of it

libs/core/include/core/byte_array/consumers.hpp
troels-im: The main reason for this pattern is that consumers have a certain function signature that is ignorant to the return value. Another pattern that could be followed would be
```
ConsumerFunction StringConsumerFactory(int type)
{
  return [type](byte_array::ConstByteArray const &str, uint64_t &pos, char quoteChar)
  {
     // ...
     return type;
   }
}
```
I don't really have a preference for one or the other. It feels like it belongs to a different PR to change it.

libs/core/include/core/yaml/document.hpp
troels-im: Agreed.

libs/core/tests/yaml/gtest/document_tests.cpp
troels-im: No, they are different. `yaml_tests` tests the equivalence between YAML defined objects and JSON defined objects. This test is testing on a more fundamental basis the individual components.

n-hutton (approved):

n-hutton (approved):

ejfitzgerald (approved):

PR #1516, opened 2019-08-16
Switch out dkg service for beacon

n-hutton:
This PR switches out the DKG for the beacon manager paradigm. There need to be subsequent refactors: 
- The 'snapshot' (is this the name we are going for) format might want thinking about
- Threshold needs to be set somewhere
- There needs to be a clear distinction between stakers and committe members which there isn't in the code currently
- Needs to be a time lag or offset  so committees prepare well before producing their first block (easy to do I believe).
- Need to look at the committe block emission timings - I believe the first staker is disproportionally likely to create the next block. Should think carefully about the dynamics of this
- Miner and light miner catch up (put entropy in the block)
- Unit tests for the block coordinator and others


libs/beacon/include/beacon/entropy.hpp
```diff
@@ -34,7 +34,7 @@ struct Entropy
 
   uint64_t       round{0};
   ConstByteArray seed{"genesis"};
-  ConstByteArray entropy{};
+  ConstByteArray entropy{"====DEFAULT====="};
```
n-hutton, line 37: @troelsfr there was a bug where requesting the entropy too early caused a segfault by reinterpret_cast-ing this entropy

libs/beacon/src/beacon_service.cpp
```diff
@@ -144,9 +144,11 @@ BeaconService::BeaconService(Endpoint &endpoint, CertificatePtr certificate,
 BeaconService::Status BeaconService::GenerateEntropy(Digest /*block_digest*/, uint64_t block_number,
                                                      uint64_t &entropy)
 {
+  FETCH_LOG_INFO(LOGGING_NAME, "Generating entropy");
+
   uint64_t round = block_number / blocks_per_round_;
 
-  if (latest_entropy_.round == round)
+  if (latest_entropy_.round == round && latest_entropy_)
```
n-hutton, line 151: Bug here

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -201,6 +205,9 @@ BlockCoordinator::BlockCoordinator(MainChain &chain, DAGPtr dag, StakeManagerPtr
     }
   });
 
+  // Initialising the BLS library
+  crypto::bls::Init();
```
n-hutton, line 216: Where should this go, in the beacon service?

ejfitzgerald: Yeah this is annoying, maybe we can simply update the BLS wrapper library to always call this before an operation?

n-hutton: In the constructor of beacon service would make sense since I would think but yeah don't really care - I was going to let Troels put it where he likes

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -1074,16 +1088,67 @@ BlockCoordinator::State BlockCoordinator::OnReset()
...
+        FETCH_LOG_INFO(LOGGING_NAME, "Block: ", block_number,
+                       " creating new aeon. Periodicity: ", AEON_PERIOD, " threshold: ", threshold,
+                       " cabinet size: ", cabinet_member_list.size());
+
+        // Disallow multiple invocations
+        static bool did_genesis_already = false;
+
+        if (block_number != 0 || !did_genesis_already)
+        {
+          beacon_->StartNewCabinet(cabinet_member_list, threshold, block_number,
```
ejfitzgerald, line 1141: `beacon_` should be of a interface class type so that we can mock it out

libs/ledger/tests/unit/random_address.hpp
```diff
@@ -47,3 +48,26 @@ fetch::ledger::Address GenerateRandomAddress(RNG &&rng)
...
+ * Generate a random identity based on a specified RNG
+ *
+ * @tparam RNG The RNG type
+ * @param rng The reference to the RNG to be used
+ * @return The generated identity
+ */
+template <typename RNG>
+fetch::crypto::Identity GenerateRandomIdentity(RNG &&rng)
+{
+  static constexpr std::size_t IDENTITY_BYTES = 32;
```
ejfitzgerald: 64

ejfitzgerald (changes requested):
.

libs/ledger/src/chain/block_coordinator.cpp
n-hutton: Noted, will implement when doing unit tests for the block coord

libs/beacon/include/beacon/entropy.hpp
```diff
@@ -43,6 +43,11 @@ struct Entropy
     return *reinterpret_cast<uint64_t const *>(entropy.pointer());
```
troels-im, line 43: Please add `assert` above the reinterpret cast.

libs/beacon/include/beacon/entropy.hpp
```diff
@@ -43,6 +43,11 @@ struct Entropy
...
   }
 
+  operator bool() const
+  {
+    return !entropy.empty();
```
troels-im, line 48: Actually, it can be argued that the entropy is valid only if it has 32 bytes.

libs/beacon/src/beacon_service.cpp
troels-im: :+1:

libs/beacon/src/beacon_service.cpp
```diff
@@ -172,6 +174,7 @@ BeaconService::Status BeaconService::GenerateEntropy(Digest /*block_digest*/, ui
   }
 
   // We found the entropy
+  assert(latest_entropy_);
```
troels-im, line 177: :+1:

libs/beacon/include/beacon/entropy.hpp
n-hutton: you're right

PR #1517, opened 2019-08-18
Fix bug in Identifier (inconsistent state)

wkaluza:
On calling the `Parse` method, the `full_` member variable was updated even if the subsequent tokenisation failed.


libs/ledger/include/ledger/identifier.hpp
```diff
@@ -37,7 +37,6 @@ class Identifier
 public:
   enum class Type
   {
-    INVALID,
```
ejfitzgerald: Feel like we should keep this type because this can be default initialised

libs/ledger/include/ledger/identifier.hpp
```diff
@@ -58,13 +57,11 @@ class Identifier
...
   ConstByteArray const &full_name() const;
   ConstByteArray        qualifier() const;
-  std::size_t           size() const;
   bool                  empty() const;
 
   Identifier GetParent() const;
 
   // Parsing
-  bool Parse(ConstByteArray const &name);
   bool Parse(ConstByteArray &&name);
```
ejfitzgerald: Recommend simply making this a value to support move or copy semantics

ejfitzgerald (changes requested):
Just some minor niggles looks good though

libs/ledger/include/ledger/identifier.hpp
wkaluza: Done

libs/ledger/include/ledger/identifier.hpp
wkaluza: Done

ejfitzgerald (approved):

PR #1518, opened 2019-08-18
Move some function definitions to source files

wkaluza:


libs/moment/include/moment/deadline_timer.hpp
```diff
@@ -19,6 +19,8 @@
 
 #include "moment/clocks.hpp"
 
+#include <cstdint>
```
pbukva, line 22: is this new include necessary?

pbukva (approved):
LGTM

libs/moment/include/moment/deadline_timer.hpp
wkaluza: It's for `uint64_t`

PR #1538, opened 2019-08-20
Serialization benchmark

XiaodongWuFetch:


libs/ml/benchmark/serialization/serialization.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include "ml/utilities/graph_builder.hpp"
+
+#include "benchmark/benchmark.h"
+
+#include <memory>
+#include <string>
+
+using SizeType = fetch::math::SizeType;
+template <typename T>
+void create_graph(std::shared_ptr<fetch::ml::Graph<typename fetch::math::Tensor<T>>> g,
```
baykaner: style CreateGraph

libs/ml/benchmark/serialization/serialization.cpp
```diff
@@ -0,0 +1,114 @@
...
+                  SizeType input_size, SizeType output_size, SizeType n_layers)
+{
+  using DataType   = T;
+  using TensorType = fetch::math::Tensor<DataType>;
+  // set up the neural net architecture
+  std::string input_name = g->template AddNode<fetch::ml::ops::PlaceHolder<TensorType>>("", {});
+
+  for (int i = 0; i < static_cast<int>(n_layers); i++)
```
baykaner, line 47: this is a weird neural net. You define n_layers fullyconnected layers but they're not connected to each other in sequence, they each take input from the placeholder. Wouldn't it be more sensible to set it up so that you use h_1 as the input name for each subsequent layer?

libs/ml/benchmark/serialization/serialization.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+  // make a graph
+  auto g = std::make_shared<fetch::ml::Graph<TensorType>>();
+  create_graph<T>(g, input_size, output_size, n_layers);
+
+  fetch::ml::GraphSaveableParams<TensorType> gsp1 = g->GetGraphSaveableParams();
+
+  for (auto _ : state)
+  {
+    fetch::serializers::MsgPackSerializer b;
```
baykaner: might as well move this line outside of the for loop as well (no need to benchmark the time cost of setting up the generic serializer

baykaner: alternatively you can wrap it up in a DoNotOptimse call

libs/ml/benchmark/serialization/serialization.cpp
```diff
@@ -0,0 +1,114 @@
...
+    b << gsp1;
+  }
+}
+
+BENCHMARK_TEMPLATE(BM_Setup_And_Serialize, float, 100, 50, 1)->Unit(benchmark::kMillisecond);
```
baykaner: please repeat these function calls for double, fp32 and fp64

libs/ml/benchmark/serialization/serialization.cpp
```diff
@@ -0,0 +1,114 @@
...
+  create_graph<T>(g, input_size, output_size, n_layers);
+
+  fetch::ml::GraphSaveableParams<TensorType> gsp1 = g->GetGraphSaveableParams();
+  fetch::serializers::MsgPackSerializer      b;
+  b << gsp1;
+  b.seek(0);
+
+  for (auto _ : state)
+  {
+    fetch::ml::GraphSaveableParams<TensorType> gsp2;
```
baykaner: doNotOptimise calls for constructing gsp2 and for b.seek(0)

baykaner (changes requested):

libs/ml/benchmark/serialization/serialization.cpp
XiaodongWuFetch: agreed

libs/ml/benchmark/serialization/serialization.cpp
```diff
@@ -46,7 +46,7 @@ void create_graph(std::shared_ptr<fetch::ml::Graph<typename fetch::math::Tensor<
   for (int i = 0; i < static_cast<int>(n_layers); i++)
   {
     std::string h_1 = g->template AddNode<fetch::ml::layers::FullyConnected<TensorType>>(
-        "FC_" + std::to_string(i), {input_name}, input_size, output_size);
+        "FC_1_No_" + std::to_string(i), {input_name}, input_size, output_size);
```
baykaner: you still don't use h_1 to specify input to the subsequent layer

baykaner (changes requested):

baykaner (approved):

PR #1542, opened 2019-08-21
Add addition telemetry information into the ledger

ejfitzgerald:


libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -182,6 +182,9 @@ BlockCoordinator::BlockCoordinator(MainChain &chain, DAGPtr dag, StakeManagerPtr
   , current_block_num_{telemetry::Registry::Instance().CreateGauge<uint64_t>(
         "ledger_latest_block_num",
         "The lastest block number that has been executed by the block coordinator")}
+  , next_block_num_{telemetry::Registry::Instance().CreateGauge<uint64_t>(
+        "ledger_next_block_num",
+        "The next block number that has been executed by the block coordinator")}
```
pbukva: _"Number of the next block which is scheduled to be executed by the block coordinator."_, or something like that

pbukva (approved):
lgtm

PR #1543, opened 2019-08-21
RBC for verification vectors

troels-im:
This PR integrates the RBC into the beacon setup.


apps/beacon/main.cpp
```diff
@@ -173,7 +174,10 @@ int main()
     }
 
     FETCH_LOG_INFO("default", "Found entropy for block: ", block_number, " as ", entropy);
-
+    if (block_number > 10)
```
troels-im, line 177: Should definitely not be there.

PR #1548, opened 2019-08-22
Improve exception safety of mutex usage

wkaluza:
There is still some code in DKG, network and the HttpConnectionManager which is unsafe, but it doesn't lend itself to an immediate fix.

### DIFF
--- a/libs/core/src/mutex.cpp
+++ b/libs/core/src/mutex.cpp
@@ -1,0 +1,160 @@
+//------------------------------------------------------------------------------
+//
+//   Copyright 2018-2019 Fetch.AI Limited
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "core/abstract_mutex.hpp"
+#include "core/logger.hpp"
+#include "core/macros.hpp"
+#include "core/mutex.hpp"
+
+#include <atomic>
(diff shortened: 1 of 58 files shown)


ejfitzgerald (changes requested):
Looks good, but can we switch to the `FETCH_LOCK` macro please

libs/network/include/network/details/idle_work_store.hpp
```diff
@@ -92,7 +92,13 @@ class IdleWorkStore
...
     if (mutex_.try_lock())
     {
-      is_due = Clock::now() >= (last_run_ + interval_);
+      try
+      {
+        is_due = Clock::now() >= (last_run_ + interval_);
+      }
+      catch (...)
+      {
+      }
```
ejfitzgerald, line 101: Should we log this case do you think?

ejfitzgerald (dismissed):

ejfitzgerald (approved):

libs/network/include/network/details/idle_work_store.hpp
wkaluza: In summary of our conversation on Slack just now, `idle_work_store.hpp` is scheduled for demolition in the near future.

libs/core/src/mutex.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+namespace {
+constexpr std::size_t DEFAULT_TIMEOUT_MS = 3000;
+}
+
+/**
+ * The debug mutex acts like a normal mutex but also contains several other checks. This code is
+ * intended to be only used in software development.
+ */
+DebugMutex::MutexTimeout::MutexTimeout(std::string filename, int const &line)
```
bipll, line 48: `int const &`

libs/core/src/mutex.cpp
```diff
@@ -0,0 +1,160 @@
...
+  LOG_STACK_TRACE_POINT;
+  FETCH_LOG_ERROR(LOGGING_NAME, "The system will terminate, mutex timed out: ", filename_, " ",
+                  line_);
+
+  // Send a sigint to ourselves since we have a handler for this
+  kill(0, SIGINT);
+}
+
+DebugMutex::DebugMutex(int line, std::string file)
+  : AbstractMutex()
```
bipll, line 96: Is this explicit default-init needed?

PR #1550, opened 2019-08-22
Dkg refactoring

jinmannwong:
Refactors crypto computation and state machine of DKG into two new files for integration into beacon service and manager. Dkg tests have been modified for the refactoring.


libs/beacon/include/beacon/beacon_setup_service.hpp
```diff
@@ -19,76 +19,62 @@
...
-#include "beacon/entropy.hpp"
 
 namespace fetch {
 namespace beacon {
 
 class BeaconSetupService
 {
 public:
-  static constexpr char const *LOGGING_NAME = "BeaconSetupService";
+  static constexpr char const *LOGGING_NAME = "DkgSetupService";
```
n-hutton: Incorrect name here

libs/beacon/src/beacon_manager.cpp
```diff
@@ -17,167 +17,617 @@
 //------------------------------------------------------------------------------
 
 #include "beacon/beacon_manager.hpp"
+#include "crypto/ecdsa.hpp"
 
 namespace fetch {
 namespace dkg {
 
-void BeaconManager::Reset(uint64_t cabinet_size, uint32_t threshold)
+bn::G2 BeaconManager::zeroG2_;
```
n-hutton, line 25: These are statics right? I think they can not be static in beacon_manager.hpp and just construct them implicitly

libs/beacon/src/beacon_manager.cpp
```diff
@@ -17,167 +17,617 @@
...
+bn::Fr BeaconManager::zeroFr_;
+bn::G2 BeaconManager::group_g_;
+bn::G2 BeaconManager::group_h_;
+
+constexpr char const *LOGGING_NAME = "DKG";
```
n-hutton: Incorrect logging name here

libs/beacon/src/beacon_manager.cpp
```diff
@@ -17,167 +17,617 @@
...
+    certificate_.reset(ptr);
+    ptr->GenerateKeys();
+  }
 
-  received_shares_.clear();
-  participants_.clear();
+  static std::once_flag flag;
+
+  std::call_once(flag, []() {
+    bn::initPairing();
```
n-hutton, line 45: If this is here it should not be in block coordinator. Also, I might consider a pattern like the BLSInitialiser - that way it hides this away

libs/beacon/src/beacon_manager.cpp
```diff
@@ -17,167 +17,617 @@
...
+    bn::mapToG2(group_g_, g);
+    bn::mapToG2(group_h_, h);
+  });
+}
 
-  verification_vectors_.resize(cabinet_size);
-  received_shares_.resize(cabinet_size);
+void BeaconManager::GenerateCoefficients()
+{
+  std::vector<bn::Fr> a_i(polynomial_degree_ + 1, zeroFr_), b_i(polynomial_degree_ + 1, zeroFr_);
```
n-hutton: style: we don't usually declare multiple on the same line

libs/beacon/src/beacon_manager.cpp
```diff
@@ -17,167 +17,617 @@
...
+                   " received complaint answer with unknown reporter index");
+    return true;
+  }
+
+  CabinetIndex reporter_index = identity_to_index_[answer.first];
+  // Verify shares received
+  bn::Fr s, sprime;
+  bn::G2 lhsG, rhsG;
+  s.clear();
+  sprime.clear();
```
n-hutton: I am confused about clear in this context - is it really neccessary to clear it immediately after constructing it? what does clearing it do?

libs/beacon/src/beacon_manager.cpp
```diff
@@ -17,167 +17,617 @@
...
+/**
+ * If in qual a member computes individual share of the secret key and further computes and
+ * broadcasts qual coefficients
+ */
+void BeaconManager::ComputeSecretShare()
 {
-  auto it = identity_to_index_.find(from);
-  assert(it != identity_to_index_.end());
+  secret_share_.clear();
+  xprime_i = 0;
```
n-hutton, line 228: is = 0 here the same as clear? It doesn't need to do it on every iteration of the qual loop?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -117,6 +145,19 @@ BeaconSetupService::State BeaconSetupService::OnIdle()
     else
     {
       // Initiating setup
+      std::set<MuddleAddress> cabinet;
+      for (auto &m : beacon_->aeon.members)
+      {
+        cabinet.insert(m.identifier());
+      }
+
+      pre_dkg_rbc_.ResetCabinet(cabinet);
```
n-hutton, line 155: I wonder if not resetting the cabinet here will have problems for people in 'observe' mode, that is, they will still accept and process messages from former cabinet members

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -128,238 +169,772 @@ BeaconSetupService::State BeaconSetupService::OnIdle()
...
     }
     else
     {
-      connected++;
+      connected.insert(m.identifier());
     }
   }
 
-  if (all_connected)
+  if (connected.size() == beacon_->aeon.members.size())
```
n-hutton, line 200: doesn't this want to be -1 since you don't count yourself as a connection?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -128,238 +169,772 @@ BeaconSetupService::State BeaconSetupService::OnIdle()
...
+ * @return True if self is in qual and qual is at least of size polynomial_degree_ + 1, false
+ otherwise
+ */
+bool BeaconSetupService::BuildQual()
+{
+  // Create set of muddle addresses
+  std::set<MuddleAddress> cabinet;
+  for (auto &m : beacon_->aeon.members)
+  {
+    cabinet.insert(m.identifier());
```
n-hutton, line 858: if we use set instead of unordered set everywhere this could be an assignment rather than loop, perhaps we should do that

libs/crypto/include/crypto/mcl_dkg.hpp
```diff
@@ -31,24 +31,42 @@
 namespace bn = mcl::bn256;
 
 namespace fetch {
-namespace dkg {
+namespace crypto {
+namespace mcl {
+
+using PrivateKey     = bn::Fr;
```
n-hutton, line 37: Might be good to use these in beacon_manager.hpp

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -205,8 +205,8 @@ BlockCoordinator::BlockCoordinator(MainChain &chain, DAGPtr dag,
     }
   });
 
-  // Initialising the BLS library
-  crypto::bls::Init();
+  // Initialising the MCL library
+  bn::initPairing();
```
n-hutton: Not sure we need this here for your changes above

libs/muddle/src/rbc.cpp
```diff
@@ -34,8 +34,10 @@ constexpr char const *LOGGING_NAME = "RBC";
  * @param threshold Threshold number of Byzantine peers
  * @param dkg
  */
-RBC::RBC(Endpoint &endpoint, MuddleAddress address, CallbackFunction call_back, uint16_t channel)
+RBC::RBC(Endpoint &endpoint, MuddleAddress address, CallbackFunction call_back, uint16_t channel,
+         bool ordered_delivery)
```
n-hutton, line 38: what is ordered delivery about?

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: Yes, this is the TODO: Fix I had above this. I have pushed this in a later commit.

libs/muddle/src/rbc.cpp
jinmannwong: The RBC has a counter for all messages sent by a node. In the original RBC protocol messages are delivered in the order of the sequence counter, which means that messages with a greater sequence counter than you expect are held back in a queue until they are can be delivered. I added this option of out of order delivery, where the sequence counter is not checked on delivery for the pre-dkg where you might not have received the original message.

libs/ledger/src/chain/block_coordinator.cpp
jinmannwong: You are right, I will remove it.

libs/crypto/include/crypto/mcl_dkg.hpp
jinmannwong: Sure.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: I inserted the nodes own identity into connected so it is correct as it is.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: This is true. I have added resets for the rbc and pre_dkg_rbc with an empty cabinet

libs/beacon/src/beacon_manager.cpp
jinmannwong: It should not be reset at the iteration of the qual loop. It is supposed to accumulate on each loop iteration. Also, yes, I believe for this type they are the same.

libs/beacon/src/beacon_manager.cpp
jinmannwong: Sure.

libs/beacon/src/beacon_manager.cpp
jinmannwong: I have removed it from block coordinator.

libs/beacon/src/beacon_manager.cpp
jinmannwong: Thanks.

libs/beacon/include/beacon/beacon_setup_service.hpp
jinmannwong: Thanks.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: Yes, I agree this would be better but would require RBC to take in a set of identities rather than just muddle addresses. I agree we should do this.

libs/beacon/src/beacon_manager.cpp
jinmannwong: I believe the clear initialises the class members to zero. It may not be necessary but it is there as a precaution.

libs/beacon/include/beacon/aeon.hpp
```diff
@@ -37,9 +35,9 @@ struct Aeon
   using Identity  = crypto::Identity;
   using TimeStamp = std::chrono::time_point<std::chrono::system_clock>;
 
-  std::unordered_set<Identity> members{};
-  uint64_t                     round_start{0};
-  uint64_t                     round_end{0};
+  std::set<Identity> members{};
```
n-hutton, line 38: Probably not you but in this file there are two atomics. I don't think they need to be atomic, plus ready isn't used

libs/beacon/src/beacon_service.cpp
```diff
@@ -70,12 +70,11 @@ BeaconService::BeaconService(Endpoint &endpoint, CertificatePtr certificate,
   , state_machine_{std::make_shared<StateMachine>("BeaconService", State::WAIT_FOR_SETUP_COMPLETION,
                                                   ToString)}
   , blocks_per_round_{blocks_per_round}
-  , verification_vec_subscription_{endpoint_.Subscribe(SERVICE_DKG, CHANNEL_VERIFICATION_VECTORS)}
+  , group_public_key_subscription_{endpoint_.Subscribe(SERVICE_DKG, CHANNEL_PUBLIC_KEY)}
```
n-hutton, line 73: Looks like the state for WAIT_FOR_VERIFICATION_VECTORS is mis-named now?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -64,45 +67,70 @@ char const *ToString(BeaconSetupService::State state)
...
-  , id_subscription_(endpoint_.Subscribe(SERVICE_DKG, CHANNEL_ID_DISTRIBUTION))
-  , rpc_client_{"BeaconSetupService", endpoint_, SERVICE_DKG, CHANNEL_RPC}
+  , shares_subscription(endpoint_.Subscribe(SERVICE_DKG, CHANNEL_SECRET_KEY))
+  , pre_dkg_rbc_{endpoint_, identity_.identifier(),
+                 [this](MuddleAddress const &from, ConstByteArray const &payload) -> void {
+                   std::set<MuddleAddress>               connections;
+                   fetch::serializers::MsgPackSerializer serializer{payload};
+                   serializer >> connections;
+
+                   if (ready_connections_.find(from) == ready_connections_.end())
```
n-hutton, line 78: there's no lock on ready_connections_ here

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -64,45 +67,70 @@ char const *ToString(BeaconSetupService::State state)
...
-        CabinetMemberDetails result;
-        serialiser >> result;
-
-        // TODO(tfr): Check signature
-        // TODO(tfr): Check address
-        std::lock_guard<std::mutex> lock(mutex_);
-        this->member_details_queue_.emplace_back(std::move(result));
-      });
+    // Dispatch the event
+    OnNewShares(from, shares);
```
n-hutton, line 125: Might be worth adding info such as the shares towards which 'round' this is so irrelevant messages can be discarded.

I note that OnNewShares  + OnNewCoefficients don't appear to be protected by a mutex

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -128,238 +169,771 @@ BeaconSetupService::State BeaconSetupService::OnIdle()
...
 {
   std::lock_guard<std::mutex> lock(mutex_);
+  dkg_state_gauge_->set(static_cast<uint8_t>(State::WAIT_FOR_READY_CONNECTIONS));
 
-  // Creating identity
-  CabinetMemberDetails member;
-  member.identity = beacon_->manager.identity();
-  member.id       = beacon_->manager.id();
-  member_details_queue_.push_back(member);
+  // Try broadcasting connections again in case they were discarded by RBCs who's cabinet was reset
```
n-hutton, line 221: This might deadlock the network in the limit right? Is there no feedback mechanism from the DKG to indicate how many peers have recieved your message?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -128,238 +169,771 @@ BeaconSetupService::State BeaconSetupService::OnIdle()
...
+    for (auto &m : beacon_->aeon.members)
+    {
+      if (m == identity_)
+      {
+        continue;
+      }
+      if (ready_connections_.find(m.identifier()) != ready_connections_.end())
+      {
+        assert(ready_connections_.at(m.identifier()) == connections_);
+        // TODO(jmw): Strategy if connections for members are different
```
n-hutton, line 238: Yeah, we should think a bit about this tomorrow perhaps - are the conditions for continuing that threshold number of members form a complete subgraph?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -128,238 +169,771 @@ BeaconSetupService::State BeaconSetupService::OnIdle()
...
-      if (counter_party == identity_)
-      {
-        ShareSubmission submission;
-        submission.from                = from;
-        submission.share               = std::move(share);
-        submission.verification_vector = verification_vector;
+    fetch::serializers::MsgPackSerializer serializer;
+    serializer.Reserve(counter.size());
+    serializer << shares;
+    endpoint_.Send(cab_i.identifier(), SERVICE_DKG, CHANNEL_SECRET_KEY, serializer.data());
```
n-hutton, line 499: This is the point we want to be encrypted right? If so I'll chat to Ed

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -128,238 +169,771 @@ BeaconSetupService::State BeaconSetupService::OnIdle()
...
 {
-  std::lock_guard<std::mutex> lock(mutex_);
-  assert(beacon_ != nullptr);
+  std::unique_lock<std::mutex> lock{mutex_};
+  dkg_state_gauge_->set(static_cast<uint8_t>(State::WAIT_FOR_SHARE));
+
+  if ((coefficients_received_.size() == beacon_->aeon.members.size() - 1) &&
+      (shares_received_.size() == beacon_->aeon.members.size() - 1))
+  {
+    lock.unlock();
```
n-hutton: shouldn't this lock protect shares_received.clear() ?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -128,238 +169,771 @@ BeaconSetupService::State BeaconSetupService::OnIdle()
...
+      BroadcastQualCoefficients();
+
+      complaints_manager_.Clear();
+      return State::WAIT_FOR_QUAL_SHARES;
     }
     else
     {
-      member_details_[member.identity] = member.id;
+      complaints_manager_.Clear();
+      // TODO(jmw): procedure failed for this node
```
n-hutton, line 323: I think it might be a good idea to reset the whole thing if minimum conditions are not met - chat tomorrow?

libs/beacon/include/beacon/beacon_manager.hpp
```diff
@@ -59,187 +52,125 @@ class BeaconManager
...
-      ptr->GenerateKeys();
-    }
-  }
+  BeaconManager(CertificatePtr = nullptr);
 
   BeaconManager(BeaconManager const &) = delete;
   BeaconManager &operator=(BeaconManager const &) = delete;
 
-  void SetCertificate(CertificatePtr cert)
+  void SetCertificate(CertificatePtr certificate)
```
n-hutton: Would you mind putting everything in a cpp that can be?

libs/beacon/include/beacon/beacon_manager.hpp
```diff
@@ -17,36 +17,29 @@
...
-#include <unordered_map>
+#include "crypto/mcl_dkg.hpp"
+#include "dkg/dkg_messages.hpp"
 
 namespace fetch {
 namespace dkg {
 
 class BeaconManager
 {
 public:
```
n-hutton, line 28: Might need some high-level comments in this file (I can do this)

libs/beacon/src/beacon_manager.cpp
```diff
@@ -17,167 +17,628 @@
...
-void BeaconManager::SetMessage(BeaconManager::ConstByteArray next_message)
+void BeaconManager::SetDkgOutput(PublicKey &public_key, PrivateKey &secret_share,
+                                 std::vector<PublicKey> & public_key_shares,
+                                 std::set<MuddleAddress> &qual)
 {
-  current_message_ = next_message;
-  signature_buffer_.clear();
-  signer_ids_.clear();
-  already_signed_.clear();
+  // crypto::mcl::Init(public_key_shares, cabinet_size_);
```
n-hutton: Undesired init?

libs/beacon/include/beacon/beacon_setup_service.hpp
```diff
@@ -19,14 +19,16 @@
 
 #include "core/service_ids.hpp"
 #include "core/state_machine.hpp"
+#include "dkg/dkg_complaints_manager.hpp"
```
n-hutton, line 22: is this moving to beacon/ ?

libs/beacon/src/beacon_service.cpp
jinmannwong: Thanks.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: Thanks.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: I am not sure what you mean by 'round' here, unless you are thinking of signature shares? The shares here are the ones that need to be encrypted and are only sent once during the DKG. Also, you are correct about OnNewShares not being protected by a mutex. I will correct that. OnNewCoefficients does not need it as it is only called from within OnDkgMessage which has a lock at the beginning.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: You are correct. This is not the way to solve this but we need some feedback mechanism between the RBC and DKG for this. Or we make the nodes send requests to peers which it has not received ready messages from.

libs/beacon/include/beacon/beacon_manager.hpp
jinmannwong: No problem.

libs/beacon/include/beacon/beacon_manager.hpp
jinmannwong: Thanks.

libs/beacon/src/beacon_manager.cpp
jinmannwong: Should have deleted that. Thanks.

libs/beacon/include/beacon/beacon_setup_service.hpp
jinmannwong: Wasn't sure about whether everything was going to be moved into the same directory so left it.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: Sure. However, this is just a single node failing and therefore does not require the DKG being reset. There are points of failure where everyone fails, which is currently not handled.

libs/beacon/include/beacon/aeon.hpp
jinmannwong: I agree that they don't seem to need to be atomic. I can change this.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: Yes, the secret shares need to be encrypted.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: Yes, thanks.

n-hutton (approved):

PR #1552, opened 2019-08-23
implemented wrapper for serializing large objects

baykaner:


libs/ml/tests/ml/serializers/ml_types.cpp
```diff
@@ -119,14 +118,12 @@ TYPED_TEST(SerializersTestNoInt, serialize_graph_saveable_params)
   TensorType tmp_prediction = g->ForwardPropagate(output);
 
   fetch::ml::GraphSaveableParams<TypeParam> gsp1 = g->GetGraphSaveableParams();
-  fetch::serializers::MsgPackSerializer     b;
-  b << gsp1;
-  b.seek(0);
+  fetch::serializers::BigSerializer b;
+  b.Serialiaze(gsp1);
```
ejfitzgerald: Lets make this wrapper have the same interface as the MsgPackSerializer i.e. streaming operators and and accessors. Then the type can simple be changed

libs/core/include/core/serializers/main_serializer.hpp
```diff
@@ -38,6 +38,34 @@
 namespace fetch {
 namespace serializers {
 
+/**
+ * When serializing large objects it is useful to use this wrapper class
+ * It ensures the right amount of memory is allocated first
+ */
+class BigSerializer
```
ejfitzgerald: Need a better name I think

ejfitzgerald (changes requested):

ejfitzgerald (approved):
LGTM

PR #1560, opened 2019-08-25
Don't move

bipll:
Somewhat fantastic may it seem
but hear my word and I can prove:
trust me you don't need those movements,
so just return don't move.

### DIFF
--- a/libs/network/tests/muddle/mrpc_stress_tests.cpp
+++ b/libs/network/tests/muddle/mrpc_stress_tests.cpp
@@ -142,7 +142,7 @@ protected:
     {
       buffer[i] = fill;
     }
-    return std::move(buffer);
+    return buffer;
   }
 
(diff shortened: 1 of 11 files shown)


libs/network/tests/muddle/mrpc_stress_tests.cpp
```diff
@@ -142,7 +142,7 @@ class MuddleRpcStressTests : public ::testing::Test
     {
       buffer[i] = fill;
     }
-    return std::move(buffer);
+    return buffer;
```
ejfitzgerald: I have seen this pattern a few times with `ConstByteArray` I suggest maybe adding braces to make it clear that a new object is being created. i.e. `return {std::move(buffer)}`.

ejfitzgerald:
Looks good, I suggest getting it to compile and lets try to get it landed

ejfitzgerald (approved):
LGTM

PR #1569, opened 2019-08-27
Feature/consensus forced catchup mechanism

n-hutton:
This PR:

- Abstracts away consensus in the block coord in a cleaner way

- modifies the genesis file format (now, there are default settings for consensus, and they can be overriden using command line flags, but if the genesis file has settings these have the highest priority).

- other bits


libs/ledger/include/ledger/chain/block_coordinator.hpp
```diff
@@ -360,31 +352,32 @@ class BlockCoordinator
...
+  telemetry::CounterPtr         new_syn_state_count_;
+  telemetry::CounterPtr         new_exec_state_count_;
+  telemetry::CounterPtr         new_wait_exec_state_count_;
+  telemetry::CounterPtr         proof_search_state_count_;
+  telemetry::CounterPtr         transmit_state_count_;
+  telemetry::CounterPtr         reset_state_count_;
+  telemetry::CounterPtr         executed_block_count_;
+  telemetry::CounterPtr         mined_block_count_;
+  telemetry::CounterPtr         executed_tx_count_;
+  telemetry::GaugePtr<uint64_t> block_height_;
```
ejfitzgerald: I suspect these are duplicates are the ones that are in master. I would not include at this stage, rather merge in `master`

libs/ledger/include/ledger/chain/block_coordinator.hpp
```diff
@@ -360,31 +352,32 @@ class BlockCoordinator
...
+  telemetry::GaugePtr<uint64_t> block_hash_;
```
ejfitzgerald, line 379: Not sure about this metric. You kind of want something that is actually a number

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -586,65 +548,15 @@ BlockCoordinator::State BlockCoordinator::OnPreExecBlockValidation()
...
-      {
-        return fail("Block signed by miner deemed invalid by the staking mechanism");
-      }
+      consensus_->UpdateCurrentBlock(*previous);  // Only update with valid blocks
+      auto result = consensus_->ValidBlock(*previous, *current_block_);
 
-      if (current_block_->weight !=
-              stake_->GetBlockGenerationWeight(*previous, current_block_->body.miner) &&
-          current_block_->weight != 0)
+      if (!(result == ConsensusInterface::Status::YES ||
```
ejfitzgerald, line 554: Reasonably confusing does it make sense to simple say == NO

libs/ledger/src/consensus/consensus.cpp
```diff
@@ -0,0 +1,203 @@
...
+  {
+    FETCH_LOG_WARN(LOGGING_NAME,
+                   "Incorrect stake weight found for block. Weight: ", current.weight);
+    return Status::NO;
+  }
+
+  return Status::YES;
+}
+
+void Consensus::Refresh()
```
ejfitzgerald, line 188: ?

ejfitzgerald (changes requested):
Just a few changes

libs/ledger/include/ledger/chain/block_coordinator.hpp
n-hutton: I was just curious when miners are all stuck on a block if it's the same block they're stuck on. Not sure what better way there is to do this.

libs/ledger/include/ledger/chain/block_coordinator.hpp
n-hutton: I only added block height - do you want me to remove this?

libs/ledger/src/chain/block_coordinator.cpp
n-hutton: Not to my mind - this is explicitly continuting even in the presence of 'unknown' (this will go away if we put the group sig in the block however)

libs/ledger/src/consensus/consensus.cpp
n-hutton: I was going to have this basically look at the main chain or other structures to see if it's got anything interesting (the suggestion that DKG has failed and you should abort). However I cut this PR short since I wanted to merge in your muddle changes and avoid conflicts

libs/ledger/include/ledger/chain/block_coordinator.hpp
ejfitzgerald: Yes it is in `master` (hopefully the merge will be easier)

libs/ledger/include/ledger/chain/block_coordinator.hpp
ejfitzgerald: Fair enough - I just think there might be some issues with prom, but lets see

libs/crypto/include/crypto/fetch_mcl.hpp
```diff
@@ -21,12 +21,14 @@
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wconversion"
 #pragma GCC diagnostic ignored "-Wpedantic"
+#pragma GCC diagnostic ignored "-Wmacro-redefined"
```
ejfitzgerald, line 24: 👍  - this is an issue but sounds good

ejfitzgerald (approved):

PR #1572, opened 2019-08-27
remove redundant copy in main serializer

XiaodongWuFetch:


ejfitzgerald (dismissed):
LGTM

libs/core/src/serializers/main_serializer.cpp
```diff
@@ -31,8 +31,8 @@
 namespace fetch {
 namespace serializers {
 
-MsgPackSerializer::MsgPackSerializer(byte_array::ByteArray const &s)
-  : data_{s.Copy()}
+MsgPackSerializer::MsgPackSerializer(byte_array::ByteArray const s)
```
bipll: Remove `const` perhaps, to not overcomplicate matters? Right now `std::move` in init-expression returns a `ByteArray const &&`.

ejfitzgerald (approved):
LGTM

PR #1584, opened 2019-08-29
Fix SGNS example

XiaodongWuFetch:
1. Fix the weight initialization of embedding matrices
2. Change the optimiser to Adam which made training way faster
3. Combined with a 250 larger learning rate

### DIFF
--- a/libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
+++ b/libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
@@ -253,8 +253,8 @@ int main(int argc, char **argv)
   std::cout << "beginning training...: " << std::endl;
 
   // Initialise Optimiser
-  fetch::ml::optimisers::AdamOptimiser<TensorType> optimiser(g, {"Input", "Context"}, "Label", error,
-                                                            tp.learning_rate_param);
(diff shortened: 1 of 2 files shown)


libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -157,6 +157,8 @@ std::string ReadFile(std::string const &path)
 
 struct TrainingParams
 {
+  // TODO (#1585) something is broken here. if u set max_word_count to sth smaller like 10000, there
+  // would be an error at the end of the sentence
```
evsmithx, line 161: It's a minor comment, but txt spk in comments makes me wince!

evsmithx (approved):

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
XiaodongWuFetch: ahh ur

XiaodongWuFetch: ah you are right, I should definitely make it more formal.

PR #1587, opened 2019-08-29
Make a simple bert that can do a simple cls task

XiaodongWuFetch:


libs/ml/examples/bert_simple_cls_example/main.cpp
```diff
@@ -0,0 +1,307 @@
...
+using StateDictType = typename fetch::ml::StateDict<TensorType>;
+using OptimiserType = typename fetch::ml::optimisers::AdamOptimiser<TensorType>;
+
+using RegType         = fetch::ml::RegularisationType;
+using WeightsInitType = fetch::ml::ops::WeightsInitialisation;
+using ActivationType  = fetch::ml::details::ActivationType;
+
+struct BERTConfig
+{
+  // the default config is for bert base uncased pretrained model
```
baykaner: what does uncased mean?

baykaner: can you give a link to where to download the dataset you're using too - you've set a specific vocab size after all

libs/ml/examples/bert_simple_cls_example/main.cpp
```diff
@@ -0,0 +1,307 @@
...
+  SizeType n_heads           = 12u;
+  SizeType ff_dims           = 3072u;
+  SizeType vocab_size        = 30522u;
+  SizeType segment_size      = 2u;
+  DataType epsilon           = static_cast<DataType>(1e-12);
+  DataType dropout_keep_prob = static_cast<DataType>(1);
+};
+
+struct BERTInterface
+{
```
baykaner: this looks like a utility function that you re-use in both the load/save example and here. move it to ml/examples/utilities/bert/utilities.hpp

libs/ml/examples/bert_simple_cls_example/main.cpp
```diff
@@ -0,0 +1,307 @@
...
+                  false);
+  }
+
+  // Do validation
+  EvaluateGraph(g, interface.inputs, classification_output, test_data.first, test_data.second);
+
+  return 0;
+}
+
+std::pair<std::vector<std::string>, std::vector<std::string>> MakeBertModel(
```
baykaner: move to utilities/bert as well

libs/ml/examples/bert_simple_cls_example/main.cpp
```diff
@@ -0,0 +1,307 @@
...
+  SizeType max_seq_len       = config.max_seq_len;
+  SizeType model_dims        = config.model_dims;
+  SizeType n_heads           = config.n_heads;
+  SizeType ff_dims           = config.ff_dims;
+  SizeType vocab_size        = config.vocab_size;
+  SizeType segment_size      = config.segment_size;
+  DataType epsilon           = config.epsilon;
+  DataType dropout_keep_prob = config.dropout_keep_prob;
+
+  std::cout << max_seq_len << std::endl;
```
baykaner: if this is an intentional print statement then change it to:
std::cout << "max_seq_len: " << max_seq_len << std::endl;

baykaner (changes requested):
nice. just move the relevant functions out into utilities and this looks good

libs/ml/examples/bert_simple_cls_example/main.cpp
XiaodongWuFetch: CASED and uncased

XiaodongWuFetch: that would be included in the fetch-script. cuz the conversion from pytorch pretrained model to txt C++ readable file happens there

libs/ml/examples/bert_simple_cls_example/main.cpp
XiaodongWuFetch: oh there are a bunch of these functions tho. things such as read tensor from files, evaluate graphs, BERTConfig, BERTInterface etc.

libs/ml/examples/bert_simple_cls_example/main.cpp
XiaodongWuFetch: no it is not

baykaner (approved):

PR #1588, opened 2019-08-29
Bert load save load example

XiaodongWuFetch:


libs/ml/examples/bert_save_load_pretrained_model_example/main.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+using DataType   = float;
+using TensorType = fetch::math::Tensor<DataType>;
+using SizeType   = typename TensorType::SizeType;
+using SizeVector = typename TensorType::SizeVector;
+
+using GraphType     = typename fetch::ml::Graph<TensorType>;
+using StateDictType = typename fetch::ml::StateDict<TensorType>;
+using OptimiserType = typename fetch::ml::optimisers::AdamOptimiser<TensorType>;
+// using OptimiserType = typename fetch::ml::optimisers::SGDOptimiser<TensorType>;
```
baykaner: left in comment

libs/ml/examples/bert_save_load_pretrained_model_example/main.cpp
```diff
@@ -0,0 +1,527 @@
...
+  return output;
+}
+
+void SaveGraphToFile(GraphType &g, std::string const file_name)
+{
+  // start serializing and writing to file
+  fetch::ml::GraphSaveableParams<TensorType> gsp1 = g.GetGraphSaveableParams();
+  std::cout << "got saveable params" << std::endl;
+
+  fetch::serializers::SizeCounter       counter;
```
baykaner: rework this to use the LargeObjectSerializeHelper in core/serializers/main_serializers.hpp

libs/ml/examples/bert_save_load_pretrained_model_example/main.cpp
```diff
@@ -0,0 +1,527 @@
...
+        state_dict, ff_dims, model_dims, file_prefix + "output_dense_weight",
+        file_prefix + "output_dense_bias",
+        "SelfAttentionEncoder_Feedforward_Feedforward_No_2_TimeDistributed_FullyConnected_Weights",
+        "SelfAttentionEncoder_Feedforward_Feedforward_No_2_TimeDistributed_FullyConnected_Bias");
+    PutWeightInFullyConnected(state_dict, model_dims, model_dims,
+                              file_prefix + "attention_output_dense_weight",
+                              file_prefix + "attention_output_dense_bias",
+                              "SelfAttentionEncoder_Multihead_Attention_MultiheadAttention_"
+                              "Final_Transformation_TimeDistributed_FullyConnected_Weights",
+                              "SelfAttentionEncoder_Multihead_Attention_MultiheadAttention_"
```
baykaner: these all seem problematic because if we change any of the default naming convention for these ops these will all break. is there no way to set these names dynamically based out the output from graph.AddNode()? I guess this is a problem with how we manage nested layers at the moment. I guess we should store the layer names as variables or otherwise make them accessible dynamically

libs/ml/examples/bert_save_load_pretrained_model_example/main.cpp
```diff
@@ -0,0 +1,527 @@
...
+            << static_cast<double>(time_span.count()) / static_cast<double>(batch_size)
+            << std::endl;
+  if (verbose)
+  {
+    for (std::size_t i = 0; i < output.shape().size(); i++)
+    {
+      std::cout << " | " << output.shape(i);
+    }
+    std::cout << "\nfirst token: \n"
+              << output.View(0).Copy().View(0).Copy().ToString() << std::endl;
```
baykaner: a comment would be nice here view.copy.view.copy is pretty unclear)

libs/ml/examples/bert_save_load_pretrained_model_example/main.cpp
```diff
@@ -0,0 +1,527 @@
...
+  assert(weight_file.is_open());
+
+  std::string weight_str;
+  getline(weight_file, weight_str);
+  weight_file.close();
+
+  return TensorType::FromString(weight_str);
+}
+
+void PutWeightInLayerNorm(StateDictType &state_dict, SizeType model_dims,
```
baykaner: can you split this into two or three files please all under examples/bert_save_load_pretrained_model_example/, for example:
examples/bert_save_load_pretrained_model_example/load_save_utilities.hpp
examples/bert_save_load_pretrained_model_example/bert_weight_setting_utilities.hpp
etc.

It would be nice if main.cpp was just the main function, the bert config and maybe one or two other core components, these utility functions are large but auxiliary and make it harder to read what the actual logic flow is

baykaner (changes requested):
looks pretty clean overall. I've requested a few simple changes

libs/ml/examples/bert_save_load_pretrained_model_example/main.cpp
XiaodongWuFetch: well it is quite troublesome to change. and actually it is quite difficult to implement cuz the add node process is hidden in subgraph.

libs/ml/examples/bert_save_load_pretrained_model_example/main.cpp
XiaodongWuFetch: sure I will move them to utilities

libs/ml/examples/bert_save_load_pretrained_model_example/main.cpp
XiaodongWuFetch: it seems that the largeobject helper does not give me access to its bytearray data like normal serializer. it can be tricky to use I m afraid

baykaner (dismissed):

baykaner (approved):

PR #1589, opened 2019-08-29
Introduce timeouts to beacon setup

n-hutton:


libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -382,58 +545,78 @@ BeaconSetupService::State BeaconSetupService::OnWaitForQualShares()
...
+        reconstruction_shares_received_.find(member) != reconstruction_shares_received_.end())
     {
-      received_all = false;
-      break;
+      received_count++;
     }
   }
-  if (received_all)
+
+  // TODO(HUT): ask Jenny about this to confirm - not checked (?)
```
n-hutton: here

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -382,58 +545,78 @@ BeaconSetupService::State BeaconSetupService::OnWaitForQualShares()
...
     }
     else if (qual_complaints_manager_.ComplaintsFind(identity_.identifier()))
     {
       FETCH_LOG_WARN(LOGGING_NAME, "Node: ", beacon_->manager.cabinet_index(),
                      " is in qual complaints");
       beacon_->manager.ComputePublicKeys();
 
-      return State::BEACON_READY;
+      // TODO(HUT): ask Jenny whether to proceed or reset here
+      /* return State::RESET; */
```
n-hutton: here

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -145,32 +153,70 @@ BeaconSetupService::State BeaconSetupService::OnIdle()
...
+  }
+
+  auto cabinet_size = static_cast<uint32_t>(cabinet.size());
+
+  coefficients_received_.clear();
+  complaints_answer_manager_.ResetCabinet(0);
+  complaints_answer_manager_.ResetCabinet(cabinet_size);
+  complaints_manager_.Clear();
+  complaints_manager_.ResetCabinet(0);
+  complaints_manager_.ResetCabinet(cabinet_size);
```
jinmannwong, line 210: complaints_manager_.Clear() and ResetCabinet(0) is unnecessary here if you call ResetCabinet(cabinet_size)

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -145,32 +153,70 @@ BeaconSetupService::State BeaconSetupService::OnIdle()
...
+  for (auto &m : beacon_->aeon.members)
+  {
+    cabinet.insert(m.identifier());
+  }
+
+  auto cabinet_size = static_cast<uint32_t>(cabinet.size());
+
+  coefficients_received_.clear();
+  complaints_answer_manager_.ResetCabinet(0);
+  complaints_answer_manager_.ResetCabinet(cabinet_size);
```
jinmannwong, line 209: ResetCabinet(0) is unnecessary here if you call ResetCabinet(cabinet_size)

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -219,109 +265,200 @@ BeaconSetupService::State BeaconSetupService::OnWaitForDirectConnections()
...
+
+  FETCH_LOG_DEBUG(LOGGING_NAME, "All connections: ", connected_peers.size());
+  FETCH_LOG_DEBUG(LOGGING_NAME, "Relevant connections: ", connections.size());
+
+  beacon_dkg_connections_gauge_->set(connections.size());
+
+  // If we get over threshold connections, send a message to all peers each time
+  // we increase over this threshold (note we won't advance if we ourselves don't get over)
+  if (connections.size() > connections_.size() && connections.size() >= require_connections)
+  {
```
jinmannwong, line 382: I don't think connections_ is filled previous to this so its size is 0. Same with the use of connections_ below in the message.

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -219,109 +265,200 @@ BeaconSetupService::State BeaconSetupService::OnWaitForDirectConnections()
...
+ */
 BeaconSetupService::State BeaconSetupService::OnWaitForShares()
 {
   std::lock_guard<std::mutex> lock(mutex_);
-  dkg_state_gauge_->set(static_cast<uint8_t>(State::WAIT_FOR_SHARE));
+  beacon_dkg_state_gauge_->set(static_cast<uint8_t>(State::WAIT_FOR_SHARES));
+
+  const bool coeff_failed  = (coefficients_received_.size() < beacon_->manager.polynomial_degree());
+  const bool shares_failed = (shares_received_.size() < beacon_->manager.polynomial_degree());
+  const bool is_ok         = !coeff_failed && !shares_failed;
```
jinmannwong: is_ok should be checking that sets coefficients_received_ and shares_received_ have an intersection which is of size at least polynomial_degree.

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -332,25 +469,42 @@ BeaconSetupService::State BeaconSetupService::OnWaitForComplaints()
 BeaconSetupService::State BeaconSetupService::OnWaitForComplaintAnswers()
 {
   std::lock_guard<std::mutex> lock(mutex_);
-  dkg_state_gauge_->set(static_cast<uint8_t>(State::WAIT_FOR_COMPLAINT_ANSWERS));
+  beacon_dkg_state_gauge_->set(static_cast<uint8_t>(State::WAIT_FOR_COMPLAINT_ANSWERS));
+
+  bool const is_ok = complaints_answer_manager_.IsFinished();
+
```
jinmannwong, line 518: Same comment here about IsFinished() as for complaints_manager_ above.

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -219,109 +265,200 @@ BeaconSetupService::State BeaconSetupService::OnWaitForDirectConnections()
...
 }
 
 BeaconSetupService::State BeaconSetupService::OnWaitForComplaints()
 {
   std::lock_guard<std::mutex> lock(mutex_);
-  dkg_state_gauge_->set(static_cast<uint8_t>(State::WAIT_FOR_COMPLAINTS));
+  beacon_dkg_state_gauge_->set(static_cast<uint8_t>(State::WAIT_FOR_COMPLAINTS));
+
+  const bool is_ok = complaints_manager_.IsFinished(beacon_->manager.polynomial_degree());
+
```
jinmannwong, line 478: Did you modify IsFinished for complaints_manager? Otherwise, it only returns true when it has received all complaint answers.

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -360,19 +514,28 @@ BeaconSetupService::State BeaconSetupService::OnWaitForComplaintAnswers()
...
+  beacon_dkg_state_gauge_->set(static_cast<uint8_t>(State::WAIT_FOR_QUAL_SHARES));
 
   std::set<MuddleAddress> diff;
   std::set<MuddleAddress> qual{beacon_->manager.qual()};
   std::set_difference(qual.begin(), qual.end(), qual_coefficients_received_.begin(),
                       qual_coefficients_received_.end(), std::inserter(diff, diff.begin()));
-  if (diff.empty())
+
+  const bool is_ok = (qual_coefficients_received_.size() >= beacon_->manager.polynomial_degree());
+
```
jinmannwong, line 564: Instead of computing the set difference here you need to check that the intersection of qual and qual_coefficients_received_ is of size manager.polynomial_degree(). Otherwise nodes not in qual can you coefficients and they will be counted.

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -382,58 +545,78 @@ BeaconSetupService::State BeaconSetupService::OnWaitForQualShares()
 BeaconSetupService::State BeaconSetupService::OnWaitForQualComplaints()
 {
   std::lock_guard<std::mutex> lock(mutex_);
-  dkg_state_gauge_->set(static_cast<uint8_t>(State::WAIT_FOR_QUAL_COMPLAINTS));
+  beacon_dkg_state_gauge_->set(static_cast<uint8_t>(State::WAIT_FOR_QUAL_COMPLAINTS));
+
+  const bool is_ok =
+      /* qual_complaints_manager_.IsFinished(beacon_->manager.qual(), identity_.identifier());*/
+      true;
```
jinmannwong: Strange that the IsFinished only caused a problem here when it did not for complaints_manager_ and complaints_answer_manager_. I think a more consistent solution is to edit IsFinished for all so that is takes in a number of complaints you want to received before you consider it finished.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: No, here you proceed to BEACON_READY. You are in complaints but you can still compute the public key and take part in the threshold signing.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: Yes, this is_ok looks fine.

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -382,58 +545,78 @@ BeaconSetupService::State BeaconSetupService::OnWaitForQualShares()
...
-  if (received_all)
+
+  // TODO(HUT): ask Jenny about this to confirm - not checked (?)
+  // bool const is_ok = (received_count >= beacon_->manager.polynomial_degree());
+
+  if (timer_to_proceed_.HasExpired())
   {
-    // Process reconstruction shares. Return if the sender is in complaints, or not in QUAL
+    // Process reconstruction shares. Fail if the sender is in complaints, or not in QUAL
     for (auto const &share : reconstruction_shares_received_)
```
jinmannwong, line 654: You are not failing here. The return (which is what used to be there) is actually referring to the continue below. Just a note that reconstruction shares from non-qual members or people in qual complaints should not be considered.

libs/beacon/src/beacon_setup_service.cpp
n-hutton: thanks

libs/beacon/src/beacon_setup_service.cpp
n-hutton: This is confusing the way I have written it - there is connections and connections_ here - I will rename to make it clearer but I think it is correct

libs/beacon/src/beacon_setup_service.cpp
n-hutton: Ok, it will become:

```
  auto intersection = (coefficients_received_ & shares_received_);
  const bool is_ok         = intersection.size() >= beacon_->manager.polynomial_degree();
```

libs/beacon/src/beacon_setup_service.cpp
n-hutton: I have changed it now, to this:

```
  if (complaints_received_.size() >= polynomial_degree)
  {
    // TODO(jmw): Add miners which did not send a complaint to complaints?
    // All miners who have received over t complaints are also disqualified
    for (auto const &node_complaints : complaints_counter_)
    {
      if (node_complaints.second > polynomial_degree)
      {
        complaints_.insert(node_complaints.first);
      }
    }
    finished_ = true;
    return true;
  }
  return false;
```

libs/beacon/src/beacon_setup_service.cpp
n-hutton: Becomes:

```
  auto intersection = (qual_coefficients_received_ & beacon_->manager.qual());
  const bool is_ok = (intersection.size() >= beacon_->manager.polynomial_degree());

```

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -332,25 +501,44 @@ BeaconSetupService::State BeaconSetupService::OnWaitForComplaints()
 BeaconSetupService::State BeaconSetupService::OnWaitForComplaintAnswers()
 {
   std::lock_guard<std::mutex> lock(mutex_);
-  dkg_state_gauge_->set(static_cast<uint8_t>(State::WAIT_FOR_COMPLAINT_ANSWERS));
+  beacon_dkg_state_gauge_->set(static_cast<uint64_t>(State::WAIT_FOR_COMPLAINT_ANSWERS));
+
+  // TODO(HUT): fix this.
```
n-hutton: Note

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -449,44 +666,116 @@ BeaconSetupService::State BeaconSetupService::OnWaitForReconstructionShares()
...
+    for (auto const &key_and_count : dry_run_public_keys_)
+    {
+      if (key_and_count.second >= beacon_->manager.polynomial_degree())
+      {
+        found_key = true;
+      }
+    }
+
+    if (!found_key)
+    {
```
jinmannwong, line 737: I think the check here should be that 1) Is consensus reach on any key and then 2) whether this key is the group public key the node has computed. If 1) fails then everyone should reset the DKG and if 2) fails then the node has just failed this particular DKG instance (but shouldn't reset since others have suceeded). In either case of failure, the node doesn't need to bother with going on to computing the group signature

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -713,6 +1004,22 @@ void BeaconSetupService::OnNewSharesPacket(muddle::Packet const &packet,
...
+
+  fetch::serializers::MsgPackSerializer serialiser(packet.GetPayload());
+
+  DryRunInfo to_receive;
+  serialiser >> to_receive;
+
+  // TODO(HUT): cabinet check
+  std::lock_guard<std::mutex> lock(mutex_);
+  dry_run_public_keys_[to_receive.public_key]++;
+  dry_run_shares_[packet.GetSender()] = to_receive.sig_share;
```
jinmannwong: I think the signature shares should be grouped together with the group public key. You should not combine signature shares for different group public keys.

CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ option(FETCH_ENABLE_COVERAGE            "Enable coverage generation"
 option(FETCH_ENABLE_METRICS             "Enable the collection of metrics"      OFF)
 option(FETCH_DISABLE_COLOUR_LOG         "Disable the colour logging"            OFF)
 option(FETCH_STATIC_LINK                "Enable static linking"                 OFF)
-option(FETCH_ENABLE_LTO                 "Enable link-time optimisation"         ON)
+option(FETCH_ENABLE_LTO                 "Enable link-time optimisation"         OFF)
```
ejfitzgerald: ?

ejfitzgerald: If this is for the docker build this can be reverted because I have fixed this issue in the latest version of the `infra-constellation` project

libs/dkg/src/dkg_complaints_manager.cpp
```diff
@@ -18,17 +18,14 @@
...
-
 namespace fetch {
 namespace dkg {
 
 constexpr char const *LOGGING_NAME = "DKGComplaints";
 
 void ComplaintsManager::ResetCabinet(uint32_t cabinet_size)
 {
-  FETCH_LOCK(mutex_);
+  std::lock_guard<std::mutex> lock{mutex_};
```
ejfitzgerald: This looks like a retrograde step?

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -1335,7 +1335,6 @@ void BlockCoordinator::Reset()
 void BlockCoordinator::EnableMining(bool enable)
 {
   mining_enabled_ = enable;
-  syncronised_    = enable;  // Super curious!
```
ejfitzgerald, line 1338: 👍

ejfitzgerald (changes requested):

libs/beacon/src/beacon_setup_service.cpp
n-hutton: Ok, I have changed it as you suggest, to this:

```
  if (timer_to_proceed_.HasExpired())
  {
    bool found_key     = false;
    bool found_our_key = false;

    for (auto const &key_and_count : dry_run_public_keys_)
    {
      if (key_and_count.second >= beacon_->manager.polynomial_degree())
      {
        found_key     = true;
        found_our_key = beacon_->manager.group_public_key() == key_and_count.first;
      }
    }

    if (!found_key)
    {
      FETCH_LOG_WARN(LOGGING_NAME, "Failed to reach consensus on group public key!");
      SetTimeToProceed(State::RESET);
      return State::RESET;
    }

    if (!found_our_key)
    {
      FETCH_LOG_WARN(LOGGING_NAME, "Other nodes didn't agree with our computed group public key!");
      SetTimeToProceed(State::RESET);
      return State::RESET;
    }

    for (auto const &share : dry_run_shares_)
    {
      beacon_->manager.AddSignaturePart(share.second.identity, share.second.signature);
    }

    bool const could_sign = beacon_->manager.can_verify() && beacon_->manager.Verify();

    if (could_sign)
    {
      SetTimeToProceed(State::BEACON_READY);
      return State::BEACON_READY;
    }
    else
    {
      FETCH_LOG_WARN(LOGGING_NAME, "Failed to complete dry run for signature signing!");
      SetTimeToProceed(State::RESET);
      return State::RESET;
    }
  }
```

Note that it is fine for failed nodes to reset since once they see a valid block being generated from the other members of the cabinet they will call abort on the DKG setup.

libs/beacon/src/beacon_setup_service.cpp
n-hutton: Doesn't this mean that someone can corrupt your signing by submitting invalid sig shares? Seems like you could only detect this combinatorially?

In any case, I have put in a fix for this.

CMakeLists.txt
n-hutton: yes, this was not intended to slip in. thanks for fixing it, I will revert this

libs/dkg/src/dkg_complaints_manager.cpp
n-hutton: I don't think I have been changing these around - @jinmannwong ?

jinmannwong: Not me either. I was confused by this too.

n-hutton: reverted to FETCH_LOCK

ejfitzgerald (approved):

PR #1591, opened 2019-08-29
DKG Manifest Entry

ejfitzgerald:


libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -507,9 +514,7 @@ BlockCoordinator::State BlockCoordinator::OnSynchronised(State current, State pr
   }
   else if (State::SYNCHRONISING == previous)
   {
-    FETCH_LOG_INFO(LOGGING_NAME, "Chain Sync complete on 0x", current_block_->body.hash.ToHex(),
-                   " (block: ", current_block_->body.block_number, " prev: 0x",
-                   current_block_->body.previous_hash.ToHex(), ")");
+    //
```
n-hutton, line 517: delete

libs/muddle/src/muddle.cpp
```diff
@@ -93,7 +93,7 @@ Muddle::Muddle(NetworkId network_id, CertificatePtr certificate, NetworkManager
 
 Muddle::~Muddle()
 {
-  MuddleRegistry::Instance().Unregister(node_address_);
+  MuddleRegistry::Instance().Unregister(this);
```
n-hutton, line 96: I am curious about what this does

n-hutton (approved):
lgtm

libs/muddle/src/muddle.cpp
ejfitzgerald: Basically just removes an element from an internal map. The internal map contains weak pointers to muddle instances. The key for the map is the pointer to the instance (and that is all that it is used for).

PR #1593, opened 2019-09-02
[BloomFilter] Fixing ReportFalsePositives(...)

pbukva:


libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1544,7 +1544,7 @@ DigestSet MainChain::DetectDuplicateTransactions(BlockHash const &starting_hash,
 
   bloom_filter_false_positive_count_->add(false_positives);
 
-  if (!bloom_filter_->ReportFalsePositives(false_positives))
+  if (bloom_filter_->ReportFalsePositives(false_positives))
```
wkaluza, line 1547: It seems this change offsets the other one - either invert the inequality in `bloom_filter.cpp` or remove the negation here, but not both.

I'd say we keep the inequality as it was - my (un)original thinking was that returning `false` corresponds to some failure or otherwise undesirable condition, but I got the logic in this line the wrong way around.

wkaluza (changes requested):

libs/ledger/src/chain/main_chain.cpp
pbukva: Both is necessary.
Based on original implementation of the `ReportFalsePositives` method, specifically its trivial case `return false;`, the intended behaviour of the of the method is to return `false` if there are *NO* false positives detected in the *trivial* case, or return `true` if false positives exceeds specific criteria. That corresponds to naming of the method.

In the *non*-trivial case where the ratio `positive_count_ / false_positive_count_` is actuallly evaluated, then false positives are supposed to be reported (return `true`) *IF* `positive_count_ / false_positive_count_ < MINIMAL_POSITIVE_TO_FALSE_POSITIVE_RATIO_CONSTANT`.

wkaluza: Fair enough. In that case please update the docstring in `bloom_filter.hpp` to reflect the new behaviour.

wkaluza (approved):

PR #1598, opened 2019-09-02
FixedPoint overflows fix by adding epsilons

MissingNO57:


baykaner (changes requested):
lets ditch this - i don't think it's the way forward. or else implement the clamping in the softmax as discussed

PR #1600, opened 2019-09-02
Fixes for JSON parsing of integer and floating point values

ejfitzgerald:


wkaluza:
Interesting. Was the old code failing, or is this cosmetic?

wkaluza (approved):

PR #1602, opened 2019-09-02
Reactor Safety Fixes

ejfitzgerald:
* Add exception handling at the bottom of the reactor
* Catch the case where bootstrap notifications fail and handle gracefully


wkaluza (approved):
Looks good

PR #1603, opened 2019-09-03
Type dispatch refactored in vm, and in smart contracts

bipll:
This commit simplifies and deduplicates vm code by getting _absolutely_ rid of numerous `case TypeIds::FieldN: lhsv.primary.fieldN = DoTheSameForEachN(rhsv.fieldN);` switches found all over the vm libs. Every such switch was replaced by a single few-lines call to a very small library function. A similar reduction was applied to smart_contract.cpp

To achieve this, a `Switch` template is introduced, similar to (though more generic than) `ApplyFunctor` found earlier in a now-closed PR (and `ApplyFunctor` is redefined to use `Switch` and be a more convenient tool), and a simple pack mechanics, contained in meta/pack.hpp, to back it up. The latter implements all the same variadic operations found earlier in meta/type_util.hpp, but does encapsulate variadic parameter packs in a special dedicated type Pack so they can be composed, combined and reused.


robertdickson:
While I appreciate that it would be nice to have a general mechanism for dispatching on type ids, in order to reduce the code repetition of boilerplate switch statements, I am very far from convinced that this PR is the answer. 
 
The solution presented here has a number of issues, the severest of which is that, after stripping away templating, the dispatch essentially boils down to a binary chop on sorted ids followed by a linear scan, with each decision incurring the heavy penalty of an if statement. Another issue, for me at any rate, is that the readability of the code at the point of use (especially in those cases where the case handler is non-generic, see e.g. the smart contract AddToParameterPack()) is much less than if a plain switch was used.

It seems to me that this PR, in the interests of avoiding a few repeated switches, will face an unnecessary efficiency penalty because its generalised nature replaces the power of a maximally efficient C++ switch statement with a logarithmic scan of ids implemented as a series of repeated if's. A few examples might serve to make the point:

a) type dispatch on numeric operations such as addition or multiplication on Float64 requires 3 if statements, on UInt64, 4 if statements
b) a constructor like State < Address > () or Array < String > () requires 6 if statements to implement the switch

If we realise that the type dispatch exemplified by a) and b) will be executing in the VM constantly, in all sorts of critical paths, it would seem that the efficient implementation of a modern compiler's plain old switch should perhaps trump other considerations such as code repetition. This code is executed _a lot_.

I believe the issue here is that the type dispatch mechanism is trying to be too general at the expense of efficiency. If a way could be found, though, to employ some kind of templated master switch that compiles in lambdas for the used cases, and compiles away unused cases, I would happily welcome it.

ejfitzgerald:
Closing because stale

PR #1612, opened 2019-09-04
Remove workaround

### LINKED ISSUE #1533: state.set does not make a deep copy of non-primitive objects
this leads to a bug where an object can be modified after being set:

e.g. 

```
function main()

   var obj = MySpecialObject();
   var state = State<MySpecialObject>("my_object");
   state.set(obj);

   obj.Foo() // modifies internal state
endfunction

Now deserialising

function main()
        var state = State<MySpecialObject>("my_object");
        var obj = state.get();
        obj.Foo() // this is now the second call to obj when it should be the first
endfunction
```

wkaluza:
Closes #1533

### DIFF
--- a/libs/vm_modules/tests/unit/ml.cpp
+++ b/libs/vm_modules/tests/unit/ml.cpp
@@ -332,16 +332,8 @@ TEST_F(MLTests, sgd_optimiser_serialisation_test)
       var state = State<Optimiser>("optimiser");
       state.set(optimiser);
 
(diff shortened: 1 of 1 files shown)


wkaluza:
@baykaner are there any more of these workarounds that we can now get rid of?

baykaner (approved):

ejfitzgerald:
Nice one @wkaluza and @baykaner

PR #1615, opened 2019-09-04
Consolidate synergetic and smart contracts

### LINKED ISSUE #1686: Comment and code mismatch
Comment and code mismatch in `synergetic_contract.cpp:82`:

```
  // loop through all the function annotations
  if (1u == fn.annotations.size())
  {
```

wkaluza:
Closes #1686


apps/constellation/constellation.cpp
```diff
@@ -272,20 +271,17 @@ Constellation::Constellation(CertificatePtr certificate, Config config)
...
+  reactor_.Attach(dag_service_->GetWeakRunnable());
 
-    auto syn_miner = std::make_unique<NaiveSynergeticMiner>(dag_, *storage_, certificate);
-    if (!reactor_.Attach(syn_miner->GetWeakRunnable()))
-    {
-      FETCH_LOG_ERROR(LOGGING_NAME, "Failed to attach synergetic miner to reactor.");
-      throw std::runtime_error("Failed to attach synergetic miner to reactor.");
-    }
-    synergetic_miner_ = std::move(syn_miner);
+  auto syn_miner = std::make_unique<NaiveSynergeticMiner>(dag_, *storage_, certificate);
```
ejfitzgerald: For the moment I would like to keep the synergetic miner controlled via feature flag.

apps/constellation/constellation.cpp
```diff
@@ -222,7 +216,7 @@ Constellation::Constellation(CertificatePtr certificate, Config config)
   , storage_(std::make_shared<StorageUnitClient>(internal_muddle_.AsEndpoint(), shard_cfgs_,
                                                  cfg_.log2_num_lanes))
   , lane_control_(internal_muddle_.AsEndpoint(), shard_cfgs_, cfg_.log2_num_lanes)
-  , dag_{GenerateDAG(cfg_.features.IsEnabled("synergetic"), "dag_db_", true, certificate)}
+  , dag_{GenerateDAG("dag_db_", true, certificate)}
```
ejfitzgerald, line 273: While I agree that this should PR should enable this by default. I would still like the option to disable it if it starts causing issues - at least in the short term

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -117,7 +108,7 @@ BlockCoordinator::BlockCoordinator(MainChain &chain, DAGPtr dag, StakeManagerPtr
   , tx_wait_periodic_{TX_SYNC_NOTIFY_INTERVAL}
   , exec_wait_periodic_{EXEC_NOTIFY_INTERVAL}
   , syncing_periodic_{NOTIFY_INTERVAL}
-  , synergetic_exec_mgr_{CreateSynergeticExecutor(features, dag_, storage_unit_)}
+  , synergetic_exec_mgr_{CreateSynergeticExecutor(dag_, storage_unit_)}
```
ejfitzgerald, line 111: See comment above but I suspect this will need to be reinstated

scripts/ci/install-test-dependencies.sh
```diff
@@ -1,9 +1,7 @@
...
-pip3 install wheel
-
-pip3 install fetchai-ledger-api==0.5.1
-pip3 install -i https://test.pypi.org/simple/ fetchai-netutils==0.0.6a1
-
-pip3 install pyyaml
+pip3 install --no-cache-dir --ignore-installed --force-reinstall wheel pyyaml requests
+# ???
+pip3 install --no-cache-dir --ignore-installed --force-reinstall git+https://github.com/wkaluza/ledger-api-py.git@wip_consolidate_smart_and_synergetic_contracts#egg=fetchai-ledger-api
+pip3 install --no-cache-dir --ignore-installed --force-reinstall --index-url https://test.pypi.org/simple/ fetchai-netutils==0.0.6a1
```
ejfitzgerald: I guess these will need to be updated before the merge

libs/ledger/src/upow/synergetic_contract.cpp
```diff
@@ -330,8 +288,7 @@ Status SynergeticContract::Complete(uint64_t block, BitVector const &shards)
 
   // setup the storage infrastructure
   CachedStorageAdapter storage_cache(*storage_);
-  StateSentinelAdapter state_sentinel{
-      storage_cache, Identifier{digest_.ToHex() + "." + std::to_string(block)}, shards};
+  StateSentinelAdapter state_sentinel{storage_cache, Identifier{digest_.ToHex()}, shards};
```
ejfitzgerald: Maybe I am wrong but this presumable sets the state space prefix as:

`<digest>.state.`

Where as we want:

`<digest>.<address>.state.`

libs/ledger/src/upow/synergetic_execution_manager.cpp
```diff
@@ -99,7 +100,7 @@ ExecStatus SynergeticExecutionManager::PrepareWorkQueue(Block const &current, Bl
     }
 
     // attempt to lookup the contract being referenced
-    auto it = work_map.find(node.contract_digest);
+    auto it = work_map.find(node.contract_digest.ToHex());
```
ejfitzgerald: I think that internal data structures should just store the binary version not the hex version

libs/ledger/src/upow/synergetic_executor.cpp
```diff
@@ -51,7 +51,7 @@ void SynergeticExecutor::Verify(WorkQueue &solutions, ProblemData const &problem
       auto const &contract_address = solution->contract_digest();
 
       // create the contract
-      contract = factory_.Create(contract_address);
+      contract = factory_.Create(contract_address.ToHex());
```
ejfitzgerald: This looks a little confusing here because we have `contract_address` which isn't the new contract address it is an alias for `contract_digest`.

I also think that we should rework the interface of this factory so that it always expects a 32 byte digest (raw) and that internally it does the `.ToHex()`

libs/ledger/tests/chaincode/smart_contract_tests.cpp
```diff
@@ -116,7 +116,7 @@ TEST_F(SmartContractTests, CheckSimpleContract)
   EXPECT_TRUE(IsIn(query_handlers, "value"));
 
   // define our what we expect the values to be in our storage requests
-  auto const expected_key      = contract_name_->full_name() + ".state.value";
+  auto const expected_key      = contract_name_->qualifier() + ".state.value";
```
ejfitzgerald: Why has this changed?

ejfitzgerald (changes requested):

scripts/ci/install-test-dependencies.sh
wkaluza: Yes, the API will need to be reviewed and released before this can be merged. I created a PR; https://github.com/fetchai/ledger-api-py/pull/34

apps/constellation/constellation.cpp
wkaluza: Done.

apps/constellation/constellation.cpp
wkaluza: Done.

libs/ledger/src/chain/block_coordinator.cpp
wkaluza: Done.

libs/ledger/src/upow/synergetic_execution_manager.cpp
wkaluza: Done.

ejfitzgerald (dismissed):

ejfitzgerald (approved):

PR #1626, opened 2019-09-05
Distributed learning export buffer

MissingNO57:
Queue timestamp
Coordinator has list of clients
Export buffer for clients


libs/math/include/math/base_types.hpp
```diff
@@ -28,9 +28,10 @@
 namespace fetch {
 namespace math {
 
-using SizeType   = uint64_t;
-using SizeVector = std::vector<SizeType>;
-using SizeSet    = std::unordered_set<SizeType>;
+using SizeType    = uint64_t;
+using PtrDiffType = int64_t;
```
baykaner, line 32: it seems to be that this is not a type that is really fundamental to the math library.
This seems relatively example specific

baykaner (changes requested):

baykaner (approved):

PR #1627, opened 2019-09-05
Muddle Tweaks

ejfitzgerald:
* Fix issue with manifest queries
* Update logging to include network name for easier debugging


apps/beacon/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 #
 # BEACON
 #
+cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
```
ejfitzgerald, line 4: Whoops this should not be here

jinmannwong (approved):
Looks good to me.

PR #1629, opened 2019-09-06
Fix for graph serialisation size bug

evsmithx:
Graph serialisation produced a huge file because the placeholder nodes saved their input data. This fixed modifies placeholders to not do this.


MissingNO57 (approved):
LGTM

PR #1631, opened 2019-09-07
Small cleanup to network/service/function.hpp.

bipll:


ejfitzgerald:
Nice catch. I am actually pretty certain this file is not used anymore. I would recommend trying to remove it actually.

ejfitzgerald:
Cancel my last it is still relevant in master at the moment, however, I am done some tidying on the consensus branch and have removed it. LGTM

ejfitzgerald (approved):

PR #1632, opened 2019-09-09
Refactoring consensus and stake manager

jinmannwong:
Move block to block management of committees from stake_manager to consensus


libs/ledger/src/consensus/consensus.cpp
```diff
@@ -27,20 +30,173 @@ using NextBlockPtr    = Consensus::NextBlockPtr;
...
+
+  // Invalid to request a committee too far ahead in time
+  assert(committee_history_.find(last_snapshot) != committee_history_.end());
+
+  if (committee_history_.find(last_snapshot) == committee_history_.end())
+  {
+    FETCH_LOG_INFO(LOGGING_NAME,
+                   "No committee history found for block: ", previous.body.block_number, " AKA ",
+                   last_snapshot);
+    // TODO(jmw): Check: should return nullptr here?
```
n-hutton: Yup, it looks like this is only used in consensus.cpp and it checks for a nullptr

n-hutton (approved):
LGTM

PR #1633, opened 2019-09-09
Move function definitions to source files

wkaluza:


libs/crypto/include/crypto/bls_base.hpp
```diff
@@ -56,46 +56,16 @@ enum
 
 void Init();
 
-inline PrivateKey PrivateKeyByCSPRNG()
-{
-  PrivateKey ret;
+PrivateKey PrivateKeyByCSPRNG();
```
ejfitzgerald: Can we revert this file

libs/ledger/include/ledger/chaincode/token_contract.hpp
```diff
@@ -74,15 +74,5 @@ class TokenContract : public Contract
   StakeUpdates stake_updates_;
 };
 
-inline void TokenContract::ClearStakeUpdates()
```
ejfitzgerald: Can we revert this file

ejfitzgerald (changes requested):

libs/ledger/include/ledger/chaincode/token_contract.hpp
wkaluza: Done

libs/crypto/include/crypto/bls_base.hpp
wkaluza: Done

wkaluza (commented):
Reverted

ejfitzgerald (approved):

PR #1634, opened 2019-09-10
Modify DKG for timeouts

jinmannwong:
- Include in complaint management the scenario of missing messages, which should give rise to complaints on the sender
- Removed complaints managers in dkg directory and moved them into beacon directory with some refactoring
- Enabled some disabled tests for the DKG


libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -440,22 +449,13 @@ BeaconSetupService::State BeaconSetupService::OnWaitForShares()
...
-
-  if (!condition_to_proceed_ && is_ok)
-  {
-    condition_to_proceed_ = true;
-    FETCH_LOG_INFO(LOGGING_NAME, "State: ", ToString(state_machine_->state()),
-                   " Ready. Seconds to spare: ", state_deadline_ - GetTime(), " of ",
-                   seconds_for_state_);
-  }
-
   if (timer_to_proceed_.HasExpired())
```
jinmannwong: I removed these as they resulted in inconsistency in the test runs - nodes would fail when they shouldn't when these were included. The tests appear to pass consistently when nothing is done before the timer has expired.

libs/beacon/include/beacon/beacon_complaints_manager.hpp
```diff
@@ -26,7 +26,7 @@
 #include <vector>
 
 namespace fetch {
-namespace dkg {
+namespace beacon {
```
n-hutton, line 29: 👍

libs/beacon/src/beacon_complaints_manager.cpp
```diff
@@ -0,0 +1,319 @@
...
+#include <mutex>
+
+namespace fetch {
+namespace beacon {
+
+constexpr char const *LOGGING_NAME = "DKGComplaints";
+
+void ComplaintsManager::ResetCabinet(MuddleAddress const &address, uint32_t threshold)
+{
+  FETCH_LOCK(mutex_);
```
n-hutton, line 31: We should at some point review which of these classes we actually need to be thread safe, probably this doesn't need to be the way it's used

libs/beacon/src/beacon_complaints_manager.cpp
```diff
@@ -0,0 +1,319 @@
...
+    {
+      complaints_from_.insert(from_id);
+    }
+  }
+}
+
+void ComplaintsManager::Finish(std::set<Identity> const &cabinet)
+{
+  FETCH_LOCK(mutex_);
+  if (!finished_)
```
n-hutton, line 78: I thought that the concept of 'finishing' was going away as you only fail on qual?

libs/beacon/src/beacon_manager.cpp
```diff
@@ -552,9 +542,9 @@ BeaconManager::AddResult BeaconManager::AddSignaturePart(Identity const & from,
                                                          Signature const &signature)
 {
   auto it = identity_to_index_.find(from.identifier());
-  assert(it != identity_to_index_.end());
+  auto iq = qual_.find(from.identifier());
```
n-hutton, line 545: iq?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -726,7 +729,7 @@ BeaconSetupService::State BeaconSetupService::OnDryRun()
 
     for (auto const &key_and_count : dry_run_public_keys_)
     {
-      if (key_and_count.second >= beacon_->manager.polynomial_degree())
+      if (key_and_count.second >= QualSize())
```
n-hutton, line 739: You swapped threshold for qualsize here? isn't the qual size variable?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -329,6 +328,12 @@ uint64_t BeaconSetupService::PreDKGThreshold()
   return ret;
 }
 
+uint32_t BeaconSetupService::QualSize()
+{
+  // Set to 2/3n for now
```
n-hutton, line 333: Won't this fall over if the threshold is higher than this?

libs/beacon/tests/dkg/dkg_resilience.cpp
```diff
@@ -463,10 +471,11 @@ struct HonestDkgMember : DkgMember
...
 
     // Setting the aeon details
-    beacon->aeon.round_start               = 0;
-    beacon->aeon.round_end                 = 10;
-    beacon->aeon.members                   = std::move(cabinet);
-    beacon->aeon.start_reference_timepoint = static_cast<uint64_t>(std::time(nullptr));
+    beacon->aeon.round_start = 0;
+    beacon->aeon.round_end   = 10;
+    beacon->aeon.members     = std::move(cabinet);
+    // Plus 5 so tests pass on first DKG attempt
```
n-hutton, line 477: 👍

libs/beacon/src/beacon_complaints_manager.cpp
jinmannwong: The finish is not about failing. It is about telling the complaints manager we have ended the wait for complaints messages, which is important as it then adds to complaints all nodes it did not get a message from.

libs/beacon/src/beacon_manager.cpp
jinmannwong: We should check that the sender is both in the cabinet and in qual.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: Yes, this is true. There should be a check that the QualSize is larger than threshold. Unfortunately this choice of qual size was a bit arbitrary so we don't have a nice analytical expression for it in terms of the threshold.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: This is what David proposed for the dry run. That we set a minimum for the qual size, which is higher than the signature threshold, and impose that we only succeed in the dry run if we meet at least minimum qual size number of people with the same group public key. QualSize() is not variable, it is some minimum we impose whereas the actual size of qual is indeed variable.

n-hutton (approved):
lgtm

PR #1636, opened 2019-09-10
Distributed w2v fixes

MissingNO57:
W2V dataloader end of file bug fix
W2V client data splitting fixed
Client learning rate decay fixed
Removed leftover code
Added client's train and test loss member variables


libs/ml/examples/word2vec_distributed_learning/main.cpp
```diff
@@ -55,9 +56,19 @@ std::vector<std::string> SplitTrainingData(std::string const &train_file,
...
+      // Last client gets the rest
+      pos = input_data.size();
+    }
+    else
+    {
+      // find next instance of space character
+      pos = input_data.find(" ", pos, 1);
+    }
+
+    client_data.push_back(input_data.substr(old_pos, pos - old_pos));
```
evsmithx, line 61: The if-else above is unnecessary now that substr uses (pos - old_pos). If the length is longer than the input string then it just goes to the end of the string.

evsmithx (dismissed):
LGTM

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -396,7 +395,7 @@ void GraphW2VLoader<T>::BufferNextSamples()
             window_size_;  // the current word start from position that allows full context window
         current_sentence_++;
       }
-      if (IsDone())
+      if (IsDone() || current_sentence_ >= data_.size())
```
evsmithx, line 397: If current_sentence_ >= data_.size() won't IsDone() be true anyway? I think the second condition here is unnecessary.

MissingNO57: IsDone is checking for data in buffer and when there are data it continues, but we can be in next non-existing sentence and crash. This was the only way I found to pass W2V tests and not be crashing on distributed W2V training.

baykaner (approved):

PR #1637, opened 2019-09-10
Ml remove graph step

baykaner:


MissingNO57 (approved):
LGTM

libs/ml/tests/ml/layers/PRelu.cpp
```diff
@@ -230,14 +230,16 @@ TYPED_TEST(PReluTest, saveparams_test)
   TypeParam loss = layer.Evaluate(error_output);
   layer.BackPropagate(error_output);
   layer.ApplyRegularisation();
-  layer.Step(DataType{0.1f});
```
evsmithx, line 233: With Step there was the option to set the learning rate - is there an equivalent with GetGradients + ApplyGradients?

evsmithx (dismissed):

libs/ml/tests/ml/layers/PRelu.cpp
```diff
@@ -230,14 +230,16 @@ TYPED_TEST(PReluTest, saveparams_test)
...
+  layer.ApplyGradients(grads);
 
   // train g2
   layer2.SetInput(label_name, labels);
   TypeParam loss2 = layer2.Evaluate(error_output);
   layer2.BackPropagate(error_output);
   layer2.ApplyRegularisation();
-  layer2.Step(DataType{0.1f});
+  auto grads2 = layer2.GetGradients();
+  layer2.ApplyGradients(grads2);
```
MissingNO57, line 250: I've just realised that you need to multiply gradient by -LEARNING_RATE otherwise you are maximising the error.

MissingNO57 (changes requested):

libs/ml/tests/ml/layers/PRelu.cpp
MissingNO57: Or use Optimiser which would do it for you.

baykaner: oh that's true. it actually makes no difference for this test; but it's good to demonstrate correct usage :-p I'll fix that

libs/ml/tests/ml/layers/PRelu.cpp
baykaner: no. but there shouldn't be.
This is because Step was effectively doing SGD without pointing that out explicitly.

If we really want to do SGD without using the SGDOptimiser there's nothing that stops us doing what @MissingNO57 proposed (e.g. graph.ApplyGradients(-LEARNING_RATE * graph.GetGradients())), but then at least we're being explicit about what we're doing.

MissingNO57 (dismissed):
LGTM

MissingNO57 (dismissed):
LGTM

MissingNO57 (dismissed):
LGTM?

MissingNO57 (approved):
LGTM!

PR #1638, opened 2019-09-10
HTTP tx submission - make bulk and JSON modes consistent

wkaluza:


libs/ledger/src/chaincode/contract_http_interface.cpp
```diff
@@ -332,22 +365,14 @@ ContractHttpInterface::SubmitTxStatus ContractHttpInterface::SubmitJsonTx(
   FETCH_LOG_DEBUG(LOGGING_NAME, "NEW TRANSACTION RECEIVED");
   FETCH_LOG_DEBUG(LOGGING_NAME, request.body());
 
-  FETCH_UNUSED(expected_contract);
```
ejfitzgerald, line 335: Let's have a chat about this F2F. However, the original intension here (and it appears to have been lost in `master`) is that based on the submission endpoint, there is a restriction on the target contract that can be submitted.

The principal idea here is that by imposing this restriction we can more closely monitor the transactions that are being submitted externally.

ejfitzgerald (changes requested):
Looks good, lets catch up regarding the `expected_contract` stuff

libs/ledger/src/chaincode/contract_http_interface.cpp
wkaluza: Ok, good to know. What caught my eye initially was the differences between how we treat 'bulk' and JSON submissions, so I would like to have a word about that too.

ejfitzgerald (approved):
As discussed - we will submit this change for the short term and then revisit adding back the restrictions into the HTTP endpoint

PR #1649, opened 2019-09-12
Feature/efficient signature building

n-hutton:
Change the way threshold sigs are built by using RPC pulls rather than broadcasts - this is due to there being a bug.


libs/beacon/include/beacon/beacon_manager.hpp
```diff
@@ -160,14 +160,13 @@ struct MapSerializer<dkg::BeaconManager::SignedMessage, D>
...
-  static uint8_t const PUBLIC_KEY = 1;
-  static uint8_t const IDENTITY   = 2;
+  static uint8_t const SIGNATURE = 0;
+  static uint8_t const IDENTITY  = 1;
 
   template <typename Constructor>
   static void Serialize(Constructor &map_constructor, Type const &member)
   {
-    auto map = map_constructor(3);
+    auto map = map_constructor(2);
```
n-hutton, line 169: @jinmannwong I think at some point you changed this serializer - it wasn't quite correct since the map_constructor needs to take an int of how many things you are serializing (changed to 2)

jinmannwong: Thanks, it was gross negligence on my part that I removed the public key in some parts and not others.

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -215,19 +215,28 @@ BeaconSetupService::State BeaconSetupService::OnReset()
   shares_received_.clear();
   dry_run_shares_.clear();
   dry_run_public_keys_.clear();
-  pre_dkg_rbc_.ResetCabinet({});
-  rbc_.ResetCabinet({});
+  pre_dkg_rbc_.Enable(false);
```
n-hutton, line 218: 'disable' the RBC when in reset mode - note signature shares etc. can still arrive (I have seen them arrive 20s after transmission)

scripts/end_to_end_test/run_end_to_end_test.py
```diff
@@ -310,7 +310,8 @@ def stop(self):
         if self._nodes:
             for n, node in enumerate(self._nodes):
                 print('Stopping Node {}...'.format(n))
-                node.stop()
+                if(node):
```
n-hutton, line 313: ignore this file

scripts/end_to_end_test/end_to_end_test_pos.yaml
```diff
@@ -5,7 +5,7 @@
 #setup_conditions:
 #    test_name : "multinode_POS_with_destaking"
 #    number_of_nodes: 4
-#    node_connections: [[0, 1], [0, 2], [0,3]]
+#    node_connections: [[0, 1], [0, 2], [0, 3], [2, 3]]
```
n-hutton, line 8: ignore this file

libs/core/src/logging.cpp
```diff
@@ -21,9 +21,28 @@
 #include "telemetry/counter.hpp"
 #include "telemetry/registry.hpp"
 
+#if defined(__GNUC__)
```
n-hutton: ignore file

libs/beacon/src/beacon_service.cpp
```diff
@@ -574,6 +610,18 @@ bool BeaconService::AddSignature(SignatureShare share)
...
+  {
+    FETCH_LOG_INFO(LOGGING_NAME, "Accidental duplicate signature added!");
+  }
+
+  if (ret == BeaconManager::AddResult::INVALID_SIGNATURE)
+  {
+    FETCH_LOG_INFO(LOGGING_NAME, "Invalid signature found!");
+    return false;
+  }
+
```
jinmannwong: The case of INVALID_SIGNATURE is already considered above.

libs/beacon/include/beacon/beacon_service.hpp
```diff
@@ -122,14 +122,20 @@ class BeaconService : public ledger::EntropyGeneratorInterface
 
   friend class BeaconServiceProtocol;
 
+  struct SignatureInformation
+  {
+    uint64_t                                               round{uint64_t(-1)};
+    std::map<BeaconManager::MuddleAddress, SignatureShare> threshold_signatures;
+  };
```
jinmannwong, line 129: Is the muddle address of the person signing stored twice here? I believe it is already in SignatureShare and therefore one of them seems redundant.

jinmannwong (approved):
Looks good. I left a two comments on minor things.

libs/beacon/src/beacon_service.cpp
n-hutton: You're right, thanks! fixed.

libs/beacon/include/beacon/beacon_service.hpp
n-hutton: It is a bit redundant but it is just for easier lookup. Instead of a map it could be a vector, or only the share could be stored as the value. For now I'll leave it for another time

PR #1651, opened 2019-09-12
initial commit

n-hutton:


libs/core/src/logging.cpp
```diff
@@ -209,9 +242,17 @@ LogRegistry::Logger &LogRegistry::GetLogger(char const *name)
   auto it = registry_.find(name);
   if (it == registry_.end())
   {
-    // create the new logger instance
-    auto logger = spdlog::stdout_color_mt(name, spdlog::color_mode::automatic);
+    // create the new logger instance - note it suppresses duplicate messages
+    auto dup_filter =
+        std::make_shared<spdlog::sinks::dup_filter_sink_mt>(std::chrono::milliseconds(100));
+    dup_filter->add_sink(std::make_shared<spdlog::sinks::stdout_color_sink_mt>());
```
ejfitzgerald: I think that it probably makes sense to have a single `stdout_color_sink_mt` for all logger instances.

ejfitzgerald (changes requested):

libs/core/src/logging.cpp
n-hutton: done

ejfitzgerald (approved):

PR #1653, opened 2019-09-12
Etch squeeze catch wrong shape

baykaner:
- fixes tensor Squeeze so that it will not squeeze a tensor with no dimensions of size 1
- changes tensor squeeze so that the first size 1 dimension (starting from the traling edge) is squeezed


MissingNO57 (approved):
LGTM

PR #1654, opened 2019-09-12
Tests for mcl crypto and beacon manager

jinmannwong:
Tests for crypto computations in mcl crypto and beacon manager. Beacon manager tests goes through dkg and threshold signing functions.


libs/beacon/src/beacon_manager.cpp
```diff
@@ -45,18 +45,7 @@ BeaconManager::BeaconManager(CertificatePtr certificate)
...
-    // Values taken from TMCG main.cpp
-    const bn::Fp2 g(
-        "1380305877306098957770911920312855400078250832364663138573638818396353623780",
-        "14633108267626422569982187812838828838622813723380760182609272619611213638781");
-    const bn::Fp2 h(
-        "6798148801244076840612542066317482178930767218436703568023723199603978874964",
-        "12726557692714943631796519264243881146330337674186001442981874079441363994424");
-    bn::mapToG2(group_g_, g);
-    bn::mapToG2(group_h_, h);
+    crypto::mcl::SetGenerators(group_g_, group_h_);
```
n-hutton, line 48: 👍

libs/beacon/src/beacon_manager.cpp
```diff
@@ -169,13 +158,7 @@ std::unordered_set<BeaconManager::MuddleAddress> BeaconManager::ComputeComplaint
...
 {
   CabinetIndex from_index = identity_to_index_[from];
-  if (identity_to_index_.find(answer.first) == identity_to_index_.end())
-  {
-    FETCH_LOG_WARN(LOGGING_NAME, "Node ", cabinet_index_,
-                   " received complaint answer with unknown reporter index");
-    return true;
-  }
-
+  assert(identity_to_index_.find(answer.first) != identity_to_index_.end());
```
n-hutton, line 161: So, just to check, assertions should be used for something that should never be able to happen. There is no way a node can submit information to get past this point is there?

libs/beacon/tests/dkg/beacon_manager_tests.cpp
```diff
@@ -0,0 +1,378 @@
...
+{
+  PrivateKey              private_key;
+  PublicKey               group_public_key;
+  std::vector<PublicKey>  public_key_shares;
+  std::set<MuddleAddress> qual;
+};
+
+TEST(beacon_manager, dkg_and_threshold_signing)
+{
+  bn::initPairing();
```
n-hutton, line 43: Is this going to do bad stuff if there is multiple tests that call initPairing?

libs/beacon/tests/dkg/beacon_manager_tests.cpp
```diff
@@ -0,0 +1,378 @@
...
+    EXPECT_TRUE(manager->qual().empty());
+    EXPECT_TRUE(manager->group_public_key() == zero.getStr());
+  }
+
+  for (auto &manager : beacon_managers)
+  {
+    manager->GenerateCoefficients();
+  }
+
+  // Checks
```
n-hutton: Need to flesh out this comment a little

libs/beacon/tests/dkg/beacon_manager_tests.cpp
```diff
@@ -0,0 +1,378 @@
...
+  for (uint32_t index = 0; index < cabinet_size; ++index)
+  {
+    std::shared_ptr<BeaconManager> manager      = beacon_managers[index];
+    std::vector<std::string>       coefficients = manager->GetCoefficients();
+    for (uint32_t elem = 0; elem < threshold; ++elem)
+    {
+      // Coefficients generated should be non-zero
```
n-hutton, line 108: 👍

libs/beacon/tests/dkg/beacon_manager_tests.cpp
```diff
@@ -0,0 +1,378 @@
...
+  // in correctly
+  beacon_managers[0]->AddShares(honest, beacon_managers[1]->GetOwnShares(my_address));
+  beacon_managers[0]->AddCoefficients(honest, beacon_managers[1]->GetCoefficients());
+  EXPECT_EQ(beacon_managers[0]->GetReceivedShares(honest),
+            beacon_managers[1]->GetOwnShares(my_address));
+
+  // Add shares and coefficients failing verification from malicious party
+  PrivateKey s_i;
+  PrivateKey sprime_i;
+  s_i.setStr(beacon_managers[2]->GetOwnShares(my_address).first);
```
n-hutton: this line is a bit confusing. Why is getownshares taking an address?
I would probably split this into two lines to make it a bit clearer what the .first refers to

libs/crypto/tests/unit/mcl_dkg_tests.cpp
```diff
@@ -0,0 +1,271 @@
...
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "crypto/mcl_dkg.hpp"
+#define MCLBN_FP_UNIT_SIZE 4
```
n-hutton: I am not sure this define here will do anything if it happens after the MCL library has been parsed and included.

n-hutton: Also, what does FP mean? Floating point?

libs/crypto/tests/unit/mcl_dkg_tests.cpp
```diff
@@ -0,0 +1,271 @@
...
+#include <cstdint>
+#include <iostream>
+#include <ostream>
+
+using namespace fetch::crypto::mcl;
+using namespace fetch::byte_array;
+
+TEST(MclTests, BaseMcl)
+{
+  const char *aa = "12723517038133731887338407189719511622662176727675373276651903807414909099441";
```
n-hutton: What are these? Variable names should be descriptive if possible

libs/crypto/tests/unit/mcl_dkg_tests.cpp
```diff
@@ -0,0 +1,271 @@
...
+  const char *ba = "13891744915211034074451795021214165905772212241412891944830863846330766296736";
+  const char *bb = "7937318970632701341203597196594272556916396164729705624521405069090520231616";
+
+  mcl::bn::initPairing();
+  mcl::bn::G2 Q(mcl::bn::Fp2(aa, ab), mcl::bn::Fp2(ba, bb));
+  mcl::bn::G1 P(-1, 1);
+
+  // Comparison
+  {
+    mcl::bn::G2 Q0;
```
n-hutton, line 39: Perhaps I should get Troels to scan this as I am not familiar with these operations

n-hutton (changes requested):

libs/beacon/src/beacon_manager.cpp
jinmannwong: It is checked in beacon_setup_service that answer.first is in identity_to_index so the check here was redundant. I put the assert there to replace it for exactly the reason you said, it should not happen.

libs/crypto/tests/unit/mcl_dkg_tests.cpp
jinmannwong: They correspond to points on the elliptic curve.

libs/crypto/tests/unit/mcl_dkg_tests.cpp
jinmannwong: They are addition and multiplication operations on the curve. This test is just a basic test to make sure mcl is working properly. We can remove it if we want.

libs/crypto/tests/unit/mcl_dkg_tests.cpp
jinmannwong: Fp is a class in the mcl library, which has a private member that is an array. You can correct that the define like doesn't doing anything and wasn't needed.

libs/beacon/tests/dkg/beacon_manager_tests.cpp
jinmannwong: GetOwnShares takes an address because it returns the pair of shares computed for and sent to a particular cabinet member. I will split into two lines.

libs/beacon/tests/dkg/beacon_manager_tests.cpp
jinmannwong: No, initPairing can be called multiple times. It is just not thread safe.

n-hutton (approved):

PR #1655, opened 2019-09-12
Miscellaneous changes

wkaluza:


libs/core/src/byte_array/const_byte_array.cpp
```diff
@@ -24,12 +24,12 @@ namespace byte_array {
 
 ConstByteArray ConstByteArray::ToBase64() const
 {
-  return ::fetch::byte_array::ToBase64(*this);
+  return fetch::byte_array::ToBase64(*this);
```
ejfitzgerald, line 27: Don’t need namespaces here at all

ejfitzgerald (commented):
Seems legit. Worth getting @baykaner to have a look through too. One minor pedantic comment

libs/core/src/byte_array/const_byte_array.cpp
wkaluza: I thought so too, but we do need them here - at least on Clang 6 without namespace qualification this call resolves to `ConstByteArray::ToBase64` and fails with `too many arguments to function call, expected 0, have 1`.

ejfitzgerald: Of course you are correct! 😁

baykaner (approved):

libs/core/src/byte_array/const_byte_array.cpp
wkaluza: Why can't the compiler just *guess* which function I mean? Stupid compiler.

PR #1656, opened 2019-09-12
Added VMGraph::Step and tests

evsmithx:
Reimplemented Step for VMGraph

### DIFF
--- a/libs/vm_modules/src/ml/graph.cpp
+++ b/libs/vm_modules/src/ml/graph.cpp
@@ -75,7 +75,7 @@ void VMGraph::Step(DataType const &lr)
   auto grads = graph_.GetGradients();
   for (auto &grad : grads)
   {
-    grad *= DataType{-lr};
+    grad *= static_cast<DataType>(-lr);
   }
   graph_.ApplyGradients(grads);
 }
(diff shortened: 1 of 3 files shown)


libs/vm_modules/src/ml/graph.cpp
```diff
@@ -70,6 +70,16 @@ void VMGraph::BackPropagate(VMPtrString const &name)
   graph_.BackPropagate(name->str);
 }
 
+void VMGraph::Step(DataType const &lr)
+{
+  auto grads = graph_.GetGradients();
+  for (auto &grad : grads)
+  {
+    grad *= DataType{-lr};
```
baykaner: style,
grad *= static_cast<DataType>(-lr);
preferred

libs/vm_modules/tests/unit/ml.cpp
```diff
@@ -492,4 +492,43 @@ TEST_F(MLTests, optimiser_set_graph_test)
   ASSERT_TRUE(toolkit.Run());
 }
 
+TEST_F(MLTests, graph_step_test)
+{
+  static char const *src = R"(
+    function main() : Fixed64
```
baykaner: you can return the loss difference tensor instead of fixedpoint64 value.
This is preferable because Fixed64 is a (necessary) hard-coding of the value set in vm_modules/math/base_types.hpp
If we decide to replace this base type ever (quite likely we would replace it with fp128 when available), we will need to update every hard-coded value

baykaner (changes requested):

baykaner (approved):

Issue #1657, opened 2019-09-13
Remove dependency on personal forks


wkaluza:
We currently use `benchmark`, `utfcpp` and `backward-cpp` from forks made by ledger team members. We should point the submodules at the original repos. if we require custom changes to the libraries, let's try to submit PRs to the respective projects or at the very least use forks owned by the `fetchai` organisation.

- [x] backward-cpp
- [x] benchmark
- [ ] utfcpp

PR #1658, opened 2019-09-13
Point forked vendor submodules at original repos

wkaluza:
Related to #1657. @pbukva do you think you could make a PR with your compilation fixes to `utfcpp` and throw it @nemtrif 's way? Then we could close that issue.


ejfitzgerald:
What testing has you done on this @wkaluza I am not sure just the compile test will work here I am sure @n-hutton and I had to make source level modifications which is why we forked them.in the first place

wkaluza:
> What testing

@ejfitzgerald I had a look at your respective branches - the long and short of it is that @n-hutton's changes are unused (he introduced a codepath that's never taken on `master`: n-hutton/backward-cpp@4dab0c1ef0b066846e7800444e9ab57aef1f0a99), and yours boiled down to changing some gbench CMake options (disable tests and installation) and removing some CMake logging. I implemented the option adjustments here in 4cf860a4438591f492bcd08baa117cf33c5f04a1.

n-hutton:
There was an used code path - it was accidentally deleted and has since been re-instated which is why this recent build against master fails. So, I would advise to revert the backward change

PR #1659, opened 2019-09-13
Feature/consensus integration

n-hutton:


libs/dkg/CMakeLists.txt
```diff
@@ -20,6 +20,3 @@ target_link_libraries(fetch-dkg
                              fetch-crypto
                              fetch-network
                              fetch-ledger)
```
n-hutton, line 22: need to move dkg_messages before deleting the dkg lib

libs/muddle/include/muddle/peer_selection_mode.hpp
```diff
@@ -0,0 +1,30 @@
...
+//
+//------------------------------------------------------------------------------
+
+namespace fetch {
+namespace muddle {
+
+enum class PeerSelectionMode
+{
+  DEFAULT,  ///< User request based routing
+  KADEMLIA  ///< Kademlia announcing and routing
```
n-hutton, line 26: just curious, what happens when half of the network are using Kademlia and the other half aren't?

ejfitzgerald: So in general we don’t want that to be the case. However, if that does end up being the case, then the system should still function at the same level performance as the previous system.

I.e. performance will largely will depend on the connections that the user requests

PR #1665, opened 2019-09-15
Extend CI static analysis to header files

wkaluza:


wkaluza:
@troelsfr Is this what you had in mind?

Rather than alter our existing static analysis scripts, I built on infrastructure provided by clang's tools, so the code came out leaner on our end. It also does not ignore headers, which was the main shortcoming of our hand-rolled solution. There is also some support for fixing check violations automatically, though it's not exposed in the CLI, as it's not 100% reliable. If you want to give autofix a go, come have a word with me; you also will need to install `clang-tools-6.0` for it to work.

I needed to explicitly disable the capitalisation checks (`readability-identifier-naming` in `.clang-tidy`), as they currently result in 2500 violations across over 500 files. As you would expect, a lot of that is in headers, which have escaped scrutiny for a long time. Autofix results in code that does not compile (like I said - unreliable). Whether we decide to phase these checks in gradually or all at once, I don't think this PR is the place for it, as the changeset is already quite large. In any case, you said that you had made substantial progress on this front on another branch, so this may be a good place to combine our efforts.

FYI @ejfitzgerald

libs/ml/include/ml/distributed_learning/distributed_learning_client.hpp
```diff
@@ -349,9 +349,9 @@ void TrainingClient<TensorType>::AddGradient(GradientType &gradient)
 template <class TensorType>
 void TrainingClient<TensorType>::AddExportGradient(GradientType &gradient)
 {
-  std::unique_lock<std::mutex> l(export_buffer_cv_mutex_);
+  std::unique_lock<std::mutex> lock(export_buffer_cv_mutex_);
```
ejfitzgerald, line 352: ??

libs/vm/include/vm/vm.hpp
```diff
@@ -157,7 +157,7 @@ class ParameterPack
 
   bool AddSingle(Variant parameter)
   {
-    // TODO: Probably should make a deep copy
+    // TODO(?): Probably should make a deep copy
```
ejfitzgerald: Issue?

scripts/ci-tool.py
```diff
@@ -114,7 +116,7 @@ def create_junit_format(output_path, data):
                     testcase,
                     'failure',
                     type=status,
-                    message='Test {}'.format(status))
+                    message=f'Test {status}')
```
ejfitzgerald: Not sure everyone is running 3.7 yet. For compat. Let’s use previous syntax

scripts/ci-tool.py
```diff
@@ -130,8 +132,8 @@ def create_junit_format(output_path, data):
 
 def build_type(text):
     if text not in BUILD_TYPES:
-        raise RuntimeError('Invalid build type {}. Choices: {}'.format(
-            text, ','.join(BUILD_TYPES)))
+        raise RuntimeError(
+            f'Invalid build type {text}. Choices: {", ".join(BUILD_TYPES)}')
```
ejfitzgerald: Again here

scripts/ci-tool.py
```diff
@@ -310,20 +313,26 @@ def main():
...
     options = {
-        'CMAKE_BUILD_TYPE': args.build_type
+        'CMAKE_BUILD_TYPE': args.build_type,
     }
 
     if args.metrics:
         options['FETCH_ENABLE_METRICS'] = 1
 
+    if args.lint:
+        options['CMAKE_EXPORT_COMPILE_COMMANDS'] = 1
```
ejfitzgerald: This option should always be enabled in the crake configuration so we shouldn’t need to set it here

ejfitzgerald (changes requested):
I really like this PR. Here my main comments:

- [x] 👍 - I like the static analysis module, however, I think we should give it a more general name. I think we should be aiming to put most of the shared ci / development logic into it such as coverage tools code etc
- [x] I want to keep general compat. with default python install out in the wild therefore our supported python version is basically 3.5+. Given that I don’t think we can update to using the f string syntax everywhere yet
- [x] I do not want to delay the integration of consensus with master so we should land this after that has handed (hopefully at the end of the day)

wkaluza:
>     (hopefully at the end of the day)

I believe you.

scripts/ci-tool.py
wkaluza: Done.

libs/ml/include/ml/distributed_learning/distributed_learning_client.hpp
wkaluza: Purely for readability. Lowercase L == ++ungood

scripts/ci-tool.py
wkaluza: But f-strings are so purdy... Done.

scripts/ci-tool.py
wkaluza: Done.

libs/vm/include/vm/vm.hpp
wkaluza: Done, #1669.

scripts/ci-tool.py
troels-im: Is it not the assumption the ci-tools are ran in docker by default?

libs/ml/include/ml/distributed_learning/distributed_learning_client.hpp
ejfitzgerald: Nope sorry I should have made a better comment - I am curious about this piece of code because there is a mutex for the condition along with another mutex. Maybe a question for @baykaner which will be ultimately update in the future

scripts/ci-tool.py
ejfitzgerald: You are right that we should make sure the docker image is using the lowest version available. We can add that too the list

ejfitzgerald (dismissed):

ejfitzgerald (dismissed):

ejfitzgerald (approved):

PR #1666, opened 2019-09-16
More beacon_setup_service tests and fixes

jinmannwong:
More tests for messages containing incorrect or non-valid inputs and code fixes for these. Removed redundant signature field in DKG messages and replaced Identity with MuddleAddress in some parts of the code.


libs/beacon/include/beacon/beacon_setup_service.hpp
```diff
@@ -227,6 +200,30 @@ class BeaconSetupService
   uint64_t         seconds_for_state_     = 0;
   uint64_t         expected_dkg_timespan_ = 0;
   bool             condition_to_proceed_  = false;
+
+  /// @name Handlers for messages
+  /// @{
+  void OnDkgMessage(MuddleAddress const &from, std::shared_ptr<DKGMessage> msg_ptr);
```
n-hutton, line 206: 👍

libs/beacon/src/beacon_manager.cpp
```diff
@@ -119,20 +119,43 @@ std::pair<BeaconManager::Share, BeaconManager::Share> BeaconManager::GetReceived
...
                                     std::vector<Coefficient> const &coefficients)
 {
-  for (uint32_t ii = 0; ii <= polynomial_degree_; ++ii)
+  if (coefficients.size() == polynomial_degree_ + 1)
   {
-    C_ik[identity_to_index_[from]][ii].setStr(coefficients[ii]);
+    for (uint32_t ii = 0; ii <= polynomial_degree_; ++ii)
+    {
+      bool c_set{false};
+      C_ik[identity_to_index_[from]][ii].setStr(&c_set, coefficients[ii].data());
```
n-hutton: Was the bool return value from this always an option in the library? Also, stylistically if there is no reason for ii I would just use i to maintain consistency

n-hutton: The setStr here, why/how could it fail? I feel like it would be great if this fn could be reduced to:

```
if (coefficients.size() == polynomial_degree_ + 1)
{
  C_ik[identity[from]] = coefficients;
}
```

Or something like that

libs/beacon/src/beacon_manager.cpp
```diff
@@ -297,37 +344,48 @@ BeaconManager::MuddleAddress BeaconManager::VerifyQualComplaint(MuddleAddress co
...
                      from_index, " for node ", victim_index);
-      return answer.first;
+      return from;
     }
     else
     {
-      FETCH_LOG_WARN(LOGGING_NAME, "Node ", cabinet_index_, " received incorrect complaint from ",
-                     from_index);
-      return from;
+      // check equation (5)
```
n-hutton: What do these numbers refer to?

libs/beacon/src/beacon_service.cpp
```diff
@@ -230,15 +230,15 @@ void BeaconService::StartNewCabinet(CabinetMemberList members, uint32_t threshol
...
+  if (members.find(identity_.identifier()) == members.end())
   {
     beacon->observe_only = true;
     FETCH_LOG_INFO(LOGGING_NAME, "Beacon in observe only mode. Members: ", members.size());
   }
   else
   {
     beacon->manager.SetCertificate(certificate_);
-    beacon->manager.Reset(members, threshold);
+    beacon->manager.NewCabinet(members, threshold);
```
n-hutton, line 241: For consistent naming here I assume

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -1088,7 +1086,7 @@ void BeaconSetupService::OnNewShares(MuddleAddress
   bool in_cabinet{false};
   for (auto &m : beacon_->aeon.members)
   {
-    if (m.identifier() == from)
+    if (m == from)
```
n-hutton: best to rename m here to member or the like - single letter variable not our style

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -1321,7 +1325,7 @@ bool BeaconSetupService::BasicMsgCheck(MuddleAddress const &              from,
...
```
n-hutton: and here

n-hutton (changes requested):

libs/beacon/src/beacon_manager.cpp
jinmannwong: Yes, the bool return value was always an option. The setStr can fail if someone sends you a string which does convert to number of the correct type i.e. in a  malformed message. There is no reason for ii, I can change it to i.

libs/beacon/src/beacon_manager.cpp
jinmannwong: They refer to equations in the DKG paper. I didn't realise I put that in, I will remove it.

libs/beacon/src/beacon_service.cpp
jinmannwong: Not for consistent naming. It was because I introduced a new function Reset() for each time the DKG is re-run with the same cabinet. I changed the existing Reset to NewCabinet to clarify the different between resetting the beacon and instating a new cabinet.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: Ok, will do.

n-hutton (approved):

PR #1678, opened 2019-09-17
Implement exponential backoff for peer discovery

ejfitzgerald:


libs/muddle/internal/peer_selector.hpp
```diff
@@ -73,7 +73,7 @@ class PeerSelector : public core::PeriodicRunnable
   {
     PeerData    peer_data{};
     std::size_t peer_index{0};  // The current peer being evaluated
-    Timepoint   last_updated{Clock::now()};
+    std::size_t consecutive_failures{0};
```
n-hutton, line 76: don't we avoid size_t? not that it matters in this case

n-hutton (dismissed):

libs/muddle/internal/peer_selector.hpp
ejfitzgerald: In general no, It gets painful when interacting with standard containers. The only place where we really don't want to use it is in the VM

n-hutton: and serialization stuff right

n-hutton (dismissed):

libs/muddle/src/peer_selector.cpp
```diff
@@ -32,18 +32,21 @@
...
+constexpr auto        MAX_ANNONCEMENT_INTERVAL = 30min;
+constexpr std::size_t MINIMUM_PEERS            = 3;
+constexpr char const *BASE_NAME                = "PeerSelector";
+constexpr std::size_t MAX_CACHE_KAD_NODES      = 20;
+constexpr std::size_t MAX_CONNECTED_KAD_NODES  = 3;
+constexpr std::size_t MAX_LOG2_BACKOFF         = 11;  // 2048
 
-static std::unordered_set<Address> operator+(std::unordered_set<Address>        input,
-                                             std::unordered_set<Address> const &other)
+std::unordered_set<Address> operator+(std::unordered_set<Address>        input,
```
bipll: Shouldn't this be turned into a library function? Looks too fundamental/simple/generic for an anonymous namespace.

libs/muddle/src/promise_runnable.cpp
```diff
@@ -20,16 +20,78 @@
...
+using PromiseState = service::PromiseState;
+
+/**
+ * Compute the desired timeout or deadline for the monitored promise
+ *
+ * @param promise The promise to be monitored
+ * @param timeout The (optional) timeout to be imposed
+ * @return The desired deadline time
+ */
+Timepoint CalculateDeadline(Promise const &promise, Duration const *timeout = nullptr)
```
bipll: Why not simply split it in two overloads rather than pass optional arg as pointer?

libs/muddle/src/promise_runnable.cpp
```diff
@@ -20,16 +20,78 @@
...
+
+          FETCH_LOG_WARN(LOGGING_NAME, "Explicitly marking the transaction as timed out");
+          ready = true;
+        }
+      }
+    }
+  }
+
+  return ready;
   return (!complete_) && promise_ && (promise_->state() != service::PromiseState::WAITING);
```
bipll: Unreachable return.

bipll (changes requested):

libs/muddle/internal/peer_selector.hpp
ejfitzgerald: Yep that is correct

libs/muddle/src/promise_runnable.cpp
ejfitzgerald: Good spot 👍

libs/muddle/src/promise_runnable.cpp
ejfitzgerald: Hmmm, honestly I was thinking that is easier to read the intention this way. I have changed it let me know what you think 😁

libs/muddle/src/peer_selector.cpp
ejfitzgerald: Done

ejfitzgerald:
@bipll - I believe that I have addresses all your comments can you double check and approve if / when you are happy

libs/muddle/src/promise_runnable.cpp
bipll: It's better this way better IMO, two one-liners against one function with default arg values and branches. It would even be shorter if not for duplicated doc comment. >_<

libs/muddle/src/peer_selector.cpp
```diff
@@ -233,10 +234,12 @@ void PeerSelector::ResolveAddresses(Addresses const &addresses)
     // make the call to the remote service
     auto promise = rpc_client_.CallSpecificAddress(address, RPC_MUDDLE_DISCOVERY,
                                                    DiscoveryService::CONNECTION_INFORMATION);
+    // lookup the peer information
+    auto const &peer_data = peers_info_[address];
 
     // wrap the promise is a task
     auto task = std::make_shared<PromiseTask>(
-        std::move(promise),
+        std::move(promise), CalculatePromiseTimeout(peer_data.consecutive_failures),
```
bipll: Nitpicking: doesn't this overload accept an lvalue promise?

libs/muddle/src/promise_runnable.cpp
```diff
@@ -20,17 +20,82 @@
...
+          promise_->Timeout();
+
+          FETCH_LOG_WARN(LOGGING_NAME, "Explicitly marking the transaction as timed out");
+          ready = true;
+        }
+      }
+    }
+  }
+
+  return ready;
```
bipll, line 96: Minor suggestion: how about this, slightly turned inside-out and thus a bit shorter and less-nested form?
```
{
  if (complete_ || !promise_)
  {
    return false;
  }

  if (promise_->state() != PromiseState::WAITING)
  {
    return true;
  }

  if (Clock::now() >= deadline_)
  {
    // signal that the promise has timed out
    promise_->Timeout();

    FETCH_LOG_WARN(LOGGING_NAME, "Explicitly marking the transaction as timed out");
    return true;
  }

  return false;
}
```

bipll (commented):
Two more comments but otherwise LGTM.

libs/muddle/src/peer_selector.cpp
ejfitzgerald: No that is a good spot

libs/muddle/src/promise_runnable.cpp
ejfitzgerald: I take your point that what was there beforehand was more complicated than needed. In general, I much prefer code with single return points. It makes code much easier to debug in the future. Please have a look at the my next changes I think they are the best of the two worlds

bipll (approved):
Ok, LGTM.
(I still have a point about return points, to be discussed later.)

n-hutton (approved):

PR #1684, opened 2019-09-19
enforcing test time evaluation in etch

baykaner:
- this results in models trained with dropout making predictions without dropout


evsmithx (approved):
LGTM

PR #1685, opened 2019-09-19
Word2vec clients shouldn't need to have the same vocab

evsmithx:
Modifying the word2vec client so that clients don't need to have the same vocabulary. Changes required:

1. Gradient updates now need to be translated from the sender's vocab to the receiver's.
2. Clients need a method for sending their vocab to other clients on request.
3. When weights are synchronised in synchronised mode the weights also need to be translated by each client.


libs/ml/examples/word2vec_distributed_learning/translator.hpp
```diff
@@ -0,0 +1,130 @@
...
+#include "ml/dataloaders/word2vec_loaders/vocab.hpp"
+
+#include <utility>
+#include <vector>
+
+namespace fetch {
+namespace ml {
+namespace distributed_learning {
+
+struct Translator
```
baykaner, line 35: a comment is needed here to explain the job of the translator

libs/ml/examples/word2vec_distributed_learning/translator.hpp
```diff
@@ -0,0 +1,130 @@
...
+    if (vocab_hash == MyVocabHash())  // no need to translate
+    {
+      mask.Fill(DataType{1});
+      return std::make_pair(gradient_update, mask);
+    }
+
+    assert(VocabKnown(vocab_hash));
+
+    std::vector<std::string> other_vocab = known_vocabs[vocab_hash];
+
```
baykaner, line 56: also assert that the other vocab size is the same size as this vocab size?

libs/ml/examples/word2vec_distributed_learning/translator.hpp
```diff
@@ -0,0 +1,130 @@
...
+    // figure out which way around the matrix is
+    bool       first_dim_embedding;
+    SizeType   embedding_size;
+    TensorType transl_gu;
```
baykaner: its not obvious to me what a transl_gu is.
might as well rename the variable to something written in full here

libs/ml/examples/word2vec_distributed_learning/translator.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+    if (gradient_update.shape()[0] == other_vocab.size())
```
baykaner, line 62: why are we permitting transposed embeddings?

libs/ml/examples/word2vec_distributed_learning/translator.hpp
```diff
@@ -0,0 +1,130 @@
...
+    {
+      first_dim_embedding = true;
+      embedding_size      = gradient_update.shape()[0];
+      transl_gu           = TensorType({embedding_size, MyVocabSize()});
+    }
+
+    for (SizeType i = 0; i < other_vocab.size(); i++)
+    {
+      std::string word     = other_vocab[i];
+      std::size_t transl_i = my_vocab->IndexFromWord(word);
```
baykaner: idx is a useful abbreviation of index. i is a less clear abbreviation

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -664,7 +667,7 @@ math::SizeType GraphW2VLoader<T>::vocab_size() const
  * @return
  */
 template <typename T>
-typename GraphW2VLoader<T>::VocabType const &GraphW2VLoader<T>::vocab() const
+std::shared_ptr<typename GraphW2VLoader<T>::VocabType> const &GraphW2VLoader<T>::GetVocab() const
```
baykaner, line 669: might as well add a using VocabPtrType = std:shared_ptr<...>

libs/ml/examples/word2vec_distributed_learning/word2vec_client.hpp
```diff
@@ -244,6 +259,64 @@ void Word2VecClient<TensorType>::TestEmbeddings(std::string const &word0, std::s
...
+/**
+ * @return pair of vocab strings and vocab hash
+ */
+template <class TensorType>
+std::pair<std::vector<std::string>, byte_array::ConstByteArray>
+Word2VecClient<TensorType>::GetVocab()
+{
+  auto vocab      = w2v_data_loader_ptr_->GetVocab();
+  auto vocab_hash = w2v_data_loader_ptr_->GetVocabHash();
+  return std::make_pair(vocab->GetReverseVocab(), vocab->GetVocabHash());
```
baykaner, line 214: according to the function names at least this GetVocab gets a ReverseVocab. That seems strange

libs/ml/include/ml/dataloaders/word2vec_loaders/w2v_dataloader.hpp
```diff
@@ -70,7 +70,7 @@ class W2VLoader : public DataLoader<fetch::math::Tensor<T>, fetch::math::Tensor<
 
   /// accessors and helper functions ///
   SizeType         Size() const override;
-  SizeType         vocab_size() const;
+  SizeType         vocab_size();
```
baykaner: why is this not const anymore?

libs/ml/examples/word2vec_simple/main.cpp
```diff
@@ -42,7 +42,7 @@ std::string ReadFile(std::string const &path)
   return std::string((std::istreambuf_iterator<char>(t)), std::istreambuf_iterator<char>());
 }
 
-void SaveEmbeddings(W2VLoader<FloatType> const &data_loader, std::string const &output_filename,
+void SaveEmbeddings(W2VLoader<FloatType> &data_loader, std::string const &output_filename,
```
baykaner: why cant the data loader be const anymore?

libs/ml/include/ml/distributed_learning/distributed_learning_client.hpp
```diff
@@ -46,22 +46,43 @@ struct ClientParams
...
+  SizeType batch_size{};
   DataType learning_rate;
 
   std::vector<std::string> inputs_names = {"Input"};
   std::string              label_name   = "Label";
   std::string              error_name   = "Error";
 };
 
+template <typename TensorType>
+struct GradientUpdate
```
baykaner: as per prior conversation, this Update struct should be a generic object containing:
- data (where you store your gadients in this case)
- data type flag
- timestamp
- client id
- hash

therefore can you rename it to "Update"
and rename grads to data
and add a comment explaining above explaining what the update struct is for.

baykaner (changes requested):

libs/ml/examples/word2vec_distributed_learning/translator.hpp
evsmithx: The two vocabularies might not be the same size, so I'm not sure what you mean? The embedding size of the two models has to match, but the translator doesn't know anything about embedding size.

libs/ml/examples/word2vec_distributed_learning/translator.hpp
evsmithx: I initially thought that the input and context embeddings would have opposite shapes, but actually the skipgram layer transposes context embeddings so they're both {embedding_size, vocab_size}. I could get rid of this now?

libs/ml/examples/word2vec_distributed_learning/word2vec_client.hpp
evsmithx: ReverseVocab is both the way to get a string from a word_id (i.e. the reverse of vocab which gets an id from a string), and also the most compact way of sending the information required for translation, so I'm not sure what to call it!

baykaner: it's just rather bizarre to see a function of the form
void GetX()
{
  return GetReverseX();
}
maybe add a comment for that last line to explain why its a good idea?

baykaner (changes requested):

baykaner (dismissed):

baykaner (dismissed):

baykaner (approved):

PR #1688, opened 2019-09-20
MNIST loader fix

MissingNO57:


evsmithx (approved):
LGTM. We really should have tests to catch this sort of thing!

PR #1689, opened 2019-09-22
Additional clang-tidy checks

### LINKED ISSUE #957: Storage Key add explicit constructor

wkaluza:
Closes #957

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,42 +1,32 @@
 Checks: "-*,\
 bugprone-*,\
+cert-dcl21-cpp,\
+cert-dcl50-cpp,\
+cert-env33-c,\
+-cert-err34-c,\
+cert-err52-cpp,\
+-cert-err58-cpp,\
+cert-err60-cpp,\
+-cert-flp30-c,\
 -clang-analyzer-security.FloatLoopCounter,\
 google-build-namespaces,\
 -google-explicit-constructor,\
 google-readability-todo,\
 google-runtime-operator,\
+misc-new-delete-overloads,\
+misc-non-copyable-objects,\
+misc-static-assert,\
 misc-suspicious-*,\
+misc-throw-by-value-catch-by-reference,\
 misc-unused-*,\
 modernize-loop-convert,\
 modernize-make-shared,\
 modernize-make-unique,\
 modernize-pass-by-value,\
 modernize-use-*,\
--performance-*,\
+performance-*,\
+readability-avoid-const-params-in-decls,\
 readability-braces-around-statements,\
 readability-container-size-empty,\
--readability-identifier-naming,\
 readability-simplify-boolean-expr,\
 readability-static-definition-in-anonymous-namespace"
-
-CheckOptions:
-  - key:             readability-identifier-naming.FunctionCase
-    value:           aNy_CasE
-  - key:             readability-identifier-naming.MemberCase
-    value:           lower_case
-  - key:             readability-identifier-naming.PublicMemberSuffix
-    value:           ''
-  - key:             readability-identifier-naming.ProtectedMemberSuffix
-    value:           '_'
-  - key:             readability-identifier-naming.PrivateMemberSuffix
-    value:           '_'
-  - key:             readability-identifier-naming.ClassCase
-    value:           CamelCase
-  - key:             readability-identifier-naming.EnumCase
-    value:           CamelCase
-  - key:             readability-identifier-naming.ParameterCase
-    value:           lower_case
-  - key:             readability-identifier-naming.UnionCase
-    value:           CamelCase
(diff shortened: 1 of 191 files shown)


libs/storage/include/storage/key.hpp
```diff
@@ -70,9 +70,7 @@ struct Key
 
   Key() = default;
 
-  // TODO(private issue 957): There are a number of implicit conversions for this key, in many
```
ejfitzgerald, line 73: Let’s make sure this issue is marked as resolved

.clang-tidy
```diff
@@ -1,42 +1,32 @@
...
+misc-throw-by-value-catch-by-reference,\
 misc-unused-*,\
 modernize-loop-convert,\
 modernize-make-shared,\
 modernize-make-unique,\
 modernize-pass-by-value,\
 modernize-use-*,\
--performance-*,\
+performance-*,\
+readability-avoid-const-params-in-decls,\
```
ejfitzgerald, line 28: 👍

.clang-tidy
```diff
@@ -1,42 +1,32 @@
...
 readability-braces-around-statements,\
 readability-container-size-empty,\
--readability-identifier-naming,\
```
ejfitzgerald, line 18: We should work out what to do here, at one point this check was enabled. Although it might be “dead” code at the moment I think it is useful to show intent

.clang-tidy
```diff
@@ -1,42 +1,32 @@
...
+cert-env33-c,\
+-cert-err34-c,\
+cert-err52-cpp,\
+-cert-err58-cpp,\
+cert-err60-cpp,\
+-cert-flp30-c,\
 -clang-analyzer-security.FloatLoopCounter,\
 google-build-namespaces,\
--google-explicit-constructor,\
+google-explicit-constructor,\
```
ejfitzgerald, line 13: I have mixed feelings about this. From a purity point of view, I agree and ideally it would have been enabled from day 1. However, this PR introduces a lot of `// NOLINT` Comments to combat the difficult cases. I think this is probably worse from a readability point of view

.clang-tidy
```diff
@@ -1,42 +1,32 @@
...
 google-runtime-operator,\
+misc-new-delete-overloads,\
+misc-non-copyable-objects,\
+misc-static-assert,\
 misc-suspicious-*,\
+misc-throw-by-value-catch-by-reference,\
 misc-unused-*,\
 modernize-loop-convert,\
 modernize-make-shared,\
 modernize-make-unique,\
```
ejfitzgerald, line 24: I think we need to enable more of these checks

ejfitzgerald (changes requested):

libs/storage/include/storage/key.hpp
wkaluza: Done

.clang-tidy
wkaluza: My recommendation is that we enforce naming conventions using an in-house script. This goes against my every instinct, but this check is simply too poorly designed. I discussed this with @troelsfr. In brief: clang-tidy supports identifier affixes based on member access modifiers. We would like to use this and mark private and protected members with underscore. Here's the problem: clang-tidy only recognises non-static and non-const members for the purposes of this check. You can specify affixes for statics and const members, but they would be applied irrespective of access being public, private or protected.

```
class Foo
{
  int       private_member_with_suffix_;
  int const private_member_without_suffix; // inconsistent
};
```

I'll take care of this in the near future; Troels wrote a script that may serve as a starting point, though it's not quite ready to be tied into CI. For now I focused on strengthening checks that directly affect API behaviour, such as explicit constructors, performance checks, etc.

.clang-tidy
wkaluza: Some types are only difficult cases because there is widespread reliance on implicit construction for these types in the code. Enabling this check and shooting down further NOLINTs in code review will at least let us prevent the problem from getting worse.

As an aside, not all the NOLINTs are in place for constructors which would be too tedious to make explicit: I deliberately allowed implicit conversions between `vm::Ptr<Base>` and `vm::Ptr<Derived>`, for example, as this is reasonable and mirrors stdlib smart pointer behaviour. Construction of `vm::Ptr<T>` from `T*`, on the other hand is made explicit.

.clang-tidy
wkaluza: I left in a few disabled ones which are my recommendations to enable next. We have a lot of `cert-err58-cpp` violations, for example, where we construct statics and globals that may throw on construction.

ejfitzgerald: `cert-err58-cpp` will be harder to remove but I would recommend enabling all the `modernize-*` checks

ejfitzgerald (approved):

PR #1690, opened 2019-09-22
noticed thread safety in files

n-hutton:


libs/core/src/logging.cpp
```diff
@@ -171,9 +171,12 @@ LogRegistry::LogRegistry() = default;
 
 void LogRegistry::Log(LogLevel level, char const *name, std::string &&message)
 {
-  if (level < global_level_)
   {
-    return;
+    FETCH_LOCK(lock_);
+    if (level < global_level_)
```
ejfitzgerald: I think that global log level should be set using an atomic instead

libs/muddle/src/router.cpp
```diff
@@ -654,8 +654,8 @@ void Router::SetKademliaRouting(bool enable)
  */
 Router::Handle Router::LookupRandomHandle(Packet::RawAddress const & /*address*/) const
 {
-  static std::random_device rd;
-  static std::mt19937       rng(rd());
+  thread_local std::random_device rd;
+  thread_local std::mt19937       rng(rd());
```
ejfitzgerald, line 658: 👍

ejfitzgerald (changes requested):

libs/muddle/src/rbc.cpp
```diff
@@ -534,6 +534,11 @@ void RBC::Deliver(SerialisedMessage const &msg, uint32_t sender_index)
   deliver_msg_callback_(miner_id, msg);
   lock_.lock();
```
bipll, line 535: Here, unlocking/locking is done manually. Wouldn't it be safer to use RAII, like with `lock_guard` (but inverse)?

n-hutton: yes, that would be better - I will make the change in another PR that I am working on currently

bipll: May I suggest UnlockGuard for the name? :)

wkaluza (approved):

libs/muddle/src/rbc.cpp
n-hutton: sure :)

PR #1691, opened 2019-09-22
Sliced heaviest chain

bipll:
Heaviest chain is now requested in genesis-to-tip order and received in portions.

I also accidentally a larger share of boilerplate for `MapSerializer`; if it looks good, can be made codebase-wide in a separate PR (currently in main_chain_rpc_protocol.hpp).


libs/crypto/include/crypto/hash.hpp
```diff
@@ -35,7 +35,6 @@ byte_array::ByteArray Hash(byte_array::ConstByteArray const &input)
                 "Use a type derived from fetch::crypto::Hasher:Interface");
   T hasher;
 
-  hasher.Reset();
```
ejfitzgerald, line 38: This is not needed any more - I remember that at the OpenSSL level it was required to call reset before starting the hash

libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -254,32 +271,6 @@ class MainChain
   telemetry::CounterPtr             bloom_filter_query_count_;
   telemetry::CounterPtr             bloom_filter_positive_count_;
   telemetry::CounterPtr             bloom_filter_false_positive_count_;
-
```
ejfitzgerald: 👍

libs/ledger/include/ledger/protocols/main_chain_rpc_protocol.hpp
```diff
@@ -84,4 +109,56 @@ class MainChainProtocol : public service::Protocol
...
+template <class T, class D, class... Fields>
+struct MapSerializerTemplate
+{
+  using Type       = T;
+  using DriverType = D;
+
+  template <class Constructor>
+  static void Serialize(Constructor &map_constructor, Type const &t)
+  {
+    value_util::Accumulate(
```
ejfitzgerald: Please can we follow the convention with this map serialiser. The key thing is that although the type indexes are incrementing currently. In the future if we want to remove any fields and we want to implement some form of backward compat. we will need the complete list of identifiers.

Also, on a personal level, this is harder to read what is going on

libs/ledger/include/ledger/protocols/main_chain_rpc_service.hpp
```diff
@@ -161,6 +162,7 @@ class MainChainRpcService : public muddle::rpc::Server,
   Address         current_peer_address_;
   BlockHash       current_missing_block_;
   Promise         current_request_;
+  BlockHash       past_recent_tip_ = GENESIS_DIGEST;
```
ejfitzgerald: Given the way that this "constant" is generated I am not so sure this is a good idea. Since it depending on init. order this will change

ejfitzgerald:
Closing because stale

PR #1693, opened 2019-09-24
ApplyRegularisation reworked

MissingNO57:


libs/ml/include/ml/ops/trainable.hpp
```diff
@@ -53,7 +53,6 @@ class Trainable
   virtual TensorType        GetGradients() const                                      = 0;
   virtual void              ResetGradients()                                          = 0;
   virtual void              ApplyGradient(TensorType const &grad)                     = 0;
-  virtual void              ApplyRegularisation()                                     = 0;
```
baykaner: why remove this? We currently only have one type of trainable (a variable). But it might as well be a requirement of a trainable that one can regularise it - right?

baykaner (changes requested):
as requested, make the applyRegularisation method protected and adjust the tests as necessary

baykaner (approved):

PR #1694, opened 2019-09-24
Feature/benchmark rbc

n-hutton:
Add alternate communication channel for DKG: PBC aka punishment broadcast channel.

Add benchmark for this, and the RBC. Fake networking included to make the benchmark times bearable and avoid problems with the number of open file handles during the benchmark.


libs/beacon/include/beacon/beacon_setup_service.hpp
```diff
@@ -85,6 +87,8 @@ class BeaconSetupService
   using MuddleEndpoint          = muddle::MuddleEndpoint;
   using MuddleInterface         = muddle::MuddleInterface;
   using RBC                     = muddle::RBC;
+  using ReliableChannel         = muddle::BroadcastChannelInterface;
```
n-hutton, line 91: Broadcast interface represents an interface for a rbc type thing

libs/beacon/src/beacon_manager.cpp
```diff
@@ -376,7 +376,7 @@ BeaconManager::MuddleAddress BeaconManager::VerifyQualComplaint(MuddleAddress co
 void BeaconManager::ComputePublicKeys()
 {
 
-  FETCH_LOG_INFO(LOGGING_NAME, "Node ", cabinet_index_, " compute public keys");
+  FETCH_LOG_INFO(LOGGING_NAME, "Node ", cabinet_index_, " compute public keys begin.");
```
n-hutton, line 381: turns out this can take quite a while

libs/beacon/src/beacon_service.cpp
```diff
@@ -487,37 +500,43 @@ BeaconService::State BeaconService::OnVerifySignaturesState()
...
-  std::lock_guard<std::mutex> lock(mutex_);
-
-  if (ret.threshold_signatures.empty())
   {
-    FETCH_LOG_INFO(LOGGING_NAME, "Peer wasn't ready when asking for signatures: ",
-                   qual_promise_identity_.identifier().ToBase64());
-    state_machine_->Delay(std::chrono::milliseconds(10));
-    return State::COLLECT_SIGNATURES;
-  }
+    std::lock_guard<std::mutex> lock(mutex_);
```
n-hutton, line 504: Only change here is to not lock when doing manager verify since the lock stops peers from pulling the sigs

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -85,48 +85,49 @@ uint64_t GetTime()
   return static_cast<uint64_t>(std::time(nullptr));
 }
 
+//#define CHANNEL_TYPE RBC
```
n-hutton, line 91: switch here between versions of RBC

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -85,48 +85,49 @@ uint64_t GetTime()
...
-                 CHANNEL_CONNECTIONS_SETUP, false}
-  , rbc_{endpoint_, identity_.identifier(),
-         [this](MuddleAddress const &from, ConstByteArray const &payload) -> void {
-           DKGEnvelope   env;
-           DKGSerializer serializer{payload};
-           serializer >> env;
-           OnDkgMessage(from, env.Message());
-         },
-         CHANNEL_RBC_BROADCAST, false}
+  , rbc_{new CHANNEL_TYPE(endpoint_, identity_.identifier(),
```
n-hutton, line 103: only use one RBC-esque channel

libs/beacon/src/beacon_service.cpp
```diff
@@ -487,37 +500,43 @@ BeaconService::State BeaconService::OnVerifySignaturesState()
...
+      AddSignature(address_sig_pair.second);
+    }
 
-  FETCH_LOG_INFO(LOGGING_NAME, "After adding, we have ", all_sigs_map.size(),
-                 " signatures. Round: ", current_entropy_.round);
+    FETCH_LOG_INFO(LOGGING_NAME, "After adding, we have ", all_sigs_map.size(),
+                   " signatures. Round: ", current_entropy_.round);
+  }  // Mutex unlocks here since verification can take some time
+
+  MilliTimer const timer{"Verify threshold signature", 100};
```
n-hutton, line 545: this prints if it takes longer than 100 ms from construction to destruction of the object

apps/beacon/main.cpp
```diff
@@ -157,8 +157,12 @@ int main()
   // Attaching the cabinet logic
   for (auto &member : committee)
   {
-    member->reactor.Attach(member->beacon_service.GetMainRunnable());
```
ejfitzgerald, line 160: I would suggest adding an overload for `Attach` to take a vector of the runnables

libs/beacon/include/beacon/block_entropy.hpp
```diff
@@ -0,0 +1,125 @@
...
+#include "ledger/chain/digest.hpp"
+#include "beacon/beacon_manager.hpp"
+
+#include "crypto/hash.hpp"
+#include "crypto/sha256.hpp"
+
+namespace fetch {
+namespace beacon {
+
+class BlockEntropyInterface
```
ejfitzgerald: Ideally these would be seperated

libs/beacon/src/beacon_manager.cpp
ejfitzgerald: I would suggest that we add a `MilliTimer` here

libs/beacon/src/beacon_service.cpp
ejfitzgerald: `FETCH_LOCK` for the win 🎆

libs/beacon/tests/dkg/dkg_resilience.cpp
```diff
@@ -607,7 +607,12 @@ void GenerateTest(uint32_t cabinet_size, uint32_t threshold, uint32_t qual_size,
   {
     for (auto &member : committee)
     {
-      member->reactor.Attach(member->GetWeakRunnable());
+      auto all_runnables = member->GetWeakRunnables();
```
ejfitzgerald: Apply new override here

libs/muddle/include/muddle/muddle_interface.hpp
```diff
@@ -275,5 +275,11 @@ MuddlePtr CreateMuddle(NetworkId const &network, network::NetworkManager const &
 MuddlePtr CreateMuddle(char const network[4], network::NetworkManager const &nm,
                        std::string const &external_address);
 
+MuddlePtr CreateMuddleFake(NetworkId const &network, ProverPtr certificate,
```
ejfitzgerald: I don't think that these should live in this header file

libs/muddle/include/muddle/question_struct.hpp
```diff
@@ -0,0 +1,175 @@
...
+    SEEN = 2,  // Signature indicating seen
+  };
+
+  QuestionStruct()                          = default;
+  QuestionStruct(QuestionStruct const &rhs) = default;
+  QuestionStruct(QuestionStruct &&rhs)      = default;
+  QuestionStruct &operator=(QuestionStruct const &rhs) = default;
+  QuestionStruct &operator=(QuestionStruct &&rhs) = default;
+
+  QuestionStruct(Digest const &question, Answer const &answer, CertificatePtr certificate,
```
ejfitzgerald, line 61: This should have a `.cpp` file ideally

libs/muddle/internal/muddle_fake.hpp
```diff
@@ -0,0 +1,546 @@
...
+  using Payload              = Packet::Payload;
+  using ConnectionPtr        = std::weak_ptr<network::AbstractConnection>;
+  using Handle               = network::AbstractConnection::connection_handle_type;
+  using ThreadPool           = network::ThreadPool;
+  using HandleDirectAddrMap  = std::unordered_map<Handle, Address>;
+  using Prover               = crypto::Prover;
+  using DirectMessageHandler = std::function<void(Handle, PacketPtr)>;
+
+  // Construction / Destruction
+  FakeMuddleEndpoint(NetworkId network_id, Address address, Prover *certificate = nullptr,
```
ejfitzgerald, line 222: Again this should be in `.cpp` files

libs/muddle/src/rbc.cpp
```diff
@@ -77,6 +77,9 @@ RBC::RBC(Endpoint &endpoint, MuddleAddress address, CallbackFunction call_back,
   });
 }
 
+RBC::~RBC()
```
ejfitzgerald: = default

ejfitzgerald (dismissed):

libs/beacon/src/beacon_manager.cpp
n-hutton: willdo

libs/beacon/src/beacon_service.cpp
n-hutton: can't tell if sarcastic but sure

apps/beacon/main.cpp
n-hutton: I did think this actually just never got around to it

libs/beacon/include/beacon/block_entropy.hpp
n-hutton: this file wasn't intended to be included

libs/muddle/include/muddle/muddle_interface.hpp
n-hutton: good point

libs/muddle/include/muddle/question_struct.hpp
n-hutton: while still in dev it's slightly faster to do in all hpp style. Will be as cpps for trying to get into master

libs/muddle/internal/muddle_fake.hpp
n-hutton: above

libs/muddle/src/rbc.cpp
n-hutton: ok

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -642,49 +645,53 @@ BeaconSetupService::State BeaconSetupService::OnWaitForReconstructionShares()
...
+      SetTimeToProceed(State::DRY_RUN_SIGNING);
+      return State::DRY_RUN_SIGNING;
     }
+  }
 
-    SetTimeToProceed(State::DRY_RUN_SIGNING);
-    return State::DRY_RUN_SIGNING;
+  if (timer_to_proceed_.HasExpired())
+  {
+    FETCH_LOG_INFO(LOGGING_NAME, "Timed out during reconstruction.");
```
jinmannwong: I don't think this if statement is correct. Currently, if a node is missing one reconstruction share it resets, when it could in fact proceed onto running reconstruction. As I mentioned the if statement should be if the timer has expired or one has collected cabinet_size - 1 reconstruction shares then do the verification and run reconstruction etc. If compute public keys is taking a long time then it should be moved into its own state.

libs/muddle/src/rbc.cpp
```diff
@@ -703,7 +711,10 @@ bool RBC::CheckTag(RBCMessage const &msg)
  */
 bool RBC::SetPartyFlag(uint32_t sender_index, TagType tag, MessageType msg_type)
 {
-  assert(parties_.size() == current_cabinet_.size());
+  if (sender_index >= parties_.size())
+  {
+    return false;
+  }
```
jinmannwong: I don't think the assert should be replaced by something else here. If the assert fails this means there is a race condition between parties being reset and a message being processed and this should be caught and fixed. If we replace this with this if statement then we will not be alerted by this at all. I don't expect the assert to fail as as the sender index is computed within the lock in OnRBCMessage.

libs/muddle/src/rbc.cpp
```diff
@@ -534,6 +537,11 @@ void RBC::Deliver(SerialisedMessage const &msg, uint32_t sender_index)
   deliver_msg_callback_(miner_id, msg);
   lock_.lock();
 
+  if (sender_index >= parties_.size())
+  {
+    return;
+  }
+
```
jinmannwong, line 543: This if statement should also appear after the deliver_msg_callback_ below in line 563.

n-hutton: good catch, It's in now

libs/muddle/include/muddle/question_struct.hpp
```diff
@@ -0,0 +1,175 @@
...
+      }
+
+      // Add info from other table to ours
+      AnswerAndSeen &  rhs_answer_and_seen = rhs.table_[address];
+      Answer const &   rhs_answer          = GetAnswer(rhs_answer_and_seen);
+      Signature const &rhs_signature       = GetSignature(rhs_answer_and_seen);
+      Seen const &     rhs_seen            = GetSeen(rhs_answer_and_seen);
+
+      if (!rhs_answer.empty())
+      {
```
jinmannwong, line 136: As we discussed earlier, this needs to be changed so that node's original table is not replaced. True of the signature below as well.

libs/muddle/include/muddle/punishment_broadcast_channel.hpp
```diff
@@ -0,0 +1,411 @@
...
+  bool ResetCabinet(CabinetMembers const &cabinet) override
+  {
+    FETCH_LOCK(lock_);
+    current_cabinet_ = cabinet;
+
+    // Set threshold depending on size of committee
+    if (current_cabinet_.size() % 3 == 0)
+    {
+      threshold_ = static_cast<uint32_t>(current_cabinet_.size() / 3 - 1);
+    }
```
jinmannwong, line 136: Note: threshold here computes the maximum number of adversaries allowed. Which in the original RBC is t <1/3n. It is not the number of messages you need to wait for to avoid inconsistency in the network.

libs/muddle/include/muddle/punishment_broadcast_channel.hpp
```diff
@@ -0,0 +1,411 @@
...
+   * If not, continue to try to resolve answers
+   */
+  bool AnsweredQuestion()
+  {
+    uint32_t messages_confirmed = 0;
+
+    for (auto const &node_plus_answers : question_.table_)
+    {
+      if (QuestionStruct::GetSeen(node_plus_answers.second).size() >= threshold_ &&
+          node_plus_answers.first != question_.self_)
```
jinmannwong, line 189: Should be current_cabinet_.size() - threshold_ here instead of threshold_

libs/muddle/include/muddle/punishment_broadcast_channel.hpp
```diff
@@ -0,0 +1,411 @@
...
+            FETCH_LOCK(lock_);
+
+            // Guard against receiving a non-matching table
+            if (recvd_question.question_ != question_.question_)
+            {
+              FETCH_LOG_DEBUG(LOGGING_NAME, "Note: ignoring non matching question");
+            }
+            else
+            {
+              answers = question_.Update(threshold_, recvd_question);
```
jinmannwong, line 282: Same as above regarding threshold_

libs/beacon/src/beacon_setup_service.cpp
n-hutton: fixed

libs/muddle/include/muddle/question_struct.hpp
n-hutton: Yup done

libs/muddle/include/muddle/punishment_broadcast_channel.hpp
n-hutton: This needs to be reworked as we discussed to obey the guarantee honest nodes will arrive at the same answer - I will have to do this in a subsequent PR

libs/muddle/include/muddle/punishment_broadcast_channel.hpp
n-hutton: I will do this in the subsequent rework

libs/muddle/include/muddle/punishment_broadcast_channel.hpp
n-hutton: rework

PR #1697, opened 2019-09-26
Save copy elision.

bipll:
At times constructs like `return std::move(local_object);` are found in our libs (in vendors as well, but I guess this can't be helped ... quickly and easily at least). This prevents copy elision and, in particular, breaks build on GCC 9.


troels-im (approved):
LGTM

PR #1698, opened 2019-09-26
Implement trusted dealer for DKG

jinmannwong:
Trusted dealer introduced for DKG to allow for benchmarking of entropy generation without DKG. Add functionality to Beacon Service to set output from trusted dealer into the next Aeon.


libs/beacon/src/beacon_service.cpp
```diff
@@ -252,6 +252,43 @@ void BeaconService::StartNewCabinet(CabinetMemberList members, uint32_t threshol
   cabinet_creator_.QueueSetup(beacon);
 }
 
+void BeaconService::StartNewCabinet(CabinetMemberList members, uint32_t threshold,
```
ejfitzgerald: There is quite a lot of duplicated code between the two `StartNewCabinet` methods. Is it worth doing a little more refactoring so that this can be reused?

jinmannwong: Yes, this did occur to me as well. I will change this.

libs/beacon/include/beacon/trusted_dealer.hpp
```diff
@@ -0,0 +1,62 @@
...
+//------------------------------------------------------------------------------
+
+#include "crypto/mcl_dkg.hpp"
+
+namespace fetch {
+namespace beacon {
+
+class TrustedDealer
+{
+  using MuddleAddress = byte_array::ConstByteArray;
```
n-hutton: Style: public: first, then private: after usually.

Looking at this, couldn't it use unordered_map<MuddleAddress, DkgOutput> without the index, instead of DKGOutput do a return {};?

I'd have to check but this pattern of returning a const ref to something within this class can often be dangerous for memory safety - is it that the MCL stuff doesn't do copy constructors well (something I have painfully noticed)?

libs/beacon/tests/dkg/beacon_manager_tests.cpp
```diff
@@ -30,16 +30,16 @@ using Certificate    = Prover;
...
-{
-  PrivateKey              private_key;
-  PublicKey               group_public_key;
-  std::vector<PublicKey>  public_key_shares;
-  std::set<MuddleAddress> qual;
-};
-
 TEST(beacon_manager, dkg_and_threshold_signing)
 {
+  struct DkgOutput
```
n-hutton: Seems like it makes sense to have this struct a part of the codebase and Get/SetDkgOutput function signature is just DkgOutput GetDkgOutput(); ?

libs/beacon/tests/dkg/trusted_dealer_test.cpp
```diff
@@ -0,0 +1,228 @@
...
+using namespace std::chrono_literals;
+
+using std::this_thread::sleep_for;
+using std::chrono::milliseconds;
+
+using Prover        = crypto::Prover;
+using ProverPtr     = std::shared_ptr<Prover>;
+using MuddleAddress = byte_array::ConstByteArray;
+
+struct DummyManifesttCache : public ManifestCacheInterface
```
n-hutton: typo here? two T's

libs/beacon/tests/dkg/trusted_dealer_test.cpp
```diff
@@ -0,0 +1,228 @@
...
+  {
+    std::cout << "- Scheduling round " << i << std::endl;
+    TrustedDealer dealer(cabinet_addresses,
+                         static_cast<uint32_t>(static_cast<double>(cabinet.size()) * threshold));
+    for (auto &member : cabinet)
+    {
+      member->beacon_service.StartNewCabinet(
+          cabinet_addresses, static_cast<uint32_t>(static_cast<double>(cabinet.size()) * threshold),
+          i * numbers_per_aeon, (i + 1) * numbers_per_aeon,
+          static_cast<uint64_t>(std::time(nullptr)), dealer.GetKeys(member->identity.identifier()));
```
n-hutton, line 247: I would do it slightly differently to this: By creating an inherited class of beacon_service that switches out the cabinet creator for TrustedDealer which effectively has a static internally that doles out keys - could chat on Monday?

The reason I say this is that doing this allows you to do nice things like mocking, plus there is no debug-only arguments to functions when you are looking at the production code

libs/crypto/include/crypto/mcl_dkg.hpp
```diff
@@ -41,34 +41,25 @@ using Generator      = bn::G2;
...
-std::vector<bn::Fr> InterpolatePolynom(std::vector<bn::Fr> const &a, std::vector<bn::Fr> const &b);
-
-// For signatures
-Signature SignShare(MessagePayload const &message, PrivateKey const &x_i);
-
-bool VerifySign(PublicKey const &y, MessagePayload const &message, Signature const &sign,
-                Generator const &G);
+struct DkgOutput
+{
+  bn::G2              group_public_key;
```
n-hutton: bn::G2 -> PublicKey ? etc.

Style: Functions first, then variables.

Did you mean for there to be two DkgOutputs structs in this PR?

libs/crypto/src/mcl_dkg.cpp
```diff
@@ -322,6 +322,70 @@ Signature LagrangeInterpolation(std::unordered_map<CabinetIndex, Signature> cons
   return res;
 }
 
+/**
+ * Generates the group public key, public key shares and private key share for a number of
```
n-hutton, line 333: 👍  nice commenting I like it

libs/beacon/tests/dkg/trusted_dealer_test.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+      pending_nodes.emplace(ii);
+    }
+    while (!pending_nodes.empty())
+    {
+      sleep_for(100ms);
+      for (auto it = pending_nodes.begin(); it != pending_nodes.end();)
+      {
+        fetch::beacon::EventCommitteeCompletedWork event;
+        if (cabinet[*it]->event_manager->Poll(event))
```
n-hutton, line 262: Btw, I don't think this actually tests the trusted dealer - it just tests that an event happens (the event could be 'failed to sign' right?)

libs/crypto/tests/unit/mcl_dkg_tests.cpp
```diff
@@ -205,21 +205,14 @@ TEST(MclDkgTests, Signing)
...
+  uint32_t threshold      = 101;
 
-  // Construct polynomial of degree threshold - 1
-  std::vector<bn::Fr> vec_a;
-  Init(vec_a, threshold);
-  for (uint32_t ii = 0; ii < threshold; ++ii)
-  {
-    vec_a[ii].setRand();
-  }
+  // outputs[i] is assigned to node with index i in the committee
```
n-hutton, line 211: I thought the committee indexing only really mattered for the RBC?

libs/crypto/tests/unit/mcl_dkg_tests.cpp
```diff
@@ -205,21 +205,14 @@ TEST(MclDkgTests, Signing)
...
+  auto outputs = TrustedDealerGenerateKeys(committee_size, threshold);
```
n-hutton, line 212: 👍

n-hutton (changes requested):

libs/beacon/tests/dkg/trusted_dealer_test.cpp
jinmannwong: Fixed.

libs/beacon/tests/dkg/trusted_dealer_test.cpp
jinmannwong: The event is that 10 rounds of the random beacon have been generated so if the signatures fail then some nodes will not trigger a EventCommitteeCompletedWork. Also, the test for the TrustedDealer is in the mcl_dkg_tests. Here, I am only testing that the StartNewCabinet with a DkgOutput passed in works.

libs/crypto/tests/unit/mcl_dkg_tests.cpp
jinmannwong: No, the DKG associates each committee member with an index according to their position in the set of muddle address. This is important for the lagrange interpolation required in computing the group signature, as well as other parts of the DKG.

libs/crypto/include/crypto/mcl_dkg.hpp
jinmannwong: Fixed the style and bn::G2. The other DkgOutputs also contains qual, which the crypto function TrustedDealerGenerateKeys does not need to know about. I will edit this to avoid confusion.

libs/beacon/tests/dkg/beacon_manager_tests.cpp
jinmannwong: As I mention below, the DkgOutput here is different from the one declared in mcl_dkg as qual is included here.

PR #1701, opened 2019-09-26
W2v test loss

evsmithx:


libs/ml/examples/model_saver.hpp
```diff
@@ -29,28 +32,175 @@ namespace ml {
...
+}
+
+/**
+ * Saves the saveparams of a graph to a file location specified by user.
+ * Is quicker than SaveModel for large graphs.
+ * @param g the graph to save
+ * @param save_location a string specifying save location
+ */
+template <typename GraphType>
+void SaveLargeModel(GraphType &g, std::string const &save_location)
```
baykaner: SaveModel and SaveLargeModel seems like a fair bit of duplicated code.
How about just having SaveLargeModel (but rename it to SaveModel).
Or if you really need both functionality, you can give it a boolean flag defaulting to large model?

libs/ml/examples/model_saver.hpp
```diff
@@ -29,28 +32,175 @@ namespace ml {
...
+void SaveObject(MLType &g, std::string const &save_location)
+{
+  std::cout << "Starting object serialization" << std::endl;
+
   fetch::serializers::MsgPackSerializer serializer;
-  serializer << g.StateDict();
+
+  serializer << g;
 
   std::fstream file(save_location, std::fstream::out);  // fba = FetchByteArray
```
baykaner: is this comment useful?

libs/ml/examples/model_saver.hpp
```diff
@@ -29,28 +32,175 @@ namespace ml {
...
+  return graph_ptr;
+}
+
+/**
+ * Saves an object to a file location specified by user.
+ * @param g the graph to save
+ * @param save_location a string specifying save location
+ */
+template <typename MLType>
+void SaveObject(MLType &g, std::string const &save_location)
```
baykaner: SaveObject and SaveModel and SaveLargeModel all look suspiciously similar. Is there some duplication here?

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -59,76 +62,10 @@ std::pair<std::string, std::string> Model(fetch::ml::Graph<TensorType> &g, SizeT
...
-
-  for (std::size_t l = 0; l < output.size(); ++l)
-  {
-    std::cout << "rank: " << l << ", "
-              << "distance, " << output.at(l).second << ": " << dl.WordFromIndex(output.at(l).first)
-              << std::endl;
-  }
-}
-
 void TestEmbeddings(Graph<TensorType> const &g, std::string const &skip_gram_name,
```
baykaner, line 65: I wonder if testEmbeddings should live in the dataloader? that seems more like a general utility function?

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -157,46 +90,49 @@ std::string ReadFile(std::string const &path)
 
 struct TrainingParams
 {
-  // TODO (#1585) something is broken here. if u set max_word_count to sth smaller like 10000, there
-  // would be an error at the end of the sentence
+  // TODO (#1585) something is broken here. If you set max_word_count to something smaller like
```
baykaner: I think this TODO is fixed right? Please check and remove as necessary

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -157,46 +90,49 @@ std::string ReadFile(std::string const &path)
...
+  // 10000, there would be an error at the end of the sentence
   SizeType max_word_count = fetch::math::numeric_max<SizeType>();  // maximum number to be trained
-  SizeType negative_sample_size = 5;     // number of negative sample per word-context pair
-  SizeType window_size          = 5;     // window size for context sampling
-  DataType freq_thresh          = 1e-3;  // frequency threshold for subsampling
-  SizeType min_count            = 5;     // infrequent word removal threshold
+  SizeType negative_sample_size = 5;      // number of negative sample per word-context pair
+  SizeType window_size          = 2;      // window size for context sampling
```
baykaner, line 98: is this usual? I don't remember what Mikolov et al. set for window size but I thought it was a max of 4 or 5? not 2?

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -157,46 +90,49 @@ std::string ReadFile(std::string const &path)
...
+  DataType freq_thresh          = 1e-3f;  // frequency threshold for subsampling
+  SizeType min_count            = 100;    // infrequent word removal threshold
```
baykaner, line 100: is this not quite high?
Unless a word occurred 100 times we ditch it?

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -157,46 +90,49 @@ std::string ReadFile(std::string const &path)
...
 
-  SizeType batch_size      = 100000;  // training data batch size
-  SizeType embedding_size  = 100;     // dimension of embedding vec
+  SizeType batch_size      = 10000;  // training data batch size
+  SizeType embedding_size  = 500;    // dimension of embedding vec
```
baykaner, line 103: pretty large default embeddings!
if these hyperparameters are to match an implementation in a paper somewhere please add a comment above them explaining this and giving a link to that paper

libs/ml/examples/word2vec_graph_evaluation/main.cpp
```diff
@@ -0,0 +1,86 @@
...
+//
+//------------------------------------------------------------------------------
+
+#include "math/tensor.hpp"
+#include "ml/core/graph.hpp"
+#include "ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp"
+#include "ml/layers/skip_gram.hpp"
+#include "word2vec_utilities.hpp"
+
+#include <model_saver.hpp>
```
baykaner: "" not <>

libs/ml/examples/word2vec_utilities.hpp
```diff
@@ -0,0 +1,215 @@
...
+{
+  using SizeType = typename TensorType::SizeType;
+  using DataType = typename TensorType::Type;
+
+  // get word vectors for word_ids
+  TensorType word1_vec = embeddings.Slice(word1, 1).Copy();
+  TensorType word2_vec = embeddings.Slice(word2, 1).Copy();
+  TensorType word3_vec = embeddings.Slice(word3, 1).Copy();
+
+  //  word1_vec /= fetch::math::L2Norm(word1_vec);
```
baykaner: you don't normalise these anymore? I thought that was important (to ensure you're really just comparing the angle between entities in embeddings space, rather than absolute distance, which may be less relevant)?

In any case, left in commented out code

libs/ml/examples/word2vec_utilities.hpp
```diff
@@ -0,0 +1,215 @@
...
+  //  word2_vec /= fetch::math::L2Norm(word2_vec);
+  //  word3_vec /= fetch::math::L2Norm(word3_vec);
+
+  TensorType                                 word4_vec = word2_vec - word1_vec + word3_vec;
+  std::vector<std::pair<SizeType, DataType>> output;
```
baykaner: lines 53 & 54 can be one line

libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -417,11 +417,7 @@ void GraphW2VLoader<T>::BufferNextSamples()
   // set up a counter to add samples to buffer
   SizeType counter = 0;
 
-  // reset all three buffers
   input_words_.Fill(cur_word_id);
-  labels_.Fill(BufferPositionUnused);
```
baykaner, line 423: as discussed, you probably shouldn't delete these

baykaner (changes requested):

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
evsmithx: It's only used here - it's similar to one in the distributed example, but the distributed w2v client needs to lock the model and console so it's not the same. Maybe we'll move it if we reuse it somewhere?

baykaner (approved):

PR #1708, opened 2019-09-26
GCC9-related fixes

bipll:
This commit tries to get rid of too many warnings issued by GCC 9.


libs/storage/include/storage/file_object.hpp
```diff
@@ -55,14 +55,7 @@ struct FileBlockType
   static_assert(BS > META_DATA_BYTES,
                 "Block size needs to exceed the min requirement for metadata");
 
-  FileBlockType()
-  {
-    // Ensures that padded bytes are not uninitialised.
-    memset(this, 0, sizeof(decltype(*this)));
```
ejfitzgerald, line 61: I have a vague suspicion that something in the code relied on the padding of the class being zeroed. Maybe we can get around this issue if we mark these structures as packed?

libs/core/include/core/bitvector.hpp
```diff
@@ -48,6 +48,8 @@ class BitVector
   BitVector(BitVector const &other);
   ~BitVector() = default;
 
+  BitVector &operator=(BitVector const &other);
```
ejfitzgerald, line 51: Worth adding a `BitVector &operator=(BitVector &&other) noexcept = default` to be clear (or maybe I have missed the point?

Also I would probably do a move constructor too

libs/storage/include/storage/file_object.hpp
bipll: In other words, something relied on UB.
> Maybe we can get around this issue if we mark these structures as packed?

Yes, a good idea.
Another way is — once we remove non-trivial ctors, technically, these structs turn into POD. Even if zeroing them is absolutely required, it shouldn't be done in their ctors (which breaks trivialness) but rather in an external function, so they are not instantiated directly anymore.

libs/core/include/core/bitvector.hpp
bipll: (Explicitly defaulted are implicitly noexcept.)

libs/storage/include/storage/file_object.hpp
n-hutton: Yeah I think packing would work @ejfitzgerald the only thing I am aware of like that would be if you have padding that you do not zero and you write it to a file, undefined behaviour sanitizer and/or address san complain (I vaguely recall). Also, why would you want random garbage in your file

bipll: The hardest part is actually the physical storage layer where all the IO is performed via whole objects simply being sent as is to `std::fstream::read/write()` and `std::memcpy()`. This would work with objects data but unfortunately not every object kept in stacks can be called such, even if packed.
I have a plan for a better temporary fix which this PR is too narrow to contain. :\

n-hutton: @bipll I am curious, what is your solution?

PR #1709, opened 2019-09-26
[Ledger] Storage Integrity tool (prototype)

pbukva:


apps/chain-tool/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+#
+# T X   G E N E R A T O R
```
ejfitzgerald: Header name

apps/chain-tool/main.cpp
```diff
@@ -0,0 +1,890 @@
...
+#include "ledger/chain/transaction_layout_rpc_serializers.hpp"
+#include "ledger/chain/transaction_rpc_serializers.hpp"
+#include "meta/log2.hpp"
+
+#include "storage/object_store.hpp"
+
+#include <fstream>
+#include <regex>
+
+#include <dirent.h>
```
ejfitzgerald, line 34: Worth wrapping this and probably moving to the vectorise library or similar (because that is sort of our platform library)

apps/chain-tool/main.cpp
```diff
@@ -0,0 +1,890 @@
...
+
+  std::string const file_name;
+
+  explicit ChainHeadStore(std::string file_name_ = std::string{DEFAULT_FILE_NAME})
+    : file_name{std::move(file_name_)}
+  {}
+
+  BlockHash GetHead() const
+  {
+    std::fstream head_store;
```
ejfitzgerald: There is already a function that basically does reading a whole file into a buffer, might be worth to reuse that

.clang-tidy
```diff
@@ -1,6 +1,6 @@
 Checks: "-*,\
 bugprone-*,\
-cert-dcl21-cpp,\
+-cert-dcl21-cpp,\
```
wkaluza: Please re-enable. https://clang.llvm.org/extra/clang-tidy/checks/cert-dcl21-cpp.html

wkaluza (changes requested):

libs/ledger/include/ledger/chain/block.hpp
```diff
@@ -46,6 +46,8 @@ class Block
   using Slice    = std::vector<TransactionLayout>;
   using Slices   = std::vector<Slice>;
   using DAGEpoch = fetch::ledger::DAGEpoch;
+  using Hash     = Digest;
```
pbukva: Eh, not quite sure how this one got here.

apps/chain-tool/main.cpp
```diff
@@ -0,0 +1,890 @@
...
+  bch.IterateChainBackward(
+      heaviest_chain,
+      [&tx_stores, &tx_count_in_blockchain, &last_reported_progress_tx_count, &tx_count_missing,
+       &tx_count_missing_accumulated, &tx_count_processed, &tx_count_stored_in_trimmed_db,
+       &trimmed_tx_stores, progress_step, print_missing_txs,
+       log2_num_of_lanes](BlockNode const &node, BlockHash const & /*block_hash*/) {
+        uint64_t slice_idx{0};
+        for (auto const &slice : node.db_record.block.body.slices)
+        {
+          if (!node.is_block_set)
```
pbukva: Eh, this should go above the for cycle

apps/chain-tool/main.cpp
```diff
@@ -0,0 +1,890 @@
...
+            Transaction tx;
+
+            bool res{false};
+            try
+            {
+              res = tx_stores[lane].Get(storage::ResourceID{tx_layout.digest()}, tx);
+            }
+            catch (...)
+            {
+              if (print_missing_txs)
```
pbukva, line 795: This printout should either be removed (since printout is handled bellow the catch block), or the mesage should be changed - that Tx fetch failed with exception (adding exception`s `what()` message).

apps/chain-tool/main.cpp
```diff
@@ -0,0 +1,890 @@
...
+              {
+                std::cerr << "EXCEPTION: Tx fetch from db failed:"
+                          << " lane = " << lane << ", tx hash = " << tx_layout.digest().ToHex()
+                          << std::endl;
+                std::cerr.flush();
```
pbukva, line 800: not necessary

apps/chain-tool/main.cpp
```diff
@@ -0,0 +1,890 @@
...
+  }
+
+  std::size_t lane{0};
+  for (auto const &count : tx_count_missing)
+  {
+    if (count > 0)
+    {
+      std::cerr << "INCONSISTENCY:"
+                << " Lane " << lane << " is missing " << count
+                << " transactions required by block-chain are missing in tx db store" << std::endl;
```
pbukva: Reformulate

.clang-tidy
ejfitzgerald: This was the result of a discussion between @pbukva and I. Lets discuss in the office, however, in brief, it looks like this check is not always that helpful in the case of post-increment operators.

ejfitzgerald (changes requested):

.clang-tidy
wkaluza: I'd be interested to hear a more detailed explanation. Also, last time I tried to define a postfix increment for an iterator, your take was that we should avoid them altogether. What changed?

libs/ledger/include/ledger/chain/block.hpp
ejfitzgerald, line 49: Think that we should just use `Digest`

libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -78,12 +78,14 @@ enum class BlockStatus
...
 
+struct BlockDbRecord;
+
 class MainChain
 {
 public:
   using BlockPtr             = std::shared_ptr<Block const>;
   using Blocks               = std::vector<BlockPtr>;
-  using BlockHash            = Digest;
+  using BlockHash            = Block::Hash;
```
ejfitzgerald, line 88: `Digest is better here`

apps/chain-tool/main.cpp
```diff
@@ -0,0 +1,890 @@
...
+
+    std::fstream head_store;
+    head_store.open(file_name, std::ios::binary | std::ios::out | std::ios::trunc);
+    head_store.seekg(0, std::ios::beg);
+    head_store.write(reinterpret_cast<char const *>(head.pointer()),
+                     static_cast<std::streamsize>(head.size()));
+  }
+};
+
+// struct MnimalisticBlockDBMetadata
```
ejfitzgerald: Dead code?

apps/chain-tool/main.cpp
```diff
@@ -0,0 +1,890 @@
...
+      s << "The root node of the chain has wrong index " << last_block_number
+        << ", expected value is 0.";
+      err_msg = s.str();
+    }
+
+    return {err_code, err_msg};
+  }
+
+private:
+  static Tree read_block_db(BlockStore &block_store)
```
ejfitzgerald: Style 🚓 : Inconsistent naming

apps/chain-tool/main.cpp
```diff
@@ -0,0 +1,890 @@
...
+      chains.emplace(curr.chain);
+    }
+
+    stack.pop_back();
+  }
+
+  return retval;
+}
+
+std::tuple<TxStores, int, std::string> open_tx_db_stores()
```
ejfitzgerald: Style 🚓 - Again here

apps/chain-tool/main.cpp
```diff
@@ -0,0 +1,890 @@
...
+  if (dp)
+  {
+    std::regex const rex{"node_storage_lane([0-9]+)_transaction\\.db"};
+    std::smatch      match;
+
+    LaneIdx lane_idx_min{std::numeric_limits<LaneIdx>::max()};
+    LaneIdx lane_idx_max{0};
+
+    dirent *entry{nullptr};
+    while ((entry = readdir(dp.get())))
```
ejfitzgerald, line 683: This would be nice to abstract away into a platform library

ejfitzgerald (changes requested):

.clang-tidy
wkaluza: Ok, having spoken to @ejfitzgerald, let's roll with it for now. After this PR is in, I will try to re-enable this check and ininvestigate why it was failing.

ejfitzgerald (dismissed):
LGTM

libs/ledger/include/ledger/chain/block.hpp
pbukva: Using the `Digest` name here will be a little problematic, since we already have publicly available type alias `Digest` defined in `fetch::ledger` namespace (declared in the `libs/ledger/include/ledger/chain/digest.hpp` file).

pbukva: I can simply remove the `Hash` declaration here.

wkaluza (approved):

PR #1710, opened 2019-09-27
Improve static analysis output

wkaluza:
Adds character offset and source code context to violation report. Also prints disabled checks in addition to enabled ones.

Example violation report:
```
Static analysis check(s) failed in:
  libs/ledger/include/ledger/chain/block_coordinator.hpp

  Line: 397 @ 13
  Message: converting integer literal to bool, use bool literal instead
  Check: modernize-use-bool-literals
  Context:
      mining_ = 2;
                ^
```


evsmithx (approved):
LGTM

PR #1712, opened 2019-09-30
reintroduce backward trace

n-hutton:
accidentally removed in commit #3f82faaa

### DIFF
--- a/libs/ledger/src/chain/block_coordinator.cpp
+++ b/libs/ledger/src/chain/block_coordinator.cpp
@@ -206,6 +206,9 @@ BlockCoordinator::BlockCoordinator(MainChain &chain, DAGPtr dag,
   state_machine_->RegisterHandler(State::RESET,                        this, &BlockCoordinator::OnReset);
   // clang-format on
 
+  int *thing = nullptr;
+  thing[9]   = 9;
+
   state_machine_->OnStateChange([this](State current, State previous) {
     FETCH_UNUSED(this);
     if (periodic_print_.Poll())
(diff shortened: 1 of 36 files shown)


libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -206,6 +206,9 @@ BlockCoordinator::BlockCoordinator(MainChain &chain, DAGPtr dag,
   state_machine_->RegisterHandler(State::RESET,                        this, &BlockCoordinator::OnReset);
   // clang-format on
 
+  int *thing = nullptr;
```
ejfitzgerald: Not sure you want to keep this in

scripts/install-test-dependencies.sh
```diff
@@ -4,3 +4,5 @@ set -e
 pip3 install --no-cache-dir --ignore-installed --force-reinstall wheel pyyaml requests
 pip3 install --no-cache-dir --ignore-installed --force-reinstall fetchai-ledger-api==0.8.0a2
 pip3 install --no-cache-dir --ignore-installed --force-reinstall ./scripts/fetchai_netutils
+
+sudo apt install libdw-dev
```
ejfitzgerald: I think that this should be part of the develop docker image

wkaluza:
Does this complement the changes you made to your fork of `backward-cpp`? They are currently unused on master.

n-hutton:
Yes, this utilizes the fetch logging when stacktracing which is what the fork is for. This was used but you removed it at one point, I presume accidentally. There was also some backtrace macros I will eventually put back in (they can be useful)

ejfitzgerald (approved):

PR #1713, opened 2019-09-30
Updating model interface

baykaner:
- reworked model interface to be in line with keras.model style
- replaced shared_ptr dataloader/optimiser with unique_ptrs

PR #1714, opened 2019-09-30
Sequential model

baykaner:
- implements sequential model
should be merged in after #1713


libs/ml/tests/ml/model/sequential_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+                                                               train_data, train_labels);
+
+  // test loss decreases
+  fetch::math::SizeType count{0};
+  while (count < n_training_steps)
+  {
+    DataType loss{0};
+    DataType later_loss{0};
+    EXPECT_TRUE(model.Train(1, loss));
+    EXPECT_TRUE(model.Train(1, later_loss));
```
evsmithx: Should there be a check here that later_loss < loss?

evsmithx (dismissed):

libs/ml/tests/ml/model/sequential_test.cpp
baykaner: yep, nice catch

evsmithx (approved):
LGTM

PR #1718, opened 2019-10-01
Benchmarks signing, verification and group signature computation for entropy

jinmannwong:
Benchmarks for crypto functions used in entropy generation for committee sizes between 50 and 500. Trusted dealer branch has been merged into this to generate keys for signing.


libs/crypto/benchmark/bls_signing_and_verification.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+using fetch::byte_array::ConstByteArray;
+using fetch::byte_array::ByteArray;
+using fetch::random::LinearCongruentialGenerator;
+
+namespace {
+using RNG = LinearCongruentialGenerator;
+RNG rng;
+
+template <std::size_t LENGTH>
```
n-hutton: Slightly unneccessary use of templete here where a parameter would do, I feel like

libs/crypto/benchmark/bls_signing_and_verification.cpp
```diff
@@ -0,0 +1,141 @@
...
+  {
+    *words++ = rng();
+  }
+
+  return ConstByteArray{buffer};
+}
+
+void SignBLSSignature(benchmark::State &state)
+{
+  bn::initPairing();
```
n-hutton: OK, we need to resolve this requirement for this to be called - I would suggest you add it to this PR. If we have a static/global object which is constructed before runtime we can guarantee that initPairing() is always called before any code is run, so long as you're linking agsint the crypto library.

I say this as it's quite error prone to have to remember to do this, plus it's almost certainly not thread safe. I had an issue with the block entropy PR as I mentioned to you - caused by effectively a race to init this in various places.

libs/crypto/benchmark/bls_signing_and_verification.cpp
```diff
@@ -0,0 +1,141 @@
...
+  auto     outputs = fetch::crypto::mcl::TrustedDealerGenerateKeys(committee_size, threshold);
+  assert(outputs.size() == committee_size);
+
+  // Randomly select index to sign
+  auto index = static_cast<uint32_t>(rng() % committee_size);
+
+  for (auto _ : state)
+  {
+    // Sign message
+    fetch::crypto::mcl::SignShare(msg, outputs[index].private_key_share);
```
n-hutton, line 73: I suppose it's fine that the message being signed each time is the same, although I might consider signing each time with a different message to guard against any caching throwing the result

libs/crypto/benchmark/bls_signing_and_verification.cpp
```diff
@@ -0,0 +1,141 @@
...
+  auto     outputs = fetch::crypto::mcl::TrustedDealerGenerateKeys(committee_size, threshold);
+
+  // Randomly select index to sign
+  auto sign_index = static_cast<uint32_t>(rng() % committee_size);
+  auto signature  = fetch::crypto::mcl::SignShare(msg, outputs[sign_index].private_key_share);
+
+  // Randomly select another index to verify
+  auto verify_index = static_cast<uint32_t>(rng() % committee_size);
+  bool check;
+  FETCH_UNUSED(check);
```
n-hutton: delete check?

libs/crypto/benchmark/bls_signing_and_verification.cpp
```diff
@@ -0,0 +1,141 @@
...
+  std::unordered_map<uint32_t, fetch::crypto::mcl::Signature> threshold_signatures;
+  while (threshold_signatures.size() < threshold)
+  {
+    auto sign_index = static_cast<uint32_t>(rng() % committee_size);
+    auto signature  = fetch::crypto::mcl::SignShare(msg, outputs[sign_index].private_key_share);
+    threshold_signatures.insert({sign_index, signature});
+  }
+
+  fetch::crypto::mcl::Signature group_sig;
+  FETCH_UNUSED(group_sig);
```
n-hutton: delete unused?

libs/crypto/benchmark/bls_signing_and_verification.cpp
```diff
@@ -0,0 +1,141 @@
...
+  for (auto _ : state)
+  {
+    // Compute group signature
+    group_sig = fetch::crypto::mcl::LagrangeInterpolation(threshold_signatures);
+  }
+  auto index = static_cast<uint32_t>(rng() % committee_size);
+  FETCH_UNUSED(index);
```
n-hutton: You don't usually test anything in a benchmark - in this case devs would be unlikely to see this as you don't usually run the benchmark in debug mode where this assert would be enabled

n-hutton (approved):
Looks good. A few minor comments

libs/crypto/benchmark/bls_signing_and_verification.cpp
jinmannwong: Ok, I can remove these unused variables which check the computation is correct.

libs/crypto/benchmark/bls_signing_and_verification.cpp
jinmannwong: Ok, I can move that inside the for loop and just pause the timings.

libs/crypto/benchmark/bls_signing_and_verification.cpp
jinmannwong: Sure.

n-hutton (approved):

PR #1727, opened 2019-10-02
Model serialisation with tests

MissingNO57:


libs/ml/include/ml/model/dnn_classifier.hpp
```diff
@@ -76,4 +84,36 @@ DNNClassifier<TensorType>::DNNClassifier(ModelConfig<DataType>        model_conf
...
+struct MapSerializer<ml::model::DNNClassifier<TensorType>, D>
+{
+  using Type                      = ml::model::DNNClassifier<TensorType>;
+  using DriverType                = D;
+  static uint8_t const BASE_MODEL = 1;
+
+  template <typename Constructor>
+  static void Serialize(Constructor &map_constructor, Type const &sp)
+  {
+    auto map = map_constructor(2);
```
evsmithx: This should be 1 not 2

libs/ml/include/ml/model/dnn_regressor.hpp
```diff
@@ -76,4 +84,36 @@ DNNRegressor<TensorType>::DNNRegressor(ModelConfig<DataType>        model_config
...
+struct MapSerializer<ml::model::DNNRegressor<TensorType>, D>
+{
+  using Type                      = ml::model::DNNRegressor<TensorType>;
+  using DriverType                = D;
+  static uint8_t const BASE_MODEL = 1;
+
+  template <typename Constructor>
+  static void Serialize(Constructor &map_constructor, Type const &sp)
+  {
+    auto map = map_constructor(2);
```
evsmithx: map_constructor(1)

libs/ml/include/ml/model/model.hpp
```diff
@@ -272,4 +276,191 @@ void Model<TensorType>::PrintStats(SizeType epoch, DataType loss, DataType test_
...
+  static uint8_t const ERROR_NODE_NAME  = 10;
+
+  static uint8_t const LOSS_SET_FLAG      = 11;
+  static uint8_t const OPTIMISER_SET_FLAG = 12;
+  static uint8_t const COMPILED_FLAG      = 13;
+
+  template <typename Constructor>
+  static void Serialize(Constructor &map_constructor, Type const &sp)
+  {
+    auto map = map_constructor(14);
```
evsmithx: 13

libs/ml/include/ml/model/model.hpp
```diff
@@ -272,4 +276,191 @@ void Model<TensorType>::PrintStats(SizeType epoch, DataType loss, DataType test_
...
+
+    map.Append(DATALOADER_TYPE, static_cast<uint8_t>(sp.dataloader_ptr_->LoaderCode()));
+
+    // serialize dataloader
+    switch (sp.dataloader_ptr_->LoaderCode())
+    {
+    case ml::LoaderType::TENSOR:
+    {
+      fetch::ml::dataloaders::TensorDataLoader<TensorType, TensorType> *loader_ptr =
+          static_cast<fetch::ml::dataloaders::TensorDataLoader<TensorType, TensorType> *>(
```
evsmithx, line 334: You can use auto here to avoid duplicating the type name

libs/ml/include/ml/model/model_config.hpp
```diff
@@ -51,4 +54,56 @@ struct ModelConfig
...
+  static uint8_t const MIN_DELTA           = 4;
+  static uint8_t const LEARNING_RATE_PARAM = 6;
+  static uint8_t const BATCH_SIZE          = 7;
+  static uint8_t const SUSBET_SIZE         = 8;
+  static uint8_t const PRINT_STATS         = 9;
+
+  template <typename Constructor>
+  static void Serialize(Constructor &map_constructor, Type const &sp)
+  {
+    auto map = map_constructor(10);
```
evsmithx: 9

libs/ml/include/ml/model/sequential.hpp
```diff
@@ -79,8 +86,47 @@ void Sequential<TensorType>::Add(Params... params)
...
+  using Type                          = ml::model::Sequential<TensorType>;
+  using DriverType                    = D;
+  static uint8_t const BASE_MODEL     = 1;
+  static uint8_t const LAYER_COUNT    = 2;
+  static uint8_t const PREV_LAYER_STR = 3;
+
+  template <typename Constructor>
+  static void Serialize(Constructor &map_constructor, Type const &sp)
+  {
+    auto map = map_constructor(4);
```
evsmithx: 3

libs/ml/tests/ml/model/dnn_regressor_test.cpp
```diff
@@ -88,24 +90,24 @@ bool RunTest(fetch::ml::optimisers::OptimiserType optimiser_type,
...
 
+  DataType loss{0};
+  model.Train(1, loss);
+
   // test loss decreases
   fetch::math::SizeType count{0};
   while (count < n_training_steps)
   {
-    DataType loss{0};
     DataType later_loss{0};
```
evsmithx: Are we trying to check here that loss decreases every training step? If so then loss and later_loss should be defined outside the loop, and after doing one training step compare the values then set loss = later_loss.

libs/ml/tests/ml/model/sequential_test.cpp
```diff
@@ -92,24 +94,23 @@ bool RunTest(fetch::ml::optimisers::OptimiserType optimiser_type,
...
   fetch::math::SizeType count{0};
   while (count < n_training_steps)
   {
-    DataType loss{0};
     DataType later_loss{0};
-    model.Train(1, loss);
     model.Train(1, later_loss);
     EXPECT_LE(later_loss, loss);
     count++;
   }
```
evsmithx, line 112: See previous comment about this code.

libs/ml/tests/ml/model/sequential_test.cpp
```diff
@@ -122,32 +123,95 @@ TYPED_TEST(ModelsTest, adagrad_sequential)
...
 }
+
 TYPED_TEST(ModelsTest, adam_sequential)
 {
   using DataType = typename TypeParam::Type;
-  ASSERT_TRUE(RunTest<TypeParam>(fetch::ml::optimisers::OptimiserType::ADAM,
-                                 static_cast<DataType>(1e-5),
-                                 static_cast<typename TypeParam::Type>(0.1)));
+  ASSERT_TRUE(RunTest<TypeParam>(fetch::ml::OptimiserType::ADAM, static_cast<DataType>(1e-3),
+                                 static_cast<DataType>(1e-2), 10));
```
evsmithx, line 138: Does this test fail with the old values for the tolerance? Could the tolerance be expressed here in terms of the datatype tolerance?

evsmithx (changes requested):

libs/ml/tests/ml/model/dnn_regressor_test.cpp
MissingNO57: I had some issues with Adam and Momentum, where a loss could jump up.

libs/ml/tests/ml/model/sequential_test.cpp
MissingNO57: It's not the same tolerance, because it is ground truth/prediction difference.

MissingNO57: It doesn't depend on datatype

baykaner: Even so, this tolerance is quite high!
Is this because default learning rate is bad or because number of training runs is too small? either way It would be preferable to have these hyperparameters set appropriately via input parameters to RunTest function call rather than compensate for bad hyperparameters with high tolerance.

libs/ml/include/ml/model/dnn_classifier.hpp
baykaner: this still needs fixing

libs/ml/include/ml/model/dnn_classifier.hpp
```diff
@@ -33,10 +34,17 @@ class DNNClassifier : public Model<TensorType>
...
   using DataLoaderPtrType = typename Model<TensorType>::DataLoaderPtrType;
 
+  DNNClassifier()                           = default;
+  DNNClassifier(DNNClassifier const &other) = default;
+  ~DNNClassifier() override                 = default;
+
   DNNClassifier(ModelConfig<DataType> model_config, std::vector<SizeType> const &hidden_layers);
+
+  template <typename X, typename D>
+  friend struct serializers::MapSerializer;
```
baykaner: friendship needed?

libs/ml/include/ml/model/dnn_regressor.hpp
```diff
@@ -33,10 +34,17 @@ class DNNRegressor : public Model<TensorType>
...
   using DataLoaderPtrType = typename Model<TensorType>::DataLoaderPtrType;
 
+  DNNRegressor()                          = default;
+  DNNRegressor(DNNRegressor const &other) = default;
+  ~DNNRegressor() override                = default;
+
   DNNRegressor(ModelConfig<DataType> model_config, std::vector<SizeType> const &hidden_layers);
+
+  template <typename X, typename D>
+  friend struct serializers::MapSerializer;
```
baykaner: friendship needed?

libs/ml/include/ml/model/model.hpp
```diff
@@ -272,4 +284,214 @@ void Model<TensorType>::PrintStats(SizeType epoch, DataType loss, DataType test_
...
+      break;
+    }
+
+    case ml::LoaderType::MNIST:
+    case ml::LoaderType::SGNS:
+    case ml::LoaderType::W2V:
+    case ml::LoaderType::COMMODITY:
+    case ml::LoaderType::C2V:
+    {
+      throw std::runtime_error("Serialization for current dataloader type not implemented yet.");
```
baykaner: perhaps we delay merging this in until after my ML Exceptions PR goes in. Then replace these runtime_error calls with InvalidMode calls?

libs/ml/include/ml/model/model.hpp
```diff
@@ -272,4 +284,214 @@ void Model<TensorType>::PrintStats(SizeType epoch, DataType loss, DataType test_
...
+      sp.optimiser_ptr_->Init();
+      break;
+    }
+
+    case ml::OptimiserType::ADAGRAD:
+    case ml::OptimiserType::ADAM:
+    case ml::OptimiserType::MOMENTUM:
+    case ml::OptimiserType::RMSPROP:
+    {
+      throw std::runtime_error("serialization for current optimiser type not implemented yet.");
```
baykaner: same as above

libs/ml/include/ml/model/model.hpp
```diff
@@ -272,4 +284,214 @@ void Model<TensorType>::PrintStats(SizeType epoch, DataType loss, DataType test_
...
+  static uint8_t const LOSS_SET_FLAG      = 11;
+  static uint8_t const OPTIMISER_SET_FLAG = 12;
+  static uint8_t const COMPILED_FLAG      = 13;
+
+  template <typename MapType>
+  static void SerializeDataLoader(MapType map, Type const &sp)
+  {
+    map.Append(DATALOADER_TYPE, static_cast<uint8_t>(sp.dataloader_ptr_->LoaderCode()));
+
+    switch (sp.dataloader_ptr_->LoaderCode())
```
baykaner, line 329: is this switch case code really necessary?
I think you might be able to just have the main Serialize() function with map.Append(). Specifically, because some cases may not be implemented yet you may want to do this:
try
{
  map.Append(sp.dataloader_)
}
catch (Exception& e)
{
   cout << e.what() << endl;
}

If it doesn't compile because it requires a Serialise function to exist for each dataloader then just create them with serialise and deserialise functions that simply throw ml::exceptions::NotYetImplementedError()

libs/ml/include/ml/model/model_config.hpp
```diff
@@ -40,6 +40,9 @@ struct ModelConfig
 
   bool print_stats = false;
 
+  template <typename X, typename D>
+  friend struct serializers::MapSerializer;
```
baykaner: is this necessary?

baykaner (changes requested):
- i think the switch cases in model serializer can be ditched in favour of try/catch
- please wait until i get in the ML Exceptions PR then replace runtime_errors with the relevant custom exceptions

libs/ml/include/ml/exceptions/exceptions.hpp
```diff
@@ -123,6 +123,31 @@ class InvalidInput : public std::runtime_error
...
+    if (!(msg_.empty()))
+    {
+      return msg_.c_str();
+    }
+    return "invalid type";
+  }
+
+private:
+  std::string msg_;
+};
```
baykaner: why not just ditch this and use the existing InvalidMode?

baykaner (changes requested):
I think we should try to avoid an unnecessary proliferation of exceptions. InvalidMode is a generic exception that I think also covers your use case for the dataloaders

baykaner (dismissed):

evsmithx (dismissed):
LGTM

baykaner (approved):

PR #1728, opened 2019-10-02
Fix unstable synergetic e2e test

wkaluza:


libs/network/src/tcp/tcp_server.cpp
```diff
@@ -133,7 +133,7 @@ void TCPServer::Start()
 
 void TCPServer::Stop()
 {
-  // TODO(EJF): Hmmmmmm...???
+  // TODO(EJF): Hmmmmmm...?
```
ejfitzgerald: Lets simple remove this

ejfitzgerald (changes requested):
Minor but lets get this in

libs/network/src/tcp/tcp_server.cpp
wkaluza: Done

ejfitzgerald (approved):

PR #1738, opened 2019-10-03
Avoid self-deletion

wkaluza:
Move the responsibility of deleting a released `vm::Object` from the `Object` class itself into `vm::Ptr`.


robertdickson (approved):
LGTM

PR #1739, opened 2019-10-03
poc auto lint

n-hutton:
poc for an interim solution to auto fixing headers + cpp files like so: ```./scripts/ci-tool.py --lint -C master Debug --fix```

Would require the docker image to install ```clang-tools-6.0```

I believe this would require you to have run ```./scripts/ci-tool.py -B Debug``` to completion at least once before


.clang-tidy
```diff
@@ -16,6 +16,16 @@ misc-*,\
 modernize-*,\
 performance-*,\
 readability-*,\
+-readability-named-parameter,\
```
n-hutton, line 19: Might be worth considering disabling these until bazil = fast builds comes online in a week or two.

```
-readability-named-parameter -> "all parameters should be named in a function"
-readability-named-parameter -> "same as above"
-misc-unused-parameters -> "all parameters must be used - somehow this is triggered even with -WError"
-performance-unnecessary-value-param -> "parameter is copied for each invocation but only used as a const reference; consider making it a const reference"
-hicpp-use-auto -> "use auto when initializing with a cast to avoid duplicating the type name"
-modernize-use-auto -> "same as above"
-performance-move-const-arg -> "passing result of std::move() as a const reference argument; no move will actually happen"
-readability-container-size-empty -> "the 'empty' method should be used to check for emptiness instead of 'size'"
-readability-implicit-bool-conversion -> "implicit conversions to bool: if(pointer)"
-readability-inconsistent-declaration-parameter-name -> "Overloading functions"
```

scripts/end_to_end_test/here_debug/end_to_end_test_default/node0/node.log
```diff
@@ -0,0 +1,126 @@
+ F E ╱     Constellation v0.9.0-alpha3-150-g64a22b96
```
wkaluza, line 1: Remove this whole directory.

scripts/ci-tool.py
```diff
@@ -165,6 +165,10 @@ def parse_commandline():
                         help='Run the etch language tests')
     parser.add_argument('--lint', action='store_true',
                         help='Run clang-tidy')
+    parser.add_argument('--fix-lint', action='store_true',
+                        help='Run clang-tidy and attempt to fix lint errors')
+    parser.add_argument('-C', '--commit', nargs=1, default=None,
```
wkaluza, line 170: Use lowercase `-c` for consistency with `apply_style.py`

scripts/fetchai_code_quality/internal/static_analysis.py
```diff
@@ -4,9 +4,107 @@
...
+                        target_cpp_file))
+
+                target_cpp_file = os.path.abspath(target_cpp_file[0])
+
+                if verbose:
+                    print("source file: {}".format(dependency_file))
+                    print("target cpp file: {}".format(target_cpp_file))
+
+                changed_files_fullpath.add(target_cpp_file)
+                # changed_files_fullpath.remove(dependency_file)
```
wkaluza, line 92: Leftover comment

.clang-tidy
```diff
@@ -16,6 +16,16 @@ misc-*,\
...
+-misc-unused-parameters,\
+-performance-unnecessary-value-param,\
+-hicpp-use-auto,\
+-modernize-use-auto,\
+-performance-move-const-arg,\
+-readability-container-size-empty,\
+-readability-implicit-bool-conversion,\
+-readability-inconsistent-declaration-parameter-name,\
+-readability-named-parameter,\
```
wkaluza, line 28: `readability-named-parameter` excluded twice.

wkaluza: How much time are we saving by disabling these checks?

wkaluza (changes requested):
I welcome consolidating the old scripts under the umbrella of `ci-tool`.

Minor changes requested. Additionally, this solution is not robust, as not all CMake generators create `depend.make` files. You may not use Ninja, but others do.

> I believe this would require you to have run ./scripts/ci-tool.py -B Debug to completion at least once before

Can we stop perpetuating this myth, please? People are wasting time running full builds just to perform static analysis. It's perfectly fine to just run a CMake configure step, and `ci-tool.py --lint` already does that for you.

scripts/end_to_end_test/here_debug/end_to_end_test_default/node0/node.log
n-hutton: done

scripts/ci-tool.py
n-hutton: Are you sure? All other flags are uppercase for this script

scripts/fetchai_code_quality/internal/static_analysis.py
n-hutton: done

n-hutton:
@wkaluza Indeed, the solution is not robust, but not for that reason - I would say it is not robust as it makes assumptions about the format of build files that for all I know can change version to version.

I do use ninja, but have to run linting in docker. I am not aware of any other code paths to ```static_analysis.py``` than via the ci tool so you would be forced to use make anyway, and if you have ccache I don't believe this should take significant time.

```Can we stop perpetuating this myth, please?``` - I don't think it is a myth - this autofix relies on compilation artifacts so you need at least one semi-recent build.

File: .clang-tidy (hunk 5 above)
Comment (n-hutton) [reply .clang-tidy:28]: Build time - don't know, probably not a lot. Developer time - a significant amount since they are fairly likely to trip the dev making them do another CI run. Additionally, they are likely to combine with other checks, with only one being emittied. The net result being that you dutifully fix style issues only to see more an hour later after linting again.

**Comment by wkaluza** [comment]
> via the ci-tool ... forced to use make

This is incorrect: see `cmake_configure` function in `ci-tool`.

File: .clang-tidy (hunk 5 above)
Comment (wkaluza) [reply .clang-tidy:28]: How does `readability-container-size-empty` 'combine with other checks'?

**Comment by wkaluza** [comment]
> Can we stop perpetuating this myth, please? - I don't think it is a myth - this autofix relies on compilation artifacts so you need at least one semi-recent build

I just checked out your branch and ran it to a fresh build path against the following (in `flatten.hpp`):

```
  ~Flatten() = default;

  virtual std::shared_ptr<OpsSaveableParams> GetOpSaveableParams()
  {
    auto ret         = std::make_shared<SPType>();
    ret->input_shape = input_shape_;
    return ret;
  }
```
Command used:
```
python3 -u ./scripts/ci-tool.py --lint --fix-lint -p build-nathan- Debug
```
Output:
```
----------
Static analysis check(s) failed in:
  libs/ml/include/ml/ops/flatten.hpp

  Line: 50 @ 3
  Message: annotate this function with 'override' or (rarely) 'final'
  Check: hicpp-use-override
  Context:
      ~Flatten() = default;
      ^

  Line: 52 @ 46
  Message: prefer using 'override' or (rarely) 'final' instead of 'virtual'
  Check: modernize-use-override
  Context:
      virtual std::shared_ptr<OpsSaveableParams> GetOpSaveableParams()
                                                 ^

  Line: 52 @ 46
  Message: prefer using 'override' or (rarely) 'final' instead of 'virtual'
  Check: hicpp-use-override
  Context:
      virtual std::shared_ptr<OpsSaveableParams> GetOpSaveableParams()
                                                 ^

  Line: 52 @ 46
  Message: 'GetOpSaveableParams' overrides a member function but is not marked 'override'
  Check: clang-diagnostic-inconsistent-missing-override
  Context:
      virtual std::shared_ptr<OpsSaveableParams> GetOpSaveableParams()
                                                 ^
```

Resulting code:
```
  ~Flatten() override = default;

  std::shared_ptr<OpsSaveableParams> GetOpSaveableParams() override
  {
    auto ret         = std::make_shared<SPType>();
    ret->input_shape = input_shape_;
    return ret;
  }
```

**Comment by n-hutton** [comment]
As far as I understand it (although I could be wrong), you DO need to have built at least once to populate depend.make files, which the tool would rely on. ONLY for the case
of autofixing.

Example (checkout commit 49f58051)

```
# Clean any build artifacts
$ rm build-debug -rf;

# Attempt to fix linting
$ ./scripts/ci-tool.py --lint -C HEAD Debug --fix
( sets up build-debug afresh etc.)
-- Configuring done
-- Generating done
-- Build files have been written to: /build/build-debug

WARNING: No non-empty depend.make files found in /build/build-debug . Did you make sure to build in this directory?

WARNING: Failed to find dependencies for the following files:
{'/build/libs/beacon/include/beacon/beacon_manager.hpp'}
There doesn't appear to be any relevant files to lint. Quitting.

# Look into build debug
$ find build-debug/ -name depend.make | xargs cat
# This may be replaced when dependencies are built.
# Empty dependencies file for mcl.
# This may be replaced when dependencies are built.
# Empty dependencies file for bn_c384_256_test.
# This may be replaced when dependencies are built.
# Empty dependencies file for mclbn384.
# This may be replaced when dependencies are built.
# Empty dependencies file for mclbn256.
# This may be replaced when dependencies are built.
# Empty dependencies file for gmock.
# This may be replaced when dependencies are built.
# Empty dependencies file for gmock_main.
# This may be replaced when dependencies are built.
# Empty dependencies file for gtest.
# This may be replaced when dependencies are built.
# Empty dependencies file for gtest_main.
# This may be replaced when dependencies are built.
...

# Run build for the first time
$ ./scripts/ci-tool.py -B Debug

# Now run auto-lint as before
$ time ./scripts/ci-tool.py --lint -C HEAD Debug --fix
...
  readability-container-size-empty
  readability-identifier-naming
  readability-implicit-bool-conversion
  readability-inconsistent-declaration-parameter-name
  readability-named-parameter

Command:
  python3 -u /usr/bin/run-clang-tidy-6.0.py -j6 -p=/build/build-debug/shadow_build_root -header-filter=.*(apps|libs).*\.hpp$ -quiet -clang-tidy-binary=/usr/bin/clang-tidy-6.0 -clang-apply-replacements-binary=/usr/bin/clang-tidy-6.0 -export-fixes=/build/build-debug/shadow_build_root/clang_tidy_fixes.yaml

Working...
Done.
Processing results...

----------
Static analysis check(s) failed in:
  libs/beacon/include/beacon/beacon_manager.hpp

  Line: 60 @ 3
  Message: single-argument constructors must be marked explicit to avoid unintentional implicit conversions
  Check: google-explicit-constructor
  Context:
      BeaconManager(CertificatePtr certificate = nullptr);
      ^


Static analysis found 1 violation(s) in 1 file(s)
Applying fixes...
Done. Note that the automated fix feature of clang-tidy is unreliable:
  * not all checks are supported,
  * applying fixes may result in invalid code.

real	2m6.432s
user	7m42.020s
sys	0m38.870s
```

So one disadvantage of this methodology is that you need to have fairly recently built using the ci tool to get correct results.

PR #1740, opened 2019-10-04
Network Stability Improvements

ejfitzgerald:


libs/muddle/internal/muddle_register.hpp
```diff
@@ -115,11 +116,10 @@ class MuddleRegister : public network::AbstractConnectionRegister
   std::string const name_;
   char const *const logging_name_{name_.c_str()};
 
-  std::atomic<Router *> router_{nullptr};
```
n-hutton, line 118: :)

libs/muddle/src/direct_message_service.cpp
```diff
@@ -130,6 +130,23 @@ void DirectMessageService::RequestDisconnect(Handle handle)
   SendMessageToConnection(handle, response);
 }
 
+void DirectMessageService::SignalConnectionLeft(Handle handle)
+{
+  FETCH_LOCK(lock_);
+
+  for (auto it = reservations_.begin(); it != reservations_.end();)
```
n-hutton, line 137: Just curious about this - what are reservations and why do multiple map to a single handle?

n-hutton (approved):
LGTM

libs/muddle/src/direct_message_service.cpp
ejfitzgerald: Reservations are important in the case where you are in the process of establishing a connection to someone. You need to record that you are already in the process of setting up a connection.

As for the reservation is never expected that there isn't a 1:1 mapping between address and handle, however, for completeness I do not terminate the search to ensure that this is the case. In the future we might support having multi path routing (UDP and TCP say) in which case this would be invalid. In any case it is expected that this map will be relatively small in size for the redundant iterations should not be too burdensome

PR #1741, opened 2019-10-04
bugfix MSE

baykaner:


MissingNO57 (dismissed):
LGTM, this is fast implementation and can be only used on last layer. We should still think about "real loss", where only difference now would be:
return {return_signal, -return_signal};

libs/ml/tests/ml/ops/loss_functions/mean_square_error_loss.cpp
```diff
@@ -68,7 +68,7 @@ TYPED_TEST(MeanSquareErrorTest, one_by_eight_dimensional_forward_test)
       {std::make_shared<TypeParam>(data1_transpose), std::make_shared<TypeParam>(data2_transpose)},
       result);
 
-  ASSERT_FLOAT_EQ(static_cast<float>(result(0, 0)), 191.18f / 8.0f / 2.0f);
+  ASSERT_FLOAT_EQ(static_cast<float>(result(0, 0)), 191.18f / 8.0f);
   // fetch::math::MeanSquareErrorLoss divided sum by number of element (ie 8 in this case)
   // and then further divide by two (cf issue 343)
```
evsmithx: Comment now obsolete

evsmithx (changes requested):

MissingNO57 (dismissed):
LGTM

MissingNO57 (approved):
LGTM

evsmithx (approved):
LGTM

PR #1745, opened 2019-10-04
Static analysis loop counter fix

cyenyxe:
Enabling a couple of static analysis checks causes the build to fail due to the use of floating-point numbers used as loop counters. This PR replaces all those with integer values, following the style of each of the existing sources as closely as possible (thus why there are different types).


ejfitzgerald (dismissed):

libs/math/tests/math/all_tests/serializers/fixed_point.cpp
```diff
@@ -41,16 +43,21 @@ TEST(FixedPointSerialisationTest, IntegerSerialisation)
 
 TEST(FixedPointSerialisationTest, DecimalSerialisation)
 {
-  for (float i(-10); i < 10; i += 0.12345f)
+  std::random_device               random_device;
```
troels-im, line 46: Out of curiosity, why not just change the for loop to a while:
```
float i = -10;
while(i < 10.f)
{
// ...
i += 0.12345f;
}
```
It would be a more natural fix to the problem as it would not change the logic of the test.

cyenyxe: Because that would have the same precision issues as the for loop, it's just that the checker may not be able to catch it...

n-hutton (approved):

PR #1746, opened 2019-10-04
Bug fix with entropy generation and benchmark

jinmannwong:
Change benchmark to take in cabinet size and number of cabinet members online. Also, add own members signature when a message is signed.


ejfitzgerald (dismissed):

n-hutton (approved):
LGTM

PR #1749, opened 2019-10-06
Feature/consensus tickets

n-hutton:
- Timestamps are now enforced by consensus
- Basic miner timing enforcement is in, needs discussion
- Committee shuffle is done based on qual, and not the stakers
- Tried to consolodate time, but needs some discussion


libs/ledger/src/consensus/consensus.cpp
```diff
@@ -151,49 +171,97 @@ uint64_t Consensus::GetBlockGenerationWeight(Block const &previous, Address cons
...
 
-  // At this point the miner will decide if they should produce a block. The first miner in the
-  // committee will wait until block_interval after the block at the HEAD of the chain, the second
-  // miner 2*block_interval and so on.
-  uint32_t time_to_wait = block_interval_ms_;
-  bool     in_committee = false;
+  // TODO(HUT): check block has been signed correctly
 
-  bool found = false;
+  // Time slot protocol: within the block period, only the heaviest weighted miner may produce a
```
n-hutton, line 226: Basic algorithm for block scheduling is here

libs/moment/include/moment/clock_interfaces.hpp
```diff
@@ -30,8 +30,10 @@ class ClockInterface
 {
 public:
   using ChronoClock = std::chrono::steady_clock;
+  using SystemClock = std::chrono::system_clock;
```
n-hutton, line 33: Turns out my way of doing it in this way does not work well - one way around this might be to template the clock interface on steady vs. system clock although the solution might look ugly.

libs/moment/include/moment/detail/steady_clock.hpp
```diff
@@ -40,6 +41,12 @@ class SteadyClock : public ClockInterface
   {
     return ChronoClock::now();
   }
+
+  Duration TimeSinceEpoch() const override
```
n-hutton: Not the solution intended to go into master here.

jinmannwong (commented):
Beacon related files and the ValidBlockTiming look good.

libs/moment/include/moment/clock_interfaces.hpp
```diff
@@ -45,8 +49,8 @@ class ClockInterface
    *
    * @return The current timestamp
    */
-  virtual Timestamp Now() const = 0;
-
+  virtual TimestampChrono NowChrono() const = 0;
```
ejfitzgerald, line 52: I think at least this should be NowSteady, but I think 5e interface needs some updates

ejfitzgerald: Let’s have a chat

ejfitzgerald (changes requested):
All looks good but I 5ink that we need to have a think about the new interface to the moment library

libs/moment/include/moment/clock_interfaces.hpp
n-hutton: Yes, let's

ejfitzgerald (approved):
As discussed lets land and make subsequent updates

PR #1752, opened 2019-10-07
W2V modifications for graph generation

MissingNO57:


libs/ml/examples/word2vec_distributed_learning/main.cpp
```diff
@@ -129,6 +130,7 @@ int main(int argc, char **argv)
     cp.data                        = {client_data[i]};
     auto client =
         std::make_shared<Word2VecClient<TensorType>>(std::to_string(i), cp, console_mutex_ptr);
+    client->SetMaxUpdates(iterations_count);
     // TODO(1597): Replace ID with something more sensible
```
evsmithx: This comment isn't in the right place any more! I think we decided not to do anything about this issue anyway, so maybe we can remove it?

libs/ml/examples/word2vec_distributed_learning/main.cpp
```diff
@@ -172,6 +174,24 @@ int main(int argc, char **argv)
       t.join();
     }
 
+    std::ofstream lossfile("losses.csv", std::ofstream::out | std::ofstream::app);
```
evsmithx: It would be good if this was a commandline arg.

evsmithx (changes requested):

evsmithx (dismissed):

libs/ml/examples/word2vec_SGNS_dataloader/main.cpp
```diff
@@ -16,6 +16,7 @@
 //
 //------------------------------------------------------------------------------
 
+#include "../../src/utilities/word2vec_utilities.cpp"
```
evsmithx: Shouldn't you still be including the .hpp file not the .cpp?

libs/ml/examples/word2vec_distributed_learning/word2vec_client.hpp
```diff
@@ -57,8 +57,27 @@ class Word2VecClient : public TrainingClient<TensorType>
...
+  DataType GetAnalogyScore()
+  {
+    // first get hold of the skipgram layer by searching the return name in the graph
+    std::shared_ptr<fetch::ml::layers::SkipGram<TensorType>> sg_layer =
+        std::dynamic_pointer_cast<fetch::ml::layers::SkipGram<TensorType>>(
+            (this->g_ptr_->GetNode(skipgram_))->GetOp());
+
+    // next get hold of the embeddings
+    std::shared_ptr<fetch::ml::ops::Embeddings<TensorType>> embeddings =
+        sg_layer->GetEmbeddings(sg_layer);
```
evsmithx: There's now utilities::GetEmbeddings for this

libs/ml/examples/word2vec_graph_evaluation/main.cpp
```diff
@@ -16,13 +16,13 @@
 //
 //------------------------------------------------------------------------------
 
+#include "../../src/utilities/word2vec_utilities.cpp"
```
evsmithx: Same here, include the .hpp file

libs/ml/include/ml/distributed_learning/distributed_learning_client.hpp
```diff
@@ -17,6 +17,7 @@
 //
 //------------------------------------------------------------------------------
 
+#include "../../../src/utilities/word2vec_utilities.cpp"
```
evsmithx: hpp

libs/ml/src/utilities/word2vec_utilities.cpp
```diff
@@ -0,0 +1,45 @@
...
+//------------------------------------------------------------------------------
+
+#include "ml/utilities/word2vec_utilities.hpp"
+
+namespace fetch {
+namespace ml {
+namespace utilities {
+
+// Timestamp for logging
+std::string GetStrTimestamp()
```
evsmithx, line 26: This should be declared in the header file

evsmithx (changes requested):

libs/ml/src/utilities/word2vec_utilities.cpp
MissingNO57: It can't be, because it is not templated so it wouldn't pass static analysis.

evsmithx (approved):

PR #1753, opened 2019-10-07
Node and subgraph backprop

baykaner:
- reworked node and subgraph backprop functions to use maps instead of vectors. Result is code that is slightly clearer and slightly more performant according to ML training benchmark:

Screenshot 2019-10-07 at 12 04 36
Screenshot 2019-10-07 at 12 05 42


evsmithx (dismissed):
LGTM

evsmithx (approved):
LGTM

PR #1756, opened 2019-10-08
add initial dmlf classes

engineerkatie:


libs/dmlf/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+target_link_libraries(fetch-dmlf
+                      PUBLIC fetch-core
+                             fetch-ml
+                             fetch-network
+                             fetch-muddle
+                             fetch-crypto
+                             fetch-math
+                             fetch-vm
+                             fetch-vm-modules
+                             pthread)
```
ejfitzgerald: Was this needed? I think that it should already with `fetch-core`

libs/dmlf/include/dmlf/queue_interface.hpp
```diff
@@ -0,0 +1,47 @@
...
+public:
+  using Bytes = byte_array::ByteArray;
+
+  QueueInterface()          = default;
+  virtual ~QueueInterface() = default;
+
+  virtual void        PushNewMessage(Bytes msg) = 0;
+  virtual std::size_t size() const              = 0;
+
+  QueueInterface(const QueueInterface &other) = delete;
```
ejfitzgerald: Style 🚓 - Would normally keep all constructors together.

libs/dmlf/include/dmlf/type_map.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <exception>
+#include <iostream>
+#include <typeindex>
+#include <typeinfo>
+#include <unordered_map>
+
+namespace fetch {
+namespace dmlf {
+
+template <typename ValueType = std::string>
```
ejfitzgerald, line 30: Is this `ValueType` likely to change in the future, if not I would recommend that we do not make it a template at this stage and that we add `.cpp` files

libs/dmlf/include/dmlf/update_interface.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include <cstdint>
+
+namespace fetch {
+namespace dmlf {
+
+class UpdateInterface
+{
+public:
+  using TimeStampType   = std::uint64_t;
```
ejfitzgerald: Style 🚓  - In general we would not include the `Type` suffix for type aliases as it is sort of redundant. That being said the Math library is a first class offender for this (which is probably where this convention has come from)

libs/dmlf/src/local_learner_networker.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "core/serializers/base_types.hpp"
+#include "dmlf/local_learner_networker.hpp"
+#include "dmlf/update_interface.hpp"
+
+#include <iostream>
+
+namespace fetch {
+namespace dmlf {
+
+LocalLearnerNetworker::LocalLearnerNetworker() = default;
```
ejfitzgerald: Not required should just be in the header file

libs/dmlf/src/local_learner_networker.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+LocalLearnerNetworker::~LocalLearnerNetworker() = default;
```
ejfitzgerald: Same here

libs/dmlf/include/dmlf/vm_wrapper_interface.hpp
```diff
@@ -0,0 +1,66 @@
...
+{
+public:
+  VmWrapperInterface()          = default;
+  virtual ~VmWrapperInterface() = default;
+
+  using OutputHandler = std::function<void(const std::string &)>;
+  using InputHandler  = std::function<std::string(void)>;
+  using Params        = std::vector<std::string>;
+  using Flags         = std::vector<std::string>;
+  using Status        = enum {
```
ejfitzgerald:
```suggestion
  enum Status  {
```

libs/dmlf/src/local_learner_networker.cpp
```diff
@@ -0,0 +1,68 @@
...
+  for (auto ind : indexes)
+  {
+    auto t = peers_[ind];
+    t->Recieve(data);
+  }
+}
+
+void LocalLearnerNetworker::Recieve(Bytes const &data)
+{
+  AbstractLearnerNetworker::NewMessage(data);
```
ejfitzgerald: This feels like it should be part of the interface. Seems like a lot of the files ultimately call this or a variant of this function

libs/dmlf/src/local_learner_networker.cpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+void LocalLearnerNetworker::PushUpdate(const std::shared_ptr<UpdateInterface> &update)
+{
+  std::vector<std::shared_ptr<LocalLearnerNetworker>> targets;
+
+  auto indexes = alg_->GetNextOutputs();
+  auto data    = update->Serialise();
+
+  for (auto ind : indexes)
```
ejfitzgerald: Worth making the following change. I understand these are just `std::size_t`'s at the moment but this would make like easier in the future if this changes?

```suggestion
  for (auto const &ind : indexes)
```

libs/dmlf/src/muddle2_learner_networker.cpp
```diff
@@ -0,0 +1,183 @@
...
+using fetch::byte_array::ByteArray;
+using fetch::byte_array::ConstByteArray;
+using fetch::byte_array::FromBase64;
+using fetch::network::NetworkManager;
+using fetch::service::Promise;
+
+using PromiseList = std::vector<Promise>;
+using SignerPtr   = std::shared_ptr<crypto::ECDSASigner>;
+
+Muddle2LearnerNetworker::Muddle2LearnerNetworkerProtocol::Muddle2LearnerNetworkerProtocol(
```
ejfitzgerald: The naming of this class is a little curious. Was there a muddle 1 networker?

libs/dmlf/src/muddle2_learner_networker.cpp
```diff
@@ -0,0 +1,183 @@
...
+    break;
+  case MuddleChannel::MULTIPLEX:
+    AbstractLearnerNetworker::NewDmlfMessage(b);
+    break;
+  default:;
+  }
+  return 0;
+}
+
+Muddle2LearnerNetworker::~Muddle2LearnerNetworker() = default;
```
ejfitzgerald: To headerfile

libs/dmlf/src/muddle2_learner_networker.cpp
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  for (auto &prom : promises)
+  {
+    prom->Wait();
+  }
+}
+
+void Muddle2LearnerNetworker::PushUpdateType(const std::string &                     type,
+                                             const std::shared_ptr<UpdateInterface> &update)
```
ejfitzgerald: This is (`std::shared_ptr<UpdateInterface>`) is used quite a lot around the place. Is it worth making a namespace level type alias for this like `using UpdateInterfacePtr = std::shared_ptr<UpdateInterface>` for this and using it

libs/dmlf/src/muddle2_learner_networker.cpp
```diff
@@ -0,0 +1,183 @@
...
+{
+  auto client = std::make_shared<RpcClient>("Client", mud_->GetEndpoint(), 1, 1);
```
ejfitzgerald: Can we move these magic numbers to `service_ids.hpp` - you remember all the fun and games we went through last time when this was not the case

libs/dmlf/src/update_interface.cpp
```diff
@@ -0,0 +1,19 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "dmlf/update_interface.hpp"
```
ejfitzgerald: Don't need this?

libs/dmlf/CMakeLists.txt
engineerkatie: removed

libs/dmlf/include/dmlf/type_map.hpp
engineerkatie: I think the intention is that this will change -- Troels has defn mentioned hashes being used at some point.

libs/dmlf/src/muddle2_learner_networker.cpp
engineerkatie: renamed

libs/dmlf/include/dmlf/update_interface.hpp
engineerkatie: dine

libs/dmlf/src/local_learner_networker.cpp
engineerkatie: done

libs/dmlf/src/local_learner_networker.cpp
engineerkatie: done

libs/dmlf/src/local_learner_networker.cpp
engineerkatie: done

libs/dmlf/src/muddle2_learner_networker.cpp
engineerkatie: done

libs/dmlf/src/muddle2_learner_networker.cpp
engineerkatie: dine

libs/dmlf/src/local_learner_networker.cpp
engineerkatie: I've made it part iof the interface and added comments about why some of the methods are simple thunks.

libs/dmlf/include/dmlf/vm_wrapper_interface.hpp
engineerkatie: dine

libs/dmlf/include/dmlf/queue_interface.hpp
engineerkatie: done

libs/dmlf/src/muddle2_learner_networker.cpp
engineerkatie: done

libs/dmlf/include/dmlf/vm_wrapper_etch.hpp
```diff
@@ -0,0 +1,83 @@
...
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "dmlf/vm_wrapper_interface.hpp"
+
+//#include "vm/generator.hpp"
```
ejfitzgerald, line 22: Dead code

ejfitzgerald (approved):

PR #1758, opened 2019-10-08
possibly unused files

n-hutton:


libs/core/include/core/containers/mapping.hpp
```diff
@@ -1,111 +0,0 @@
-#pragma once
```
ejfitzgerald: Can we keep this please. This might not be useful currently, however, very useful in general

ejfitzgerald (changes requested):
I would like to keep the mapping an suggest @baykaner will want to keep some of the maths / ml stuff at least

libs/core/include/core/containers/mapping.hpp
n-hutton: yup

baykaner (dismissed):
don't remove the math files. I will raise a task to write tests for them.
Also, I need konstantinos to weigh in on the vectorise fiels.

👍

libs/network/include/network/muddle/rpc/server.hpp
```diff
@@ -1,154 +0,0 @@
-#pragma once
```
ejfitzgerald, line 1: When did these get reintroduced have we had a merge blunder somewhere?

libs/miner/include/miner/block_optimiser.hpp
```diff
@@ -1,457 +0,0 @@
-#pragma once
```
ejfitzgerald: Can we keep the annealer miner stuff. We have an open task to reinstate this

ejfitzgerald (changes requested):

libs/network/include/network/muddle/rpc/server.hpp
n-hutton: You may have introduced it in commit 1bc800ac

libs/miner/include/miner/block_optimiser.hpp
n-hutton: The following files have been saved from certain doom: 
libs/miner/include/miner/annealer_miner.hpp
 libs/miner/include/miner/block_optimiser.hpp
 libs/miner/include/miner/optimisation/binary_annealer.hpp

ejfitzgerald: 👍

libs/network/include/network/muddle/rpc/server.hpp
ejfitzgerald: Oh the nostalgia - that was when it originally came into `master` I was pretty sure that I binned / moved these files around with Muddle V2. Oh well you got the stragglers! 😁

libs/miner/include/miner/instance/binary_problem.hpp
```diff
@@ -1,201 +0,0 @@
-#pragma once
```
ejfitzgerald: Please can we save this one too

ejfitzgerald (changes requested):
Just one more file to save and then lets let @fetch-bot do its thang!

libs/miner/include/miner/instance/binary_problem.hpp
n-hutton: Yup done

ejfitzgerald (approved):

PR #1761, opened 2019-10-08
Variable freezing

MissingNO57:


libs/ml/include/ml/core/graph.hpp
```diff
@@ -523,6 +526,37 @@ bool Graph<TensorType>::SetRegularisation(std::string node_name, RegPtrType regu
...
+{
+  for (auto &t : trainable_lookup_)
+  {
+    auto tmp = std::dynamic_pointer_cast<ops::Trainable<TensorType>>(t.second->GetOp());
+    tmp->SetFrozenState(frozen_state);
+  }
+}
+
+/**
+ * Set variable freezing for specified trainable by it's name
```
evsmithx: its

libs/ml/include/ml/ops/variable.hpp
```diff
@@ -147,8 +154,11 @@ class Variable : public DataHolder<T>, public Trainable<T>
...
   void ApplyGradient(TensorType const &grad) override
   {
-    ApplyRegularisation();
-    this->data_->InlineAdd(grad);
+    if (!this->value_frozen_)
+    {
+      ApplyRegularisation();
+      this->data_->InlineAdd(grad);
+    }
     ResetGradients();
```
evsmithx: I think that ResetGradients should be inside the if-statement? If a variable has gradients from before it was frozen and then it is frozen and then ApplyGradients is called then the gradients from before the freeze will be deleted. I think ResetGradients should only be called if the gradients were applied?

libs/ml/tests/ml/graph/graph.cpp
```diff
@@ -233,6 +233,97 @@ TYPED_TEST(GraphTest, applying_regularisation_all_trainables)
...
+  // Create a graph with a single weights node
+  fetch::ml::Graph<TensorType> g;
+
+  std::string weights = g.template AddNode<fetch::ml::ops::Weights<TensorType>>("Weights", {});
+
+  g.SetInput(weights, data_1);
+
+  // Freeze variable
+  g.SetFrozenState(weights, true);
+  g.SetFrozenState(weights, true);
```
evsmithx: Duplicated statement?

libs/ml/include/ml/core/graph.hpp
```diff
@@ -523,6 +526,37 @@ bool Graph<TensorType>::SetRegularisation(std::string node_name, RegPtrType regu
...
 
+/**
+ * Set variable freezing for all trainables in graph
+ * @tparam TensorType
+ * @param frozen_state true=freeze variables, false=unfreeze variables
+ */
+template <typename TensorType>
+void Graph<TensorType>::SetFrozenState(bool frozen_state)
+{
+  for (auto &t : trainable_lookup_)
```
baykaner: this doesn't work for any interesting graphs. That's because trainable_lookup_ doesn't store all of the trainables; only the trainables in THIS graph.
You need to call GetTrainables which makes a recursive call to collect all the trainables.

libs/ml/include/ml/core/graph.hpp
```diff
@@ -523,6 +526,37 @@ bool Graph<TensorType>::SetRegularisation(std::string node_name, RegPtrType regu
...
+/**
+ * Set variable freezing for specified trainable by it's name
+ * @tparam TensorType
+ * @param node_name name of specific trainable
+ * @param frozen_state true=freeze variables, false=unfreeze variables
+ */
+template <typename TensorType>
+bool Graph<TensorType>::SetFrozenState(std::string node_name, bool frozen_state)
+{
+  NodePtrType t             = trainable_lookup_.at(node_name);
```
baykaner: as above

evsmithx (changes requested):

libs/ml/tests/ml/graph/graph.cpp
```diff
@@ -233,6 +233,97 @@ TYPED_TEST(GraphTest, applying_regularisation_all_trainables)
...
+  using TensorType = TypeParam;
+  using DataType   = typename TypeParam::Type;
+
+  TensorType data_1 = TensorType::FromString("1, -2, 3, -4, 5, -6, 7, -8");
+  TensorType data_2 = TensorType::FromString("3, 0, 5, -2, 7, -4, 9, -6");
+
+  // Create a graph with a single weights node
+  fetch::ml::Graph<TensorType> g;
+
+  std::string weights = g.template AddNode<fetch::ml::ops::Weights<TensorType>>("Weights", {});
```
baykaner, line 248: This test doesn't catch the problems I've highlighted above.
Add another test that makes a more realistic graph (e.g. several fully connected layers)

baykaner (changes requested):

libs/ml/include/ml/core/graph.hpp
```diff
@@ -523,6 +526,51 @@ bool Graph<TensorType>::SetRegularisation(std::string node_name, RegPtrType regu
...
+template <typename TensorType>
+bool Graph<TensorType>::SetFrozenState(std::string node_name, bool frozen_state)
+{
+  OpPtrType target_node_op = GetNode(node_name)->GetOp();
+  auto *trainable_ptr = dynamic_cast<fetch::ml::ops::Trainable<TensorType> *>(target_node_op.get());
+  auto *graph_ptr     = dynamic_cast<fetch::ml::Graph<TensorType> *>(target_node_op.get());
+
+  if (trainable_ptr)
+  {
+    // it's definitely a trainable
```
baykaner: I don't think you need this comment :-p

libs/ml/include/ml/core/graph.hpp
```diff
@@ -523,6 +526,51 @@ bool Graph<TensorType>::SetRegularisation(std::string node_name, RegPtrType regu
...
+    trainable_ptr->SetFrozenState(frozen_state);
+  }
+  else if (graph_ptr)
+  {
+    // it's definitely a graph
```
baykaner: or this

baykaner (changes requested):
one more complication:
how does frozenness interact with shared layers?
can we please have a test that demonstrates this. Presumably the desired behaviour is that freezing a shared layer automatically freezes other shared layers?

libs/ml/include/ml/core/graph.hpp
MissingNO57: I thought it would be clearer, but ok.

libs/ml/include/ml/core/graph.hpp
MissingNO57: I thought it would be clearer, but ok.

evsmithx (approved):

baykaner (approved):

PR #1763, opened 2019-10-08
Making w2v sgns dataloader templated on TensorType

evsmithx:
Making w2v sgns dataloader templated on TensorType rather than DataType, to match the rest of the ml library.


baykaner (approved):
Yup, nice. thanks for cleaning things up

PR #1764, opened 2019-10-08
Feature/add and check signatures consensus

n-hutton:


libs/core/include/core/fetch_backward.hpp
```diff
@@ -50,3 +50,18 @@
 #if defined(__GNUC__)
 #pragma GCC diagnostic pop
 #endif
+
```
n-hutton: Unrelated but might be useful

apps/constellation/constellation.cpp
```diff
@@ -121,6 +121,9 @@ uint16_t LookupLocalPort(Manifest const &manifest, ServiceIdentifier::Type servi
 std::shared_ptr<ledger::DAGInterface> GenerateDAG(std::string const &db_name, bool load_on_start,
                                                   Constellation::CertificatePtr certificate)
 {
+  FETCH_UNUSED(db_name);
```
ejfitzgerald: Why is this necessary?

libs/beacon/include/beacon/block_entropy.hpp
```diff
@@ -28,27 +28,35 @@ namespace beacon {
...
+  // The hash of the above (when new committee) note, this could be implicit. Is not
+  // serialized.
+  Digest digest;
+
+  // In the case of a new cabinet, personal signatures of the hash
+  // from qual members
+  Confirmations confirmations;
+
+  // Signature of the previous entropy, used as the entropy
+  GroupSignature group_signature = GroupSignature{};
```
ejfitzgerald:
```suggestion
  GroupSignature group_signature{};
```

ejfitzgerald: If needed at all?

libs/beacon/src/beacon_manager.cpp
```diff
@@ -644,14 +644,15 @@ bool BeaconManager::Verify(Signature const &signature)
...
  */
-bool BeaconManager::Verify(std::string const &group_public_key, MessagePayload const &message,
-                           std::string const &signature)
+bool BeaconManager::Verify(byte_array::ConstByteArray const &group_public_key,
+                           MessagePayload const &            message,
+                           byte_array::ConstByteArray const &signature)
 {
   PublicKey tmp;
-  tmp.setStr(group_public_key);
+  tmp.setStr(std::string(group_public_key));
```
ejfitzgerald, line 657:
```suggestion
  tmp.setStr(static_cast<std::string>(group_public_key));
```

ejfitzgerald: (I would prefer)

libs/beacon/src/beacon_manager.cpp
```diff
@@ -644,14 +644,15 @@ bool BeaconManager::Verify(Signature const &signature)
...
 
   Signature tmp2;
-  tmp2.setStr(signature);
+  tmp2.setStr(std::string(signature));
```
ejfitzgerald, line 660:
```suggestion
  tmp2.setStr(static_cast<std::string?(signature));
```

libs/beacon/src/beacon_service.cpp
```diff
@@ -340,8 +340,28 @@ BeaconService::State BeaconService::OnCompleteState()
   beacon_entropy_last_generated_->set(index);
   beacon_entropy_generated_total_->add(1);
 
-  // Populate the block entropy structure appropriately TODO(HUT): this.
-  // block_entropy_being_created_->group_signature = active_exe_unit_->manager.GroupSignature();
+  // Populate the block entropy structure appropriately
+  block_entropy_being_created_->group_signature =
+      active_exe_unit_->manager.GroupSignature().getStr();
+
+  std::cerr << "pre-verify 0" << std::endl;  // DELETEME_NH
```
ejfitzgerald: Delete?

libs/core/include/core/fetch_backward.hpp
ejfitzgerald: Worth adding to logging instead?

libs/crypto/tests/unit/mcl_dkg_tests.cpp
```diff
@@ -29,7 +29,7 @@ using namespace fetch::byte_array;
 
 TEST(MclTests, BaseMcl)
 {
-  fetch::crypto::mcl::details::MCLInitialiser();
+  // fetch::crypto::mcl::details::MCLInitialiser();
```
ejfitzgerald: Remove these?

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -1106,20 +1112,25 @@ BlockCoordinator::State BlockCoordinator::OnReset()
 
   if (next_block_)
   {
+    std::cerr << "wasn " << next_block_->body.block_number << std::endl;  // DELETEME_NH
```
ejfitzgerald: Remove?

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -1106,20 +1112,25 @@ BlockCoordinator::State BlockCoordinator::OnReset()
...
+    std::cerr << "wasc" << std::endl;  // DELETEME_NH
     block = current_block_.get();
   }
   else
   {
     FETCH_LOG_ERROR(LOGGING_NAME, "Unable to find a previously executed block!");
   }
 
+  // TODO(HUT): figure out where/when to do this
+  // block->UpdateDigest();
```
ejfitzgerald: Already done after collecting final merkle tree hash?

ejfitzgerald (changes requested):
Seems good to me just a couple of niggles

libs/beacon/include/beacon/block_entropy.hpp
n-hutton: A relic from when I tried to use the MCL stuff direct (a mistake). Good spot will change.

libs/beacon/src/beacon_manager.cpp
n-hutton: this is constructing a string, not casting it

libs/beacon/src/beacon_manager.cpp
n-hutton: see below

libs/core/include/core/fetch_backward.hpp
n-hutton: ok

libs/crypto/src/mcl_dkg.cpp
```diff
@@ -33,6 +33,8 @@ namespace mcl {
 
 std::atomic<bool> details::MCLInitialiser::was_initialised{false};
 
+details::MCLInitialiser execute_me_before_runtime;
```
n-hutton: Note: so long as this is linked, it should call the MCL init just before main is called. This might cause undefined behaviour if any MCL functions are called in other constructors so let's avoid this.

libs/ledger/src/chain/block_coordinator.cpp
n-hutton: Uhhh yes. I would say good spot but it was like a sore thumb

libs/ledger/src/chain/block_coordinator.cpp
n-hutton: There is a corner case when exceptions are thrown, this might not be true. Will open ticket.

libs/ledger/src/consensus/consensus.cpp
```diff
@@ -398,6 +430,37 @@ NextBlockPtr Consensus::GenerateNextBlock()
...
+
+  return fetch::crypto::Verifier::Verify(block.body.miner_id, block.body.hash,
+                                         block.miner_signature);
+}
+
+/**
+ * Given a block entropy, determine whether it has been signed off on
+ * by enough qualified stakers.
+ */
+bool Consensus::EnoughQualSigned(BlockEntropy const &block_entropy) const
```
n-hutton, line 453: Holding off on this part until your stake changes are in @ejfitzgerald

apps/constellation/constellation.cpp
n-hutton: It's not, I think I deleted it at the same time you commented ahah

libs/beacon/src/beacon_service.cpp
n-hutton: done

libs/beacon/src/dkg_output.cpp
```diff
@@ -22,7 +22,6 @@ using fetch::beacon::DkgOutput;
 
 DkgOutput::DkgOutput()
 {
-  fetch::crypto::mcl::details::MCLInitialiser();
```
n-hutton: Since we call this in constellation main it is probably safe not to include these @jinmannwong

ejfitzgerald (dismissed):

libs/beacon/src/beacon_manager.cpp
ejfitzgerald: So does the above, but I am not too phased either way

ejfitzgerald (approved):

PR #1766, opened 2019-10-09
Reset cache for shared ops

baykaner:
reworked optimiser's ApplyGradient methods so that instead of explicitly calling trainable_ptr->ApplyGradient they compute all gradients and then provide this vector to graph->ApplyGradient so that the graph can handle the shared layers appropriately


libs/ml/include/ml/core/graph.hpp
```diff
@@ -547,6 +546,10 @@ void Graph<TensorType>::ApplyGradients(std::vector<TensorType> &grad)
     auto grad_it = grad.begin();
     ApplyGradients(grad_it);
 
+    //    for (auto const &t : trainable_lookup_)
+    //    {
+    //      ResetGraphCache(false, t.second);
+    //    }
```
evsmithx: Commented code

libs/ml/include/ml/core/graph.hpp
```diff
@@ -547,6 +546,10 @@ void Graph<TensorType>::ApplyGradients(std::vector<TensorType> &grad)
...
     for (auto const &t : nodes_)
     {
       // TODO(#1554) - we should only reset the cache for trained nodes, not all nodes
```
evsmithx: Presumably this comment is now obsolete?

libs/ml/tests/ml/ops/embeddings.cpp
```diff
@@ -136,6 +136,8 @@ TYPED_TEST(EmbeddingsTest, backward)
   // Get a copy of the gradients and check that they were zeroed out after Step
   TensorType grads_copy = e.GetGradientsReferences();
 
+  std::cout << "grads_copy.ToString(): " << grads_copy.ToString() << std::endl;
```
evsmithx: Debugging code?

evsmithx (changes requested):

evsmithx (approved):
LGTM

PR #1767, opened 2019-10-09
Mark slow tests as such

wkaluza:


ejfitzgerald (approved):

libs/dmlf/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 #
-# F E T C H   D I S T R I B U T E D   M L  F R A M E W O R K   L I B R A R Y
+# F E T C H   D I S T R I B U T E D   M L   F R A M E W O R K   L I B R A R Y
```
n-hutton, line 2: Extra space here for some reason - not that it matters ofc

n-hutton (approved):

libs/dmlf/CMakeLists.txt
wkaluza: Yes, three spaces instead of two. It makes it consistent with all the other headers of this kind.

PR #1772, opened 2019-10-09
Add serialization to vm::Variant, only for variant of Primitive types

lrahmani:


libs/vm/include/vm/variant.hpp
```diff
@@ -532,4 +534,210 @@ struct AnyFloatingPoint : Variant
...
+      return;
+    }
+
+    // primitive type variant
+    if (variant.IsPrimitive())
+    {
+      switch (variant.type_id)
+      {
+      case vm::TypeIds::Bool:
+      case vm::TypeIds::UInt8:
```
ejfitzgerald, line 569: Since underlying message pack serialisation only really cares about unsigned and signed types. Might make sense to reflect this in this code and case all unsigned values to `uint16_t` and all signed values to `int64_t`

libs/vm/tests/variant/serialization.cpp
```diff
@@ -0,0 +1,138 @@
...
+  Reset();
+  primitive.Set(f64);
+  SerializeAs(Float64);
+  ASSERT_EQ(variant_out.Get<double>(), f64);
+
+  Reset();
+  primitive.Set(f64);
+  SerializeAs(Fixed64);
+  ASSERT_EQ(variant_out.Get<double>(), f64);
+}
```
ejfitzgerald, line 176: Worth adding a test case for an object (to check that is throws)?

ejfitzgerald (changes requested):
Looks good just a minor suggestion re msgpack

libs/vm/tests/variant/serialization.cpp
lrahmani: yes, that is true, it was in the plan and forget about it

libs/vm/include/vm/variant.hpp
lrahmani: I am not sure doing so will simplify the code, as we will have to add `static_cast`s for every type in the `Primitive` field at decentralization time

libs/vm/tests/variant/serialization.cpp
```diff
@@ -0,0 +1,165 @@
...
+  }
+};
+
+TEST_F(VariantSerialization, primitive_types_serialization)
+{
+  primitive.Set(bl);
+  SerializeAs(Bool);
+  ASSERT_EQ(variant_out.Get<bool>(), bl);
+
+  Reset();
```
wkaluza: Please don't do this. This kind of fixture reuse makes tests very hard to maintain and causes more trouble than it is worth. Split this test up into suitably named individual test cases (one per primitive type). Then each one will start off with a fresh fixture, and the `Reset` function can be safely deleted.

wkaluza (changes requested):

libs/vm/tests/variant/serialization.cpp
```diff
@@ -0,0 +1,165 @@
...
+  primitive.Set(f32);
+  SerializeAs(Float32);
+  ASSERT_EQ(variant_out.Get<float>(), f32);
+
+  Reset();
+  primitive.Set(f32);
+  SerializeAs(Fixed32);
+  ASSERT_EQ(variant_out.Get<float>(), f32);
```
wkaluza: Also, shouldn't this be `Get<fixed_point::fp32_t>`, compared for equality against `fixed_point::fp32_t`? Same for Fixed64 below.

wkaluza (changes requested):

wkaluza (dismissed):

ejfitzgerald (approved):

PR #1774, opened 2019-10-09
Feature/shadow most of rest

n-hutton:


libs/math/src/math/linalg/blas/gemv_n_vector.cpp
```diff
@@ -129,7 +129,7 @@ void Blas<S, Signature(_y <= _alpha, _A, _x, _n, _beta, _y, _m),
   {
     for (j = 0; j < int(a.width()); ++j)
     {
-      Type temp = alpha * x[jx];
+      temp = alpha * x[jx];
```
ejfitzgerald, line 132: Change of behaviour?

libs/vm/CMakeLists.txt
```diff
@@ -20,3 +20,7 @@ target_link_libraries(fetch-vm
                              fetch-math
                              fetch-core
                              fetch-ledger)
+
+if (_is_clang_compiler)
+  target_compile_options(fetch-vm PRIVATE -Wshadow)
```
ejfitzgerald, line 25: It is not clear if this has been enabled for all libraries. If it has is it worth simple adding the configuration at the top level?

libs/math/src/math/linalg/blas/gemv_n_vector.cpp
n-hutton: No, it does not look like it to me?

libs/vm/CMakeLists.txt
n-hutton: That was my intention for the next PR - not all libraries are enabled which I would do next. Then think about GCC enabling (but probably not)

ejfitzgerald (approved):

PR #1776, opened 2019-10-10
Add serialization to dmfl::ExecutionResult and dmlf::ExecutionErrorMessage

lrahmani:
Note: This pr shouldn't be merged until pr #1772 is.

- all `dmlf` targets compiles `$ make -j10 $(make help | grep dmlf | awk '{print $2 }')`
- all `dmlf` tests pass

### DIFF
--- a/libs/dmlf/include/dmlf/execution_error_message.hpp
+++ b/libs/dmlf/include/dmlf/execution_error_message.hpp
@@ -73,6 +73,7 @@ public:
     return message_;
   }
 
+protected:
 private:
   Stage       stage_;
   Code        code_;
@@ -97,9 +98,8 @@ public:
   template <typename Constructor>
   static void Serialize(Constructor &map_constructor, Type const &exec_err_msg)
   {
-    uint32_t stage, code;
-    stage = static_cast<uint32_t>(exec_err_msg.stage_);
-    code  = static_cast<uint32_t>(exec_err_msg.code_);
+    uint32_t const stage = static_cast<uint32_t>(exec_err_msg.stage_);
+    uint32_t const code  = static_cast<uint32_t>(exec_err_msg.code_);
 
     auto map = map_constructor(3);
     map.Append(STAGE, stage);
--- a/libs/dmlf/include/dmlf/execution_error_message.hpp
+++ b/libs/dmlf/include/dmlf/execution_error_message.hpp
@@ -17,6 +17,9 @@
 //
 //------------------------------------------------------------------------------
 
+#include "core/serializers/base_types.hpp"
+#include "core/serializers/main_serializer.hpp"
+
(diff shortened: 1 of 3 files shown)


libs/dmlf/include/dmlf/execution_error_message.hpp
```diff
@@ -72,4 +80,45 @@ class ExecutionErrorMessage
...
+    auto map = map_constructor(3);
+    map.Append(STAGE, stage);
+    map.Append(CODE, code);
+    map.Append(MESSAGE, exec_err_msg.message_);
+  }
+
+  template <typename MapDeserializer>
+  static void Deserialize(MapDeserializer &map, Type &exec_err_msg)
+  {
+    uint32_t stage, code;
```
ejfitzgerald, line 112: Nothing particularly wrong here but I would always initialise all variables, just in case the code gets updated in the future

libs/dmlf/include/dmlf/execution_error_message.hpp
```diff
@@ -72,4 +80,45 @@ class ExecutionErrorMessage
...
+
+  static uint8_t const STAGE   = 1;
+  static uint8_t const CODE    = 2;
+  static uint8_t const MESSAGE = 3;
+
+  template <typename Constructor>
+  static void Serialize(Constructor &map_constructor, Type const &exec_err_msg)
+  {
+    uint32_t stage, code;
+    stage = static_cast<uint32_t>(exec_err_msg.stage_);
```
ejfitzgerald: Just a personal thing too, I would normally either move the casting to the serialiser call site or I would do something like the following:

```c++
    uint32_t const stage = static_cast<uint32_t>(exec_err_msg.stage_);
```

libs/dmlf/include/dmlf/execution_error_message.hpp
```diff
@@ -17,23 +17,29 @@
...
+#include "core/serializers/main_serializer.hpp"
+
 #include <string>
 
 namespace fetch {
 namespace dmlf {
 
 class ExecutionErrorMessage
 {
+  template <typename T, typename D>
```
ejfitzgerald: Again just a minor style thing. The general project convention is to in general just keep three sections in a class:

- public
- protected
- private

ejfitzgerald:
Just minor comments @lrahmani feel free to take or leave them 😁

libs/dmlf/include/dmlf/execution_error_message.hpp
lrahmani: considered all of the comments :+1:

ejfitzgerald (dismissed):

ejfitzgerald (approved):

PR #1778, opened 2019-10-10
Alpha TX Syncing Issues

ejfitzgerald:
- Remove "timeout" processing for sub-trees
- Fix thready safety issue around last processed block hash
- Add additional monitoring telemetry


libs/ledger/src/storage_unit/transaction_store_sync_protocol.cpp
```diff
@@ -132,6 +141,8 @@ uint64_t TransactionStoreSyncProtocol::ObjectCount()
 TransactionStoreSyncProtocol::TxArray TransactionStoreSyncProtocol::PullSubtree(
     byte_array::ConstByteArray const &rid, uint64_t bit_count)
 {
+  telemetry::FunctionTimer telemetry_timer{*pull_subtree_histogram_};
```
n-hutton, line 144: what does this function do? I would like to understand this telemetry histogram stuff

n-hutton (dismissed):
Seems like this is mainly debug/telemetry right?

n-hutton:
Fix thready safety issue around last processed block hash - can you elaborate on this?

ejfitzgerald:
> Fix thready safety issue around last processed block hash - can you elaborate on this?

Basically the `last_executed_block_hash_` in the execution manager was not protected with a mutex even though it was accessed across 3 different threads. It is now protected by a sync state variable

libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
```diff
@@ -274,67 +312,77 @@ TransactionStoreSyncService::State TransactionStoreSyncService::OnResolvingSubtr
...
-      auto pending = pending_subtree_.GetPending();
-      for (auto &req : pending)
-      {
-        roots_to_sync_.push(promise_id_to_roots_[req.second.id()]);
-      }
+      // we are waiting for the in flight sub tree requests to terminate
+      state_machine_->Delay(10ms);
+      return State::RESOLVING_SUBTREE;
     }
   }
```
pbukva: I think it would be better to avoid duplication of the logic and simply do:
```cpp
bool const is_subtree_sync_complete = roots_to_sync_.empty() && counts.pending == 0
if (!is_subtree_sync_complete)
{
  state_machine_->Delay(10ms);
  return State::QUERY_SUBTREE;
}
```

That will simply execute `OnQuerySubtree()` where all of the logic is already implemented.

libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
```diff
@@ -234,6 +265,9 @@ TransactionStoreSyncService::State TransactionStoreSyncService::OnQuerySubtree()
 
     promise_id_to_roots_[promise.id()] = root;
     pending_subtree_.Add(root, promise);
+
+    subtree_requests_total_->increment();
+    ++requests_made;
```
pbukva, line 270: I'm putting comment on this line only because gihub review does not allow me to put it on unchanged code.
The comment is about the `FETCH_LOG_INFO(...)` call bellow, which should be rather relocated to `OnResolvingSubtree()` method. Reason beying that we should print this when we actually received data for subtrees, as opposed to current situation where the print is done a bit prematurely when we send requests (thus it is missleading).

libs/ledger/src/storage_unit/transaction_store_sync_protocol.cpp
ejfitzgerald: This is the equivalent of MilliTimer for telemetry. It records start and end time of function and submits.the value to this histogram metric. The histogram metric will then store the value on one of its configured buckets and them this is later samples by prom. Essentially this gives us sample points over time to see how long this function is taking

pbukva (changes requested):
as per my inline comments

libs/ledger/src/execution_manager.cpp
```diff
@@ -588,8 +591,13 @@ void ExecutionManager::MonitorThreadEntrypoint()
           FETCH_LOCK(idle_executors_lock_);
           if (!idle_executors_.empty())
           {
+            Address last_block_miner{};
+            summary_.Apply([&last_block_miner](Summary const &summary) {
```
pbukva, line 595: I think it will be better to use rather `SynchronisedState<T>::Get()` method in order to **get** the value out of the thread-safe state - code will be more readable and also computationally cheaper as well at least for `T = Summary` & **no** need to construct lambda with **capture**.

```cpp
Address last_block_miner {summary_.Get().last_block_miner};
```

This comment would apply to all **get** operations from `summary_` state.

libs/ledger/src/storage_unit/transaction_store_sync_protocol.cpp
n-hutton: awesome

libs/ledger/src/execution_manager.cpp
n-hutton: Minor note: I prefer ```=``` to construction to indicate the type returned from Get is an Address

ejfitzgerald: So I deliberately did it this way because `.Get()` has been removed from the `master` equivalent. Therefore I was planning on making my life a little easier in the future merge

pbukva (approved):
lgtm

PR #1781, opened 2019-10-11
Squeeze op with tests

MissingNO57:


libs/ml/include/ml/ops/squeeze.hpp
```diff
@@ -0,0 +1,129 @@
...
+  {
+    assert(inputs.size() == 1);
+    assert(output.shape() == this->ComputeOutputShape(inputs));
+
+    output.Copy((*inputs.at(0)));
+    output.Squeeze();
+  }
+
+  /**
+   * elementwise Squeeze gradient is 1/x * error:
```
baykaner: if this is true, where is the division by x happening?

libs/ml/include/ml/ops/squeeze.hpp
```diff
@@ -0,0 +1,129 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "math/standard_functions/log.hpp"
```
baykaner: do you need this?

libs/ml/tests/ml/ops/squeeze.cpp
```diff
@@ -0,0 +1,182 @@
...
+  TensorType data = TensorType::FromString("1, -2, 4, -10, 100");
+  data.Reshape({1, 5});
+  TensorType error = TensorType::FromString("1, 1, 1, 2, 0");
+  error.Reshape({5});
+
+  auto shape = error.shape();
+
+  fetch::ml::ops::Squeeze<TypeParam> op;
+
+  std::vector<TensorType> prediction =
```
baykaner: pedantic point but this is an error_signal, not a prediction

libs/ml/tests/ml/ops/squeeze.cpp
```diff
@@ -0,0 +1,182 @@
...
+  TensorType new_prediction(op.ComputeOutputShape({std::make_shared<const TensorType>(data)}));
+  new_op.Forward(vec_data, new_prediction);
+
+  // test correct values
+  EXPECT_TRUE(
+      new_prediction.AllClose(prediction, static_cast<DataType>(0), static_cast<DataType>(0)));
+  fetch::math::state_clear<DataType>();
+}
+
+TYPED_TEST(SqueezeTest, saveparams_backward_test)
```
baykaner, line 189: please add a test that also adds the op to a graph and then serialises and deserialises the graph.

I suspect it will fail because you will need to add a little bit to the graph_builder code

baykaner (changes requested):

libs/ml/include/ml/ops/squeeze.hpp
MissingNO57: Copy/paste error

libs/ml/tests/ml/ops/squeeze.cpp
MissingNO57: You were right, it's fixed now.

libs/ml/tests/ml/ops/squeeze.cpp
```diff
@@ -0,0 +1,227 @@
...
+{
+};
+
+using BothTypes = ::testing::Types<fetch::math::Tensor<fetch::fixed_point::fp32_t>,
+                                   fetch::math::Tensor<fetch::fixed_point::fp64_t>,
+                                   fetch::math::Tensor<float>, fetch::math::Tensor<double>>;
+
+TYPED_TEST_CASE(SqueezeTest, BothTypes);
+
+TYPED_TEST(SqueezeTest, forward_test)
```
baykaner: I should have mentioned this before. It would be nice to extend this test slightly to demonstrate that squeeze actually removes the first 1 dimensional axis encountered starting with the trailing dimension.
So, if your original tensor is shape {1, 6, 1}; squeezing once gives {1, 6}, then squeezing again gives {6}, and squeezing a third time will throw.

baykaner (changes requested):
sorry, just one more small thing!

libs/ml/include/ml/ops/squeeze.hpp
```diff
@@ -0,0 +1,128 @@
...
+  {
+    FETCH_UNUSED(me);
+    assert(me.get() == this);
+
+    auto copyshare = std::make_shared<MyType>(*this);  // calls default copy constructor of MyType
+
+    return copyshare;
+  }
+  /**
+   * Squeeze removes dimensions with size 1
```
baykaner: this is a vague and misleading description.
Squeeze removes the first size 1 dimension encountered starting at the trailing edge. If no dimensions are size 1, it will throw.
It might be worth changing this latter behaviour in the op, relative to the maths. In the maths maybe we want to throw if we can't squeeze, whereas in the ML we might want to do nothing if we can't squeeze. For now we don't have to change behaviour, but this comment could use a little more detail

baykaner (dismissed):

baykaner (approved):

PR #1784, opened 2019-10-11
Centralise serialisation of mcl types

jinmannwong:
Define serialisation/deserialisation of mcl types in mcl_dkg with exception throw if the string can not be mapped to the appropriate. Note, mcl types in block entropy have been left as strings as comparision operator required for map keys have not been defined for mcl public key - own definition can be added in a later PR.


libs/beacon/src/beacon_manager.cpp
```diff
@@ -88,7 +88,7 @@ std::vector<BeaconManager::Coefficient> BeaconManager::GetCoefficients()
   std::vector<Coefficient> coefficients;
   for (uint32_t k = 0; k <= polynomial_degree_; k++)
   {
-    coefficients.push_back(C_ik[cabinet_index_][k].getStr());
+    coefficients.push_back(C_ik[cabinet_index_][k]);
```
ejfitzgerald:
```suggestion
    coefficients.emplace_back(C_ik[cabinet_index_][k]);
```

ejfitzgerald (changes requested):
Just a minor suggestion, but otherwise looks good to me 😁

ejfitzgerald (dismissed):

libs/beacon/src/beacon_manager.cpp
```diff
@@ -654,10 +596,20 @@ bool BeaconManager::Verify(byte_array::ConstByteArray const &group_public_key,
   }
 
   PublicKey tmp;
-  tmp.setStr(std::string(group_public_key));
-
   Signature tmp2;
-  tmp2.setStr(std::string(signature));
+
```
n-hutton, line 600: 👍

libs/crypto/include/crypto/mcl_dkg.hpp
```diff
@@ -152,4 +153,96 @@ std::vector<DkgKeyInformation> TrustedDealerGenerateKeys(uint32_t committee_size
 
 }  // namespace mcl
 }  // namespace crypto
+
+namespace serializers {
+template <typename D>
+struct ArraySerializer<crypto::mcl::Signature, D>
```
n-hutton, line 159: Side note: Signature here refers to the group signature right? Is it worth renaming this type to GroupSignature?

n-hutton (approved):
LGTM

libs/crypto/include/crypto/mcl_dkg.hpp
jinmannwong: It is for signature shares and the group signature as they are both of the same type.

PR #1785, opened 2019-10-12
S{implifi|ynchroniz}ed Wrappers

bipll:
This commit simplifies classes `Protected` and `Waitable`.

Current implementation is incomplete in the following sense. In tests/unit/thread_safe_wrappers_tests.cpp there's a test named `nonconst_protect_on_nonconst_type_allows_read_and_write_access`. Its code looks as:
```cpp
    Wrapper<Type, std::mutex> protected_value(initial_value);
    
    protected_value.ApplyVoid([this](auto &payload) {
      EXPECT_FALSE(is_read_only(payload));
      
      EXPECT_EQ(payload, initial_value);
      payload = new_value;
    
      EXPECT_EQ(payload, new_value);
    });
```
It works as expected, seemingly, but if we just add a similar call next to it:
```cpp
    protected_value.Apply([this](auto &payload) {
      EXPECT_FALSE(is_read_only(payload));

      EXPECT_EQ(payload, initial_value);
      payload = new_value;

      EXPECT_EQ(payload, new_value);
      return payload;
    });
```
the test fails to compile: inside the closure, `payload` is not actually settable (see #1786 as a working example of a broken build).

This raises a minor issue: explicit invocation of `ApplyVoid` might conceal the actual constness of functions invoked and arguments passed.

### DIFF
--- a/libs/core/include/core/synchronisation/protected.hpp
+++ b/libs/core/include/core/synchronisation/protected.hpp
@@ -28,13 +28,13 @@ namespace fetch {
 template <typename T, typename M = Mutex>
 class Protected
 {
-private:
+protected:
   mutable M mutex_;
   T         payload_;
 
 public:
   template <typename... Args>
-  explicit Protected(Args &&... args);
+  explicit constexpr Protected(Args &&... args);
   ~Protected()                 = default;
   Protected(Protected const &) = delete;
   Protected(Protected &&)      = delete;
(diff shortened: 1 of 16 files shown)


ejfitzgerald (approved):
Looks good to me,  worth running by @wkaluza though

libs/core/include/core/synchronisation/protected.hpp
```diff
@@ -43,58 +43,26 @@ class Protected
   Protected &operator=(Protected &&) = delete;
 
   template <typename Handler>
-  void ApplyVoid(Handler &&handler)
+  constexpr decltype(auto) Apply(Handler &&handler)
```
wkaluza, line 46: `decltype(auto)` is not appropriate here. Use `auto` instead to avoid leaking references to the protected payload.

libs/core/include/core/synchronisation/protected.hpp
```diff
@@ -43,58 +43,26 @@ class Protected
...
-
     FETCH_LOCK(mutex_);
 
-    handler(payload_);
+    return std::forward<Handler>(handler)(payload_);
   }
 
   template <typename Handler>
-  void ApplyVoid(Handler &&handler) const
+  constexpr decltype(auto) Apply(Handler &&handler) const
```
wkaluza, line 54: Use `auto` return type.

libs/core/include/core/synchronisation/protected.hpp
```diff
@@ -28,13 +28,13 @@ namespace fetch {
 template <typename T, typename M = Mutex>
 class Protected
 {
-private:
+protected:
```
wkaluza, line 31: I'd prefer if we stuck with the original design, which uses encapsulation instead of inheritance.

libs/core/include/core/synchronisation/waitable.hpp
```diff
@@ -23,16 +23,29 @@
...
 #include <functional>
 #include <mutex>
+#include <type_traits>
 #include <utility>
 
 namespace fetch {
 
 template <typename T, typename M = std::mutex>
-class Waitable
+class Waitable : Protected<T, M>
```
wkaluza, line 32: See comment above on encapsulation vs inheritance. Please revert.

wkaluza (changes requested):
This change brings this code closer to the original design and I think it is a step in the right direction.

Two points:
* I'm wary of using inheritance for no other reason than code reuse, as it is known to result in code which is less maintainable and more difficult to reason about. There is ample literature on the subject.
* As designed, there is a risk of calls `Apply` leaking references which would make the payload accessible without adequate synchronisation. This defeats the purpose of having these wrappers in the first place.

ejfitzgerald:
Closing because stale

PR #1787, opened 2019-10-13
Sccache based CI builds

ejfitzgerald:
This PR enabled `sccache` based CI builds which adds shared caching of compiled output to the Jenkins build. This improves compile time generally, especially in the case of very similar builds.


CMakeLists.txt
```diff
@@ -76,17 +76,22 @@ endif (FETCH_DEBUG_SANITIZER)
 
 # Enable CCACHE - compiler caching to reduce build times
 if (FETCH_ENABLE_CCACHE)
+  find_program(SCCACGE_FOUND sccache)
   find_program(CCACHE_FOUND ccache)
 
-  if (CCACHE_FOUND)
+  if (SCCACGE_FOUND)
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE sccache)
+    # set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK sccache)
```
n-hutton: delete ?

n-hutton: Also, there appears to be a typo: 'sscacge'

n-hutton (changes requested):

CMakeLists.txt
ejfitzgerald: done

scripts/ci-tool.py
```diff
@@ -305,6 +307,34 @@ def test_language(build_root):
     subprocess.check_call(cmd)
 
 
+def run_sccache_server(sccache_path):
```
wkaluza, line 310: How does this behave if the server is already running? See comment below.

scripts/ci-tool.py
```diff
@@ -386,6 +426,20 @@ def main():
         fetchai_code_quality.static_analysis(
             project_root, build_root, args.fix, concurrency, args.commit, verbose=False)
 
+    if (args.build or args.lint or args.all) and sccache_path:
+        subprocess.check_call([sccache_path, '-s'])
+
+        # stop the server
+        stop_sscache_server(sccache_path)
```
wkaluza, line 433: This won't be reached if the build fails at any point.

ejfitzgerald: That is correct - this feature in only intended for CI builds currently - so the what happens when if the server crashes or fails it not really an issue because everything is containerised.

If you run this locally ultimately it will also cause a compilation failure

n-hutton (approved):

wkaluza (approved):

PR #1788, opened 2019-10-14
[minor] [dmlf]Add serialization to SourceFile(s)

lrahmani:
For dmlf, as it'll be an argument in the muddle rpc.

Please let me know if I need to move the serialization code elsewhere in the code base (for e.g. to avoid including serialization headers everywhere)


libs/vm/include/vm/common.hpp
```diff
@@ -297,4 +300,34 @@ struct SourceFile
...
+  using Type       = fetch::vm::SourceFile;
+  using DriverType = D;
+
+  static uint8_t const FILENAME = 1;
+  static uint8_t const SOURCE   = 2;
+
+  template <typename Constructor>
+  static void Serialize(Constructor &map_constructor, Type const &source_file)
+  {
+    auto map = map_constructor(3);
```
wkaluza:
```suggestion
    auto map = map_constructor(2);
```

wkaluza (changes requested):

wkaluza (dismissed):
LGTM

wkaluza (approved):

PR #1789, opened 2019-10-14
Notarisations phase 1

jinmannwong:
Notarisations set up for cabinet members. Members notarise blocks which have been verified by the block coordinator and then fetch other notarisations at a particular block height from peers. Each member creates a chain of notarised blocks.

Notarisation catch up via signature broadcast and feedback into main chain is missing.

### DIFF
--- a/libs/beacon/src/beacon_manager.cpp
+++ b/libs/beacon/src/beacon_manager.cpp
@@ -623,13 +623,25 @@ BeaconManager::AddResult BeaconManager::AddSignaturePart(Identity const & from,
   return AddResult::SUCCESS;
 }
 
+BeaconManager::Signature BeaconManager::ComputeGroupSignature(
+    std::unordered_map<MuddleAddress, Signature> sig_shares)
+{
+  std::unordered_map<uint32_t, Signature> points;
+  for (auto const &address_sig_pair : sig_shares)
+  {
+    uint32_t index = identity_to_index_[address_sig_pair.first];
+    points.insert({index, address_sig_pair.second});
+  }
+  return crypto::mcl::LagrangeInterpolation(points);
+}
+
 /**
  * @brief verifies the group signature.
  */
 bool BeaconManager::Verify()
 {
   group_signature_ = crypto::mcl::LagrangeInterpolation(signature_buffer_);
-  return Verify(group_signature_);
+  return VerifyGroupSignature(current_message_, group_signature_);
 }
 
 /**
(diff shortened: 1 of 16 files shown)


libs/beacon/src/beacon_manager.cpp
```diff
@@ -695,6 +712,20 @@ BeaconManager::SignedMessage BeaconManager::Sign()
   return smsg;
 }
 
+/**
+ * @brief signs the any message.
```
n-hutton, line 716: 'the any' ?

libs/crypto/include/crypto/mcl_dkg.hpp
```diff
@@ -152,4 +153,36 @@ std::vector<DkgKeyInformation> TrustedDealerGenerateKeys(uint32_t committee_size
 
 }  // namespace mcl
 }  // namespace crypto
+
```
n-hutton, line 156: This crosses over with your other PR right?

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -1055,6 +1064,13 @@ BlockCoordinator::State BlockCoordinator::OnProofSearch()
     // update the digest
     next_block_->UpdateDigest();
 
+    // Notify notarisation of new block
+    // TODO(JMW): Check block is within notarisation window and previous block is notarised?
+    if (notarisation_)
+    {
+      notarisation_->NotariseBlock(current_block_->body);
```
n-hutton, line 1071: Looks like this should be next_block_ here, not current_block_ - as an aside, I didn't realise that even with upgraded consensus, it would still do PoW - I would suggest to rewrite it as:

```
  // No need to hash if using BLS consensus
  bool const should_proceed = consensus_ || miner_->Mine(*next_block_, 100);

  if (should_proceed)
  {
    // update the digest
    next_block_->UpdateDigest();

    // Notify notarisation of new block
    // TODO(JMW): Check block is within notarisation window and previous block is notarised?
    if (notarisation_)
    {
      notarisation_->NotariseBlock(current_block_->body);
    }
...
```

Or I can do it in another PR

libs/ledger/src/protocols/notarisation_service.cpp
```diff
@@ -0,0 +1,392 @@
...
+  , state_machine_{std::make_shared<StateMachine>("NotarisationService", State::KEY_ROTATION,
+                                                  StateToString)}
+  , chain_{main_chain}
+{
+  // Attaching the protocol
+  rpc_server_ = std::make_shared<Server>(endpoint_, SERVICE_MAIN_CHAIN, CHANNEL_RPC);
+  rpc_server_->Add(RPC_NOTARISATION, &notarisation_protocol_);
+
+  // clang-format off
+      state_machine_->RegisterHandler(State::KEY_ROTATION, this, &NotarisationService::OnKeyRotation);
```
n-hutton, line 46: Unneccessary indent here

libs/ledger/src/protocols/notarisation_service.cpp
```diff
@@ -0,0 +1,392 @@
...
+    return;
+  }
+
+  // If first block of the aeon then qual should match
+  if (block.block_number == active_exe_unit_->aeon.round_start)
+  {
+    assert(block.block_entropy.qualified == active_exe_unit_->manager.qual());
+  }
+
+  // Determine rank of miner in qual
```
n-hutton, line 304: Since this logic also happens in consensus.cpp, it might be worth thinking about making the assertion that the BC will only update the notarisation service with ever heavier blocks

libs/ledger/src/protocols/notarisation_service.cpp
```diff
@@ -0,0 +1,392 @@
...
+  // Set highest notarised block rank for this block height
+  previous_notarisation_rank_[block.block_number] = miner_rank;
+}
+
+std::vector<std::weak_ptr<core::Runnable>> NotarisationService::GetWeakRunnables()
+{
+  return {state_machine_};
+}
+
+void NotarisationService::NewAeonExeUnit(SharedAeonExecutionUnit const &keys)
```
n-hutton, line 341: Might be worth thinking about doing a full copy of this due to thread safety. It looks safe to me currently, but if you were to do 'setmessage' in this class it would be unsafe to also be generating entropy

libs/ledger/tests/consensus/notarisation_tests.cpp
```diff
@@ -0,0 +1,232 @@
...
+  {
+    auto port_number = static_cast<uint16_t>(10000 + i);
+    nodes.emplace_back(new NotarisationNode(port_number, i));
+    cabinet.insert(nodes[i]->address());
+  }
+
+  TrustedDealer dealer(cabinet, threshold);
+
+  // Connect muddles together (localhost for this example)
+  for (uint32_t ii = 0; ii < committee_size; ii++)
```
n-hutton, line 140: ```ii``` :)

n-hutton (changes requested):

PR #1793, opened 2019-10-14
Do-notation for synchronised wrappers

bipll:
This commit turns synchronised access to `Protected`/`Waitable` values upside down by introducing `LockedRef()` method to designate areas in code to be run under locked mutex. Monadically speaking, this can be regarded as `do`-notation for synced wrappers, as opposed to `fmap`-style `Apply()` and `ApplyVoid()`.


wkaluza (changes requested):
This approach has already been explored and rejected in the run-up to PR #1289. 'LockedRefs' (or 'access handles', as I called them) are an unnecessary risk due to the potential of accidentally releasing and re-acquiring the underlying mutex, when a single acquisition-release was intended. Adapted, here is an example of such accidental misuse that was offered at the time:

```
Protected<std::vector<int>> v;

// ...

for (auto it = v.LockedRef()->begin(); it != v.LockedRef()->end(); ++it)
{
  // ...
}
```

bipll:
> This approach has already been explored and rejected in the run-up to PR #1289.

Can you give more detail please? I haven't seen much of this in 1289.

> Protected<std::vector<int>> v;
> 
> // ...
> 
> for (auto it = v.LockedRef()->begin(); it != v.LockedRef()->end(); ++it)
> {
>   // ...
> }
> ```

```cpp
class NothingIsSafe {
    using EverythingIsFlawed = Protected<int, std::mutex>;
    EverythingIsFlawed this_cant_be_helped_;
    std::function<void()> callback_of_death_;
public:
    void NeverRunThis() {
        callback_of_death_ = [this](){
            this_cant_be_helped_.ApplyVoid([](auto &/*useless_arg*/){});
        };
    }
    void NowItHangs() {
        this_cant_be_helped_.ApplyVoid([this](auto &/*forsaken*/){
            if(callback_of_death_) {
                // Any last will?
                callback_of_death_();
            }
        });
    }
};
```
I vaguely remember reading that at some point half of mutex locking in Facebook's code looked like `std::lock_guard(mutex_);`; nevertheless this can hardly be regarded as an argument against using `lock_guard` at all. The unfortunate fact that _someone_ can break something by _misusing_ a construct is not world's best rationale IMO.

Can you provide a more elaborate example of RAII mislocking? (your example is very simple, so due to its brevity could also be an example of "hardly introduced/easily caught in review".)

My idea was that `LockedRef()` should be used following very simple rules, basically either a single entry per statement in expressions, or at the beginning of a block, in place of lambda header in `Apply.*()` invocation (regard this as C++ flavor of `let`), and only that.

Note also that current implementation (with Applies) is still not complete as can be seen in #1786.

wkaluza:
> Can you give more detail please? I haven't seen much of this in 1289.

There is an early version of that approach on the branch `bak_protect` in my fork of `ledger`. It should give you an idea of the similarities between our approaches. As for your counterexample, I see a deterministic deadlock as a less severe problem than a race condition due to mislocking. They're not really comparable in terms of debugging effort. Just because no code is perfectly safe (smart pointers and integers aren't, after all) does not mean we should admit constructs that are less safe than they have to be. We found managed scoped access using RAII handles to be less safe than using `Apply`.

> Note also that current implementation (with Applies) is still not complete as can be seen in #1786.

This is an issue relevant to your other PR #1785, not this one. If you can get `Apply` to behave without requiring explicit lambda return types, that would be genuinely excellent, but `Apply` won't be made better by adding another, completely orthogonal implementation that uses `LockedRef` instead.

ejfitzgerald:
Closing because stale

PR #1796, opened 2019-10-14
Example of variable serialisations for the VM execution

troels-im:
This PR introduces an example of how to do variable (both objects and primitives) serialisation for the VM. It also introduces convenience functionality that allows automatic conversion of C++ types into Etch types if the appropriate converters are defined.


robertdickson (approved):
LGTM

PR #1799, opened 2019-10-15
Dev/serialise variants

engineerkatie:
Simply add serialisers to the variant type.

The tests exercise most of the tricky bits. I've omitted tests for the fixed-point system in case you end up having to alter that bit.


libs/variant/include/variant/variant.hpp
```diff
@@ -576,4 +584,150 @@ void Variant::IterateObject(Function const &function) const
...
+        deserializer >> k;
+        Deserialize(deserializer, v);
+        var[k] = v;
+      }
+      break;
+    }
+    default:
+    {
+      throw std::runtime_error{"Variant has unknown type."};
+    }
```
pbukva: should be relocated after the switch(...){...} block and  change `break;` to `return;`

libs/variant/include/variant/variant.hpp
```diff
@@ -203,6 +205,12 @@ class Variant
   ConstByteArray string_;                 ///< The string value of the variant
   VariantList    array_;                  ///< The array value of the variant
   VariantObject  object_;                 ///< The object value of the variant
+
+  template <typename T, typename D>
+  friend struct serializers::MapSerializer;
+
+  template <typename T, typename D>
+  friend struct serializers::ForwardSerializer;
```
pbukva: This is not necessary - public interface of the Variant type offers all what is needed to do (de)serialisation with zero (or virtually zero) performance impact when compiler optimisation enabled in release mode.

libs/variant/include/variant/variant.hpp
```diff
@@ -576,4 +584,150 @@ void Variant::IterateObject(Function const &function) const
...
+    }
+    case Type::Type::STRING:
+    {
+      serializer << var.string_;
+      break;
+    }
+    default:
+    {
+      throw std::runtime_error{"Variant has unknown type."};
+    }
```
pbukva: should be relocated after the switch(...){...} block and change break; to return;

libs/variant/include/variant/variant.hpp
```diff
@@ -576,4 +584,167 @@ void Variant::IterateObject(Function const &function) const
...
+      return;
+    }
+
+    case Type::Type::NULL_VALUE:
+    {
+      return;
+    }
+    case Type::Type::INTEGER:
+    {
+      serializer << var.As<int>();
```
pbukva: `var.As<int64_t>()`

libs/variant/include/variant/variant.hpp
```diff
@@ -576,4 +584,167 @@ void Variant::IterateObject(Function const &function) const
...
+      return;
+    }
+    case Type::Type::FLOATING_POINT:
+    {
+      serializer << var.As<double>();
+      return;
+    }
+    case Type::Type::FIXED_POINT:
+    {
+      serializer << var.As<int>();
```
pbukva: either `var.As<fixed_point::fp64_t>()` (preferably) or potentially `var.As<uint64_t>`

libs/variant/include/variant/variant.hpp
```diff
@@ -576,4 +584,167 @@ void Variant::IterateObject(Function const &function) const
...
+      return;
+    }
+    case Type::Type::BOOLEAN:
+    {
+      serializer << var.As<bool>();
+      return;
+    }
+    case Type::Type::STRING:
+    {
+      serializer << var.As<std::string>();
```
pbukva: std::string will work, but it would be better to use byte_array::ConstByteArray (no performance hit caused by conversion string -> byte aray)

libs/variant/include/variant/variant.hpp
```diff
@@ -576,4 +584,167 @@ void Variant::IterateObject(Function const &function) const
...
+  static void Deserialize(Deserializer &deserializer, Type &var)
+  {
+    int typecode;
+    deserializer >> typecode;
+    auto type = static_cast<Type::Type>(typecode);
+    switch (type)
+    {
+    case Type::Type::UNDEFINED:
+      {
+        var = fetch::variant::Variant::Undefined();
```
pbukva: `fetch::variant::Variant` an be reduced to just `Variant` here and other places

libs/variant/include/variant/variant.hpp
```diff
@@ -576,4 +584,167 @@ void Variant::IterateObject(Function const &function) const
...
+    case Type::Type::FLOATING_POINT:
+      {
+        double tmp;
+        deserializer >> tmp;
+        var = tmp;
+        return;
+      }
+    case Type::Type::FIXED_POINT:
+      {
+        int64_t tmp;
```
pbukva: I would suggest to use the `fixed_point::fp64_t tmp;`, There will be no performance hit, since all what `FixedPoint<32,32>` class contains is just `uint64_t` as data member.

libs/variant/include/variant/variant.hpp
```diff
@@ -576,4 +584,167 @@ void Variant::IterateObject(Function const &function) const
...
+    case Type::Type::BOOLEAN:
+      {
+        bool tmp;
+        deserializer >> tmp;
+        var = tmp;
+        return;
+      }
+    case Type::Type::STRING:
+      {
+        std::string tmp;
```
pbukva: std::string -> ConstByteArray

libs/variant/include/variant/variant.hpp
```diff
@@ -576,4 +584,167 @@ void Variant::IterateObject(Function const &function) const
...
+    {
+      uint64_t    count;
+      deserializer >> count;
+      var = fetch::variant::Variant::Array(count);
+      for (std::size_t i = 0; i < count; i++)
+      {
+        Type v;
+        Deserialize(deserializer, v);
+        var[i]=v;
+      }
```
pbukva:
```suggestion
      {
        Deserialize(deserializer, var[i]);
      }
```

libs/variant/include/variant/variant.hpp
```diff
@@ -576,4 +584,167 @@ void Variant::IterateObject(Function const &function) const
...
+      return;
+    }
+    case Type::Type::OBJECT:
+    {
+      uint64_t    count;
+      deserializer >> count;
+      var = fetch::variant::Variant::Object();
+      for (uint64_t i = 0; i < count; i++)
```
pbukva, line 729: Just comment: uint64_t is alright, but std::size_t is standard type used for indexing, and it is also type used in `Variant::operator[](std::size_t)`

libs/variant/include/variant/variant.hpp
```diff
@@ -576,4 +584,167 @@ void Variant::IterateObject(Function const &function) const
...
+      {
+        Type                       v;
+        byte_array::ConstByteArray k;
+        deserializer >> k;
+        Deserialize(deserializer, v);
+        var[k] = v;
+      }
```
pbukva:
```suggestion
     {
        byte_array::ConstByteArray k;
        deserializer >> k;
        Deserialize(deserializer, var[k]);
      }
```

libs/variant/include/variant/variant.hpp
engineerkatie: done

libs/variant/include/variant/variant.hpp
engineerkatie: done!

libs/variant/include/variant/variant.hpp
engineerkatie: done

libs/variant/include/variant/variant.hpp
engineerkatie: done

libs/variant/include/variant/variant.hpp
engineerkatie: done

libs/variant/include/variant/variant.hpp
engineerkatie: done

libs/variant/include/variant/variant.hpp
engineerkatie: Hmm. My compiler disagrees...

libs/variant/include/variant/variant.hpp
engineerkatie: done

libs/variant/include/variant/variant.hpp
engineerkatie: done

libs/variant/include/variant/variant.hpp
engineerkatie: It's mostly for the type match with the deserialised counter -- because there's no ser/des for size_t

libs/variant/include/variant/variant.hpp
engineerkatie: ok

libs/variant/include/variant/variant.hpp
engineerkatie: ok

libs/variant/include/variant/variant.hpp
engineerkatie: OK, a quick "using" makes it work.

libs/variant/include/variant/variant.hpp
```diff
@@ -591,8 +591,9 @@ template <typename D>
 struct ForwardSerializer<fetch::variant::Variant, D>
 {
 public:
-  using Type       = fetch::variant::Variant;
+  using Type       = variant::Variant;
   using DriverType = D;
+  using Variant    = fetch::variant::Variant;
```
pbukva, line 592: redundant, can be sorted later

pbukva (dismissed):
LGTM

libs/variant/include/variant/variant.hpp
```diff
@@ -596,9 +593,9 @@ struct ForwardSerializer<fetch::variant::Variant, D>
   template <typename Serializer>
   static void Serialize(Serializer &serializer, Type const &var)
   {
-    auto typecode = static_cast<int>(var.type());
+    auto typecode = static_cast<int>(var.type_);
```
pbukva: this forces friend class. The public interface method `type()` gives the same performance in optimized release build as direct access.

libs/variant/include/variant/variant.hpp
```diff
@@ -204,11 +204,8 @@ class Variant
   VariantList    array_;                  ///< The array value of the variant
   VariantObject  object_;                 ///< The object value of the variant
 
-  // template <typename T, typename D>
-  // friend struct serializers::MapSerializer;
-  //
-  // template <typename T, typename D>
-  // friend struct serializers::ForwardSerializer;
+  template <typename T, typename D>
+  friend struct serializers::ForwardSerializer;
```
pbukva: as per my comment https://github.com/fetchai/ledger/pull/1799/files/13070940bbb67988bae8ed82f74a91172d67c1d4..2a313955fd5198274a21c962d455cef7e91403c3#r335450485

pbukva (approved):
lgtm

PR #1804, opened 2019-10-15
Contract payable addresses

wkaluza:
See https://github.com/fetchai/ledger-api-py/pull/41 for corresponding SDK changes.

Contracts now have addresses of their own, composed as SHA256(concat(owner_address, nonce)). A contract can check its balance and transfer funds using new Etch free functions `balance()` and `transfer(Address, UInt64)`.

Tested using e2e tests.


libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -113,6 +116,8 @@ class Contract
   Contract &operator=(Contract const &) = delete;
   Contract &operator=(Contract &&) = delete;
 
+  void updateContractContext(ContractContext context);
```
ejfitzgerald: i guess?

```suggestion
  void UpdateContractContext(ContractContext context);
```

libs/ledger/include/ledger/upow/synergetic_contract.hpp
```diff
@@ -84,6 +85,9 @@ class SynergeticContract
   void Attach(StorageInterface &storage);
   void Detach();
 
+  void                   updateContractContext(ContractContext const &context);
```
ejfitzgerald:
```suggestion
  void                   UpdateContractContext(ContractContext const &context);
```

libs/ledger/src/chaincode/contract.cpp
```diff
@@ -229,5 +228,8 @@ Contract::TransactionHandlerMap const &Contract::transaction_handlers() const
   return transaction_handlers_;
 }
 
+Contract::Contract()  = default;
```
ejfitzgerald: Haha I guess you know my feelings on these 😁

ejfitzgerald (changes requested):

libs/ledger/src/chaincode/contract.cpp
wkaluza: This is necessary to keep the unique ptr happy. We should prefer this over using a shared ptr, given we use LTO now.

wkaluza: ok, sorted out in conversation - this is only necessary when employing a unique_ptr-to-implementation pattern

libs/ledger/include/ledger/upow/synergetic_contract.hpp
wkaluza: Done.

libs/ledger/include/ledger/chaincode/contract.hpp
wkaluza: Done.

ejfitzgerald (approved):
LGTM

PR #1806, opened 2019-10-16
Fixed incorrect use of std::numeric_limits with fixed-point

evsmithx:
For fixed-point values fetch::math::numeric_lowest<Type>() should be used.


MissingNO57 (approved):
LGTM

PR #1807, opened 2019-10-16
Dev/dmlf muddle

engineerkatie:
This adds the muddle comms system and a test.


libs/dmlf/tests/dmlf_muddle/dmlf_muddle.cpp
```diff
@@ -0,0 +1,216 @@
+//------------------------------------------------------------------------------
```
JuanBesa, line 1: We put the tests in tests/execution maybe this could also got there?

engineerkatie: There's also an argument that they should go in tests/networking...

JuanBesa (dismissed):
I don't really understand the muddle part, but it looks good to me.

libs/dmlf/include/dmlf/execution_workload.hpp
```diff
@@ -0,0 +1,62 @@
...
+    , state_name_(std::move(state_name))
+    , worker_(std::move(worker))
+  {}
+  virtual ~ExecutionWorkload() = default;
+
+  ExecutionWorkload(ExecutionWorkload const &other) = default;
+  ExecutionWorkload &operator=(ExecutionWorkload const &other) = default;
+
+  bool operator==(ExecutionWorkload const &other) = delete;
+  bool operator<(ExecutionWorkload const &other)  = delete;
```
JuanBesa, line 51: These could be deleted

JuanBesa (dismissed):

JuanBesa (approved):

PR #1809, opened 2019-10-17
Changes execution interface to use Ledge Variant instead of Vm Variant

JuanBesa:
This changes the entire execution interface. To do si it added conversion functions inside basic_vm_engine and changed the outward interfaces.


lrahmani (approved):
LGTM (as an initial version)

PR #1811, opened 2019-10-18
remove bottleneck in peer discovery

n-hutton:
I noticed that during consensus deployment there was a peer discovery issue. This fixes that, DKG + block gen now proceeds smoothly.

The issue was the connection left callback gets hit a lot while other threads are trying to acquire the mutex for simply finding out the directly connected set for example. This mutex would be held for a long time (40s+), causing the DKG state machine to hang on muddle->GetDirectlyConnectedPeers(). Some investigation needs to go into why this could delay for such a long time.

Possibly related: the periodic refresh for the peer register, how frequently does this happen? I have a suspicion it is mashing the mutex also


libs/muddle/src/muddle_register.cpp
```diff
@@ -287,7 +287,7 @@ void MuddleRegister::Enter(WeakConnectionPtr const &ptr)
  */
 void MuddleRegister::Leave(ConnectionHandle handle)
 {
-  FETCH_LOCK(lock_);
+  lock_.lock();
```
bipll: As an alternative, this can be done by enclosing `lock_.lock(); ... lock_.unlock();` part in braces, keeping `FETCH_LOCK` intact.

n-hutton: You could do this, I think it seems cleaner not to, since you would need to declare the callback_copy outside the locked scope, then assign it inside the locked scope for use outside the locked scope right?

bipll: Sure, but you would also have the guarantee lock is released no matter what, should you need one.
Or an alternative would be to use a `std::unique_lock` instead of `lock_guard`.

n-hutton: You're right, I will push this change

ejfitzgerald (approved):

PR #1814, opened 2019-10-18
Improvements/beacon setup timeouts

jinmannwong:
Changed implementation of time slots in DKG to allow for easy addition of states in the state machine. Rather than dividing total DKG time by a choice of total number of slots, a time per slot is computed from previous DKG timings, which allows addition of new states without reducing the time per slot. Also removed some redundant methods and variables.


libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -939,7 +964,6 @@ void BeaconSetupService::OnDkgMessage(MuddleAddress const &              from,
     auto ptr = std::dynamic_pointer_cast<FinalStateMessage>(msg_ptr);
     if (ptr != nullptr)
     {
-      // TODO(HUT): might be unnecessary to check given guarantee of reliable broadcast
       if (final_state_payload_.find(from) == final_state_payload_.end())
```
jinmannwong: Note on previous comment: not true, the RBC does prevent a node from broadcasting multiple messages of the same type.

libs/beacon/src/beacon_service.cpp
```diff
@@ -344,14 +344,6 @@ BeaconService::State BeaconService::OnCompleteState()
...
 
-  // Check when in debug mode that the block entropy signing has gone correctly
-  if (!dkg::BeaconManager::Verify(block_entropy_being_created_->group_public_key,
-                                  block_entropy_previous_->EntropyAsSHA256(),
-                                  block_entropy_being_created_->group_signature))
-  {
-    FETCH_LOG_WARN(LOGGING_NAME, "Failed to verify freshly signed entropy!");
-  }
-
   // Save it for querying
```
jinmannwong, line 352: This test fails on all blocks except for the block at the start of the aeon as only in this block is the public key actually set correctly. Also the group signature is already verified correctly in the state previous to using manager.Verify()

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -106,6 +108,27 @@ BeaconSetupService::BeaconSetupService(MuddleInterface &muddle, Identity identit
 
   // Set subscription for receiving shares
   shares_subscription_->SetMessageHandler(this, &BeaconSetupService::OnNewSharesPacket);
+
+  // Set time slots for each state
+  static std::once_flag flag;
```
n-hutton: I would avoid using statics where possible - I think this would be unsafe were you to run multiple beacon setup services at once, during testing perhaps. Perhaps try constexpr?

libs/beacon/tests/unit/dkg_resilience.cpp
```diff
@@ -638,7 +636,7 @@ void GenerateTest(uint32_t cabinet_size, uint32_t threshold, uint32_t qual_size,
   }
 }
 
-TEST(dkg_setup, DISABLED_bad_messages)
+TEST(dkg_setup, bad_messages)
```
n-hutton, line 639: 👍

libs/crypto/src/mcl_dkg.cpp
```diff
@@ -33,29 +33,33 @@ namespace mcl {
...
-  group_g.clear();
+  generator_g.clear();
 
-  const bn::Fp2 g("1380305877306098957770911920312855400078250832364663138573638818396353623780",
-                  "14633108267626422569982187812838828838622813723380760182609272619611213638781");
-
-  bn::mapToG2(group_g, g);
+  if (string_to_hash.empty())
+  {
+    string_to_hash = "Fetch.ai Elliptic Curve Generator G";
```
n-hutton, line 42: Oh, this is David's suggestion is it? I presume any string here would do.

libs/crypto/src/mcl_dkg.cpp
```diff
@@ -33,29 +33,33 @@ namespace mcl {
...
 
-  bn::mapToG2(group_g, g);
-  bn::mapToG2(group_h, h);
+  if (string_to_hash.empty())
+  {
+    string_to_hash = "Fetch.ai Elliptic Curve Generator G";
+  }
+  if (string_to_hash2.empty())
+  {
+    string_to_hash2 = "Fetch.ai Elliptic Curve Generator h";
```
n-hutton: small h here is inconsistent - I presume intentional?

n-hutton (changes requested):

libs/crypto/src/mcl_dkg.cpp
jinmannwong: Yes, any string will go. Just so that no one thinks we have set the values specifically so we can rig the beacon.

libs/crypto/src/mcl_dkg.cpp
jinmannwong: Haha, not at all. Just a typo, which I will fix.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: I used static here because I only wanted time_slots_in_dkg_ to be computed once, otherwise on each creation of a beacon_setup_service it would accumulate. I could alternatively make time_slots_in_dkg_ not static in beacon_setup_service_, but the removal of this would mean each beacon_setup_service would set the values in time_slot_map_ again which is unnecessary. Constexpr does not work for the std::once_flag as it needs to be non-const to be passed into std::call_once.

n-hutton: This is true, for tests there would be as many copies of this map as there are alive beacon setup services, however it is a pretty small amount of memory. You wouldn't need a call once if it was a constexpr right because by definition it would only invoke once

n-hutton (approved):
LGTM

PR #1815, opened 2019-10-18
Add methods for aggregate signatures

jinmannwong:
Add methods to mcl crypto to compute and verify aggregate signatures - required for notarisations.


libs/crypto/src/mcl_dkg.cpp
```diff
@@ -396,6 +396,147 @@ std::vector<DkgKeyInformation> TrustedDealerGenerateKeys(uint32_t committee_size
...
+  bn::Fr coefficient;
+  coefficient.clear();
+
+  // Reserve first 48 bytes for some fixed value for hygenic reuse of the
+  // hashing function
+  std::string concatenated_keys = "BLS Aggregation";
+  while (concatenated_keys.length() < 48)
+  {
+    concatenated_keys += "0";
+  }
```
ejfitzgerald: It would be worth to preallocate the `concatentated_keys` buffer to avoid expensive memory allocations

```suggestion
  std::string concatenated_keys = "BLS Aggregation";
  concatenated_keys.reserve(<insert estimation of final string>);
  while (concatenated_keys.length() < 48)
  {
    concatenated_keys.push_back('0');
  }
```

ejfitzgerald (changes requested):
Just a minor request from my side

libs/crypto/src/mcl_dkg.cpp
jinmannwong: Done!

ejfitzgerald (approved):

PR #1816, opened 2019-10-19
Fix @init return bug

wkaluza:


scripts/install-test-dependencies.sh
```diff
@@ -2,5 +2,7 @@
 set -e
 
 pip3 install --no-cache-dir --ignore-installed --force-reinstall wheel pyyaml requests
-pip3 install --no-cache-dir --ignore-installed --force-reinstall fetchai-ledger-api==0.8.0a2
+# ???
+pip3 install  --no-cache-dir --ignore-installed --force-reinstall --editable git+https://github.com/wkaluza/ledger-api-py.git@pr_tx_status#egg=fetchai-ledger-api
```
cyenyxe: Reference to personal instead of upstream repo. If this is deliberate, document why and add a TODO for future replacement.

libs/ledger/src/tx_status_http_interface.cpp
```diff
@@ -85,7 +85,7 @@ constexpr PublicTxStatus Convert(TransactionStatus       tx_processing_pipeline_
     case ContractExecutionStatus::RESOURCE_FAILURE:
     case ContractExecutionStatus::TX_LOOKUP_FAILURE:
     case ContractExecutionStatus::INEXPLICABLE_FAILURE:
```
cyenyxe, line 87: I'd find easier to understand the flow if there was a default after these, instead of a `break` and the equivalent of a default outside the `switch`.

scripts/end_to_end_test/smart_contract_tests/ledger_context.py
```diff
@@ -73,10 +78,13 @@ def run(options):
     contract = Contract(CONTRACT_TEXT)
 
     # deploy the contract to the network
-    api.sync(api.contracts.create(entity1, contract, 2000))
+    status = api.sync(api.contracts.create(entity1, contract, 2000))[0]
+
+    block_number = status.exit_code
```
cyenyxe, line 83: Rename to `block_number_state` maybe?

cyenyxe (changes requested):
A couple of minor style-related comments, but the one about `pip install` could be problematic.

libs/ledger/src/tx_status_http_interface.cpp
wkaluza: We avoid using `default` in switch statements on enums - the presence of `default` prevents most compilers from warning us if there's a value we didn't cover. It's more explicit and therefore less error-prone to be forced to enumerate all the enum values, even if some of them are no-ops.

scripts/end_to_end_test/smart_contract_tests/ledger_context.py
wkaluza: This value does not come from a state variable - it's returned directly by `@init`.

scripts/install-test-dependencies.sh
wkaluza: This is deliberate and temporary - see explanation on Slack.

cyenyxe (dismissed):

libs/http/src/http_client.cpp
```diff
@@ -116,7 +116,7 @@ bool HttpClient::Request(HTTPRequest const &request, HTTPResponse &response)
   // check the status code
   auto const raw_status_code = static_cast<uint16_t>(response.status());
 
-  return ((200 <= raw_status_code) && (300 > raw_status_code));
+  return 200 <= raw_status_code && 300 > raw_status_code;
```
ejfitzgerald: Parenthesis are free dude. I think it was clearer before the change

ejfitzgerald (changes requested):
Looks good to me, just minor nit picks

libs/http/src/http_client.cpp
wkaluza: Done.

ejfitzgerald (dismissed):

ejfitzgerald (approved):

PR #1820, opened 2019-10-20
Add DKG support to the manifest

ejfitzgerald:


libs/shards/tests/unit/manifest_parse_tests.cpp
```diff
@@ -0,0 +1,84 @@
...
+    "dkg": { "uri": "tcp://127.0.0.1:8002", "port": 50002},
+    "lanes": [
+        { "uri": "tcp://127.0.0.1:8010", "port": 50010},
+        { "uri": "tcp://127.0.0.1:8012", "port": 50012}
+    ]
+}
+)";
+
+  Manifest manifest{};
+  ASSERT_TRUE(manifest.Parse(FULL_MANIFEST));
```
wkaluza, line 46: Please move to fixture; split this into five tests, as indicated by comments.

libs/shards/tests/unit/manifest_entry_tests.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+  Address const address{"<muddle address>"};
+  entry.UpdateAddress(address);
+  EXPECT_EQ(entry.address(), address);
+}
+
+TEST(ManifestEntryTests, CheckUriWithLocalPortConstruction)
+{
+  Uri           uri{"tcp://127.0.0.1:1234"};
+  ManifestEntry entry{uri, 4321};
```
wkaluza, line 72: Curious: what use case does this represent? As in, when would we want to provide a URI with one port and then override it? in the manifest with another?

wkaluza (changes requested):
Looks ok except for test layout. Should be good to go once static analysis is passing.

libs/shards/tests/unit/manifest_entry_tests.cpp
ejfitzgerald: It is covering the use case where you bind to a local port, but announce the connection on another point

libs/shards/tests/unit/manifest_parse_tests.cpp
ejfitzgerald: Lets discuss

wkaluza (approved):

PR #1823, opened 2019-10-21
add params passing and a test

engineerkatie:


libs/dmlf/include/dmlf/execution/execution_params.hpp
```diff
@@ -0,0 +1,23 @@
...
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "variant/variant.hpp"
+#include <vector>
+
+using ExecutionParams = std::vector<fetch::variant::Variant>;
```
lrahmani: I would rather alias `Variant` than `vector<Variant>`, as the user will have to handle `Variants` anyway 

```cpp
using Parameter = fetch::variant::Variant;
using ExecutionParams = std::vector<Parameter>
```
user code can than do 

```cpp
uto p3 = client->client_->Run(SERVER_PUB, "exe1", "state1", "add", {Parameter{"hi"}, Parameter{2}});
```

Then we could make `Parameter` what ever we would it like to be, if that is the reason for having an alias in the first place

engineerkatie: I know what you mean but my thinking was that the function we're calling takes a list of parameters, so the "top level" vector represents the actual parameter list (some of which may then themselves be vectors).

engineerkatie: Oh, no, hang on, I think I've misunderstood you.

engineerkatie: Yes, I have, I see what you mean. Yes.

lrahmani: sorry, I was not very clear in my comment.
Yes, I meant that `Run` method doesn't need to abstract away the fact that parameters are a list/vector of elements. The elements could be named Variant, Parameter, Variable or anything else, it doesn't matter as long as we provide a practical (generic, flexible) interface to set and get values of different types out of them.

engineerkatie: Done!

libs/dmlf/include/dmlf/remote_execution_client.hpp.bak
```diff
@@ -0,0 +1,84 @@
+#pragma once
```
lrahmani: note: can be deleted from this web page

libs/dmlf/include/dmlf/remote_execution_host.hpp.bak
```diff
@@ -0,0 +1,83 @@
+#pragma once
```
lrahmani: note: can be deleted from this web page

libs/dmlf/include/dmlf/execution/execution_interface.hpp.bak
```diff
@@ -0,0 +1,60 @@
+#pragma once
```
lrahmani: note: can be deleted from this web page

lrahmani (commented):
lgtm

libs/dmlf/include/dmlf/remote_execution_client.hpp.bak
engineerkatie: (sigh)

Gotta add these to the gitignore...

lrahmani (approved):

PR #1827, opened 2019-10-22
Fixed Embedding bug causing 0 peers distributed w2v not to train

MissingNO57:


cyenyxe:
Is there a way we can have tests for this?

evsmithx (approved):

PR #1828, opened 2019-10-22
ReduceMean Op with tests

MissingNO57:


libs/ml/include/ml/ops/reduce_mean.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+    Broadcast(
+        [](DataType const &x, DataType const &y, DataType &z) {
+          FETCH_UNUSED(y);
+          z = x;
+        },
+        error_signal, ret_error_signal, ret_error_signal);
+
+    // Average error signal along specified axis
+    fetch::math::Divide(ret_error_signal, static_cast<DataType>(inputs.at(0)->shape().at(axis_)),
```
baykaner: why don't you include the division inside the broadcast operation? that way you can save iterating over the tensor twice can't you?

libs/ml/include/ml/utilities/graph_builder.hpp
```diff
@@ -385,6 +386,13 @@ void BuildNodeAndInsertTrainables(NodeSaveableParams<T> const &nsp, std::string
     g->AddTrainable(node, name);
     break;
   }
+  case ops::ReduceMean<T>::OpCode():
```
baykaner, line 389: nice, well remembered! 👍 

we should add a test that builds a graph with one of every type of op, serialises it, then deserialises it. That should help us avoid missing these in the future. I'll add a task on jira to this effect.

baykaner (changes requested):
one minor point of optimisation

libs/math/include/math/tensor_broadcast.hpp
```diff
@@ -172,5 +172,43 @@ bool Broadcast(F function, const Tensor<T, C> &a, const Tensor<T, C> &b, Tensor<
...
+    a_range.push_back({0, i});
+  }
+
+  for (auto &i : ret.shape())
+  {
+    ret_range.push_back({0, i});
+  }
+
+  ConstTensorSliceIterator<T, C> it_a(a, a_range);
+  TensorSliceIterator<T, C>      it_ret(ret, ret_range);
```
baykaner: most of this code is replicated in the overloaded Broadcast implementation.
Make an implementation function inside an anonymous namespace and invoke that from both Broadcast functions.
That way you don't have to repeat anything

baykaner (changes requested):

baykaner (approved):

PR #1830, opened 2019-10-22
Feature/notarisations

jinmannwong:
Notarisation keys are constructed and distributed at notarisations at the end of the DKG and added to the information in the first block of each aeon.


libs/ledger/src/protocols/notarisation_service.cpp
```diff
@@ -0,0 +1,399 @@
...
+  }
+
+  if (!notarisation_unit)
+  {
+    return;
+  }
+
+  // Determine rank of miner in qual
+  auto entropy_ranked_cabinet = Consensus::QualWeightedByEntropy(
+      notarisation_unit->notarisation_members(), block.block_entropy.EntropyAsU64());
```
jinmannwong: From previous review comment regarding having the block coordinator only send the notarisation service ever higher ranked blocks: I left this here as I don't think the block coordinator should keep track of what ranked blocks the notarisation service has notarised.

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -912,6 +916,13 @@ BlockCoordinator::State BlockCoordinator::OnPostExecBlockValidation()
       dag_->CommitEpoch(current_block_->body.dag_epoch);
     }
 
+    // Notify notarisation of new valid block
+    // TODO(JMW): Check block is within notarisation window and previous block is notarised?
+    if (notarisation_)
+    {
+      notarisation_->NotariseBlock(current_block_->body);
+    }
+
```
jinmannwong: Review comment from previous PR "Looks like this should be next_block_ here, not current_block_ - as an aside, I didn't realise that even with upgraded consensus, it would still do PoW - I would suggest to rewrite it as:" - I don't think it should be next_block here. We want to notarise the block that has just been validated and executed. Also, did not add in suggested lines as I did not understand how they fit in here.

libs/beacon/include/beacon/beacon_setup_service.hpp
```diff
@@ -75,14 +76,16 @@ class BeaconSetupService
     WAIT_FOR_RECONSTRUCTION_SHARES,
     COMPUTE_PUBLIC_SIGNATURE,
     DRY_RUN_SIGNING,
+    WAIT_FOR_NOTARISATION_KEYS,
```
n-hutton: 👍

libs/beacon/include/beacon/notarisation_manager.hpp
```diff
@@ -0,0 +1,83 @@
...
+  std::pair<Signature, std::vector<bool>> ComputeAggregateSignature(
+      std::unordered_map<MuddleAddress, Signature> const &cabinet_signatures);
+  bool VerifyAggregateSignature(MessagePayload const &                         message,
+                                std::pair<Signature, std::vector<bool>> const &aggregate_signature);
+  /// @}
+
+  /// Helper functions
+  /// @{
+  uint32_t                       Index(MuddleAddress const &member) const;
+  uint64_t                       round_start() const;
```
n-hutton: snake case variable names not our coding style (but not enforced by CI it seems :) )

libs/beacon/include/beacon/notarisation_manager.hpp
```diff
@@ -0,0 +1,83 @@
...
+  void      SetAeonDetails(uint64_t round_start, uint64_t round_end, uint32_t threshold,
+                           std::map<MuddleAddress, PublicKey> const &cabinet_public_keys);
+  /// @}
+
+  /// Construction and verification of aggregate signatures
+  /// @{
+  Signature Sign(MessagePayload const &message);
+  bool      Verify(MessagePayload const &message, Signature const &signature,
+                   MuddleAddress const &member);
+  std::pair<Signature, std::vector<bool>> ComputeAggregateSignature(
```
n-hutton: It would be nice to have these as usings

libs/beacon/include/beacon/notarisation_manager.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+private:
+  // Aeon details
+  uint64_t                                    round_start_{0};
+  uint64_t                                    round_end_{0};
+  uint32_t                                    threshold_{0};
+  std::set<MuddleAddress>                     notarisation_members_{};
+  std::unordered_map<MuddleAddress, uint32_t> identity_to_index_{};
+
+  static Generator generator_;
```
n-hutton: Neccessary to have as static?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -724,6 +768,54 @@ BeaconSetupService::State BeaconSetupService::OnDryRun()
...
 }
 
+BeaconSetupService::State BeaconSetupService::OnWaitForNotarisationKeys()
+{
+  FETCH_LOCK(mutex_);
+  beacon_dkg_state_gauge_->set(static_cast<uint64_t>(State::WAIT_FOR_NOTARISATION_KEYS));
+
+  if (state_machine_->previous_state() != State::WAIT_FOR_NOTARISATION_KEYS)
+  {
+    assert(!notarisation_manager_);
```
n-hutton: Why assert notarisation manager isn't there?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -724,6 +768,54 @@ BeaconSetupService::State BeaconSetupService::OnDryRun()
...
+    notarisation_manager_ = std::make_shared<NotarisationManager>();
+    NotarisationManager::PublicKey notarisation_public_key = notarisation_manager_->GenerateKeys();
+    ConstByteArray                 signature = certificate_->Sign(notarisation_public_key.getStr());
+
+    // Insert own signed notarisation key into entropy cabinet details
+    auto signed_notarisation_key = std::make_pair(notarisation_public_key.getStr(), signature);
+    beacon_->block_entropy.member_details.insert(
+        {certificate_->identity().identifier(), signed_notarisation_key});
```
n-hutton: Looks like we pass in certificate, and identity into the constructor. Could refactor it to only pass in the certificate, and set the identifier on construction so as to only need identity_raw_ or something like that at this point.

I do need to chat to Ed about how to detangle the identity/address palava

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -724,6 +768,54 @@ BeaconSetupService::State BeaconSetupService::OnDryRun()
...
+                                          beacon_->manager.polynomial_degree() + 1,
+                                          notarisation_keys_received_);
+
+    SetTimeToProceed(State::BEACON_READY);
+    return State::BEACON_READY;
+  }
+
+  if (timer_to_proceed_.HasExpired())
+  {
+    // TODO(JMW): If we do not have enough notarisation keys (how many is enough?) then we should
```
n-hutton: Have you asked David this?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -1159,6 +1258,33 @@ void BeaconSetupService::OnReconstructionShares(SharesMessage const &shares_msg,
...
+ * Handler for signed notarisation keys, which verifies the ECDSA signature on the message
+ *
+ * @param key_msg Pointer of NotarisationKeyMessage
+ * @param from Muddle address of sender
+ */
+void BeaconSetupService::OnNotarisationKey(NotarisationKeyMessage const &key_msg,
+                                           MuddleAddress const &         from)
+{
+  auto iter = notarisation_keys_received_.find(from);
+  if (beacon_->manager.qual().find(from) != beacon_->manager.qual().end() &&
```
n-hutton: Might be worth adding a beacon manager method so you could do ```if (beacon_->InQual(from) ...``` or something to that effect.

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -1302,12 +1428,18 @@ void BeaconSetupService::SetBeaconReadyCallback(CallbackFunction callback)
   callback_function_ = std::move(callback);
 }
 
+void BeaconSetupService::SetNotarisationCallback(NotarisationCallbackFunction callback)
```
n-hutton, line 1517: Not so clear on how this works - chat irl?

libs/beacon/src/notarisation_manager.cpp
```diff
@@ -0,0 +1,131 @@
...
+namespace fetch {
+namespace ledger {
+NotarisationManager::Generator NotarisationManager::generator_;
+
+NotarisationManager::NotarisationManager()
+{
+  static std::once_flag flag;
+
+  std::call_once(flag, []() {
+    fetch::crypto::mcl::details::MCLInitialiser();
```
n-hutton: Don't we guarantee to do this at main() or before the test runs?

libs/beacon/src/notarisation_manager.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+uint64_t NotarisationManager::round_end() const
+{
+  return round_end_;
+}
+uint32_t NotarisationManager::threshold() const
+{
+  return threshold_;
+}
+std::set<NotarisationManager::MuddleAddress> const &NotarisationManager::notarisation_members()
```
n-hutton: Style here. Also, returning a set by reference can be unsafe - have you checked usages of this?

libs/ledger/include/ledger/consensus/consensus.hpp
```diff
@@ -61,6 +61,8 @@ class Consensus final : public ConsensusInterface
   void            SetCommitteeSize(uint64_t size);
   void            SetDefaultStartTime(uint64_t default_start_time);
 
+  static WeightedQual QualWeightedByEntropy(BlockEntropy::Cabinet const &cabinet, uint64_t entropy);
```
n-hutton: 👍

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -612,6 +614,8 @@ BlockCoordinator::State BlockCoordinator::OnPreExecBlockValidation()
     }
   }
 
+  // TODO(JMW): Check notarisation is correct
```
n-hutton: :)

libs/ledger/src/chain/block_coordinator.cpp
n-hutton: this notarisation is correct. However I think I was commenting on the other one (line 1067) - in the block coord. current_block is the block you are verifying, whereas next_block is the block you are creating.

libs/ledger/src/protocols/notarisation_service.cpp
n-hutton: It wouldn't, however the block coordinator will always try to switch to ever heavier branches so this will be the net effect right?

n-hutton (changes requested):

libs/ledger/src/chain/block_coordinator.cpp
jinmannwong: Sorry, I see. That second one is actually a mistake as I misunderstood the point of the two pipes and have removed the notarisation from that part.

libs/beacon/include/beacon/notarisation_manager.hpp
jinmannwong: I have added this in a later commit.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: Not needed, it was to help me when I was moving the states around to make sure some part of notarisation manager was not being set somewhere before that I had forgotten about.

libs/ledger/src/protocols/notarisation_service.cpp
jinmannwong: The heavier weighted chain does not necessary contain a heavier ranked block for a particular block height.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: Sure, I can change the constructor.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: Sure.

libs/ledger/src/chain/block_coordinator.cpp
n-hutton: I think you would want to notarize there though, since this is your HEAD.

Probably best that the notarization is invisible to the BC, although it might be advantageous to know that you are catching up and thus do not need to notarize.

jinmannwong: Yes, you are right. You would want to notarise the new block mined.

libs/beacon/src/notarisation_manager.cpp
jinmannwong: Putting this in the constructors means that you can use the objects without knowing that the MCLInitialiser needs to be called beforehand. Seems more user friendly.

libs/beacon/include/beacon/beacon_setup_service.hpp
```diff
@@ -139,10 +148,11 @@ class BeaconSetupService
 
   /// Setup management
   /// @{
-
-  void QueueSetup(const SharedAeonExecutionUnit &beacon);
+  void StartNewCabinet(CabinetMemberList members, uint32_t threshold, uint64_t round_start,
```
n-hutton, line 151: Curious, why did you switch this away from a queue structure?

libs/beacon/include/beacon/notarisation_manager.hpp
n-hutton: bump

libs/beacon/include/beacon/notarisation_manager.hpp
```diff
@@ -0,0 +1,88 @@
...
+  bool                          CanSign() const;
+  uint64_t                      round_start() const;
+  uint64_t                      round_end() const;
+  uint32_t                      threshold() const;
+  std::set<MuddleAddress> const notarisation_members() const;
+  /// @}
+
+private:
+  // Aeon details
+  uint64_t                                    round_start_{0};
```
n-hutton, line 76: It seems like unneccessary duplication to have this here and in the aeon?

libs/beacon/include/beacon/trusted_dealer_beacon_service.hpp
```diff
@@ -26,16 +27,30 @@ namespace beacon {
...
+  using CabinetMemberList         = BeaconSetupService::CabinetMemberList;
+  using CallbackFunction          = BeaconSetupService::CallbackFunction;
+  using SharedNotarisationManager = TrustedDealer::SharedNotarisationManager;
+  using CabinetNotarisationKeys   = TrustedDealer::CabinetNotarisationKeys;
+
+  TrustedDealerSetupService(MuddleInterface &muddle, ManifestCacheInterface &manifest_cache,
+                            CertificatePtr const &certificate);
+
+  void StartNewCabinet(
+      CabinetMemberList members, uint32_t threshold, uint64_t round_start, uint64_t round_end,
```
n-hutton: We should probably refactor this to take less arguments if possible (threshold could be a class variable at least)

libs/beacon/src/beacon_complaints_manager.cpp
```diff
@@ -44,74 +44,91 @@ void ComplaintsManager::AddComplaintAgainst(MuddleAddress const &complaint_addre
...
+  // Only keep elements of complaints in cabinet
+  auto cabinet_complaints = complaints & cabinet;
+  complaints_received_.insert({from, cabinet_complaints});
 }
 
 void ComplaintsManager::Finish(std::set<MuddleAddress> const &cabinet)
 {
   FETCH_LOCK(mutex_);
-  if (!finished_)
+
```
n-hutton, line 67: Is this a bugfix?

libs/beacon/src/beacon_service.cpp
```diff
@@ -63,9 +61,8 @@ BeaconService::BeaconService(MuddleInterface &               muddle,
   , beacon_entropy_current_round_{telemetry::Registry::Instance().CreateGauge<uint64_t>(
         "beacon_entropy_current_round", "The current round attempting to generate for.")}
 {
-
   // Attaching beacon ready callback handler
-  cabinet_creator_.SetBeaconReadyCallback([this](SharedAeonExecutionUnit beacon) {
+  beacon_setup.SetBeaconReadyCallback([this](SharedAeonExecutionUnit beacon) {
```
n-hutton, line 84: Good rename. I think in a subsequent PR one of us should refactor these callbacks away. I don't really like callbacks since they break encapsulation and in this instance the wait for setup completion could simply poll a beacon setup queue instead as it's polling anyway.

libs/beacon/src/beacon_service.cpp
```diff
@@ -109,58 +106,12 @@ BeaconService::Status BeaconService::GenerateEntropy(uint64_t block_number, Bloc
...
-  beacon->aeon.block_entropy_previous    = prev_entropy;
-
-  cabinet_creator_.QueueSetup(beacon);
-}
-
 BeaconService::State BeaconService::OnWaitForSetupCompletionState()
 {
   FETCH_LOCK(mutex_);
 
+  active_exe_unit_.reset();
```
n-hutton, line 138: 👍

libs/beacon/src/beacon_service.cpp
```diff
@@ -355,9 +306,9 @@ BeaconService::State BeaconService::OnCompleteState()
   // If there is still entropy left to generate, set up and go around the loop
   if (block_entropy_being_created_->block_number < active_exe_unit_->aeon.round_end)
   {
-    block_entropy_previous_                    = std::move(block_entropy_being_created_);
-    block_entropy_being_created_               = std::make_shared<BlockEntropy>();
-    *block_entropy_being_created_              = *block_entropy_previous_;
+    block_entropy_previous_      = std::move(block_entropy_being_created_);
+    block_entropy_being_created_ = std::make_shared<BlockEntropy>();
+    block_entropy_being_created_->SelectCopy(*block_entropy_previous_);
```
n-hutton, line 362: This is better.

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -414,6 +427,41 @@ BeaconSetupService::State BeaconSetupService::OnWaitForReadyConnections()
   return State::WAIT_FOR_READY_CONNECTIONS;
 }
 
+BeaconSetupService::State BeaconSetupService::OnWaitForNotarisationKeys()
+{
+  FETCH_LOCK(mutex_);
+  beacon_dkg_state_gauge_->set(static_cast<uint64_t>(State::WAIT_FOR_NOTARISATION_KEYS));
+
+  if (!condition_to_proceed_ && valid_dkg_members_.size() == beacon_->aeon.members.size())
```
n-hutton, line 438: Let's chat tomorrow - I don't follow what is going on

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -631,7 +678,7 @@ BeaconSetupService::State BeaconSetupService::OnWaitForReconstructionShares()
       for (auto const &elem : share.second)
       {
         // Check person who's shares are being exposed is a member of qual
-        if (beacon_->manager.qual().find(elem.first) != beacon_->manager.qual().end())
+        if (beacon_->manager.InQual(elem.first))
```
n-hutton, line 684: 👍

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -755,6 +810,24 @@ BeaconSetupService::State BeaconSetupService::OnDryRun()
...
+  // Populate beacon entropy with notarisation keys of qual
+  if (notarisation_callback_function_)
+  {
+    std::map<MuddleAddress, NotarisationManager::PublicKey> qual_notarisation_keys;
+    for (auto const &member : beacon_->manager.qual())
+    {
+      assert(notarisation_key_msgs_.find(member) != notarisation_key_msgs_.end());
+      auto notarisation_msg = notarisation_key_msgs_.at(member);
+      beacon_->block_entropy.aeon_notarisation_keys.insert(
+          {member, std::make_pair(notarisation_msg.PublicKey(), notarisation_msg.Signature())});
```
n-hutton: Seems like someone can cause you to throw here on the at if they get into qual but never sent you the notarisation info right? Also, since this is added at the last stage doesn't this leave it open to the miner not populating notarisation or populating it incorrectly causing the chain to stop?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -1227,14 +1356,15 @@ void BeaconSetupService::CheckComplaintAnswers()
...
 bool BeaconSetupService::BuildQual()
 {
-  // Create set of muddle addresses
-  std::set<MuddleAddress> cabinet;
-  for (auto &member : beacon_->aeon.members)
-  {
-    cabinet.insert(member);
-  }
-  beacon_->manager.SetQual(complaint_answers_manager_.BuildQual(cabinet));
+  beacon_->manager.SetQual(complaint_answers_manager_.BuildQual(valid_dkg_members_));
```
n-hutton, line 1377: 👍

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -1227,14 +1356,15 @@ void BeaconSetupService::CheckComplaintAnswers()
...
   std::set<MuddleAddress> qual = beacon_->manager.qual();
+
+  // Check there are no members in qual that are not in valid_dkg_members
+  std::set<MuddleAddress> diff;
```
n-hutton: How could we have qual that wasn't composed of valid DKG members? Does our set helper functions not allow this to be ```auto diff = qual & valid_dkg_members_```?

libs/beacon/src/block_entropy.cpp
```diff
@@ -20,14 +20,19 @@
 
 using fetch::beacon::BlockEntropy;
 
-BlockEntropy::BlockEntropy() = default;
+BlockEntropy::BlockEntropy()
+{
+  crypto::mcl::details::MCLInitialiser();
```
n-hutton: I think it is better practice to ensure that MCL initialiser will be called before it would be used anywhere so the user doesn't have to concern themselves as to whether it has been done.

The clear here is worth thinking about whether it can be refactored away/wrapped in a convenience class - it's sort of anti-RAII to initialise other classes in your own constructor

libs/beacon/src/block_entropy.cpp
```diff
@@ -58,5 +63,10 @@ void BlockEntropy::HashSelf()
 
 bool BlockEntropy::IsAeonBeginning() const
 {
-  return !qualified.empty();
+  bool is_beginning = !confirmations.empty();
+  if (is_beginning)
+  {
+    assert(!aeon_notarisation_keys.empty());
```
n-hutton: Seems like this breaks the idea that we can choose whether we want to enable notarisation in main net. It could also be condensed to ```assert(confirmations.empty() == aeon_not...keys.empty()) ```?

libs/beacon/src/notarisation_manager.cpp
n-hutton: I think that relying on this would at least make the tests non thread safe due to static races as I mentioned the other day - happy to discuss

libs/beacon/tests/unit/dkg_resilience.cpp
```diff
@@ -564,11 +539,18 @@ void GenerateTest(uint32_t cabinet_size, uint32_t threshold, uint32_t qual_size,
...
+  // Create previous entropy
+  BlockEntropy prev_entropy;
+  prev_entropy.group_signature = "Hello";
+
   std::this_thread::sleep_for(std::chrono::milliseconds(100));
   // Reset cabinet for rbc in pre-dkg sync
   for (uint32_t ii = 0; ii < cabinet_size; ii++)
   {
-    cabinet_members[ii]->QueueCabinet(cabinet_addresses, threshold);
+    cabinet_members[ii]->StartNewCabinet(
```
n-hutton: ii :)

libs/constellation/src/constellation.cpp
```diff
@@ -267,9 +283,11 @@ Constellation::Constellation(CertificatePtr certificate, Config config)
                                                                *muddle_, cfg_.log2_num_lanes))
   , dag_{GenerateDAG("dag_db_", true, certificate)}
   , beacon_network_{CreateBeaconNetwork(cfg_, certificate, network_manager_)}
-  , beacon_{CreateBeaconService(cfg_, *beacon_network_, *shard_management_, certificate)}
+  , beacon_setup_{CreateBeaconSetupService(cfg_, *beacon_network_, *shard_management_, certificate)}
```
n-hutton, line 300: In a subsequent PR we may as well let consensus own and setup the beacon setup and beacon service since constellation does not really use it.

libs/crypto/src/mcl_dkg.cpp
```diff
@@ -407,14 +407,14 @@ bn::Fr SignatureAggregationCoefficient(PublicKey const &             notarisatio
  * @return Pair consisting of aggregate signature and a vector indicating who's signatures were
  * aggregated
  */
-std::pair<Signature, std::vector<bool>> ComputeAggregateSignature(
+AggregateSignature ComputeAggregateSignature(
```
n-hutton, line 429: 👍

libs/crypto/src/mcl_dkg.cpp
```diff
@@ -407,14 +407,14 @@ bn::Fr SignatureAggregationCoefficient(PublicKey const &             notarisatio
...
     std::unordered_map<uint32_t, Signature> const &signatures,
     std::vector<PublicKey> const &                 public_keys)
 {
   Signature aggregate_signature;
   aggregate_signature.clear();
-  std::vector<bool> signers;
-  signers.resize(public_keys.size(), false);
+  SignerRecord signers;
+  signers.resize(public_keys.size(), 0);
```
n-hutton, line 435: Worth thinking about using a bitset here if the signer record is serialized and stored for all time

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1363,15 +1363,19 @@ bool MainChain::DetermineHeaviestTip()
     // find the heaviest item in our tip selection
     auto it = std::max_element(
         tips_.begin(), tips_.end(), [](TipsMap::value_type const &a, TipsMap::value_type const &b) {
-          auto        a_weight{a.second.total_weight}, b_weight{b.second.total_weight};
+          auto        a_total_weight{a.second.total_weight}, b_total_weight{b.second.total_weight};
           auto const &a_hash{a.first}, &b_hash{b.first};
+          auto        a_weight{a.second.weight}, b_weight{b.second.weight};
 
-          return a_weight < b_weight || (a_weight == b_weight && a_hash < b_hash);
+          return a_total_weight < b_total_weight ||
```
n-hutton, line 1391: 👍 Using weight as a tiebreak right? Since it's eqivalent to rank - def worth putting in a comment here to explain why this is important

libs/ledger/src/consensus/consensus.cpp
```diff
@@ -167,30 +160,55 @@ Block GetBeginningOfAeon(Block const &current, MainChain const &chain)
...
   return ret;
 }
 
-uint64_t Consensus::GetBlockGenerationWeight(Block const &previous, chain::Address const &address)
+bool Consensus::VerifyNotarisation(Block const &block) const
 {
-  auto const cabinet = GetCabinet(previous);
-
-  if (!cabinet)
+  // Genesis is not notarised so the body of blocks with block number 1 do
```
n-hutton, line 174: 👍

libs/ledger/src/consensus/consensus.cpp
```diff
@@ -167,30 +160,55 @@ Block GetBeginningOfAeon(Block const &current, MainChain const &chain)
...
+  if (notarisation_ && block.body.block_number > 1)
   {
-    FETCH_LOG_WARN(LOGGING_NAME, "Unable to determine block generation weight");
-    return 0;
+    // Try to verify with notarisation units in notarisation_
+    auto previous_notarised_block = chain_.GetBlock(block.body.previous_hash);
+    auto result = notarisation_->Verify(previous_notarised_block->body.block_number,
+                                        previous_notarised_block->body.hash,
+                                        block.body.block_entropy.block_notarisation);
+    if (result == NotarisationResult::CAN_NOT_VERIFY)
```
n-hutton, line 183: Could you explain this tomorrow?

n-hutton (changes requested):

libs/beacon/include/beacon/beacon_setup_service.hpp
jinmannwong: It is not switched away from a queue structure. Maybe QueueNewCabinet is a more appropriate name, but the change is mainly that the aeon_execution_unit it now created in beacon_setup_service itself and then pushed onto the queue rather than having it pre-constructed elsewhere and then pass to the beacon_setup_service to just push onto the queue.

libs/beacon/include/beacon/notarisation_manager.hpp
jinmannwong: Yes, it is duplicated. The reason being that the beacon service and notarisation service should control the destruction of their respective cryptographic keys in the aeon execution unit and notarisation manager, which individually need to contain information about the aeon they are valid in.

libs/beacon/src/beacon_complaints_manager.cpp
jinmannwong: Not a bug fix but to ensure correct usage. The beacon setup service should not be calling this more than once.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: No, the assert is there for debugging. The change to include valid_dkg_members is so that you do not process any messages from people who have not sent you notarisation keys so they should never be part of your qual. If someone doesn't populate with notarisations or with incorrect ones then the aeon beginning block produced by that person is just invalid. As long as there is one honest member in qual then there will be a aeon beginning block.

libs/beacon/src/beacon_setup_service.cpp
jinmannwong: There shouldn't be any qual that is not composed of valid dkg members. This is put in as a check for when I was introducing valid_dkg_members_. I will rewrite it in terms of the set help functions so that it can all be contained inside the assert.

libs/beacon/src/block_entropy.cpp
jinmannwong: Yes, you are correct. The keys are already checked in ValidBlock in Consensus so I will remove it here.

libs/crypto/src/mcl_dkg.cpp
jinmannwong: I looked into the bitset but it seemed to require you to know the number of bits required at compile time.

libs/ledger/src/chain/main_chain.cpp
jinmannwong: Sure.

libs/beacon/tests/unit/dkg_resilience.cpp
jinmannwong: Man, I thought I removed all of these!

libs/beacon/include/beacon/trusted_dealer_beacon_service.hpp
jinmannwong: Ok, sure. I guess we can also set the aeon length as a class variable and then we only need to have aeon start as an argument.

libs/beacon/src/beacon_manager.cpp
```diff
@@ -27,31 +27,21 @@
...
-bn::G2 BeaconManager::group_h_;
-
 constexpr char const *LOGGING_NAME = "BeaconManager";
+// To set elliptic curve generator in static verify function
+constexpr char const *generator_string = "Fetch.ai Elliptic Curve Generator G";
 
 BeaconManager::BeaconManager(CertificatePtr certificate)
-  : certificate_{std::move(certificate)}
+  : group_g_{generator_string}
+  , group_h_{"Fetch.ai Elliptic Curve Generator H"}
```
n-hutton: It's a bit odd here that generator_string is a constexpr for g, but h pastes it in directly. Also, in my mind the beacon manager is really just trying to hold keys, having the generator string here and in a few other places makes it unclear what is responsible for setting these.

I think in mcl_dkg it could be something like

```
void SetGenerator(Generator &generator_g, std::string string_to_hash = "Fetch.ai Elliptic Curve Generator G")
{
  assert(!string_to_hash.empty());
  bn::hashAndMapToG2(generator_g, string_to_hash);
}
```

And here it would be cleaner if it were

```
group_g_{mcl::DefaultGeneratorG()}
```

libs/beacon/src/beacon_manager.cpp
```diff
@@ -600,7 +588,8 @@ bool BeaconManager::Verify(byte_array::ConstByteArray const &group_public_key,
     return false;
   }
 
-  return crypto::mcl::VerifySign(tmp, message, tmp2, group_g_);
+  Generator generator_tmp{generator_string};
```
n-hutton: this needs to be tmp because of the static issue? Is there a high overhead to constructing a generator for every verify?

libs/ledger/src/protocols/notarisation_service.cpp
```diff
@@ -274,8 +274,11 @@ NotarisationService::State NotarisationService::OnComplete()
     notarisation_collection_height_ = notarised_chain_height_ + 1;
   }
 
-  // If we have new notarisation keys, or we have notarised all blocks for this aeon
-  if (new_keys || notarised_chain_height_ >= active_notarisation_unit_->round_end())
+  // If we have new notarisation keys and the next block to notarise is the end of the aeon then we
```
n-hutton, line 277: 👍

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1367,6 +1367,9 @@ bool MainChain::DetermineHeaviestTip()
           auto const &a_hash{a.first}, &b_hash{b.first};
           auto        a_weight{a.second.weight}, b_weight{b.second.weight};
 
+          // The weight, which is related to the rank of the miner producing the block is used here
```
n-hutton: A little confusing here for newcomers I would suspect. Suggest: 

```
The weight, which is related to the rank of the miner producing the block is used here
to tie break chains with equivalent total weight, choosing the weight of the tips as a tiebreaker.
This is important for consensus.
```

n-hutton (changes requested):
Sorry, there are minor comments still

libs/ledger/src/chain/main_chain.cpp
jinmannwong: Sure.

libs/beacon/src/beacon_manager.cpp
jinmannwong: Yes, the tmp is here as this function is static and the generator is no longer static. From some quick benchmarks it adds around 10% onto the verification time, which is approx 0.15ms.

jinmannwong: I am happy to stick with the solution in the deployment updates PR. This was something implemented before I saw that.

n-hutton (approved):

PR #1835, opened 2019-10-23
Distributed networking clients examples

MissingNO57:


libs/dmlf/examples/boston_housing_multiprocess_distributed_learning/main.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "dmlf/networkers/muddle_learner_networker.hpp"
+#include "dmlf/simple_cycling_algorithm.hpp"
+#include "math/matrix_operations.hpp"
+#include "math/tensor.hpp"
+#include "ml/dataloaders/ReadCSV.hpp"
+#include "ml/dataloaders/tensor_dataloader.hpp"
+#include "ml/exceptions/exceptions.hpp"
+#include "ml/ops/loss_functions/cross_entropy_loss.hpp"
+#include "ml/optimisation/adam_optimiser.hpp"
+#include "ml/optimisation/sgd_optimiser.hpp"
```
evsmithx: I don't think you're using SGD optimiser

libs/dmlf/include/dmlf/update.hpp
```diff
@@ -22,6 +22,7 @@
 #include "crypto/hash.hpp"
 #include "crypto/sha256.hpp"
 #include "dmlf/update_interface.hpp"
+#include "ml/dataloaders/word2vec_loaders/vocab.hpp"
```
baykaner, line 25: i think this is fine for now, but at some point in the future we'll probably rework this such that we have an abstract update and inherited updates; in this case a W2VUpdate. Then the serialisers will have the same inheritance structure.

libs/dmlf/examples/word2vec_multiprocess_distributed_learning/main.cpp
```diff
@@ -0,0 +1,143 @@
...
+  std::string train_file            = argv[1];
+  client_params.analogies_test_file = argv[2];
+  std::string output_csv_file       = argv[3];
+  std::string config                = std::string(argv[4]);
+  int         instance_number       = std::atoi(argv[5]);
+
+  // Distributed learning parameters:
+  SizeType number_of_rounds = 1000;
+  SizeType number_of_peers  = 3;
+  // have been processed in total by the clients
```
evsmithx: I think this is in the wrong place!

evsmithx (dismissed):

libs/dmlf/examples/word2vec_multiprocess_distributed_learning/main.cpp
evsmithx: (The comment, I mean)

libs/dmlf/examples/boston_housing_multiprocess_distributed_learning/main.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+using namespace fetch::ml::ops;
+using namespace fetch::ml::layers;
+using namespace fetch::ml::distributed_learning;
+
+using DataType         = fetch::fixed_point::FixedPoint<32, 32>;
+using TensorType       = fetch::math::Tensor<DataType>;
+using VectorTensorType = std::vector<TensorType>;
+using SizeType         = fetch::math::SizeType;
+
```
baykaner, line 49: lets add a comment here explaining what is the overall goal of this example. And then lets do the same for the other 5 examples so that it's totally clear what is the difference for each one

libs/dmlf/examples/mnist_multiprocess_distributed_learning/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+using namespace fetch::ml::ops;
+using namespace fetch::ml::layers;
+using namespace fetch::ml::distributed_learning;
+
+using DataType         = fetch::fixed_point::FixedPoint<32, 32>;
+using TensorType       = fetch::math::Tensor<DataType>;
+using VectorTensorType = std::vector<TensorType>;
+using SizeType         = fetch::math::SizeType;
+
+std::shared_ptr<TrainingClient<TensorType>> MakeClient(
```
baykaner: it looks like this function is re-used in several different examples. Can we make it a general utility that gets re-used?

libs/dmlf/examples/word2vec_multiprocess_distributed_learning/main.cpp
```diff
@@ -0,0 +1,143 @@
...
+  client_params.word2 = "queen";
+  client_params.word3 = "father";
+
+  // calc the true starting learning rate
+  client_params.starting_learning_rate = static_cast<DataType>(client_params.batch_size) *
+                                         client_params.starting_learning_rate_per_sample;
+  client_params.ending_learning_rate = static_cast<DataType>(client_params.batch_size) *
+                                       client_params.ending_learning_rate_per_sample;
+  client_params.learning_rate_param.starting_learning_rate = client_params.starting_learning_rate;
+  client_params.learning_rate_param.ending_learning_rate   = client_params.ending_learning_rate;
```
baykaner: presumably some of this is copy paste from the other word2vec example? We should try to avoid repeated code. Can you unify these together into some kind of utility/config function call somewhere?

baykaner (changes requested):

libs/dmlf/include/dmlf/distributed_learning/utilities/mnist_client_utilities.hpp
```diff
@@ -0,0 +1,78 @@
...
+#include "ml/ops/placeholder.hpp"
+#include "ml/optimisation/adam_optimiser.hpp"
+
+namespace fetch {
+namespace dmlf {
+namespace distributed_learning {
+namespace utilities {
+
+template <typename TensorType>
+std::shared_ptr<fetch::dmlf::distributed_learning::TrainingClient<TensorType>> MakeMNISTClient(
```
baykaner, line 33: I was imagining a more general MakeClient method for which we specify the architecture of the model. However, it's probably fine to leave it as it is and simply rework it as part of a refactor that lets clients have models instead of separate graph, dataloader, and optimiser

baykaner (dismissed):

baykaner (approved):

PR #1837, opened 2019-10-23
Added predict binding to etch model & implemented Boston etch example

baykaner:


libs/vm-modules/examples/boston_house_prices/main.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+  auto module = std::make_shared<fetch::vm::Module>();
+
+  fetch::vm_modules::System::Bind(*module);
+
+  fetch::vm_modules::ml::BindML(*module);
+
+  fetch::vm_modules::CreatePrint(*module);
+
+  module->CreateFreeFunction("read_csv", &read_csv);
```
troels-im: Could we rename to `readCSV`? This would follow the style of the other functions in the library.

libs/vm-modules/examples/boston_house_prices/script/boston.etch
```diff
@@ -0,0 +1,32 @@
...
+
+// evaluate performance
+var loss = model.evaluate();
+print(loss);
+
+// make predictions on all test data
+var predictions = model.predict(test_data);
+print(predictions.at(0u64, 0u64));
+
+endfunction
```
troels-im, line 32: This looks great :+1:

libs/vm-modules/examples/boston_house_prices/main.cpp
```diff
@@ -0,0 +1,140 @@
...
+  module->CreateFreeFunction("read_csv", &read_csv_no_transpose);
```
troels-im: Could we also add a `readCSVFromString`? The reason being that this would be a way to add the dataset to a smart contract.

troels-im (dismissed):
A few minor comments, but looks great!

libs/vm-modules/examples/boston_house_prices/main.cpp
baykaner: task raised on jira ML-168 for adding that as a general utility (rather than making a special function for this example and then never using it)

MissingNO57 (approved):
LGTM

PR #1839, opened 2019-10-24
Use hex encoding for digest

wkaluza:


n-hutton (approved):
LGTM - might be worth mentioning in the style guide use hex for digests, b64 for keys

PR #1843, opened 2019-10-24
reworked distributed boston example to read config file instead of ta…

baykaner:
…king command line arguments


MissingNO57 (approved):
LGTM

PR #1851, opened 2019-10-25
Maxpool op with tests

MissingNO57:


libs/ml/include/ml/ops/max_pool.hpp
```diff
@@ -0,0 +1,188 @@
...
+  using DataType      = typename TensorType::Type;
+  using ArrayPtrType  = std::shared_ptr<TensorType>;
+  using VecTensorType = typename Ops<T>::VecTensorType;
+  using SPType        = OpMaxPool1DSaveableParams<T>;
+  using MyType        = MaxPool<TensorType>;
+
+  MaxPool(SizeType const kernel_size, SizeType const stride_size)
+    : kernel_size_{kernel_size}
+    , stride_size_{stride_size}
+    , max_pool_1d_(kernel_size, stride_size)
```
baykaner: so in tensorflow the way this presumably gets handled is that whenever you add an op you're supposed to specify the relevant shape of the input (except the batch size which can be 'None'); so presumably they can just deduce whether it's a 1D or 2D maxpool based on the input data shape.

We could do the same thing (or just have a flag for 1D or 2D mode)? Then we wouldn't have to initialise both the 1D and 2D max pool ops on construction? In fact the MaxPool Op can just store a shared_ptr to an Ops, which can be dynamically cast to the correct type on construction, right?

In the future we may also wish to separate graph definitions and graph compilation more thoroughly, and this will require us to pre-specify input dimensions for ops anyway.

libs/ml/include/ml/ops/max_pool.hpp
```diff
@@ -0,0 +1,188 @@
...
+  {
+    FETCH_UNUSED(me);
+    assert(me.get() == this);
+
+    auto copyshare = std::make_shared<MyType>(*this);  // calls default copy constructor of MyType
+
+    return copyshare;
+  }
+  /**
+   * Applies 1D max pooling of kernel_size_ for each channel described here:
```
baykaner: or 2D max pooling?

libs/ml/include/ml/ops/max_pool.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+    default:
+    {
+      throw fetch::ml::exceptions::InvalidMode("Unsupported data shape");
+    }
+    }
+  }
+
+  /**
+   * Computes gradient of 1D max pooling of kernel_size_ for each channel described here:
```
baykaner: or 2D?

libs/ml/include/ml/serializers/ml_types.hpp
```diff
@@ -1896,6 +1907,45 @@ struct MapSerializer<ml::OpMaxPool1DSaveableParams<TensorType>, D>
...
+template <typename TensorType, typename D>
+struct MapSerializer<ml::OpMaxPoolSaveableParams<TensorType>, D>
+{
+  using Type       = ml::OpMaxPoolSaveableParams<TensorType>;
+  using DriverType = D;
+
+  static uint8_t const BASE_OPS    = 1;
+  static uint8_t const OP_CODE     = 2;
+  static uint8_t const KERNEL_SIZE = 3;
+  static uint8_t const STRIDE_SIZE = 4;
```
baykaner, line 1920: If you make the change I suggested before you might also need to serialise something to indicate whether the shared_ptr<Ops> should be dynamically cast to 1D or 2D

libs/ml/tests/unit/ops/max_pool.cpp
```diff
@@ -0,0 +1,729 @@
...
+    }
+  }
+
+  fetch::ml::ops::MaxPool<TensorType> op(3, 2);
+
+  TensorType prediction(op.ComputeOutputShape({std::make_shared<const TensorType>(data)}));
+  op.Forward({std::make_shared<const TensorType>(data)}, prediction);
+
+  // test correct values
+  ASSERT_TRUE(prediction.AllClose(gt, DataType{1e-5f}, DataType{1e-5f}));
```
baykaner: why 1e-5 tolerance? cant we use function tolerance?

libs/ml/tests/unit/ops/max_pool.cpp
```diff
@@ -0,0 +1,729 @@
...
+    gt(0, i, 0) = static_cast<DataType>(gt_input1[i]);
+    gt(0, i, 1) = static_cast<DataType>(gt_input2[i]);
+  }
+
+  fetch::ml::ops::MaxPool<TensorType> op(3, 2);
+  std::vector<TensorType>             prediction =
+      op.Backward({std::make_shared<const TensorType>(data)}, error);
+
+  // test correct values
+  ASSERT_TRUE(prediction[0].AllClose(gt, DataType{1e-5f}, DataType{1e-5f}));
```
baykaner: tolerance again

baykaner (changes requested):

baykaner (approved):

PR #1852, opened 2019-10-25
OneHot math function implemented with tests

MissingNO57:


libs/math/include/math/one_hot.hpp
```diff
@@ -0,0 +1,109 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "core/assert.hpp"
```
baykaner: doesn't look like you used this

libs/math/include/math/one_hot.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+#include <cassert>
```
baykaner: or this?

libs/math/include/math/one_hot.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+namespace fetch {
+namespace math {
+
+/**
+ * One hot function
+ * @tparam ArrayType
+ * @param ret
+ * @param indices
```
baykaner: please fill out these comment fields. they're not all obvious (e.g. depth)

libs/math/include/math/one_hot.hpp
```diff
@@ -0,0 +1,109 @@
...
+ * @param depth
+ * @param on_value
+ * @param off_value
+ * @param axis
+ */
+template <typename ArrayType>
+void OneHot(ArrayType &ret, ArrayType const &indices, typename ArrayType::SizeType depth,
+            typename ArrayType::Type on_value  = typename ArrayType::Type{1.0},
+            typename ArrayType::Type off_value = typename ArrayType::Type{0.0}, SizeType axis = 0)
```
baykaner: I can imagine using this function and specifying axis much more often than specifying the on value and off value. Since C++ syntax requires all previous optional arguments to be filled out up until the last one you set, it makes more sense to have axis be the first optional argument, and then have on_value and off_value after that

libs/math/include/math/one_hot.hpp
```diff
@@ -0,0 +1,109 @@
...
+{
+  auto ind_it = indices.begin();
```
baykaner, line 44: you should probably add an assertion checking that ret has shape().size() == (indices.shape().size() + 1)

libs/math/include/math/one_hot.hpp
```diff
@@ -0,0 +1,109 @@
...
+  auto ret_it = ret.Slice().begin();
+
+  if (axis != 0)
+  {
+    // Move the axis we want one-hot to the front
+    // to make it iterable in the inner most loop.
+    ret_it.MoveAxisToFront(axis);
+  }
+
+  while (ind_it.is_valid())
```
baykaner, line 54: please add a comment to explain how this while loop works. does it preallocate off_value everywhere and then overwrite some elements? or does it just fill each element correctly first time?

libs/math/include/math/one_hot.hpp
baykaner: can you also add an assertion that checks that depth == the size of the final dimension of ret?

baykaner (changes requested):

evsmithx (dismissed):

baykaner (approved):

PR #1853, opened 2019-10-26
Dmlf training client model rework

baykaner:
- do not merge until after #1835 

- reworks distributed training clients to own a ml::SequentialModel
- rework dmlf examples to use json config loading


libs/dmlf/examples/boston_housing_multiprocess_distributed_learning/main.cpp
```diff
@@ -104,18 +89,21 @@ int main(int argc, char **argv)
...
   client->SetNetworker(networker);
 
   // Create loss csv file
-  std::string results_filename = results_dir + "/fetch_" + std::to_string(number_of_clients) +
-                                 "_Adam_" + std::to_string(float(learning_rate)) + "_" +
+
+  std::string results_filename = results_dir + "/fetch_" + std::to_string(n_clients) + "_Adam_" +
+                                 std::to_string(float(client_params.learning_rate)) + "_" +
                                  std::to_string(seed) + "_FC3.csv";
   std::ofstream lossfile(results_filename, std::ofstream::out);
```
evsmithx: results_filename and lossfile are no longer used

libs/dmlf/examples/boston_housing_distributed_learning/main.cpp
```diff
@@ -156,19 +141,6 @@ int main(int argc, char **argv)
...
-    for (auto &c : clients)
-    {
-      std::cout << "\t"
-                << static_cast<double>(utilities::Test(c->GetModel(), data_tensor, label_tensor));
-      lossfile << ","
-               << static_cast<double>(utilities::Test(c->GetModel(), data_tensor, label_tensor));
-    }
-    std::cout << std::endl;
-    lossfile << std::endl;
-
```
evsmithx, line 171: lossfile and results_file needs to be removed from other places

libs/dmlf/examples/word2vec_distributed_learning/main.cpp
```diff
@@ -119,41 +117,36 @@ void SynchroniseWeights(std::vector<std::shared_ptr<TrainingClient<TensorType>>>
 
 int main(int argc, char **argv)
 {
-  // This example will create multiple local distributed clients with CBOW Word2Vec model and trains
+  // This example will create multiple local distributed clients with Word2Vec model and trains
```
evsmithx, line 120: trains -> train

libs/dmlf/include/dmlf/distributed_learning/distributed_learning_client.hpp
```diff
@@ -63,37 +48,35 @@ class TrainingClient
...
+  void SetNetworker(std::shared_ptr<fetch::dmlf::AbstractLearnerNetworker> networker_ptr);
 
   virtual void Run();
 
   void Train();
 
   virtual void Test();
 
-  virtual std::shared_ptr<GradientType> GetGradients();
+  virtual std::shared_ptr<fetch::dmlf::Update<TensorType>> GetGradients();
```
evsmithx: Should be GradientType? Or remove `using GradientType     = fetch::dmlf::Update<TensorType>;` from above.

libs/ml/include/ml/model/model.hpp
```diff
@@ -256,6 +267,48 @@ void Model<TensorType>::UpdateConfig(ModelConfig<DataType> &model_config)
...
+typename Model<TensorType>::ModelOptimiserType *Model<TensorType>::GetOptimiser()
+{
+  return optimiser_ptr_.get();
+}
+
+template <typename TensorType>
+typename Model<TensorType>::GraphType *Model<TensorType>::GetGraph()
+{
+  return graph_ptr_.get();
+}
```
evsmithx: I'm sure there's a reason for it, but why do these functions return raw pointers?

evsmithx (changes requested):

libs/dmlf/examples/word2vec_distributed_learning/main.cpp
baykaner: i don't think so. that comment is written in the future tense.

libs/dmlf/include/dmlf/distributed_learning/distributed_learning_client.hpp
baykaner: oh I didn't know that using existed. That's an extremely confusing using because an Update and a gradient are fundamentally different things; i'll rename it.

libs/ml/include/ml/model/model.hpp
baykaner: because the model actually OWNS (via a unique ptr) its graph, dataloader, and optimiser. This is good for ordinary use cases (because we shouldn't mess with model internals without a good reason); however distributed ML is a special case where we probably do need to mess with these internals. however, we can't just std::move them out of the model (because this would break the model).

As a compromise here I've exported raw pointers so we can use these internals; but this should indicate to any developer that doing this implies you know exactly what you're doing!

I don't really like this solution much, but I didn't like the alternatives either. Here are some for example:
1. rewrite distributed client functions to only interact with the model (however, this probably requires adding extra functions to model that only exist for this one use case),
2. change models internals to shared_ptrs (but this implies it's fine to extract and manipulate model internals, which it probably isn't)
3. make distributed client a friend of model and use the unique_ptrs directly (perhaps this is a better option?)

Perhaps I will rewrite this as a friend and see if it's cleaner

baykaner: i see that we have already made distributed client a friend of graph, so we might as well do the same for model; then we can ditch those raw pointer accessors.
nice catch :-)

evsmithx: I think another solution would be to return a const ref to a unique pointer? But making model a friend of graph sounds better. Unless that would create a circular dependency?

evsmithx (dismissed):

libs/dmlf/include/dmlf/distributed_learning/utilities/mnist_client_utilities.hpp
```diff
@@ -37,39 +37,31 @@ std::shared_ptr<fetch::dmlf::distributed_learning::TrainingClient<TensorType>> M
...
+  model_ptr->template Add<fetch::ml::layers::FullyConnected<TensorType>>(
+      10u, 10u, fetch::ml::details::ActivationType::RELU);
+  model_ptr->template Add<fetch::ml::layers::FullyConnected<TensorType>>(
+      10u, 10u, fetch::ml::details::ActivationType::SOFTMAX);
 
   // Initialise DataLoader
-  std::shared_ptr<fetch::ml::dataloaders::MNISTLoader<TensorType, TensorType>> dataloader_ptr =
-      std::make_shared<fetch::ml::dataloaders::MNISTLoader<TensorType, TensorType>>(images, labels);
+  auto dataloader_ptr =
+      std::make_unique<fetch::ml::dataloaders::MNISTLoader<TensorType, TensorType>>(images, labels);
```
evsmithx, line 51: Should this be a shared_ptr?

libs/ml/include/ml/model/model.hpp
```diff
@@ -99,6 +114,8 @@ class Model
                           DataType test_loss = fetch::math::numeric_max<DataType>());
 
 private:
+  friend class dmlf::distributed_learning::TrainingClient<TensorType>;
```
evsmithx, line 117: Should this still be here? I thought shared_ptrs removed the need for the friend class.

libs/ml/include/ml/model/model.hpp
```diff
@@ -30,6 +30,13 @@
 #include <utility>
 
 namespace fetch {
+namespace dmlf {
+namespace distributed_learning {
+template <class TensorType>
+class TrainingClient;
+}  // namespace distributed_learning
+}  // namespace dmlf
```
evsmithx, line 38: If friend class is no longer required (see below) I think we can delete this?

evsmithx (dismissed):

evsmithx (approved):

PR #1856, opened 2019-10-28
Dev/var converter for params

engineerkatie:
Prototype code for the var converter so that Troels can develop on top of it.


JuanBesa (commented):
I'll approve it.. but needs a lot of polish...

JuanBesa (approved):

PR #1860, opened 2019-10-29
MsgPack compliance for VM arrays, maps and strings.

troels-im:
This PR refactors the VM serialalers for Array, Map and String to follow the standard of MsgPack. As a consequence, these types will take up less space in the state database and it becomes possible to serialise C++ types and pass them directly as parameters.


libs/core/include/core/serializers/map_interface.hpp
```diff
@@ -33,19 +33,20 @@ class MapInterface
     : serializer_{serializer}
     , size_{size}
   {}
-
-  template <typename V>
-  void Append(uint8_t key, V const &val)
-  {
-    ++pos_;
-    if (pos_ > size_)
+  /*
```
troels-im: Dead code - delete

robertdickson (dismissed):
This is great stuff, let's get it in.

robertdickson (approved):
LGTM

PR #1861, opened 2019-10-29
One Hot Op with tests

MissingNO57:


libs/ml/include/ml/ops/one_hot.hpp
```diff
@@ -0,0 +1,134 @@
...
+public:
+  using TensorType    = T;
+  using SizeType      = typename TensorType::SizeType;
+  using DataType      = typename TensorType::Type;
+  using ArrayPtrType  = std::shared_ptr<TensorType>;
+  using VecTensorType = typename Ops<T>::VecTensorType;
+  using SPType        = OpOneHotSaveableParams<T>;
+  using MyType        = OneHot<TensorType>;
+
+  explicit OneHot(SizeType depth, SizeType axis = 0, DataType on_value = DataType{1.0},
```
baykaner, line 50: please add a function comment that explains each of these parameters. I particularly find 'depth' to be an unintuitive parameter without some description

baykaner (changes requested):

baykaner (dismissed):

evsmithx (approved):

PR #1865, opened 2019-10-29
Bugfix/http destruction

troels-im:
This PR fixes a segfault that occurs in the HTTP server when the client keeps a connection open longer than the life time of the server.


ejfitzgerald (changes requested):
Looks good can you make sure `constellation` is updated now that it needs to be constructed with `std::make_shared`. I suggest that it might be worth looking through the examples too

ejfitzgerald (approved):
(i'll read all of it next time 😁 )

PR #1873, opened 2019-10-30
Etch readcsv

baykaner:


libs/math/include/math/utilities/ReadCSV.hpp
```diff
@@ -71,7 +68,9 @@ TensorType ReadCSV(std::string const &filename, math::SizeType const cols_to_ski
     ++row;
   }
 
-  TensorType weights({row - rows_to_skip, col - cols_to_skip});
+  // we swap rows and columns because rows are ordinarily data samples, which should be the trailing
+  // dimension of our tensor
+  TensorType weights({col - cols_to_skip, row - rows_to_skip});
```
evsmithx, line 73: This seems like a confusing default! It's helpful for us but I think it'll confuse users.

libs/dmlf/examples/boston_housing_multiprocess_distributed_learning/main.cpp
```diff
@@ -76,8 +76,8 @@ int main(int argc, char **argv)
 
   // Load data
 
-  TensorType data_tensor  = fetch::ml::dataloaders::ReadCSV<TensorType>(data_file).Transpose();
-  TensorType label_tensor = fetch::ml::dataloaders::ReadCSV<TensorType>(labels_file).Transpose();
+  TensorType data_tensor  = fetch::math::utilities::ReadCSV<TensorType>(data_file).Transpose();
+  TensorType label_tensor = fetch::math::utilities::ReadCSV<TensorType>(labels_file).Transpose();
```
evsmithx: This transposes in both versions - now that ReadCSV transposes by default does this still  have the correct behaviour? The same comment applies to all the other dmlf and ml examples that use this.

evsmithx (changes requested):

libs/math/include/math/utilities/ReadCSV.hpp
baykaner: actually i think its a good default; the reason is that usually if you have a csv data file you will have feature columns and data rows. In our library data dimension has to be the trailing dimensions (i.e. the second dimension in these 2D tensors).

This way will just work the way people expect normally I think; the code isn't intuitive to look at, but it's definitely a sensible default (for reference, all 3 examples where we used readCSV we had to transpose the data)

libs/dmlf/examples/boston_housing_multiprocess_distributed_learning/main.cpp
baykaner: good catch. shouldn't be a transpose anymore

evsmithx (dismissed):

MissingNO57 (dismissed):
LGTM

evsmithx (approved):

PR #1875, opened 2019-10-30
MsgPack compliance for ledger Variant

troels-im:
This PR introduces MsgPack compliance for the ledger variant. As a consequence it is now possible to load either JSON or YAML and convert them into native C++ types through serialization, or use them them as arguments for an Etch script.

### DIFF
--- a/libs/dmlf/include/dmlf/execution/execution_result.hpp
+++ b/libs/dmlf/include/dmlf/execution/execution_result.hpp
@@ -103,8 +103,17 @@ public:
   template <typename Constructor>
   static void Serialize(Constructor &map_constructor, Type const &exec_res)
   {
-    auto map = map_constructor(3);
-    map.Append(OUTPUT, exec_res.output_);
+    auto             map    = map_constructor(3);
+    variant::Variant output = exec_res.output_;
+
+    if (output.IsUndefined())
+    {
+      // MsgPack does not support undefined
+      // unless extended.
+      output = static_cast<std::string>("");
+    }
+
+    map.Append(OUTPUT, output);
     map.Append(ERROR, exec_res.error_);
     map.Append(CONSOLE, exec_res.console_);
   }
--- a/libs/dmlf/tests/execution/local_executor_tests.cpp
+++ b/libs/dmlf/tests/execution/local_executor_tests.cpp
@@ -31,8 +31,8 @@
 namespace {
 
 using fetch::dmlf::BasicVmEngine;
-using fetch::dmlf::ExecutionResult;
 using fetch::dmlf::ExecutionEngineInterface;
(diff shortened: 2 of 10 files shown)


libs/dmlf/include/dmlf/execution/execution_result.hpp
```diff
@@ -103,8 +103,17 @@ struct MapSerializer<fetch::dmlf::ExecutionResult, D>
   template <typename Constructor>
   static void Serialize(Constructor &map_constructor, Type const &exec_res)
   {
-    auto map = map_constructor(3);
-    map.Append(OUTPUT, exec_res.output_);
+    auto             map    = map_constructor(3);
+    variant::Variant output = exec_res.output_;
+
+    if (output.IsUndefined())
```
lrahmani, line 109: the `IsUndefined` check should't be done at `ExecutionResult` serialization, but rather in the `variant::Variant` serializer/deserializer

libs/dmlf/tests/execution/local_executor_tests.cpp
```diff
@@ -131,6 +131,8 @@ TEST_F(LocalExecutorTests, delete_state)
 
 TEST_F(LocalExecutorTests, create_executable)
 {
+  // TODO(tfr): It is unclear why running the same code twice would yeild two different results
```
lrahmani: executable names are unique (similarly to state names)

libs/dmlf/tests/execution/local_executor_tests.cpp
```diff
@@ -131,6 +131,8 @@ TEST_F(LocalExecutorTests, delete_state)
...
+  // Some explanation would be good.
   result_promise = CreateExecutable("HelloWorld", SourceFile{"hello_world.etch", hello_world_etch});
```
lrahmani, line 136: executable names are unique (similarly to state names)

libs/dmlf/include/dmlf/execution/execution_result.hpp
troels-im: No, it should be done here - it does not belong in the serialisation.

libs/dmlf/tests/execution/local_executor_tests.cpp
troels-im: Please add comments in the code to explain this as it is non-obvious when reading the test.

robertdickson (approved):
looks good to me

libs/dmlf/include/dmlf/execution/execution_result.hpp
lrahmani: in my understanding Undefined is a reflection of the variant being empty which is part of the variant state not the ExecutionResult

PR #1877, opened 2019-10-31
Bug fix in beacon tests

jinmannwong:
Fixed test so all nodes have the same DKG start time and also fix setting on timer in beacon setup service so that time slots increase with each failure properly.


libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -1412,21 +1412,23 @@ void BeaconSetupService::SetTimeToProceed(BeaconSetupService::State state)
...
+    uint64_t dkg_time               = expected_dkg_time_s;
+    uint16_t failures               = 0;
+    double   adjusted_time_per_slot = time_per_slot;
 
     while (next_start_point < current_time)
     {
       failures++;
       next_start_point += dkg_time;
-      time_per_slot += 0.5 * time_per_slot;
+      adjusted_time_per_slot += 0.5 * adjusted_time_per_slot;
```
n-hutton, line 1436: 👍

n-hutton (approved):
LGTM

PR #1879, opened 2019-11-01
Fix nullptr executor synergetic execution manager

qati:
This PR fixes the issue found in  SynergeticExecutionManager::ExecuteItem method, where we end up in this method with empty executors list (LDGR-496). Also adds extra logging and telemetry to TransientObjectStore (related to LDGR-495 issue).


libs/ledger/src/upow/synergetic_execution_manager.cpp
```diff
@@ -158,14 +163,24 @@ bool SynergeticExecutionManager::ValidateWorkAndUpdateState(std::size_t num_lane
...
+    if (executors_.empty())
+    {
+      FETCH_LOG_WARN(LOGGING_NAME, "Executors empty, can't execute item!");
+      auto ptr          = std::make_shared<WorkItem>();
+      ptr->problem_data = std::move(problem_data);
+      ptr->work_queue   = std::move(queue);
+      solution_stack_.emplace_back(std::move(ptr));
+      no_executor_count_->increment();
+      return;
+    }
```
ejfitzgerald: I suspect this is just a racy thing. I would not add the element to the solution stack. I would sit this thread in a loop for a short period of time. If this fails then I would fail hard.

ejfitzgerald (changes requested):
Looks great, I would have changed the logic a little. I propose making a simple wait loop in this thread such that we can continue if possible. 

If you change it to a loop, it might also be worth adding a telemetry for the number of loops that it took for example

libs/ledger/src/upow/synergetic_execution_manager.cpp
qati: Changed the logic

ejfitzgerald (approved):

PR #1882, opened 2019-11-01
Math Test sanitising

baykaner:
- unified test types
- in some cases more types, or better tolerance is now used

### DIFF
--- a/cmake/BuildTools.cmake
+++ b/cmake/BuildTools.cmake
@@ -149,6 +149,7 @@ function (_internal_add_fetch_test
     target_link_libraries(${name} PRIVATE ${library} gmock gmock_main)
     target_include_directories(${name}
                                PRIVATE ${FETCH_ROOT_VENDOR_DIR}/googletest/googletest/include)
+    target_include_directories(${name} PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/${directory}")
     target_include_directories(${name}
                                PRIVATE ${FETCH_ROOT_VENDOR_DIR}/googletest/googlemock/include)
 
--- a/libs/math/tests/unit/activation_functions/sigmoid.cpp
+++ b/libs/math/tests/unit/activation_functions/sigmoid.cpp
@@ -32,7 +32,7 @@ class SigmoidTest : public ::testing::Test
 {
 };
 
-TYPED_TEST_CASE(SigmoidTest, FloatingTypes);
+TYPED_TEST_CASE(SigmoidTest, TensorFloatingTypes);
 
 template <typename ArrayType>
 ArrayType RandomArrayNegative(std::size_t n)
--- a/libs/math/tests/unit/activation_functions/sigmoid.cpp
+++ b/libs/math/tests/unit/activation_functions/sigmoid.cpp
@@ -20,9 +20,6 @@
 #include "math/activation_functions/sigmoid.hpp"
 #include "test_types.hpp"
 
-#include <iomanip>
-#include <iostream>
-
 namespace fetch {
 namespace math {
 namespace test {
--- a/libs/math/tests/unit/activation_functions/sigmoid.cpp
+++ b/libs/math/tests/unit/activation_functions/sigmoid.cpp
@@ -16,19 +16,23 @@
 //
 //------------------------------------------------------------------------------
 
-#include "math/activation_functions/sigmoid.hpp"
-#include "math/tensor.hpp"
-
 #include "gtest/gtest.h"
+#include "math/activation_functions/sigmoid.hpp"
+#include "test_types.hpp"
+
+#include <iomanip>
+#include <iostream>
+
+namespace fetch {
+namespace math {
+namespace test {
 
 template <typename T>
 class SigmoidTest : public ::testing::Test
 {
 };
 
-using MyTypes = ::testing::Types<fetch::math::Tensor<float>, fetch::math::Tensor<double>,
-                                 fetch::math::Tensor<fetch::fixed_point::FixedPoint<32, 32>>>;
-TYPED_TEST_CASE(SigmoidTest, MyTypes);
+TYPED_TEST_CASE(SigmoidTest, FloatingTypes);
 
 template <typename ArrayType>
 ArrayType RandomArrayNegative(std::size_t n)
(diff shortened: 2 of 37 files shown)


cmake/BuildTools.cmake
```diff
@@ -149,6 +149,7 @@ function (_internal_add_fetch_test
     target_link_libraries(${name} PRIVATE ${library} gmock gmock_main)
     target_include_directories(${name}
                                PRIVATE ${FETCH_ROOT_VENDOR_DIR}/googletest/googletest/include)
+    target_include_directories(${name} PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/${directory}")
```
evsmithx, line 152: I'm not sure what this does?

libs/math/tests/unit/activation_functions/sigmoid.cpp
```diff
@@ -16,19 +16,23 @@
...
 //------------------------------------------------------------------------------
 
+#include "gtest/gtest.h"
 #include "math/activation_functions/sigmoid.hpp"
-#include "math/tensor.hpp"
+#include "test_types.hpp"
 
-#include "gtest/gtest.h"
+#include <iomanip>
+#include <iostream>
```
evsmithx: Required imports?

libs/math/tests/unit/clustering/knn_test.cpp
```diff
@@ -18,41 +18,41 @@
...
-  EXPECT_NEAR(double(output.at(1).second), double(4), 1e-4);
   EXPECT_EQ(output.at(2).first, SizeType(2));
-  EXPECT_NEAR(double(output.at(2).second), double(14.6969384), 1e-4);
   EXPECT_EQ(output.at(3).first, SizeType(3));
-  EXPECT_NEAR(double(output.at(3).second), double(16.6132477), 1e-4);
+
+  EXPECT_TRUE(output.at(0).second - double(2) < function_tolerance<DataType>());
+  EXPECT_TRUE(output.at(1).second - double(4) < function_tolerance<DataType>());
+  EXPECT_TRUE(output.at(2).second - double(14.6973) < function_tolerance<DataType>());
+  EXPECT_TRUE(output.at(3).second - double(16.6132477) < function_tolerance<DataType>());
```
evsmithx: Why has this changed from EXPECT_NEAR? The error messages from EXPECT_NEAR print both values and the tolerance which is helpful for debugging

libs/math/tests/unit/metrics/l2_norm.cpp
```diff
@@ -16,37 +16,44 @@
...
+  test_array[2] = DataType(3);
+  test_array[3] = DataType(-4);
+  test_array[4] = DataType(5);
+  test_array[5] = DataType(-6);
+  test_array[6] = DataType(7);
+  test_array[7] = DataType(-8);
 
   // initialise to non-zero just to avoid correct value at initialisation
-  typename TypeParam::Type score(0);
+  DataType score(0);
```
evsmithx: Incorrect comment?

libs/math/tests/unit/test_types.hpp
```diff
@@ -0,0 +1,58 @@
...
+using FloatingTypes =
+    ::testing::Types<float, double, fetch::fixed_point::fp32_t, fetch::fixed_point::fp64_t>;
+using HighPrecisionFloatingTypes = ::testing::Types<float, double, fetch::fixed_point::fp64_t>;
+
+using IntAndFloatingTypes =
+    ::testing::Types<int32_t, int64_t, float, double, fetch::fixed_point::fp32_t,
+                     fetch::fixed_point::fp64_t>;
+using UnsignedIntAndFloatingTypes =
+    ::testing::Types<uint32_t, int32_t, uint64_t, int64_t, float, double,
+                     fetch::fixed_point::fp32_t, fetch::fixed_point::fp64_t>;
```
evsmithx, line 36: Judging by the name I was surprised that this contained int32_t. Maybe we should call it FloatIntAndUIntTypes?

evsmithx (changes requested):
It's a lot cleaner than it was before!

libs/math/tests/unit/distance/entropy_tests.cpp
```diff
@@ -48,5 +43,9 @@ TYPED_TEST(EntropyTest, entropy)
   A.Set(SizeType{2}, DataType(0.3));
   A.Set(SizeType{3}, DataType(0.4));
```
MissingNO57, line 44: I know these are just tests, but shouldn't it be static_cast<DataType>(0.3)?

libs/math/tests/unit/distance/euclidean_distance.cpp
```diff
@@ -47,15 +43,15 @@ TYPED_TEST(EuclideanTest, simple_test)
...
   A.Set(SizeType{2}, DataType(3));
   A.Set(SizeType{3}, DataType(4));
-  EXPECT_NEAR(double(Euclidean(A, A)), 0, (double)function_tolerance<DataType>());
+  EXPECT_NEAR(double(distance::Euclidean(A, A)), 0, (double)function_tolerance<DataType>());
 
   ArrayType B = ArrayType(4);
   B.Set(SizeType{0}, static_cast<DataType>(1));
   B.Set(SizeType{1}, DataType(2));
   B.Set(SizeType{2}, DataType(3));
   B.Set(SizeType{3}, DataType(2));
```
MissingNO57, line 52: I know these are just tests, but shouldn't it be static_cast(0.3)?

libs/math/tests/unit/distance/euclidean_distance.cpp
```diff
@@ -96,7 +92,7 @@ TYPED_TEST(EuclideanTest, matrix_euclidean_test)
   B.Set(SizeType{2}, SizeType{2}, DataType(-5));
   B.Set(SizeType{2}, SizeType{3}, DataType(-6));
```
MissingNO57, line 93: I know these are just tests, but shouldn't it be static_cast(-5)?

libs/math/tests/unit/distance/entropy_tests.cpp
baykaner: in this case it's a DataType constructor call, so it should be equivalent.

evsmithx (approved):

MissingNO57 (approved):
LGTM

PR #1884, opened 2019-11-01
Math TOP K function with tests

MissingNO57:


libs/math/include/math/top_k.hpp
```diff
@@ -0,0 +1,129 @@
...
+ * @param data input
+ * @param k number of k-highest values
+ * @param sorted TRUE=descending order, FALSE=ascending order
+ */
+template <typename ArrayDataType, typename ArrayIndicesType>
+void TopK(ArrayDataType &ret_data, ArrayIndicesType &ret_indices, ArrayDataType const &data,
+          typename ArrayDataType::SizeType k, bool sorted = true)
+{
+  using DataType   = typename ArrayDataType::Type;
+  using IndiceType = typename ArrayIndicesType::Type;
```
baykaner: by the way, the singular of 'indices' is 'index'

libs/math/include/math/top_k.hpp
```diff
@@ -0,0 +1,129 @@
...
+  // Set itaration order
+  data_it.MoveAxisToFront(axis);
+  ret_data_it.MoveAxisToFront(axis);
+  ret_indices_it.MoveAxisToFront(axis);
+
+  // Prepare vector for sorting values
+  std::vector<std::pair<SizeType, DataType>> vec;
+  vec.resize(axis_size);
+
+  SizeType i = 0, k_minus_one = k - 1;
```
baykaner: style wise I think we're suppose to put this kind of statement on two lines for clarity:
SizeType i = 0;
SizeType k_mins_one = k - 1;

libs/math/tests/unit/gtest/top_k.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "gtest/gtest.h"
+
+using namespace fetch::math;
+
+template <typename T>
+class TopKTest : public ::testing::Test
+{
+};
+
+using MyTypes = ::testing::Types<fetch::math::Tensor<float>, fetch::math::Tensor<double>,
```
baykaner: once #1882 goes in this file will need a few slight alterations:
- include test_types & don't include tensor
- namespace everything inside fetch::math::test
- remove MyTypes and use FloatingTypes
this makes it easier for us to test all the floating types we care about without missing any out from each list of types in each test file (and avoids weird bugs from not being properly namespaced)

baykaner (changes requested):

libs/math/include/math/top_k.hpp
MissingNO57: Of course, it is...

libs/math/tests/unit/gtest/top_k.cpp
MissingNO57: I did required changes and waiting for your PR to be merged.

baykaner (approved):
looks great, nice!

Issue #258, opened 2018-10-02
HTTP server can break irrecoverably during operation


n-hutton:
During TX gen I noticed that the http server can be put into a irrecoverable state. This is confirmed casually by running the example http server and hammering it with transactions until it emits an asio error: 'reached EOF', in which case it will stop responding to http requests.

PR #1887, opened 2019-11-03
Bugfix/http destruction nh

n-hutton:
This fixes a use after free in the http server. Note: the HTTP client has no close method, so more than one use of it in the test will cause it to wait forever.

The main way to solve the use after free here is to make the connection manager 'live' on the asio callback stack, thus allowing it the raw socket access it is doing in the HTTPConnection. This means that calls to the connection manager must now be posted.

### DIFF
--- a/libs/http/include/http/connection.hpp
+++ b/libs/http/include/http/connection.hpp
@@ -55,19 +55,12 @@ public:
   }
 
   ~HTTPConnection() override
-  {
-    /*manager_.Leave(handle_);*/
-    std::cerr << "HTTPDEST 1" << std::endl;  // DELETEME_NH
-  }
+  {}
 
   void Start()
   {
     is_open_ = true;
-    // handle_  = manager_.Join(shared_from_this());
-    // if (is_open_)
-    //{
     ReadHeader();
-    //}
   }
 
   void Send(HTTPResponse const &response) override
@@ -233,6 +226,7 @@ public:
   void Close()
   {
     is_open_ = false;
+    CloseConnnection();
     manager_.Leave(handle_);
   }
 
--- a/libs/http/include/http/connection.hpp
+++ b/libs/http/include/http/connection.hpp
@@ -54,8 +54,7 @@ public:
                     socket_.remote_endpoint().address().to_string());
   }
 
-  ~HTTPConnection() override
-  {}
+  ~HTTPConnection() override = default;
 
   void Start()
   {
--- a/libs/http/include/http/connection.hpp
+++ b/libs/http/include/http/connection.hpp
@@ -56,17 +56,18 @@ public:
 
   ~HTTPConnection() override
   {
-    manager_.Leave(handle_);
+    /*manager_.Leave(handle_);*/
+    std::cerr << "HTTPDEST 1" << std::endl;  // DELETEME_NH
   }
 
   void Start()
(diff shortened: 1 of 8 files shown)


libs/http/include/http/connection.hpp
```diff
@@ -55,18 +55,12 @@ class HTTPConnection : public AbstractHTTPConnection,
   }
 
   ~HTTPConnection() override
-  {
-    manager_.Leave(handle_);
-  }
+  {}
```
n-hutton: The manager owns the connections, so no need to tell it that it is leaving, I think

libs/http/include/http/connection.hpp
```diff
@@ -222,9 +216,17 @@ class HTTPConnection : public AbstractHTTPConnection,
     asio::async_write(socket_, *buffer_ptr, cb);
   }
 
+  void CloseConnnection() override
+  {
+    std::error_code dummy;
```
n-hutton, line 220: Make sure to cleanly close the connection when the connection manager is finished with it

libs/http/include/http/server.hpp
```diff
@@ -95,29 +95,76 @@ class HTTPServer : public AbstractHTTPServer
         acceptor->close(dummy);
       }
     });
+
+    // Since the connection manager contains a reference to this class, we must guarantee it has
```
n-hutton, line 99: As the comment describes, this is a  destruction race otherwise.

libs/http/include/http/server.hpp
```diff
@@ -95,29 +95,76 @@ class HTTPServer : public AbstractHTTPServer
...
+    std::shared_ptr<uint64_t> ref_counter = std::make_shared<uint64_t>();
 
-      // allow initiating class to post closes to these
-      socRef   = soc;
-      accepRef = accep;
+    {
+      auto        ref_counter_copy = ref_counter;
+      HTTPServer &server_ref       = *this;
+
+      networkManager_.Post([&socRef, &accepRef, &manager, &threadMan, port, ref_counter_copy,
```
n-hutton: Due to these references being passed in, block until there is no possibility the closure can be executed (either io_service dies, was never started, or it is executed)

libs/http/include/http/connection.hpp
troels-im: Connection is self-owning and it may well out-live the connection manager. This is what caused a problem to start with and what the PR #1885 fixes. If you look at how the connection is setup done in the server, you will see that the connection always passes shared self-pointers into lambdas that are passed to ASIO.

n-hutton: It does do this yes, but I think there will be a race on destruction/socket closure - manager contains strong pointers to the connections in a map, and it lives in server.hpp so it is very possible they will destruct after asio is gone, causing bad stuff to happen.

I don't follow why the connections would need to out live the connection manager but happy to catch up whenever you are free.

Looking into the E2E failure - it seems that in both of our PRs there is some issue closing connections (E2E fails sending http calls) - I vaguely suspect that ```asio::async_read_until(socket_, *buffer_ptr, "\r\n\r\n", cb);``` might hang forever.

n-hutton:
Closes #258

troels-im (approved):
LGTM

PR #1888, opened 2019-11-03
Rearrange and add null support to msgpack

ejfitzgerald:


libs/variant/tests/unit/serialization.cpp
```diff
@@ -108,3 +108,124 @@ TEST(VariantSerialization, ArrayArraySerialisation)
   EXPECT_EQ(y[3][2], 9);
   EXPECT_EQ(y[3][3], 10);
 }
+
+TEST(VariantSerialization, VariantInt)
```
ejfitzgerald, line 112: Theses tests have come from `libs/core/tests/serializers/variants.cpp` file

libs/variant/include/variant/variant.hpp
```diff
@@ -598,12 +598,12 @@ struct ForwardSerializer<fetch::variant::Variant, D>
     {
     case Type::Type::UNDEFINED:
     {
-      // TODO(tfr): Work out if msgpack has support for undefined
-      return;
+      throw std::runtime_error{"Unsupported variant type for serialisation"};
```
troels-im, line 601: 👍

libs/variant/include/variant/variant.hpp
```diff
@@ -598,12 +598,12 @@ struct ForwardSerializer<fetch::variant::Variant, D>
...
     }
     case Type::Type::NULL_VALUE:
     {
-      // TODO(tfr): Work out if msgpack has support for null
+      serializer.Allocate(sizeof(uint8_t));
+      serializer.WriteByte(static_cast<uint8_t>(TypeCodes::NIL));
```
troels-im: I think it would be preferable if this could be made less manual, i.e.
```serializer << nullptr;```

File: libs/variant/tests/unit/serialization.cpp [hunk 4]
```diff
@@ -108,3 +108,124 @@ TEST(VariantSerialization, ArrayArraySerialisation)
...
+  Variant           null1, null2;
+
+  null1 = Variant::Null();
+  stream << null1;
+  stream.seek(0);
+  stream >> null2;
+  EXPECT_EQ(null2.IsNull(), true);
+}
+
+TEST(VariantSerialization, DISABLED_VariantArray)
```
troels-im, line 160: I think this one can be re-enabled as Null should now work?

libs/variant/tests/unit/serialization.cpp
```diff
@@ -108,3 +108,124 @@ TEST(VariantSerialization, ArrayArraySerialisation)
...
+  stream.seek(0);
+  stream >> arr2;
+  EXPECT_EQ(arr2.IsArray(), true);
+  EXPECT_EQ(arr2[0].IsNull(), true);
+  EXPECT_EQ(arr2[1].As<int>(), 123456);
+  EXPECT_EQ(arr2[2].As<float>(), 1.25);
+  EXPECT_EQ(arr2[3].As<std::string>(), "123456");
+}
+
+TEST(VariantSerialization, DISABLED_VariantArrayOfArray)
```
troels-im, line 180: Same for this one?

libs/variant/tests/unit/serialization.cpp
```diff
@@ -108,3 +108,124 @@ TEST(VariantSerialization, ArrayArraySerialisation)
...
+  {
+    EXPECT_EQ(arr2[i].IsArray(), true);
+    for (std::size_t j = 0; j < 4; j++)
+    {
+      EXPECT_EQ(arr2[i][j].As<int>(), (i == j) ? 1 : 0);
+    }
+  }
+}
+
+TEST(VariantSerialization, DISABLED_VariantObject)
```
troels-im, line 214: And this one.

troels-im (changes requested):
I'd suggest we do not expose the MsgPack codes to the user.

libs/variant/include/variant/variant.hpp
ejfitzgerald: Done

libs/variant/tests/unit/serialization.cpp
ejfitzgerald: These tests do not work. I think there is significant overlap here with the work that @engineerkatie is doing. I will raise an issue to fix these issues but I think it is out of the scope of this PR at this time

troels-im: :+1:

n-hutton (approved):
Looks fine to me

troels-im (approved):
LGTM

PR #1890, opened 2019-11-04
Features/ml top k op

MissingNO57:


libs/ml/include/ml/ops/top_k.hpp
```diff
@@ -0,0 +1,192 @@
...
+    assert(me.get() == this);
+
+    auto copyshare = std::make_shared<MyType>(*this);  // calls default copy constructor of MyType
+
+    return copyshare;
+  }
+
+  /**
+   * Returns array of k-highest values
+   * for input array of shape [x,y,z] and value k, return array would be of shape [x,y,k]
```
evsmithx: Is this correct? I though the last dimension was the batch dimension and this was going to use the second from last?

MissingNO57: Sorry, it's fixed now

libs/math/include/math/top_k.hpp
```diff
@@ -41,7 +41,17 @@ void TopK(ArrayDataType &ret_data, ArrayIndicesType &ret_indices, ArrayDataType
...
-  SizeType axis      = data.shape().size() - 1;
+  SizeType axis;
+
+  if (data.shape().size() > 1)
+  {
+    axis = data.shape().size() - 2;
+  }
+  else
+  {
+    axis = data.shape().size() - 1;
```
evsmithx: This would be more simply put as axis = 0!

libs/math/include/math/top_k.hpp
```diff
@@ -117,8 +127,19 @@ std::pair<ArrayDataType, ArrayIndicesType> TopK(ArrayDataType const &
...
-  ret_shape.at(ret_shape.size() - 1) = k;
+  SizeType axis;
+
+  if (data.shape().size() > 1)
+  {
+    axis = data.shape().size() - 2;
+  }
+  else
+  {
+    axis = data.shape().size() - 1;
```
evsmithx: Also here!

libs/math/include/math/top_k.hpp
```diff
@@ -41,7 +41,17 @@ void TopK(ArrayDataType &ret_data, ArrayIndicesType &ret_indices, ArrayDataType
 
   assert(k <= data.shape().at(data.shape().size() - 1));
 
-  SizeType axis      = data.shape().size() - 1;
+  SizeType axis;
+
+  if (data.shape().size() > 1)
```
baykaner, line 48: I think you should add a comment explaining this design choice. This lives in the math library, but the design choice is driven by the machine learning libraries assumption that the trailing dimension is the data dimension, right?
That should be spelled out

libs/ml/include/ml/ops/top_k.hpp
```diff
@@ -0,0 +1,200 @@
...
+  std::shared_ptr<OpsSaveableParams> GetOpSaveableParams() override
+  {
+    SPType sp{};
+    sp.k      = k_;
+    sp.sorted = sorted_;
+
+    return std::make_shared<SPType>(sp);
+  }
+
+  std::shared_ptr<fetch::ml::ops::Ops<TensorType>> MakeSharedCopy(
```
baykaner: this is inside fetch::ml::ops namespace already right? no need for explicit namespacing then?

libs/ml/include/ml/ops/top_k.hpp
```diff
@@ -0,0 +1,200 @@
...
+namespace fetch {
+namespace ml {
+namespace ops {
+
+template <class T>
+class TopK : public fetch::ml::ops::Ops<T>
+{
+public:
+  using TensorType     = T;
+  using SizeType       = typename TensorType::SizeType;
```
baykaner: this is not allowed because it's shadowing.
you should do this:
using SizeType = fetch::math::SizeType;

libs/ml/include/ml/ops/top_k.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+    // 1D
+    case 1:
+    {
+      for (SizeType i{0}; i < error_signal.shape().at(0); i++)
+      {
+        ret_signal.At(indices_.At(i)) = error_signal.At(i);
+      }
+    }
+    break;
```
baykaner: it looks slightly strange to have the break; outside of the braces

libs/ml/tests/unit/ops/top_k.cpp
```diff
@@ -0,0 +1,242 @@
...
+#include "ml/ops/top_k.hpp"
+#include "ml/serializers/ml_types.hpp"
+#include "vectorise/fixed_point/fixed_point.hpp"
+
+#include "gtest/gtest.h"
+
+#include <memory>
+#include <vector>
+
+template <typename T>
```
baykaner, line 35: add namespace:
fetch::ml::test

libs/ml/tests/unit/ops/top_k.cpp
```diff
@@ -0,0 +1,242 @@
...
+  // test correct values
+  ASSERT_EQ(prediction.shape(), gt.shape());
+  ASSERT_TRUE(prediction.AllClose(gt));
+}
+
+TYPED_TEST(TopKOpTest, saveparams_test)
+{
+  using TensorType    = TypeParam;
+  using DataType      = typename TypeParam::Type;
+  using SizeType      = typename TypeParam::SizeType;
```
baykaner: better to use fetch::math::SizeType rather than getting this from the Tensor

evsmithx (dismissed):

baykaner (changes requested):

evsmithx (approved):

baykaner (approved):

PR #1891, opened 2019-11-04
Synergetic contracts end to end test

qati:
This PR extends the current run_end_to_end_test functionality to support testing synergetic contracts and dag syncronization. Also introduces Id in Runnable, so exceptions caught by Reactor are more trackable.


libs/core/src/reactor.cpp
```diff
@@ -181,8 +181,8 @@ void Reactor::Monitor()
       }
       catch (std::exception const &ex)
       {
-        FETCH_LOG_WARN(LOGGING_NAME, "The reactor caught an exception! ", name_,
-                       " error: ", ex.what());
+        FETCH_LOG_WARN(LOGGING_NAME, "The reactor caught an exception in ", runnable->GetId(), "! ",
```
cyenyxe, line 184: It is a good idea then for the `Runnable` class to return an empty string by default?

n-hutton: Indeed - why not pure virtual?

libs/ledger/include/ledger/storage_unit/transient_object_store.hpp
```diff
@@ -139,8 +146,11 @@ template <typename O>
 TransientObjectStore<O>::TransientObjectStore(uint32_t log2_num_lanes)
   : log2_num_lanes_(log2_num_lanes)
   , rids(batch_size_)
-  , state_machine_{
-        std::make_shared<core::StateMachine<Phase>>("TransientObjectStore", Phase::Populating)}
+  , state_machine_{std::make_shared<core::StateMachine<Phase>>("TransientObjectStore",
+                                                               Phase::Populating)}
+  , cache_rid_removed_{telemetry::Registry::Instance().CreateCounter(
+        "ledger_storage_transient_rid_removed_total",
+        "The number of needed rids which were removed from cache.")}
```
n-hutton: Why can they get removed from the cache? I forget

libs/ledger/src/upow/synergetic_execution_manager.cpp
```diff
@@ -158,14 +163,24 @@ bool SynergeticExecutionManager::ValidateWorkAndUpdateState(std::size_t num_lane
...
     FETCH_LOCK(lock_);
+    if (executors_.empty())
+    {
+      FETCH_LOG_WARN(LOGGING_NAME, "Executors empty, can't execute item!");
+      auto ptr          = std::make_shared<WorkItem>();
+      ptr->problem_data = std::move(problem_data);
+      ptr->work_queue   = std::move(queue);
+      solution_stack_.emplace_back(std::move(ptr));
+      no_executor_count_->increment();
+      return;
```
n-hutton, line 204: Seems Ed has a comment on this on your other PR

scripts/end_to_end_test/end_to_end_synergetic_test.yaml
```diff
@@ -0,0 +1,30 @@
...
+  pos_mode: True
+  max_test_time: 600
+
+steps:
+  - wait_network_ready: {"sleep": 5}
+  - create_wealth: {"amount": 100000000, "nodes": [0]}
+  - create_contract: {"name": "contract001", "fee_limit": 10000, "nodes": [0]}
+  - wait_for_blocks: {"nodes": [0], "num": 5}
+  - run_contract: {"contract_name": "contract001", "nodes": [0], "wait_for_blocks": 15}
+  - stop_nodes: { "nodes": [0], "remove_dbs": 1}
```
n-hutton: remove dbs True?

scripts/end_to_end_test/run_end_to_end_test.py
```diff
@@ -413,6 +433,133 @@ def add_node(parameters, test_instance):
...
+    nodes = parameters["nodes"]
+    amount = parameters["amount"]
+
+    for node_index in nodes:
+        node_host = "localhost"
+        node_port = test_instance._nodes[node_index]._port_start
+
+        api = LedgerApi(node_host, node_port)
+
+        # create the entity from the node's private key
```
n-hutton, line 448: 👍

scripts/end_to_end_test/run_end_to_end_test.py
```diff
@@ -413,6 +433,133 @@ def add_node(parameters, test_instance):
...
+        entity = Entity(get_nodes_private_key(test_instance, node_index))
+        api.sync(api.tokens.wealth(entity, amount))
+
+
+def create_contract(parameters, test_instance):
+    from smart_contract_tests.synergetic_utils import ContractTestHelper
```
n-hutton: Perhaps put this import at the top of the tests (?)

n-hutton: This might be confusing naming due to this only being valid for creating synergetic contracts.

scripts/end_to_end_test/run_end_to_end_test.py
```diff
@@ -413,6 +433,133 @@ def add_node(parameters, test_instance):
...
+            output(
+                f"Synergetic contract ({contract_name}) execution failed on node {node_index}!")
+            raise Exception(
+                f"Synergetic contract ({contract_name}) execution failed on node {node_index}!")
+        else:
+            output(
+                f"Synergetic contract ({contract_name}) executed on node {node_index} ")
+
+
+def wait_for_blocks(parameters, test_instance):
```
n-hutton, line 507: I would probably consider adding this and chain sync etc. to the instance class itself. This as it will break if used with the dmlf stuff. Does it wait for X blocks, or until it sees X block?

Worth adding a small class comment for them, too.

scripts/end_to_end_test/smart_contract_tests/synergetic_utils.py
```diff
@@ -0,0 +1,78 @@
...
+@query
+function query_result() : Int32
+  use solution;
+
+  return solution.get(-1);
+endfunction
+"""
+
+
+class ContractTestHelper:
```
n-hutton: Suggest: SynergeticContra...

scripts/fetchai_netutils/fetch/testing/testcase.py
```diff
@@ -224,6 +237,21 @@ def restart_node(self, index):
         self.start_node(index)
         time.sleep(3)
 
+    def remove_db_files(self, index):
+        root = os.path.abspath(os.path.join(
+            self._workspace, 'node{}'.format(index)))
+        pattern = ["*.db"]
+        for p in pattern:
+            [os.remove(x) for x in glob.iglob(root+'/**/' + p, recursive=True)]
```
n-hutton: Best to avoid raw concat '+' for paths I think

scripts/end_to_end_test/end_to_end_synergetic_test.yaml
```diff
@@ -0,0 +1,30 @@
+# End-to-end test scenarios
+
+---
+test_description: >
```
n-hutton, line 4: Is this the test case which fails?

scripts/end_to_end_test/run_end_to_end_test.py
```diff
@@ -413,6 +433,133 @@ def add_node(parameters, test_instance):
...
+        output(
+            f"Waiting for {wait_for_blocks_num} blocks on node {node_index}")
+        api = LedgerApi(node_host, node_port)
+        api.wait_for_blocks(wait_for_blocks_num)
+
+
+def verify_chain_sync(parameters, test_instance):
+    from fetch.cluster.chain import ChainSyncTesting
+    max_trials = parameters["max_trials"]
+    sleep_time = parameters["sleep_between_trials"]
```
n-hutton: I wouldn't pass these in, rather it can be calculated based on the block time. It is fine to take a long time I think since the test has a timeout.

n-hutton (changes requested):
Looks good on the whole, minor comments

libs/core/src/reactor.cpp
qati: made pure virtual

libs/ledger/include/ledger/storage_unit/transient_object_store.hpp
qati: I don't know why or where the removal happened, I seen this issue only once

libs/ledger/src/upow/synergetic_execution_manager.cpp
qati: yes, fixed in other PR

scripts/end_to_end_test/smart_contract_tests/synergetic_utils.py
qati: 👍

scripts/fetchai_netutils/fetch/testing/testcase.py
qati: 👍

scripts/end_to_end_test/end_to_end_synergetic_test.yaml
qati: yes

scripts/end_to_end_test/run_end_to_end_test.py
qati: Max trials is now an optional parameter, sleep time is block interval * 1.2

scripts/end_to_end_test/run_end_to_end_test.py
qati: Moved. I think waits for X blocks.

n-hutton (dismissed):
LGTM

n-hutton (approved):

PR #1893, opened 2019-11-05
Feature/implement fp128

### LINKED ISSUE #1464: do not enable hex output of FixedPoint numbers by default

### LINKED ISSUE #1599: check inline fixed point operators /= *= += for handling +/-inf

markos:
* Implementation of int class, and in particular the Int<256> specialisation which is needed for FixedPoint<64, 64> aka fp128_t. Note: this is different than the UInt class as that one does not handle signed numbers.
* Addition of testcase for Int class.
* Refactoring of the FixedPoint test case to avoid code duplication and use TYPED_TEST for all the available FixedPoint types (fp32_t/fp64_t/fp128_t).
* minor fix: hex output for FixedPoints is enabled only in Debug builds.

Todo: 
* Provide Serializer/Deserialiser for fp128 along with Etch bindings.
* Increase precision in fp128 functions by providing specializations with extra terms in the polynomial approximations.
* Optimise for performance.

Closes: #1599 
Closes: #1464


libs/math/tests/CMakeLists.txt
```diff
@@ -11,3 +11,5 @@ include(${FETCH_ROOT_CMAKE_DIR}/BuildTools.cmake)
 setup_compiler()
 
 fetch_add_slow_test(math-unit-tests fetch-math unit/)
+fetch_add_slow_test(math-fp-unit-tests fetch-math unit/fixed_point)
```
baykaner: these are a subdirectory under unit, so wont they be included in the math-unit-tests anyway?

libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -25,77 +25,50 @@
 #include <cstdint>
 #include <limits>
 
+static const int64_t N{10000};
+
 using namespace fetch::fixed_point;
 
+using MyFPTypes = ::testing::Types<fetch::fixed_point::fp32_t, fetch::fixed_point::fp64_t,
```
baykaner: please rename this to FloatingPointTypes and move this to math/tests/unit/test_types.hpp

libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -25,77 +25,50 @@
...
+                                   fetch::fixed_point::fp128_t>;
+
```
baykaner: please wrap all the tests in the fetch/math/test namespace

libs/math/tests/unit/vectors/avx_tests.cpp
```diff
@@ -112,10 +112,9 @@ using MyTypes = ::testing::Types<fetch::vectorise::VectorRegister<float, 32>,
                                  fetch::vectorise::VectorRegister<fetch::fixed_point::fp64_t, 64>,
                                  fetch::vectorise::VectorRegister<double, 64>>;
 
-using MyFPTypes = ::testing::Types<fetch::vectorise::VectorRegister<float, 32>,
-                                   fetch::vectorise::VectorRegister<fetch::fixed_point::fp32_t, 32>,
-                                   fetch::vectorise::VectorRegister<fetch::fixed_point::fp64_t, 64>,
-                                   fetch::vectorise::VectorRegister<double, 64>>;
+using MyFPTypes =
```
baykaner, line 121: again please move definitions of groups of types for testing into math/test/unit/test_types.hpp
this might be called PrimitiveFloatTypes or something like that

libs/math/tests/unit/vectors/avx_tests.cpp
```diff
@@ -112,10 +112,9 @@ using MyTypes = ::testing::Types<fetch::vectorise::VectorRegister<float, 32>,
...
+    ::testing::Types<fetch::vectorise::VectorRegister<fetch::fixed_point::fp32_t, 32>,
+                     fetch::vectorise::VectorRegister<fetch::fixed_point::fp64_t, 64>>;
 #endif
 
 TYPED_TEST_CASE(VectorRegisterTest, MyTypes);
```
baykaner, line 126: please add namespace around tests fetch/math/test

libs/vectorise/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ setup_compiler()
 # ------------------------------------------------------------------------------
 
 setup_library(fetch-vectorise)
-target_link_libraries(fetch-vectorise INTERFACE fetch-meta pthread)
+target_link_libraries(fetch-vectorise INTERFACE fetch-meta pthread fetch-core)
```
baykaner: this looks like a circular dependency!!!!
core depends on vectorise right?
we should not permit this

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
```diff
@@ -670,11 +699,15 @@ constexpr FixedPoint<I, F>::FixedPoint(T n, meta::IfIsInteger<T> * /*unused*/)
 {
   if (CheckOverflow(static_cast<NextType>(n)))
   {
+    std::cout << "this = " << FromBase(data_) << std::endl;
```
baykaner: I don't think we want to do this in release mode.
there are cases we overflow is recoverable, and printing to console probably shouldn't be done in release unless it's already a catastrophic failure

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
```diff
@@ -670,11 +699,15 @@ constexpr FixedPoint<I, F>::FixedPoint(T n, meta::IfIsInteger<T> * /*unused*/)
...
+    std::cout << "overflow" << std::endl;
     fp_state |= STATE_OVERFLOW;
     data_ = MAX;
   }
   else if (CheckUnderflow(static_cast<NextType>(n)))
   {
+    std::cout << "this = " << FromBase(data_) << std::endl;
```
baykaner: same as for overflow

libs/vectorise/tests/CMakeLists.txt
```diff
@@ -16,5 +16,8 @@ fetch_add_test(vectorise_exact_exponents_gtest fetch-vectorise meta/gtest)
 fetch_add_test(vectorise_memory_gtest fetch-vectorise memory)
 target_link_libraries(vectorise_memory_gtest PRIVATE fetch-core)
 
+fetch_add_test(vectorise_int_gtest fetch-vectorise int)
+target_link_libraries(vectorise_int_gtest PRIVATE fetch-core)
```
baykaner, line 20: again i'm concerned these look like circular dependencies?

libs/vectorise/include/vectorise/uint/int.hpp
```diff
@@ -0,0 +1,1298 @@
...
+  explicit constexpr operator __int128_t() const;
+  explicit constexpr operator __uint128_t() const;
+
+  ////////////////////////////
+  /// assignment operators ///
+  ////////////////////////////
+
+  constexpr Int &operator=(Int const &v);
+  template <typename ArrayType>
+  meta::IfHasIndex<ArrayType, Int> &operator=(ArrayType const &v);  // NOLINT
```
baykaner, line 123: what problem does the NOLINT solve here?

libs/vectorise/include/vectorise/uint/int.hpp
```diff
@@ -0,0 +1,1298 @@
...
+
+  __uint128_t carry = 0, terms[WIDE_ELEMENTS] = {};
+  terms[0] = products[0][0];
+  carry    = static_cast<WideType>(terms[0] >> WIDE_ELEMENT_SIZE);
+  terms[1] = products[0][1] + products[1][0] + carry;
+  carry    = static_cast<WideType>(terms[1] >> WIDE_ELEMENT_SIZE);
+  terms[2] = products[0][2] + products[1][1] + products[2][0] + carry;
+  carry    = static_cast<WideType>(terms[2] >> WIDE_ELEMENT_SIZE);
+  terms[3] = products[0][3] + products[1][2] + products[2][1] + products[3][0] + carry;
+  // TODO(?): decide what to do with overflow if carry > 0
```
baykaner: leftover todo
please either raise an issue or remove if redundant

libs/vectorise/include/vectorise/uint/int.hpp
```diff
@@ -0,0 +1,1298 @@
...
+      array.Append(u.ElementAt(i));
+    }
+  }
+
+  template <typename ArrayDeserializer>
+  static void Deserialize(ArrayDeserializer &array, Type &u)
+  {
+    if (array.size() != u.elements())
+    {
+      // TODO(?): Throw
```
baykaner: throwing sounds sensible to me. you may want to make a custom exception that explains what happened here.

baykaner (changes requested):

libs/math/tests/unit/fixed_point/fixed_point.cpp
markos: I explicitly named them as FixedPointTypes instead and moved to the file as you suggested.

libs/math/tests/unit/fixed_point/fixed_point.cpp
markos: done

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
markos: accidentally left out, removed.

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
markos: done

libs/math/tests/unit/vectors/avx_tests.cpp
markos: ok, however, their definition depends on whether AVX2 is enabled, I'll try to find a better solution for this

libs/math/tests/unit/vectors/avx_tests.cpp
markos: done

libs/math/tests/CMakeLists.txt
markos: yes, but I wanted them to run separately for speed reasons, compiling and running every unit test each time is very time-consuming IMHO, but you are right, I should exclude them from the main fetch-math test.

libs/vectorise/include/vectorise/uint/int.hpp
markos: removed, overflow is taken care of in the FixedPoint class that uses Int.

libs/vectorise/include/vectorise/uint/int.hpp
markos: this is also the case in the existing UInt class -where this code was copied from. I will investigate further

libs/math/tests/CMakeLists.txt
baykaner: yep, well in the short term feel free to move all the other math unit tests into a new subdirectory so they don't get run twice

markos: apparently one can just call the test with --gtest_filter=FixedPoint* and skip running the rest of the tests. Quite a time saver, I will disable building the extra binaries

libs/vectorise/tests/CMakeLists.txt
markos: circular depencenty removed, problem solved.

libs/vectorise/include/vectorise/uint/int.hpp
markos: I would like to tackle this in a separate issue and not as part of this PR.

libs/math/tests/unit/vectors/avx_tests.cpp
markos: I will also keep those as they are right now, these affect only VectorRegister types, for only the vector_tests -and used nowhere else. Furthermore, they are under an platform specific #ifdef AVX2, and would need modification for other architectures. I will investigate another way to fix this.

libs/vectorise/include/vectorise/uint/int.hpp
markos: it's a false alarm: 

> Line: 123 @ 3
>   Message: operator=() should return 'Int&'
>   Check: misc-unconventional-assign-operator
>   Context:
>       meta::IfHasIndex<ArrayType, Int> &operator=(ArrayType const &v);

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
```diff
@@ -697,12 +749,16 @@ template <typename T>
 constexpr FixedPoint<I, F>::FixedPoint(T n, meta::IfIsFloat<T> * /*unused*/)
   : data_(static_cast<typename FixedPoint<I, F>::Type>(n * ONE_MASK))
 {
-  if (CheckOverflow(static_cast<NextType>(n) * static_cast<NextType>(ONE_MASK)))
+  if (IsMinusZero(n))
+  {
+    data_ = 0;
```
denis-trapeznikov, line 754: Do you ignore the sign?

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
```diff
@@ -475,9 +495,29 @@ class FixedPoint : public BaseFixedpointType
...
 
+template <typename T>
+static inline bool IsMinusZero(T n);
+
+template <>
+inline bool IsMinusZero<float>(float const n)
+{
+  // Have to dereference the pointer as char const *, else we get a strict-aliasing error
+  uint32_t n_int = *(reinterpret_cast<uint8_t const *>(&n));
+  return (n_int == 0x80000000);
```
denis-trapeznikov, line 510: Even putting aside that C++ `float` is not necessarily an IEEE-754, endianness is ignored here.

libs/vectorise/include/vectorise/uint/uint.hpp
```diff
@@ -448,7 +449,7 @@ constexpr meta::IfIsUnsignedInteger<T, bool> UInt<S>::operator>=(T other) const
 
 ///////////////////////
 /// unary operators ///
-//////////////////////1/
+///////////////////////
 
 template <uint16_t S>
 constexpr UInt<S> &UInt<S>::operator++()
```
denis-trapeznikov, line 455: My understanding is that `constrexpr`ness of this class (and others alike) is mostly ruined by the `const`ness of `_0`, `_1` and `max`. This morning I've tried to see if this issue can be addressed and found this possible recipe: `_0`, `_1` and `max` should be turned into `constexpr static` member function (and `private` as well, they don't seem to be used anywhere else) that simply return a rvalue.

denis-trapeznikov (changes requested):

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
markos: currently we do not even test on big-endian systems. And on all systems that we care to support, IEEE754 is supported.

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
markos: -0.0 is the same as +0.0, however in fixed-point comparing to -0.0 means comparing to a big integer, which leads to an unwanted underflow. Hence we need to detect and remove -0.0 value.

libs/vectorise/include/vectorise/uint/uint.hpp
markos: fixing issues with the UInt class is not currently the scope of this PR, and it has taken a bit longer than expected already, however I will gladly do this in a next PR to fix this and similar issues.

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
denis-trapeznikov: > -0.0 is the same as +0.0

I remember reading a text years ago where they were apparently distinct, something about modelling something in gas dynamics, sorry, can't seem to retrieve it now.

markos: in terms of FixedPoint numbers, they are definitely not distinct. The problem is that while the FPU does that in hardware, and -0.0 is 0x80000000 (for 32-bit floats), you can do things like -1.0 * 0.0 and still get 0x80000000. However in our Int class, if we allow negative zero, we will have as a result of that operation, 0xffffffffffffffffffffffffffffffff, which causes an underflow. Since negative zero has no usefulness in our case, I opted to remove support for it entirely. You are probably refering  to some physical problems which use a limit to zero from the negative x. In floating point types this may have an application as the exponents are quite large compared to a fixed point, but even with Int<256> we don't reach such exponents.

libs/core/include/core/serializers/base_types.hpp
```diff
@@ -839,5 +841,64 @@ struct ForwardSerializer<fixed_point::FixedPoint<I, F>, D>
...
+struct ForwardSerializer<fixed_point::FixedPoint<64, 64>, D>
+{
+public:
+  using Type       = fixed_point::FixedPoint<64, 64>;
+  using DriverType = D;
+
+  template <typename Interface>
+  static void Serialize(Interface &interface, Type const &n)
+  {
+    interface << static_cast<uint64_t>(n.Data()) << static_cast<uint64_t>(n.Data() >> 64);
```
baykaner: can we put parentheses around "static_cast<uint64_t>(n.Data() >> 64" so that the order of operations is crystal clear? it took a few seconds to figure out what was happening here

libs/core/include/core/serializers/base_types.hpp
```diff
@@ -839,5 +841,64 @@ struct ForwardSerializer<fixed_point::FixedPoint<I, F>, D>
...
+      array.Append(u.ElementAt(i));
+    }
+  }
+
+  template <typename ArrayDeserializer>
+  static void Deserialize(ArrayDeserializer &array, Type &u)
+  {
+    if (array.size() != u.elements())
+    {
+      // TODO(?): Throw
```
baykaner: leftover comment / todo without issue number

libs/dmlf/tests/execution/basic_engine_tests.cpp
```diff
@@ -1442,7 +1442,7 @@ TEST(BasicVmEngineDmlfTests, AddFixed)
       "add", "state", "add", Params{LedgerVariant(fp64_t(4.5)), LedgerVariant(fp32_t(5.5))});
   EXPECT_TRUE(result.succeeded());
   // std::cout << result.error().message() << '\n';
-  EXPECT_EQ(result.output().As<fp64_t>(), 9.5);
+  EXPECT_EQ(result.output().As<fp64_t>(), 10.0);
```
baykaner, line 1492: do you understand why this value change is necessary? (I dont)

libs/dmlf/tests/execution/basic_engine_tests.cpp
```diff
@@ -1442,7 +1442,7 @@ TEST(BasicVmEngineDmlfTests, AddFixed)
       "add", "state", "add", Params{LedgerVariant(fp64_t(4.5)), LedgerVariant(fp32_t(5.5))});
   EXPECT_TRUE(result.succeeded());
   // std::cout << result.error().message() << '\n';
```
baykaner: please ditch this commented out code while you're there

libs/math/include/math/tensor.hpp
```diff
@@ -2599,13 +2599,12 @@ bool Tensor<T, C>::AllClose(Tensor const &o, Type const &relative_tolerance,
...
     ++it2;
 
-    T abs_e1 = e1;
-    fetch::math::Abs(abs_e1, abs_e1);
-    T abs_e2 = e2;
-    fetch::math::Abs(abs_e2, abs_e2);
-    T abs_diff = e1 - e2;
-    fetch::math::Abs(abs_diff, abs_diff);
-    T tolerance = std::max(absolute_tolerance, std::max(abs_e1, abs_e2) * relative_tolerance);
+    T abs_e1, abs_e2, abs_diff;
```
baykaner: i believe this might be contrary to our style guide (I think we are supposed to have one declaration per line for clarity)

libs/math/src/math/linalg/blas/gemm_nn_vector.cpp
```diff
@@ -124,6 +124,16 @@ template class Blas<float, Signature(_C <= _alpha, _A, _B, _beta, _C),
                     Computes(_C <= _alpha * _A * _B + _beta * _C),
                     platform::Parallelisation::VECTORISE>;
 
+// template class Blas<fetch::fixed_point::FixedPoint<16, 16>, Signature(_C <= _alpha, _A, _B,
```
baykaner: left in commented out code?

libs/math/src/math/linalg/blas/gemm_nt_vector.cpp
```diff
@@ -126,6 +126,14 @@ template class Blas<float, Signature(_C <= _alpha, _A, _B, _beta, _C),
                     Computes(_C <= _alpha * _A * T(_B) + _beta * _C),
                     platform::Parallelisation::VECTORISE>;
 
+template class Blas<
+    fetch::fixed_point::FixedPoint<16, 16>, Signature(_C <= _alpha, _A, _B, _beta, _C),
+    Computes(_C <= _alpha * _A * T(_B) + _beta * _C), platform::Parallelisation::VECTORISE>;
+
+template class Blas<
```
baykaner, line 133: no dottranspose implementation for fp128?

libs/math/src/math/linalg/blas/gemm_tn_vector.cpp
```diff
@@ -106,6 +106,14 @@ template class Blas<float, Signature(_C <= _alpha, _A, _B, _beta, _C),
                     Computes(_C <= _alpha * T(_A) * _B + _beta * _C),
                     platform::Parallelisation::VECTORISE>;
 
+template class Blas<
+    fetch::fixed_point::FixedPoint<16, 16>, Signature(_C <= _alpha, _A, _B, _beta, _C),
+    Computes(_C <= _alpha * T(_A) * _B + _beta * _C), platform::Parallelisation::VECTORISE>;
+
+template class Blas<
```
baykaner, line 113: missing the fp128 implementation here?

libs/math/src/math/linalg/blas/gemm_tt_novector.cpp
```diff
@@ -102,6 +102,14 @@ template class Blas<float, Signature(_C <= _alpha, _A, _B, _beta, _C),
                     Computes(_C <= _alpha * T(_A) * T(_B) + _beta * _C),
                     platform::Parallelisation::NOT_PARALLEL>;
 
+template class Blas<
+    fetch::fixed_point::FixedPoint<16, 16>, Signature(_C <= _alpha, _A, _B, _beta, _C),
+    Computes(_C <= _alpha * T(_A) * T(_B) + _beta * _C), platform::Parallelisation::NOT_PARALLEL>;
+
+template class Blas<
```
baykaner, line 109: again?

libs/math/src/math/linalg/blas/gemm_tt_vector.cpp
```diff
@@ -110,6 +110,14 @@ template class Blas<float, Signature(_C <= _alpha, _A, _B, _beta, _C),
                     Computes(_C <= _alpha * T(_A) * T(_B) + _beta * _C),
                     platform::Parallelisation::VECTORISE>;
 
+template class Blas<
```
baykaner, line 113: and here?

libs/math/src/math/linalg/blas/gemv_t_vector.cpp
```diff
@@ -166,12 +166,26 @@ template class Blas<double, Signature(_y <= _alpha, _A, _x, _n, _beta, _y, _m),
 template class Blas<float, Signature(_y <= _alpha, _A, _x, _n, _beta, _y, _m),
                     Computes(_y <= _alpha * T(_A) * _x + _beta * _y),
                     platform::Parallelisation::VECTORISE>;
+template class Blas<
```
baykaner, line 169: and here also?

libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -753,7 +786,7 @@ TYPED_TEST(TranscendentalTest, Exp)
   TypeParam tolerance = TypeParam::TOLERANCE;
   for (size_t i{0}; i < N; i++)
   {
-    r                = static_cast<double>(random()) / static_cast<double>(RAND_MAX);
+    r                = static_cast<double>(std::rand()) / static_cast<double>(RAND_MAX);
```
baykaner, line 789: can you just reassure me that it's fine to use std::rand() here rather than our LFG/LCG seeded by a fixed seed (i.e. isn't it possible to generate different random number here on another machine and fail the test?)

baykaner (changes requested):

libs/math/src/math/linalg/blas/gemm_tn_vector.cpp
markos: fp128 does not -yet- support vectorisation, so it's intentionally omitted.

libs/math/src/math/linalg/blas/gemm_tt_vector.cpp
markos: same, no support for fp128 in vectorisation.

libs/math/src/math/linalg/blas/gemm_tt_novector.cpp
markos: this is an actual omission, fixed.

libs/math/include/math/tensor.hpp
markos: fixed.

libs/dmlf/tests/execution/basic_engine_tests.cpp
markos: It's adding two numbers, fp64_t(4.5) and fp32_t(5.5). The result was -erroneously- 9.5, but it can be explained, there was no proper conversion between the fp types before, now the conversion is taking place and the numbers are converted to the proper type and added, so fp32_t(5.5) is *also* 5.5 as an fp64_t. And the result is correctly 10.0. The original test was just using a wrong value.

libs/core/include/core/serializers/base_types.hpp
markos: fixed.

libs/math/src/math/linalg/blas/gemv_t_vector.cpp
markos: fp128 does not -yet- support vectorisation, so it's intentionally omitted.

libs/dmlf/tests/execution/basic_engine_tests.cpp
markos: fixed

libs/core/include/core/serializers/base_types.hpp
markos: fixed

libs/math/tests/unit/fixed_point/fixed_point.cpp
markos: It's possible, but these functions are well tested to be very robust and uniformly random, if our test fails then that would be an actual problem in Exp() and not the test suite.

libs/math/src/math/linalg/blas/gemm_nt_vector.cpp
markos: fp128 does not -yet- support vectorisation, so it's intentionally omitted.

JuanBesa (dismissed):
Looks good to me

libs/math/src/math/linalg/blas/gemm_nn_vector.cpp
markos: no, it's just that gcc8 complains on those particular definitions, other compilers pass.

markos: fixed.

baykaner (approved):

PR #1896, opened 2019-11-05
Dev/vm return array

JuanBesa:
This PR adds support for the following to the BasicExecutionEngine:

- Calling etch functions whose signature requires Array<primitive>0
- Calling etch functions whose function requires Array<Array<.....<Array<Primitive>...>>
- Calling etch functions that return Array<Primitive> or Array...<Array<Primitive>...>
- FixedPoint base 32 and 64


libs/dmlf/include/dmlf/execution/basic_vm_engine.hpp
```diff
@@ -60,20 +61,28 @@ class BasicVmEngine : public ExecutionEngineInterface
...
-
 private:
+  using Serializer = serializers::MsgPackSerializer;
+
+  struct ExecutionContext
+  {
+    ExecutionContext(VM *vm, Executable *executable);
+    ~ExecutionContext();
+
+    VM *vm_;
```
denis-trapeznikov, line 74: Ends in underscore yet is not declared private.

libs/dmlf/src/basic_vm_engine.cpp
```diff
@@ -389,117 +196,267 @@ bool BasicVmEngine::HasState(std::string const &name) const
...
-  case fetch::vm::TypeIds::UInt32:
-  case fetch::vm::TypeIds::Int64:
-  {
-    return VmVariant(ledgerVariant.As<int>(), typeId);
-  }
-  case fetch::vm::TypeIds::Float32:
+  auto serializationError = [&](std::string &&errorMessage) {
+    return ExecutionResult{LedgerVariant{},
+                           Error{Error::Stage::ENGINE, Error::Code::SERIALIZATION_ERROR,
+                                 "Error(" + id + ") in output after running. " + errorMessage},
```
denis-trapeznikov: Usually, there's no need in passing arguments by rvalue-ref, pass-by-value is more preferable. Note that `errorMessage` is actually an lvalue in lambda's body, `std::move(errorMessage)` should rather be used.

libs/dmlf/src/basic_vm_engine.cpp
```diff
@@ -34,8 +33,8 @@ namespace fetch {
 namespace dmlf {
 
 // This must match the type as defined in variant::variant.hpp
-using fp64_t = fetch::fixed_point::fp64_t;
-using fp32_t = fetch::fixed_point::fp32_t;
+using fp64_t = fixed_point::fp64_t;
+using fp32_t = fixed_point::fp32_t;
```
denis-trapeznikov: Nitpicking: simply `using fixed_point::fp(64|32)_t;`

libs/dmlf/src/basic_vm_engine.cpp
```diff
@@ -127,224 +126,34 @@ ExecutionResult BasicVmEngine::Run(Name const &execName, Name const &stateName,
 {
   if (!HasExecutable(execName))
   {
-    return EngineError(Error::Code::BAD_EXECUTABLE, "No executable " + execName);
+    return EngineError(Error::Code::BAD_EXECUTABLE, "Error: No executable " + execName);
```
denis-trapeznikov, line 129: Wouldn't N be better lower-cased?

libs/dmlf/src/basic_vm_engine.cpp
```diff
@@ -127,224 +126,34 @@ ExecutionResult BasicVmEngine::Run(Name const &execName, Name const &stateName,
...
-
-  // Convert and check function signature
-  auto const *func = exec->FindFunction(entrypoint);
-
+  auto &      exec  = executables_[execName];
+  auto &      state = states_[stateName];
+  auto const *func  = exec->FindFunction(entrypoint);
   if (func == nullptr)
   {
     return EngineError(Error::Code::RUNTIME_ERROR, entrypoint + " does not exist");
```
denis-trapeznikov: No "Error: " in errormsg?

libs/dmlf/src/basic_vm_engine.cpp
```diff
@@ -389,117 +196,267 @@ bool BasicVmEngine::HasState(std::string const &name) const
...
   {
-    return ledgerVariant.IsInteger();
+    for (auto const &p : params)
+    {
+      serializer << p;
+    }
   }
-  case fetch::vm::TypeIds::Float32:
-  case fetch::vm::TypeIds::Float64:
+  catch (std::exception &ex)
```
denis-trapeznikov: Catched by non-const ref, here and below.

libs/dmlf/src/basic_vm_engine.cpp
```diff
@@ -389,117 +196,267 @@ bool BasicVmEngine::HasState(std::string const &name) const
...
   {
-    return ledgerVariant.IsFloatingPoint();
+    return Error(
+        Error::Stage::ENGINE, Error::Code::SERIALIZATION_ERROR,
+        errorPrefix + "Serializing input before running: Threw error " + std::string(ex.what()));
```
denis-trapeznikov, line 227: Is `std::string(` needed here?
Btw, `errorPrefix` could be declared non-const above and then `move`d here, and on other sites.

libs/dmlf/src/basic_vm_engine.cpp
```diff
@@ -389,117 +196,267 @@ bool BasicVmEngine::HasState(std::string const &name) const
...
+  ExecutionContext executionContext(&vm, exec);
+  for (std::size_t i = 0; i < numParameters; ++i)
+  {
+    auto type_id = func->variables[i].type_id;
+    if (type_id <= vm::TypeIds::PrimitiveMaxId)
+    {
+      VmVariant param;
+
+      serializer >> param.primitive.i64;
+      param.type_id = type_id;
```
denis-trapeznikov: Technically, this is an UB.

libs/dmlf/src/basic_vm_engine.cpp
```diff
@@ -389,117 +196,267 @@ bool BasicVmEngine::HasState(std::string const &name) const
...
+    }
+    case fetch::vm::TypeIds::Int8:
+    case fetch::vm::TypeIds::UInt8:
+    case fetch::vm::TypeIds::Int16:
+    case fetch::vm::TypeIds::UInt16:
+    case fetch::vm::TypeIds::Int32:
+    case fetch::vm::TypeIds::UInt32:
+    case fetch::vm::TypeIds::Int64:
+    {
+      output = vmVariant.Get<int>();
```
denis-trapeznikov: UB.

libs/dmlf/src/basic_vm_engine.cpp
```diff
@@ -389,117 +196,267 @@ bool BasicVmEngine::HasState(std::string const &name) const
...
+    }
+    catch (...)
+    {
+      return serializationError(" Deserializing output after running. No details.");
+    }
+
+    if (output.IsArray())  // Convert inner type from int to fixedpoint if necessary
+    {
+      auto GetInnermostTypeId = [&vm](LedgerVariant const &ledgerVariant,
+                                      VmVariant const &    vmVariant) {
```
denis-trapeznikov: The opening brace should probably be placed in the next line, to be in-line with other formatting rules.

libs/dmlf/src/basic_vm_engine.cpp
```diff
@@ -389,117 +196,267 @@ bool BasicVmEngine::HasState(std::string const &name) const
...
+        }
+
+        auto result = vmVariant.type_id;
+        if (depth >= 1)
+        {
+          auto currentTypeId = vmVariant.type_id;
+          for (int i = 0; i < depth - 1; ++i)
+          {
+            currentTypeId = vm.GetTypeInfo(currentTypeId).template_parameter_type_ids[0];
+          }
```
denis-trapeznikov: This loop can be merged with the one above.

libs/dmlf/src/basic_vm_engine.cpp
```diff
@@ -389,117 +196,267 @@ bool BasicVmEngine::HasState(std::string const &name) const
...
+
+        return result;
+      };
+
+      auto const innermostType = GetInnermostTypeId(output, vmVariant);
+
+      if (innermostType == vm::TypeIds::Fixed64)
+      {
+        std::function<void(LedgerVariant &)> toFixed64;
+        toFixed64 = [&toFixed64](LedgerVariant &var) {
```
denis-trapeznikov: If a local function needs reference to itself for recursion, does it make sense to define it in the namespace (and `LedgerVariant` alongside)?

denis-trapeznikov (dismissed):

libs/dmlf/src/basic_vm_engine.cpp
JuanBesa: Why is it UB? Which part? I agree that it is unsafe, but this depends on the definition of typeIds that the VM uses.

libs/dmlf/include/dmlf/execution/basic_vm_engine.hpp
JuanBesa: Actually I think it shouldn't be a struct, so I'll change it to a class.

libs/dmlf/src/basic_vm_engine.cpp
JuanBesa: The std::string is needed in some of the others so I kept it as is for consistency. 
I prefer to declare it const as I think it declares the intent in a better way.

libs/dmlf/src/basic_vm_engine.cpp
JuanBesa: I agree in general, but in this case I know that the lambda will be called to at the end of the function, so I'll save the extra move. 
I added the move though, good catch.

libs/dmlf/src/basic_vm_engine.cpp
JuanBesa: Can you explain more?

libs/dmlf/src/basic_vm_engine.cpp
JuanBesa: I am not sure what you are saying here. Could you clarify?

denis-trapeznikov: Oh, sorry, disregard `LedgerVariant` in comment. I mean, perhaps it makes sense to define `ToFixed64()` outside of `PrepOutput()`, to get rid of `std::function` object only instantiated to be recursive. :)

JuanBesa: Oh, I see. Mmmm well it is just to get around the lambda not being able to defined as recursive.  I don't think it is too costly and I think it nicely keeps things local.

libs/dmlf/src/basic_vm_engine.cpp
JuanBesa: I think it is clearer as is.

lrahmani (dismissed):
LGTM, we'll need some refactoring later

libs/dmlf/src/basic_vm_engine.cpp
JuanBesa: Actually I think it is better if we can save it. I'll update that

lrahmani (approved):

denis-trapeznikov (approved):

PR #1897, opened 2019-11-05
Vm contract to contract

robertdickson:
Etch compiler and VM opcode support for contract-to-contract calls via contract interfaces and contract reference variables


troels-im (approved):
LGTM

PR #1898, opened 2019-11-05
Remove mnist dataloader

evsmithx:
MNIST data is now loaded using a helper function and TensorDataloader, which simplifies the interface in a lot of places.


libs/ml/examples/mnist/main.cpp
```diff
@@ -39,16 +40,16 @@ using SizeType   = fetch::math::SizeType;
...
 int main(int ac, char **av)
 {
   DataType                      learning_rate{0.01f};
   SizeType                      subset_size{100};
   SizeType                      epochs{10};
-  SizeType                      batch_size{10};
+  SizeType                      batch_size{13};
   fetch::ml::RegularisationType regulariser = fetch::ml::RegularisationType::L1;
-  DataType                      reg_rate{0.01f};
+  DataType                      reg_rate{0.00f};
```
MissingNO57: Why you changed it like that? Shouldn't it be then RegularisationType:None?

libs/dmlf/include/dmlf/distributed_learning/utilities/mnist_client_utilities.hpp
```diff
@@ -18,11 +18,12 @@
 //------------------------------------------------------------------------------
 
 #include "dmlf/distributed_learning/distributed_learning_client.hpp"
-#include "ml/dataloaders/mnist_loaders/mnist_loader.hpp"
+#include "ml/dataloaders/tensor_dataloader.hpp"
```
baykaner: I think with a little careful thought you could perhaps delete this file altogether?

The main difference between this and the boston_client_utilities is that Boston takes in a Tensor and MNist takes in a filename. If you invoke the mnist file reading utilities you've created in the example main.cpp files perhaps you can just invoke the Boston client utilities and be done with it? (you may need to make some other small changes, such as specifying the model architecture somewhere)

libs/ml/examples/mnist/main.cpp
```diff
@@ -39,16 +40,16 @@ using SizeType   = fetch::math::SizeType;
...
-using DataLoaderType = typename fetch::ml::dataloaders::MNISTLoader<TensorType, TensorType>;
+using DataLoaderType = typename fetch::ml::dataloaders::TensorDataLoader<TensorType, TensorType>;
 
 int main(int ac, char **av)
 {
   DataType                      learning_rate{0.01f};
   SizeType                      subset_size{100};
   SizeType                      epochs{10};
-  SizeType                      batch_size{10};
+  SizeType                      batch_size{13};
```
baykaner: why?

libs/ml/include/ml/meta/ml_type_traits.hpp
```diff
@@ -35,7 +35,6 @@ enum class OptimiserType : uint8_t
 enum class LoaderType : uint8_t
 {
   TENSOR,
-  MNIST,
```
baykaner, line 38: nice! i forgot these existed

libs/ml/include/ml/utilities/mnist_utilities.hpp
```diff
@@ -0,0 +1,146 @@
...
+  {
+    file.read(&label_char, 1);
+    labels.At(0, i) = static_cast<DataType>(static_cast<uint8_t>(label_char));
+  }
+
+  return labels;
+}
+
+template <typename TensorType>
+TensorType convert_labels_to_onehot(TensorType labels)
```
baykaner, line 133: right, since you have this functionality, it's not obvious why you need to include the one hot op in the graph in the example. Can't you just preprocess the labels with this and then not bother with the one-hot op? (nice that it works anyway though)

libs/vm-modules/include/vm_modules/ml/utilities/mnist_utilities.hpp
```diff
@@ -0,0 +1,40 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "ml/utilities/mnist_utilities.hpp"
```
baykaner, line 20: I'm not sure that you need any of these includes in the header?

libs/vm-modules/src/ml/utilities/mnist_utilities.cpp
```diff
@@ -0,0 +1,66 @@
...
+namespace vm_modules {
+
+namespace math {
+class VMTensor;
+}
+
+namespace ml {
+namespace utilities {
+
+fetch::vm::Ptr<fetch::vm_modules::math::VMTensor> load_mnist_images(
```
baykaner: a bit pedantic but you can drop some of the explicit namespacing in this file. everything is within the fetch scope so you don't need that. you can also drop the fetch::vm_modules:: namespacing

baykaner (changes requested):

libs/ml/examples/mnist/main.cpp
evsmithx: When I'm debugging a graph then using prime numbers is helpful because it makes it clear which parameters affect the size of a tensor! But it's debug code and I'll remove it.

libs/ml/include/ml/utilities/mnist_utilities.hpp
evsmithx: Yep, I'm doing that everywhere else!

libs/dmlf/include/dmlf/distributed_learning/utilities/mnist_client_utilities.hpp
```diff
@@ -47,8 +48,13 @@ std::shared_ptr<fetch::dmlf::distributed_learning::TrainingClient<TensorType>> M
       10u, 10u, fetch::ml::details::ActivationType::SOFTMAX);
 
   // Initialise DataLoader
+  auto mnist_images = fetch::ml::utilities::read_mnist_images<TensorType>(images);
+  auto mnist_labels = fetch::ml::utilities::read_mnist_labels<TensorType>(labels);
+  mnist_labels      = fetch::ml::utilities::convert_labels_to_onehot(mnist_labels);
```
MissingNO57: Even better might be to modify the graph to use Ops:OneHot, it would save us some memory,

libs/ml/examples/mnist_model/main.cpp
```diff
@@ -59,7 +60,12 @@ int main(int ac, char **av)
   model_config.graph_save_location = "/tmp/mnist_model";
 
   // setup dataloader with 20% test set
-  auto data_loader_ptr = std::make_unique<DataLoaderType>(av[1], av[2]);
+  auto mnist_images = fetch::ml::utilities::read_mnist_images<TensorType>(av[1]);
+  auto mnist_labels = fetch::ml::utilities::read_mnist_labels<TensorType>(av[2]);
+  mnist_labels      = fetch::ml::utilities::convert_labels_to_onehot(mnist_labels);
```
MissingNO57, line 65: Again we can be using node with ops:OneHot instead, please discuss with Khan.

libs/ml/include/ml/utilities/mnist_utilities.hpp
MissingNO57, line 133: This function would become redundant if you use ops::OneHot

MissingNO57 (changes requested):
Just a few suggestions, apart from that it looks great!

libs/vm-modules/include/vm_modules/ml/utilities/mnist_utilities.hpp
evsmithx: I'm using Module and VMTensor in the .cpp file, so I've left them in but tidied some other bits.

libs/dmlf/include/dmlf/distributed_learning/utilities/mnist_client_utilities.hpp
evsmithx: Haha, @baykaner literally just requested the opposite change! Less memory but more processing; which do we care about more?

libs/ml/examples/mnist/main.cpp
evsmithx: The model doesn't train with regularisation on, but you're right that I could remove it entirely.

libs/dmlf/include/dmlf/distributed_learning/utilities/mnist_client_utilities.hpp
baykaner: two arguments for getting my way:
1. it simplifies the graph, which means mnist tutorials can be more concise (i.e. take a look at tensorflow mnist tutorial - they're not adding any onehot op to the graph)
2. processing is probably more important than memory in general anyway (and anyway why repeatedly process something that you can do only once!?)

MissingNO57 (approved):
lgtm

libs/ml/examples/mnist/main.cpp
baykaner: fancy!

baykaner (approved):

PR #1900, opened 2019-11-05
Feature/genesis file updates

n-hutton:
Now that the genesis file creation is in ledger-api-py, it can cut down/make the E2E tests cleaner


scripts/fetchai_netutils/fetch/genesis/genesis.py
```diff
@@ -69,21 +68,14 @@ def create_record(address, balance, stake):
...
-    args = parse_commandline()
-
+def create_genesis_file(stake_percentage, max_cabinet_size, addresses):
     # build up the stake information
     individual_balance = TOTAL_SUPPLY // len(args.addresses)
-    individual_stake = (min(args.stake_percentage, 100)
+    individual_stake = (min(stake_percentage, 100)
                         * individual_balance) // 100
 
     max_cabinet = 0
```
denis-trapeznikov: Is `max_cabinet` still needed?

n-hutton: Yes. By the way, I accidentally included the file here, it has actually moved to ledger-api-py now

PR #1901, opened 2019-11-06
Benchmark aggregate signatures and implementation improvements

jinmannwong:
Add benchmarks for aggregate signatures used in notarisations and improve implementation to improve time for constructing and verifying aggregate signatures.


libs/crypto/src/mcl_dkg.cpp
```diff
@@ -400,14 +405,12 @@ PrivateKey SignatureAggregationCoefficient(PublicKey const &             notaris
 {
   PrivateKey coefficient;
 
+  std::string concatenated_keys;
+  concatenated_keys.reserve(48 + (cabinet_notarisation_keys.size() + 1) * 310);
```
n-hutton: Seems unclear where 48 and 310 are from? Could they be constexpr somewhere? I am not sure I understand the hygenic reuse comment

n-hutton (changes requested):
Minor change requested

libs/crypto/src/mcl_dkg.cpp
jinmannwong: The reserving of the 48 bytes was a figure David gave and the 310 is the size of each public key. I can make these const expressions. The hygenic reuse comment refers to the fact that the hash function (which is also used in the DKG) is being used here for a different purpose and to be completely safe we reserve some bytes at the front so there is no way two of the same strings could be hashed with the same hash function in the different contexts.

libs/crypto/benchmark/aggregate_signing.cpp
```diff
@@ -0,0 +1,235 @@
...
+    // Verify aggregate signature
+    auto aggregate_public_key = fetch::crypto::mcl::ComputeAggregatePublicKey(
+        aggregate_signature.second, notarisation_public_keys);
+    fetch::crypto::mcl::VerifySign(aggregate_public_key, msg, aggregate_signature.first, generator);
+  }
+}
+
+}  // namespace
+
+BENCHMARK(SignatureAggregationCoefficient)->RangeMultiplier(2)->Range(50, 500);
```
n-hutton, line 232: 👍

libs/crypto/src/mcl_dkg.cpp
```diff
@@ -31,7 +31,8 @@ namespace fetch {
 namespace crypto {
 namespace mcl {
 
-std::atomic<bool> details::MCLInitialiser::was_initialised{false};
+std::atomic<bool>  details::MCLInitialiser::was_initialised{false};
+constexpr uint16_t public_key_byte_size = 310;
```
n-hutton: This should be uppercase

n-hutton (changes requested):
Minor change

n-hutton (approved):

PR #1902, opened 2019-11-06
Compare not against genesis

denis-trapeznikov:
Throughout our code, `previous_hash` is compared a lot against `GENESIS_HASH` to identify a block as genesis one. Even putting aside that this is a really strong assumption, a better style would be to isolate this logic in a separate method.


libs/ledger/src/chain/block.cpp
```diff
@@ -90,11 +90,16 @@ void Block::UpdateDigest()
...
-  if (body.previous_hash != chain::GENESIS_DIGEST)
+  if (!IsGenesis())
   {
     body.timestamp = GetTime(clock_);
   }
 }
 
+bool Block::IsGenesis() const
+{
+  return body.previous_hash == chain::GENESIS_DIGEST;
```
ejfitzgerald, line 101: I suspect this check will not work with POS (which will become the default consensus mechanism shortly). Have a chat with @n-hutton about the best way to handle this.

ejfitzgerald (commented):
I like the idea of the change, there is a complication with the way that PoS works in regards to the `chain::GENESIS_BLOCK` which we would need to address. 

I would have a chat with @n-hutton about this

libs/ledger/src/chain/block.cpp
denis-trapeznikov: This is as applied to current master so once the definition of genesis changes only this method would presumably need to change.

ejfitzgerald: Oh yes - this is a great change. It is just at the moment on `master` the behaviour would not be correct if you used the `-pos` flag on the command line.

Actually I think this is highlighting a larger bug. So this is excellent!

ejfitzgerald (approved):

PR #1906, opened 2019-11-07
Revert "Use hex encoding for digest (#1839)"

n-hutton:
This reverts commit 300464af3260a77041ba5283e8a087257b0f3f89.

Broken in latest python api


ejfitzgerald:
So I think that we keep this change but make the corresponding change in the ledger SDK. I have talked to @wkaluza and we have a plan

PR #1907, opened 2019-11-07
Adaptive Bloom filter based on valid_until

wkaluza:
We currently use single main chain Bloom filter regardless of chain size. This PR introduces a progressive mechanism where a two partially overlapping Bloom filters coexist and only deal with txs for which valid_until is near (within 20-40k blocks) the head of the chain.


libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -233,16 +234,15 @@ class MainChain
...
-  bool const                        enable_bloom_filter_;
-  telemetry::GaugePtr<std::size_t>  bloom_filter_queried_bit_count_;
-  telemetry::CounterPtr             bloom_filter_query_count_;
-  telemetry::CounterPtr             bloom_filter_positive_count_;
-  telemetry::CounterPtr             bloom_filter_false_positive_count_;
+  mutable References references_;
+  TipsMap            tips_;          ///< Keep track of the tips
+  HeaviestTip        heaviest_;      ///< Heaviest block/tip
+  LooseBlockMap      loose_blocks_;  ///< Waiting (loose) blocks
+  mutable std::unordered_map<Block::Index, std::unique_ptr<BasicBloomFilter>> bloom_filters_;
```
denis-trapeznikov: `std::unique_ptr` is probably excess here (I think it even was with the older Bloom filter).

libs/ledger/include/ledger/miner/transaction_layout_queue.hpp
```diff
@@ -44,15 +44,16 @@ class TransactionLayoutQueue
...
+  ConstIterator                         cbegin() const;
+  ConstIterator                         begin() const;
+  Iterator                              begin();
+  ConstIterator                         cend() const;
+  ConstIterator                         end() const;
+  Iterator                              end();
+  std::size_t                           size() const;
+  bool                                  empty() const;
+  DigestSet const &                     digests() const;
+  std::unordered_set<TransactionLayout> TxLayouts() const;
```
denis-trapeznikov: Worth a typedef, to be in-line with those above?

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1549,15 +1547,15 @@ DigestSet MainChain::DetectDuplicateTransactions(BlockHash const &starting_hash,
   }
 
   DigestSet potential_duplicates{};
-  for (auto const &digest : transactions)
+  for (auto const &tx_layout : transactions)
   {
-
-    std::pair<bool, std::size_t> const result = bloom_filter_->Match(digest);
+    std::pair<bool, std::size_t> const result =
```
denis-trapeznikov, line 1655: Too huge a return type, would `auto` be better here?

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1625,5 +1617,18 @@ constexpr char const *ToString(BlockStatus status)
...
 }
 
+BasicBloomFilter &MainChain::SelectBloomFilter(chain::TransactionLayout const &tx_layout) const
+{
+  Block::Index const index     = tx_layout.valid_from() / 1000000;
+  auto               filter_it = bloom_filters_.find(index);
+
+  if (filter_it == bloom_filters_.end())
+  {
+    bloom_filters_.emplace(std::make_pair(index, std::make_unique<BasicBloomFilter>()));
```
denis-trapeznikov: `std::make_pair` is not needed here.

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1625,5 +1617,18 @@ constexpr char const *ToString(BlockStatus status)
   }
 }
 
+BasicBloomFilter &MainChain::SelectBloomFilter(chain::TransactionLayout const &tx_layout) const
+{
+  Block::Index const index     = tx_layout.valid_from() / 1000000;
```
denis-trapeznikov: This actually divides transactions into continuous 'generations' one million blocks each. Close yet distinct, in terms of block number, transactions map to the same Bloom bucket.
Would it possibly make sense to better separate them?

denis-trapeznikov (dismissed):

libs/ledger/src/chain/main_chain.cpp
wkaluza: A Bloom filter is inherently probabilistic: there isn't a scheme that would guard us completely from 'close' transactions sharing a bucket. That said, this design has a bigger problem, so I'm marking it as WIP again for the moment. I may be able to do something about bucketing along the way.

libs/ledger/include/ledger/chain/main_chain.hpp
wkaluza: It was there as a quick workaround, because the BF is not moveable. It probably should be - fixed.

libs/ledger/include/ledger/miner/transaction_layout_queue.hpp
wkaluza: Done.

libs/ledger/src/chain/main_chain.cpp
wkaluza: Done.

libs/ledger/src/chain/main_chain.cpp
wkaluza: Done.

wkaluza:
Closing temporarily due to mac-mini disk space issues.

libs/chain/include/chain/transaction_validity_period.hpp
```diff
@@ -0,0 +1,55 @@
...
+                                      ? tx.valid_until() - DEFAULT_TX_VALIDITY_PERIOD
+                                      : 0u;
+  auto const valid_from = tx.valid_from() == 0 ? default_valid_from : tx.valid_from();
+
+  if (tx.valid_until() - valid_from > MAXIMUM_TX_VALIDITY_PERIOD)
+  {
+    return Transaction::Validity::INVALID;
+  }
+
+  if (tx.valid_until() <= block_index)
```
ejfitzgerald:
```suggestion
  if (tx.valid_until() < block_index)
```

I believe the rest of the system checks the validity range in the form [valid_from, valid_util). Please double check however.

libs/core/include/core/byte_array/const_byte_array.hpp
```diff
@@ -98,6 +99,17 @@ class ConstByteArray
     assert(start_ + length_ <= data_.size());
   }
 
+  explicit ConstByteArray(std::istream &in)
+  {
+    in.seekg(0, std::ios::end);
```
ejfitzgerald, line 108: Worth adding the fail bits check in the case that we found here or not worth it do you think?

scripts/fetchai_netutils/fetch/cluster/instance.py
```diff
@@ -1,8 +1,7 @@
-import os
-import subprocess
 import fnmatch
 import json
-
+import os
```
ejfitzgerald, line 3: Out of interest there are a number of updates on these javascript files is this because of an updated set of formatting rules or something else (pycharm) ?

ejfitzgerald (changes requested):
Looks really good. The only thing I would think would be a good addition is to add a unit test for the progressive bloom filter and move the overlap period to the constructor so we can test than. What do you think?

scripts/fetchai_netutils/fetch/cluster/instance.py
wkaluza: Automated changes by pycharm, nothing to worry about. There was a merge that's gone a bit wonky.

libs/chain/include/chain/transaction_validity_period.hpp
wkaluza: I think this is now the only place where this check is performed.

I checked with the Input Data Formats doc, you are correct: valid_until is defined as the block number from which the tx becomes invalid.

ejfitzgerald (dismissed):

ejfitzgerald (approved):

PR #1908, opened 2019-11-07
Manage contract context using RAII

wkaluza:
The current model of executing a contract assumes the following order of events:

 * Attach ledger context to contract
 * Dispatch transaction
 * Detach context

As things stand, `Detach` will not be called if dispatching the transaction throws an exception. This PR ensures that `Detach` is always called by delegating the `Attach`-`Detach` calls to a RAII wrapper.


libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -291,5 +293,23 @@ StateAdapter::Status Contract::SetStateRecord(T const &record, ConstByteArray co
   return state().Write(std::string{key}, data.pointer(), data.size());
 }
 
+class ContractAttachHelper
```
denis-trapeznikov: NB: `Helper` is not very descriptive here. Even smth like `Attacher` would probably be more readable.

libs/ledger/benchmark/executor_bench.cpp
```diff
@@ -80,9 +80,9 @@ void Executor_BasicBenchmark(benchmark::State &state)
 
     TokenContract tokens{};
 
-    tokens.Attach({&tokens, tx->contract_address(), &adapter, 0});
+    fetch::ledger::ContractContext      context{&tokens, tx->contract_address(), &adapter, 0};
+    fetch::ledger::ContractAttachHelper raii(tokens, context);
```
denis-trapeznikov: One of minor issues I see in this implementation is that context attachment is done in two statements in most cases. Even though one of them is simply an object instantiation, still RAII is usually perceived as something more compact, one statement to lock context until the end of the block.
One possible alternative would be something like
```cpp
template<class... Args> ContractAttachHelper(Contract &contract, Args &&...args)
    : contract_(contract)
{
    contract_.Attach(ContractContext(std::forward<Args>(args)...));
}
```
Next, contexts here in most use-cases are often short-living locals. `Contract::Attach()` could attach to them by-value rather than copy each time.

denis-trapeznikov (changes requested):

libs/ledger/include/ledger/chaincode/contract.hpp
wkaluza: Done.

libs/ledger/benchmark/executor_bench.cpp
wkaluza: I switched to passing by value; I disagree on the template constructor - it needlessly increases coupling between the `Context` and the `Attacher`. I would rather keep the `Attacher` class simple, along with the stepwise constructing and then attaching a context. To do otherwise seems to sacrifice clarity for brevity.

ejfitzgerald (commented):
Looks good to me, however, I was wondering, do you think that it make sense to combine the attacher with the context. It looks like are always used together?

wkaluza:
They are used together, but they represent distinct concepts. I would prefer to keep Context as a simple struct, and have the Attacher do what it's doing now. This reduces coupling and separates concerns better.

wkaluza:
Closing temporarily to free up mac-mini disk space.

libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -291,5 +292,15 @@ StateAdapter::Status Contract::SetStateRecord(T const &record, ConstByteArray co
   return state().Write(std::string{key}, data.pointer(), data.size());
 }
 
+class ContractContextAttacher
```
n-hutton: Could we have the class in contract_context_attacher.hpp? Optimally with a class-level comment on its utility

n-hutton (changes requested):
Looks good, just a minor suggestion on file consistency

libs/ledger/include/ledger/chaincode/contract.hpp
wkaluza: Done.

n-hutton (approved):
LGTM

denis-trapeznikov (approved):

PR #1909, opened 2019-11-07
Sparse gradient for Embedding updates

MissingNO57:


libs/ml/include/ml/core/graph.hpp
```diff
@@ -207,6 +212,13 @@ class Graph
   //  void RecursiveApply(ValType &val,
   //                      void (Graph<TensorType>::*subgraph_func)(ValType &) const) const;
```
baykaner: commented out code?

libs/ml/include/ml/core/graph.hpp
```diff
@@ -1157,5 +1239,39 @@ void Graph<TensorType>::RecursiveApply(ValType &val, GraphFunc graph_func) const
   }
 }
 
+template <typename TensorType>
```
baykaner, line 1254: function comment might be nice

libs/ml/include/ml/optimisation/lazy_adam_optimiser.hpp
```diff
@@ -0,0 +1,371 @@
...
+
+/**
+ * LazyAdam is a variant of the Adam optimizer that handles sparse updates more efficiently.
+ * The original Adam algorithm maintains moving-average accumulators for each trainable variable;
+ * the accumulators are updated at every step. This class provides lazier handling of gradient
+ * updates for sparse variables. It only updates moving-average accumulators for sparse variable
+ * indices that appear in the current batch, rather than updating the accumulators for all indices.
+ * Compared with the original Adam optimizer, it can provide large improvements in model training
+ * throughput for some applications. However, it provides slightly different semantics than the
+ * original Adam algorithm, and may lead to different empirical results.
```
baykaner, line 41: might as well add the URL for your reference in this comment

libs/ml/include/ml/optimisation/lazy_adam_optimiser.hpp
```diff
@@ -0,0 +1,371 @@
...
+  , sparsity_threshold_(sparsity_threshold)
+  , epsilon_(epsilon)
+{
+  Init();
+}
+
+// private
+
+template <class T>
+void LazyAdamOptimiser<T>::ApplyLogic(SizeType batch_size, TensorType &gradient_tensor,
```
baykaner, line 137: apply logic means what exactly?
Please add a function comment describing the responsibility of this function

libs/ml/include/ml/optimisation/lazy_adam_optimiser.hpp
```diff
@@ -0,0 +1,371 @@
...
+
+    else
+    {
+      // Sparse apply gradient
+      // if number_of_rows_to_update*sparsity_threshold_ <= total_rows
+
+      for (SizeType update_index : rows.at(rows.size() - 1))
+      {
+
+        auto       gradient_slice        = gradient_it->Slice(update_index, 1);
```
baykaner: I guess all this slicing and copying really eats into the benefits of the sparse update. Presumably there is no easy way to:
1. use views, and
2. reduce copying

libs/ml/include/ml/optimisation/lazy_adam_optimiser.hpp
baykaner: e.g. updates momenta?

libs/ml/include/ml/optimisation/lazy_adam_optimiser.hpp
```diff
@@ -0,0 +1,371 @@
...
+ * the accumulators are updated at every step. This class provides lazier handling of gradient
+ * updates for sparse variables. It only updates moving-average accumulators for sparse variable
+ * indices that appear in the current batch, rather than updating the accumulators for all indices.
+ * Compared with the original Adam optimizer, it can provide large improvements in model training
+ * throughput for some applications. However, it provides slightly different semantics than the
+ * original Adam algorithm, and may lead to different empirical results.
+ * @tparam T TensorType
+ */
+template <class T>
+class LazyAdamOptimiser : public Optimiser<T>
```
baykaner: doesn't it make more sense to inherit from AdamOptimiser and then just override the relevant methods?

libs/ml/include/ml/optimisation/lazy_adam_optimiser.hpp
```diff
@@ -0,0 +1,371 @@
...
+    ++gradient_it;
+    ++trainable_it;
+  }
+
+  // calling apply gradients on the graph ensures that the node caches are reset properly
+  this->graph_->ApplySparseGradients(this->gradients_, rows);
+}
+
+template <class T>
+void LazyAdamOptimiser<T>::ResetCache()
```
baykaner: seems like this method would be redundant if you inherit from AdamOptimiser instead of Optimiser

libs/ml/include/ml/optimisation/sgd_optimiser.hpp
```diff
@@ -57,6 +58,9 @@ class SGDOptimiser : public Optimiser<T>
   }
 
 private:
+  // ApplyGradientSparse if number_of_rows_to_update*sparsity_threshold_ <= total_rows
+  SizeType sparsity_threshold_ = 4;
```
baykaner, line 62: it would be nice to have more explanation about this. Is this number empirically derived from benchmarking results? If so it would be good to point that out and state which benchmark illustrates this

libs/ml/tests/unit/optimisation/sparse_optimisers.cpp
```diff
@@ -0,0 +1,130 @@
...
+  EXPECT_NEAR(static_cast<double>(weights[0].At(4, 4)), -0.15662828390486538,
+              static_cast<double>(fetch::math::function_tolerance<DataType>()) *
+                  static_cast<double>(gt.size()));
+  EXPECT_NEAR(static_cast<double>(weights[0].At(8, 32)), -0.02471873932518065,
+              static_cast<double>(fetch::math::function_tolerance<DataType>()) *
+                  static_cast<double>(gt.size()));
+  EXPECT_NEAR(static_cast<double>(weights[0].At(0, 9)), 0.082687103189527988,
+              static_cast<double>(fetch::math::function_tolerance<DataType>()) *
+                  static_cast<double>(gt.size()));
+}
```
baykaner, line 200: perhaps i missed it but do you have anything that tests that the indices not used in a particular batch did not get updated? (i.e., that it's really sparse)

libs/ml/include/ml/utilities/sparse_tensor_utilities.hpp
```diff
@@ -0,0 +1,72 @@
...
+ */
+template <class TensorType>
+void SparseAdd(TensorType const &src, TensorType &dst,
+               std::unordered_set<fetch::math::SizeType> const &update_rows,
+               fetch::math::SizeType                            sparsity_threshold = 4)
+{
+  using SizeType = fetch::math::SizeType;
+
+  // Normal apply
+  // if number_of_rows_to_update*sparsity_threshold_ > total_rows
```
baykaner: can you put a space between variable names and * in the comment for readability please

baykaner (changes requested):
really nice work. a few minor things mostly

libs/ml/benchmark/embeddings/embeddings.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+  SizeType batch_size           = B;
+  SizeType embedding_dimensions = S;
+  SizeType n_datapoints         = D;
+  SizeType n_epochs             = E;
+
+  auto learning_rate = DataType{0.1f};
+
+  // Prepare data and labels
+  TensorType data({1, batch_size});
```
baykaner, line 82: it's actually possible to access multiple embeddings for each datapoint; but that's probably not important for now

libs/ml/benchmark/embeddings/embeddings.cpp
```diff
@@ -0,0 +1,204 @@
...
+  data.FillUniformRandomIntegers(0, static_cast<int64_t>(n_datapoints));
+  gt.FillUniformRandom();
+
+  for (auto _ : state)
+  {
+    // make a graph
+    std::string input_name;
+    std::string label_name;
+    std::string error_name;
+    auto        g = PrepareTestGraph<TensorType>(embedding_dimensions, n_datapoints, input_name,
```
baykaner: you seem to be including graph construction inside the benchmarking loop. Do you actually want to be benchmarking this?

libs/ml/benchmark/embeddings/embeddings.cpp
```diff
@@ -0,0 +1,204 @@
...
+  label_name = g->template AddNode<fetch::ml::ops::PlaceHolder<TypeParam>>("", {});
+  error_name = g->template AddNode<fetch::ml::ops::MeanSquareErrorLoss<TypeParam>>(
+      "Error", {output_name, label_name});
+
+  return g;
+}
+
+template <typename T, fetch::math::SizeType B, fetch::math::SizeType S, fetch::math::SizeType D,
+          fetch::math::SizeType E>
+void BM_Setup_And_Train_Embeddings_Adam(benchmark::State &state)
```
baykaner: please add a function comment here explaining the point of this benchmark. Is it to get an idea of the overall time cost of training a simple embeddings model under different hyperparameters?

libs/ml/benchmark/embeddings/embeddings.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include "benchmark/benchmark.h"
+
+#include <memory>
+#include <string>
+
+//////////////////////////
+/// reusable functions ///
+//////////////////////////
+
+template <typename TypeParam>
```
baykaner, line 47: probably we should put this all inside a fetch/ml/benchmark namespace

libs/ml/benchmark/embeddings/embeddings.cpp
```diff
@@ -0,0 +1,204 @@
...
+    // Do optimisation
+    for (SizeType i = 0; i < n_epochs; ++i)
+    {
+      data.FillUniformRandomIntegers(0, static_cast<int64_t>(n_datapoints));
+      optimiser.Run({data}, gt);
+    }
+  }
+}
+
+BENCHMARK_TEMPLATE(BM_Setup_And_Train_Embeddings_Adam, float, 1, 500, 10000, 10)
```
baykaner: maybe write a nested for-loop to try out these conditions? there's a lot of them!

libs/ml/benchmark/embeddings/embeddings.cpp
```diff
@@ -0,0 +1,204 @@
...
+  using TensorType = fetch::math::Tensor<DataType>;
+
+  fetch::SetGlobalLogLevel(fetch::LogLevel::ERROR);
+
+  SizeType batch_size           = B;
+  SizeType embedding_dimensions = S;
+  SizeType n_datapoints         = D;
+  SizeType n_epochs             = E;
+
+  auto learning_rate = DataType{0.1f};
```
baykaner: maybe can remove some code repeated between these two benchmarks with a free function?

baykaner (changes requested):

baykaner (approved):

PR #1910, opened 2019-11-07
Improvement/clean genesis

denis-trapeznikov:
In this commit, genesis block receives a cleaner semantics perhaps. GENESIS_HASH, which is not a constant, is now used as a true genesis block hash, and that latter's previous is simply filled with zeroes.


libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -319,11 +319,11 @@ BlockCoordinator::State BlockCoordinator::OnSynchronising()
   FETCH_UNUSED(current_dag_epoch);
 
   // initial condition, the last processed block is empty
-  if (chain::GENESIS_DIGEST == last_processed_block)
+  if (chain::ZERO_HASH == last_processed_block)
```
ejfitzgerald: I suspect that these should probably be genesis digest. I think that in the non-proof of stake configuration we should probably update the genesis block hash too

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1518,7 +1518,7 @@ MainChain::BlockHash MainChain::GetHeadHash()
 
 void MainChain::SetHeadHash(BlockHash const &hash)
 {
-  assert(hash.size() == 32);
+  assert(hash.size() == chain::HASH_SIZE);
```
ejfitzgerald, line 1558: 👍

PR #1918, opened 2019-11-08
Remove block body concept

n-hutton:
The block has a body struct that serves no obvious use, related to ticket LDGR-9


libs/ledger/src/chain/block.cpp
```diff
@@ -34,7 +34,7 @@ Block::Block() = default;
 bool Block::operator==(Block const &rhs) const
 {
   // Invalid to compare blocks with no block hash
-  return (!this->body.hash.empty()) && (this->body.hash == rhs.body.hash);
+  return (!this->hash.empty()) && (this->hash == rhs.hash);
```
denis-trapeznikov, line 37: Can this statement be changed into something more compact, like `return !hash.empty() && hash == rhs.hash;`? >_<

ejfitzgerald (approved):

PR #1920, opened 2019-11-08
Model conv layers etch

baykaner:
- implements conv1d & conv2d layers for model in etch


MissingNO57 (approved):
LGTM, we should also add support for MaxPooling layer.

PR #1923, opened 2019-11-08
Improvement/ stutter block removal

jinmannwong:
Add capability for main chain to remove stutter blocks (malicious in consensus as it means one miner has produced multiple blocks in one blocks round), identified as blocks of equal height and weight, from tips and preventing them from being tips in the future.


libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -90,6 +91,9 @@ class MainChain
   using BlockHash            = Block::Hash;
   using BlockHashes          = std::vector<BlockHash>;
   using BlockHashSet         = std::unordered_set<BlockHash>;
+  using BlockNumber          = uint64_t;
```
n-hutton, line 95: 👍

libs/ledger/src/consensus/consensus.cpp
```diff
@@ -370,6 +389,13 @@ void Consensus::UpdateCurrentBlock(Block const &current)
     return;
   }
 
+  // If previous block is stutter block in main chain then delete notarisations for this block
+  // to prevent peers from building on this block
+  if (notarisation_ && chain_.IsStutterBlock(current_block_.block_number, current_block_.weight))
+  {
+    notarisation_->RemoveNotarisation(current_block_.block_number, current_block_.hash);
```
n-hutton, line 413: Presumably this implies if everyone in the network were to accidentally stutter, it would never progress (probably fine)

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -621,6 +630,24 @@ bool MainChain::GetPathToCommonAncestor(Blocks &blocks, BlockHash tip, BlockHash
   return success;
 }
 
+/**
```
n-hutton, line 685: 👍

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -621,6 +630,24 @@ bool MainChain::GetPathToCommonAncestor(Blocks &blocks, BlockHash tip, BlockHash
...
+ * Return whether, for a given block number and weight, whether different blocks with the same
+ * weight have been seen
+ *
+ * @param block_number Block number being queried
+ * @param block_weight Weight being queried
+ * @return Whether pair (block_number, block_weight) pair correspond to stutter blocks
+ */
+bool MainChain::IsStutterBlock(BlockNumber block_number, BlockWeight block_weight) const
+{
+  if (stutter_blocks_.find(block_number) == stutter_blocks_.end() ||
```
n-hutton: I think this could become:

```
  if(stutter_blocks_[block_number].find(block_weight) == stutter_blocks_[block_number].end())
  {
    return false;
  }

  return stutter_blocks_[block_number][block_weight] == true;

```

n-hutton: This would possibly create an empty map at block_number - doesn't seem like it would be an issue though

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -893,7 +926,34 @@ void MainChain::WriteToFile()
       IntBlockPtr current_file_head = std::make_shared<Block>();
       IntBlockPtr block_head        = block;
 
-      LoadBlock(GetHeadHash(), *current_file_head);
+      BlockHash head_hash = GetHeadHash();
```
n-hutton, line 1003: Can you explain the reason for this? Maybe IRL

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -940,18 +1000,18 @@ void MainChain::WriteToFile()
  */
 void MainChain::TrimCache()
 {
-  static const uint64_t CACHE_TRIM_THRESHOLD = 2 * chain::FINALITY_PERIOD;
+  static const BlockNumber CACHE_TRIM_THRESHOLD = 2 * chain::FINALITY_PERIOD;
```
n-hutton: Can this be a static constexpr of MainChain?

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -992,6 +1052,18 @@ void MainChain::TrimCache()
...
+  auto stutter_map_back = stutter_blocks_.rbegin();
+  if (stutter_map_back != stutter_blocks_.rend() && stutter_map_back->first > CACHE_TRIM_THRESHOLD)
+  {
+    BlockNumber const stutter_threshold = stutter_map_back->first - CACHE_TRIM_THRESHOLD;
+    auto              stutter_map_iter  = stutter_blocks_.begin();
+    while (stutter_threshold >= stutter_map_iter->first)
+    {
+      stutter_map_iter = stutter_blocks_.erase(stutter_map_iter);
+    }
+  }
```
n-hutton, line 1114: Could this be:

```
// Cache trim must be positive and stutter blocks must be ordered
while(stutter_blocks_.size() > CACHE_TRIM_THRESHOLD)
{
  stutter_blocks_.erase(stutter_blocks_.begin());
}
```

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1105,6 +1177,33 @@ bool MainChain::UpdateTips(IntBlockPtr const &block)
   assert(block->total_weight != 0);
   assert(block->block_number != 0);
 
+  if (enable_stutter_removal_)
+  {
+    // if tip is from miner who has previously produced a block for this round
+    // do not add new block to tips
+    if (stutter_blocks_.find(block->block_number) != stutter_blocks_.end() &&
```
n-hutton: Could this be something like:

```
// Update accordingly depending on whether this is the first time or more
if (stutter_blocks_[block->block_number].find(block->weight) == stutter_blocks_[block->block_number].end())
{
  stutter_blocks_[block->block_number][block->weight] = false;
}
else
{
  stutter_blocks_[block->block_number][block->weight] = true;
}

if(stutter_blocks_[block->block_number][block->weight])
{
  // Remove any existing tips with same height and weight
  for (auto const &tip : tips_)
  {
    if (tip.second.block_number == block->block_number && tip.second.weight == block->weight)
    {
      auto old_heaviest = heaviest_;
      RemoveTip(std::make_shared<Block>(*GetBlock(tip.first)));
      return old_heaviest.hash == heaviest_.hash;
    }
  }
}
```

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1105,6 +1177,33 @@ bool MainChain::UpdateTips(IntBlockPtr const &block)
...
+      stutter_blocks_.at(block->block_number).at(block->weight) = true;
+
+      // Remove any existing tips with same height and weight
+      for (auto const &tip : tips_)
+      {
+        if (tip.second.block_number == block->block_number && tip.second.weight == block->weight)
+        {
+          auto old_heaviest = heaviest_;
+          RemoveTip(std::make_shared<Block>(*GetBlock(tip.first)));
+          return old_heaviest.hash == heaviest_.hash;
```
n-hutton, line 1251: This should return false here, I think (attempting to add a stutter block could only give equal or worse weighted tip). It is also a return in a loop - the comment should indicate that this is intentional and there should be only one tip with the same height and weight if coded correctly.

This RemoveTip looks like you could change the function signature to avoid constructing a smart pointer in-place

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1205,6 +1304,19 @@ BlockStatus MainChain::InsertBlock(IntBlockPtr const &block, bool evaluate_loose
   // we expect only non-loose blocks here
   assert(!block->is_loose);
 
+  // if removing stutter blocks then need to ensure weight to correct
```
n-hutton: I think that the blocks will be checked when they arrive from the network, the main chain can rely on the weight being correctly populated

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1352,12 +1464,79 @@ bool MainChain::AddTip(IntBlockPtr const &block)
 {
   FETCH_LOCK(lock_);
 
+  if (enable_stutter_removal_)
```
n-hutton, line 1505: This is a bit of a weird one - for whatever reason, this fn is only called with the genesis block, and tips are manually added elsewhere (not ideal - should probably be refactored in future).

Is the flow not such that you identify stutter blocks and instead of adding them as a tip, you remove any other tips that are stutters?

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1352,12 +1464,79 @@ bool MainChain::AddTip(IntBlockPtr const &block)
...
 
   return DetermineHeaviestTip();
 }
 
+/**
+ * Removes a new tip for the given block and inserts in previous
+ *
+ * @param block The block for the tip to be removed
+ */
+bool MainChain::RemoveTip(IntBlockPtr const &block)
```
n-hutton: I feel like this could call a function GetFirstNonStutterBlock(&block) which would walk down the chain starting at block X and return the first non stutter block. You set this as a tip if it is not already (might create a tip where there is no actual fork but that is ok... I think). You could also reuse this walk in the loop in ReindexTips, I think.

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1442,23 +1621,70 @@ bool MainChain::ReindexTips()
...
-        (total_weight == max_total_weight && block_number == max_block_number &&
-         weight == max_weight && hash > max_hash))
+    auto &      block{*block_entry.second};
+    BlockWeight total_weight{block.total_weight};
+    BlockWeight weight{block.weight};
+    BlockNumber block_number{block.block_number};
+
+    // check if block is stutter block and if so loop until a non-stutter block is found and replace
+    // as tip
+    bool replaced{true};
```
n-hutton: As above comment

libs/ledger/src/consensus/consensus.cpp
```diff
@@ -159,7 +169,7 @@ Block GetBeginningOfAeon(Block const &current, MainChain const &chain)
 
   // Walk back the chain until we see a block specifying an aeon beginning (corner
   // case for true genesis)
-  while (!ret.block_entropy.IsAeonBeginning() && current.block_number != 0)
+  while (!ret.block_entropy.IsAeonBeginning() && ret.block_number != 0)
```
n-hutton, line 173: Good catch

libs/ledger/src/consensus/consensus.cpp
```diff
@@ -197,17 +207,32 @@ bool Consensus::VerifyNotarisation(Block const &block) const
   return true;
 }
 
-uint64_t Consensus::GetBlockGenerationWeight(Block const &previous, chain::Address const &address)
+Consensus::WeightedQual QualWeightedByEntropy(Consensus::BlockEntropy::Cabinet const &cabinet,
```
n-hutton, line 227: Did anything actually change here or is the diff confused?

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1442,23 +1621,70 @@ bool MainChain::ReindexTips()
...
+              break;
+            }
+          }
+        }
+      }
+    }
+
+    if (replaced)
+    {
+      new_tips[hash] = Tip{total_weight, weight, block_number};
```
n-hutton: Just below this, there is:

```
      // check if this tip is the current heaviest
      if (total_weight > max_total_weight ||
          (total_weight == max_total_weight && block_number > max_block_number) ||
          (total_weight == max_total_weight && block_number == max_block_number &&
           weight > max_weight) ||
          (total_weight == max_total_weight && block_number == max_block_number &&
           weight == max_weight && hash > max_hash))
```

this should probably be refactored - I think the whole function could become something of the form:

```
// Clear old tips and repopulate
tips_.clear();
std::set<Block> block_tips;

for (auto const &block_entry : block_chain_)
{
  // Loose blocks can't be tips
  if (block_entry.second->is_loose)
  {
    continue;
  }

  // Blocks with a forward reference can't be tips (exception for stutter blocks)
  if(HasForwardRef(block_entry))
  {
    continue;
  }

  // If this is a stutter block, it walk down until we find a non-stutter to use as the tip (otherwise insert block_entry)
  block_tips.insert(GetFirstNonStutter(block_entry));
}

// just for safety
if(tips_.empty())
{
  FETCH_LOG_WARN(LOGGING_NAME, "tips empty somehow!");
  AddTip(genesis);
}

// Sanity check - is heaviest_ in tips_?

// Loop here populating tips_ with block_tips
//...

// Set heaviest_
}
```

libs/ledger/src/protocols/notarisation_service.cpp
```diff
@@ -384,6 +384,21 @@ NotarisationService::AggregateSignature NotarisationService::GetAggregateNotaris
...
 }
 
+void NotarisationService::RemoveNotarisation(BlockNumber const &block_number,
+                                             BlockHash const &  block_hash)
+{
+  FETCH_LOCK(mutex_);
+
+  if (notarisations_being_built_.find(block_number) != notarisations_being_built_.end())
+  {
+    notarisations_being_built_[block_number].erase(block_hash);
```
n-hutton: Would it be ok to have an empty map allowing just:

```
    notarisations_being_built_[block_number].erase(block_hash);
    notarisations_built_[block_number].erase(block_hash);
```

Looks like these aren't trimmed at any point - might be good to do similar to in the block coordinator (unless I misunderstand and they are trimmed via removenotarisation)

libs/ledger/tests/notarisation/notarisation_tests.cpp
```diff
@@ -82,7 +82,7 @@ struct NotarisationNode
     , reactor{"ReactorName" + std::to_string(index)}
     , muddle_certificate{CreateNewCertificate()}
     , muddle{muddle::CreateMuddleFake("Test", muddle_certificate, network_manager, "127.0.0.1")}
-    , chain{false, ledger::MainChain::Mode::IN_MEMORY_DB}
+    , chain{false, ledger::MainChain::Mode::IN_MEMORY_DB, true}
```
n-hutton: Seems like we should go for the stutter thing since the tests all pass? Can we review IRL the contents of the main chain tests since it is a lot of code?

n-hutton (changes requested):
Some changes that might be worth discussing

libs/ledger/src/consensus/consensus.cpp
jinmannwong: Yes, this is true.

libs/ledger/src/chain/main_chain.cpp
jinmannwong: I don't tend to like creating empty entries like this in the maps, but can change it.

libs/ledger/src/chain/main_chain.cpp
jinmannwong: We could trim it like this but it means we might be leaving entries in the map that are older than the finality period.

libs/ledger/src/chain/main_chain.cpp
jinmannwong: It is not necessarily false as the removal of the stutter block from tips could have changed the heaviest weight tip (even if it is a lower weight to the previous stutter tip). I will add the comment and change RemoveTips to avoid the smart pointer construction.

libs/ledger/src/chain/main_chain.cpp
jinmannwong: Yes, this is true and I agree the usage of this function is a bit weird. I added this here to be safe in case it was later called not just for genesis. Indeed, AddTip should not be called on something that is a stutter block. I will change this to an assert instead to ensure correct usage.

libs/ledger/src/chain/main_chain.cpp
jinmannwong: Doing it this way you would have to loop through the set of tips again to find the heaviest, no? Whereas as it is now it is determined when a tip is added to new_tips so it can just be set at the end.

libs/ledger/src/chain/main_chain.cpp
jinmannwong: Yes, I agree that this walking down the chain to find the first non-stutter block should go in a different function. I am not sure what you mean by the second part so let's discuss in person.

libs/ledger/src/chain/main_chain.cpp
jinmannwong: Yes, this the logic in this is better.

libs/ledger/src/chain/main_chain.cpp
jinmannwong: Yes, but since this is not done yet I have just put this in as a temporary measure. I can remove it in preparation for the changes that will come.

libs/ledger/src/consensus/consensus.cpp
jinmannwong: I refactored so to include a function ShuffledCabinetRank and there is also a bug fix where GetBlockGenerationWeight did not shuffle the cabinet before computing the weight.

libs/ledger/src/protocols/notarisation_service.cpp
jinmannwong: They are trimmed but in PR #1901

libs/ledger/src/chain/main_chain.cpp
jinmannwong: Actually, changing this to use [] instead of find breaks the fact that the function does not alter the main chain. Since this function is used to query the main chain from consensus I am leaving it as it is.

libs/ledger/src/chain/main_chain.cpp
jinmannwong: The return of UpdateTips is used to determine whether the updated chain should be written to file. Since moving to a lower weight tip does not necessarily correspond to a shorter chain I have left the return as it is.

libs/ledger/src/consensus/consensus.cpp
n-hutton: awesome! I did see that at one point. By shuffle the cabinet you mean shuffle qual though, I assume.

jinmannwong: Yes, I mean shuffle qual.

n-hutton (approved):

PR #1926, opened 2019-11-08
String mutability fix

wkaluza:
String literals are pooled by the compiler as an optimisation. Unless care is taken, mutating a string could result in defects. For example, the following code

```
function main()
  var s1 = "hello";
  var s2 = "hello";

  printLn(s1);
  printLn(s2);

  s1.reverse();

  printLn(s1);
  printLn(s2);
endfunction
```

yields the following output before the changes in this PR:

```
hello
hello
olleh
olleh
```


libs/vm/include/vm/string.hpp
```diff
@@ -63,6 +63,12 @@ struct String : public Object
   std::string str;
   bool        is_literal;
   int32_t     length;
+
+private:
+  bool is_modifiable()
+  {
+    return RefCount() == 1;
+  }
```
robertdickson: Can we change this to IsTemporary() instead please?

robertdickson (approved):
Looks great, let's get this in.

PR #1930, opened 2019-11-10
Fix chain-tool startup crash

ejfitzgerald:
MCL was not initialised in the chain tool causing it to crash on startup


n-hutton (approved):
LGTM - sorry this is annoying. Hopefully it did not take too much time to debug. I wonder if we could/should add a runtime check in the mcl to see if the inializer has been called.

PR #1931, opened 2019-11-10
Fix with multi lane syncing

ejfitzgerald:
Fix issue with the number of lanes no being correctly populated in the block causing the sync to fail


libs/ledger/tests/chain/block_coordinator_tests.cpp
```diff
@@ -76,8 +76,9 @@ fetch::Digest GENESIS_DIGEST =
 fetch::Digest GENESIS_MERKLE_ROOT =
     fetch::byte_array::FromBase64("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=");
 
-constexpr std::size_t NUM_LANES  = 1;
-constexpr std::size_t NUM_SLICES = 1;
+constexpr uint32_t    LOG2_NUM_LANES = 0;
```
n-hutton, line 79: why is this constrxpr 0 here

scripts/end_to_end_test/end_to_end_test.yaml
```diff
@@ -190,3 +190,37 @@ steps:
...
 ...
+
+---
+test_description: >
+  Single multi-lane standalone node can mine transactions which are then queried
+
+setup_conditions:
+  test_name: "multi_lane_standalone"
+  number_of_nodes: 1
+  lanes: 4
```
n-hutton, line 201: 👍

scripts/end_to_end_test/run_end_to_end_test.py
```diff
@@ -112,13 +112,16 @@ def setup_test(test_yaml, test_instance):
...
+    num_lanes = yaml_extract(test_yaml, 'lanes', expected=False,
+                             expect_type=int, default=1)
 
     test_instance._number_of_nodes = number_of_nodes
     test_instance._node_load_directory = node_load_directory
     test_instance._node_connections = node_connections
     test_instance._nodes_are_mining = mining_nodes
     test_instance._max_test_time = max_test_time
     test_instance._pos_mode = pos_mode
+    test_instance._lanes = num_lanes
```
n-hutton, line 124: I didn't follow it though in the E2E - does it def start with the right number of lanes?

libs/ledger/tests/chain/block_coordinator_tests.cpp
ejfitzgerald: Only because the other constant `NUM_LANES` was `constexpr`. I guess the impact is minimal in either case

scripts/end_to_end_test/run_end_to_end_test.py
ejfitzgerald: Yes

n-hutton (approved):

PR #1936, opened 2019-11-11
Improvement/next hash to block

denis-trapeznikov:
This commit flattens the `BlockDbRecord` struct by, basically, eliminating it. `next_hash` is now a member of `Block` itself but is only set when there is an unambiguous unique forward reference.


libs/ledger/src/chain/main_chain.cpp
```diff
@@ -146,6 +145,88 @@ BlockStatus MainChain::AddBlock(Block const &blk)
   return status;
 }
 
+/**
+ * Internal: add a parent-child forward reference if it is unknown yet.
```
n-hutton, line 149: 👍

libs/ledger/include/ledger/chain/block.hpp
```diff
@@ -113,27 +114,29 @@ struct MapSerializer<ledger::Block, D>
...
-    auto map = map_constructor(16);
+    auto map = map_constructor(17);
     map.Append(NONCE, block.nonce);
     map.Append(PROOF, block.proof);
     map.Append(WEIGHT, block.weight);
     map.Append(TOTAL_WEIGHT, block.total_weight);
     map.Append(MINER_SIGNATURE, block.miner_signature);
     map.Append(HASH, block.hash);
     map.Append(PREVIOUS_HASH, block.previous_hash);
+    map.Append(NEXT_HASH, block.next_hash);
```
n-hutton, line 140: To avoid bugs/attacks, this should be cleared when recieveing it in main_chain_rpc_service (the only place blocks can arrive from peers I think)

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -146,6 +145,88 @@ BlockStatus MainChain::AddBlock(Block const &blk)
...
+    if (parent_block || LookupBlockFromCache(parent, parent_block))
+    {
+      parent_block->next_hash = BlockHash{};
+    }
+    break;
+  default:;
+  }
+  references_.emplace_hint(siblings.first, std::forward<ParentHash>(parent),
+                           std::forward<ChildHash>(child));
+}
```
n-hutton, line 197: This is hard to follow if you are jumping into this code. Could I suggest the following (and for the forget also):
```
  // Get {begin, end} iterator to all known siblings of this block
  auto siblings = forward_references_.equal_range(parent);

  auto &fwd_siblings_begin = siblings.first;
  auto &fwd_siblings_end = siblings.second;

  // Find any instances in this where the value is our child
  auto sibling_it = std::find_if(fwd_siblings_begin, fwd_siblings_end,
                               [&child](auto const &ref) { return ref.second == child; });

  if (sibling_it != fwd_siblings_end)
  {
    // this child has been already referred to
    return;
  }

  // count how many forward references are already known for our hash - note
  // 2 or more fills the block with an empty hash
  switch (references_.count(parent))
  {
  case 0:  // this one will be an unique forward ref
    if (parent_block || LookupBlockFromCache(parent, parent_block))
    {
      parent_block->next_hash = child;
    }
    break;
  case 1:  // that forward ref from parent is no longer unique
    if (parent_block || LookupBlockFromCache(parent, parent_block))
    {
      parent_block->next_hash = BlockHash{};
    }
    break;
  default:;
  }
  references_.emplace_hint(fwd_siblings_begin, std::forward<ParentHash>(parent),
                           std::forward<ChildHash>(child));
```
Stylistically, I think we prefer = to {} construction.

libs/ledger/src/chain/main_chain.cpp
n-hutton: Might add: optionally update a block with the relevant forward information

libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -216,6 +213,10 @@ class MainChain
   BlockMap::size_type UncacheBlock(BlockHash const &hash) const;
   void                KeepBlock(IntBlockPtr const &block) const;
   bool LoadBlock(BlockHash const &hash, Block &block, BlockHash *next_hash = nullptr) const;
+  template <class ParentHash, class ChildHash>
```
n-hutton: I wouldn't make these templates unless absolutely neccessary, which I don't think they are. Could I suggest ```void CacheReference(BlockHash const &parent...)```

libs/ledger/include/ledger/chain/block.hpp
```diff
@@ -60,6 +60,7 @@ class Block
   // Block core information
   Digest         hash;               ///< The hash of the block
   Digest         previous_hash;      ///< The hash of the previous block
+  Digest         next_hash;          ///< The hash of the next block in heaviest chain, if any
```
n-hutton:
```suggestion
  Digest         next_hash;          ///< The hash of the next block in heaviest chain, if any. Left empty when there is no or multiple next blocks
```

n-hutton (changes requested):
Seems like this is the precursor to switching the main chain sync to use forward references instead, so there aren't any unit tests (?).

I agree with the execution of this, although the code could be made easier to follow with some variable renaming (I think I might rename references_ to forward_refrences_). 

things I note: do you know why main chain has no private members? And why the References is mutable? We could consider making the next_hash private and the main chain a friend so as to make it clear the main chain has responsibility to correctly populate this hash.

libs/ledger/include/ledger/chain/block.hpp
n-hutton: On second thought, it might make more sense to put this clearing in the AddBlock of main chain

n-hutton:
When the main chain removes a block, it might need to change the references_ etc. - Is this done?

libs/ledger/include/ledger/chain/block.hpp
denis-trapeznikov: TLDR: I'm not sure about either.

Previous-to-next relation should be easily verifiable: `∀hash . GetBlock(GetBlock(hash)->next_hash)->previous_hash == hash`; consequently, setting `next_hash` should not be any more malicious than setting `prev` one (which we do trust, I assume). One point at which clearing `next_hash` temporarily could, in theory, make sense is dropping it from the heaviest block of last successfully retrieved chunk if we were unable to retrieve the continuation, i.e. the subsequent chain blocks, from any source at all, which would probably signify that the whole last chunk was something forged by a malicious peer (who opted later to not send any continuation, for its secret malicious purpose) but I'm not sure yet how should this situation be handled at all.

libs/ledger/include/ledger/chain/main_chain.hpp
denis-trapeznikov: > Could I suggest void CacheReference(BlockHash const &parent...)

Sure. The original implementation was templated to accommodate for various [lr]value combinations but as it turns out this method is called from a single site as of now and they are pretty lvalue there.

libs/ledger/src/chain/main_chain.cpp
denis-trapeznikov: > I think we prefer = to {} construction.

It's my personal idiosyncrasy: I am afraid to use copy initialization when copy elision is not standard yet. :-\

denis-trapeznikov:
> things I note: do you know why main chain has no private members? And why the References is mutable? We could consider making the next_hash private and the main chain a friend so as to make it clear the main chain has responsibility to correctly populate this hash.

I spot such things every now and then in various classes, makes me wonder too. Oftentimes, even class members are not renamed, the identifiers keep having trailing `_` besides being public.

libs/ledger/include/ledger/chain/block.hpp
n-hutton: It seems to my mind bugprone that you would allow peers to populate this field. As you point out, the last block of a chunk could be invalid, so the main chain needs to track potentially suspicious hashes that are stored within it for the next pull of blocks. But, all next block hashes could also be invalid in that they could have a competing next block, which should cause this field to be empty (?).

libs/ledger/src/chain/main_chain.cpp
n-hutton: Does this not do RVO which is standard, in this instance?

denis-trapeznikov:
> When the main chain removes a block, it might need to change the references_ etc. - Is this done?

Well, it was half a year ago:

```cpp
bool MainChain::RemoveBlock(BlockHash const &hash)
{
  FETCH_LOCK(lock_);

  // Step 0. Manually set heaviest to a block we still know is valid
  auto block_before_one_to_del = GetBlock(hash);
  heaviest_                    = HeaviestTip{};
  heaviest_.Update(*block_before_one_to_del);

  // Step 1. Remove this block and the whole its progeny from the cache
  BlockHashSet invalidated_blocks;
  if (!RemoveTree(hash, invalidated_blocks))
```

libs/ledger/src/chain/main_chain.cpp
denis-trapeznikov: Quite likely.

RVO is more about constructing a returned value, not using it.

PR #1937, opened 2019-11-11
etch type system improvements

robertdickson:
This PR streamlines type handling in the Etch compiler via the new ConvertNode() method which determines if a type can be converted to another one, and which handles null and initialiser lists in a unified way.


denis-trapeznikov:
Can you add a more descriptive name?

libs/vm/src/analyser.cpp
```diff
@@ -1265,9 +1197,14 @@ bool Analyser::AnnotateInplaceModuloOp(ExpressionNodePtr const &node)
...
     return false;
   }
-  if (lhs->type != rhs->type || !MatchType(lhs->type, any_integer_type_))
+  if (!ConvertNode(lhs, any_integer_type_) || !ConvertNode(rhs, any_integer_type_))
   {
-    AddError(node->line, "integral operands expected");
+    AddError(node->line, "matching integral operands expected");
+    return false;
+  }
+  if (lhs->type != rhs->type)
```
denis-trapeznikov, line 1205: This basically does the same as the the previous version, but now with an extra `ConvertNode`. Currently even the error messages are the same.

libs/vm/src/analyser.cpp
```diff
@@ -1300,13 +1237,12 @@ bool Analyser::AnnotateLHSExpression(ExpressionNodePtr const &node, ExpressionNo
...
-  for (std::size_t i = 1; i <= num_supplied_indexes; ++i)
+  std::size_t const      num_indexes = lhs->children.size() - 1;
+  ExpressionNodePtrArray index_nodes;
+  for (std::size_t i = 1; i <= num_indexes; ++i)
   {
-    NodePtr const &   supplied_index      = lhs->children[i];
-    ExpressionNodePtr supplied_index_node = ConvertToExpressionNodePtr(supplied_index);
-    supplied_index_nodes.push_back(supplied_index_node);
+    ExpressionNodePtr index_node = ConvertToExpressionNodePtr(lhs->children[i]);
+    index_nodes.push_back(index_node);
```
denis-trapeznikov: Could even be done in a single statement.

libs/vm/src/analyser.cpp
```diff
@@ -1319,11 +1255,9 @@ bool Analyser::AnnotateLHSExpression(ExpressionNodePtr const &node, ExpressionNo
 
   FunctionGroupPtr setter_fg = ConvertToFunctionGroupPtr(setter_symbol);
 
-  TypePtrArray           dummy;
-  ExpressionNodePtrArray setter_supplied_nodes = supplied_index_nodes;
-  setter_supplied_nodes.push_back(lhs);
-  FunctionPtr setter_f =
-      FindFunction(container_node->type, setter_fg, setter_supplied_nodes, dummy);
+  ExpressionNodePtrArray setter_index_nodes = index_nodes;
```
denis-trapeznikov: An excess copy.

libs/vm/src/analyser.cpp
```diff
@@ -1602,83 +1536,38 @@ bool Analyser::AnnotateEqualityOp(ExpressionNodePtr const &node)
...
-      // Type-uninferable nulls will be transformed to boolean false
-      lhs->type = bool_type_;
-      rhs->type = bool_type_;
+      AddError(node->line, "incompatible types");
+      return false;
     }
   }
+  assert(lhs->type == rhs->type);
+  bool const enabled = IsOperatorEnabled(lhs->type, op);
+  if (!enabled)
```
denis-trapeznikov: In `if` statements above, such tests are done directly in parentheses, with no extra `bool` variables.

libs/vm/src/analyser.cpp
```diff
@@ -1972,23 +1865,20 @@ bool Analyser::AnnotateInvokeOp(ExpressionNodePtr const &node)
...
-    NodePtr const &   supplied_parameter      = node->children[i];
-    ExpressionNodePtr supplied_parameter_node = ConvertToExpressionNodePtr(supplied_parameter);
-    if (!AnnotateExpression(supplied_parameter_node))
+    ExpressionNodePtr parameter_node = ConvertToExpressionNodePtr(node->children[i]);
+    if (!AnnotateExpression(parameter_node))
     {
       return false;
     }
-    supplied_parameter_nodes.push_back(supplied_parameter_node);
+    parameter_nodes.push_back(parameter_node);
```
denis-trapeznikov: `std::move(parameter_node)`? (Here and in other loci.)

denis-trapeznikov (changes requested):

libs/vm/src/analyser.cpp
robertdickson: ConvertNode() is now the centralised place for all type checks, conversions of null and init lists to concrete types and checking parameters against prototypes.

libs/vm/src/analyser.cpp
robertdickson: Agreed.

libs/vm/src/analyser.cpp
robertdickson: Agreed.

libs/vm/src/analyser.cpp
robertdickson: Yes it should be. I will scan for other std::move() opportunities.

libs/vm/src/analyser.cpp
robertdickson: Indeed, I will fix.

robertdickson: Done.

libs/vm/src/analyser.cpp
robertdickson: Done

libs/vm/src/analyser.cpp
robertdickson: Done.

libs/vm/src/analyser.cpp
robertdickson: Done.

wkaluza (approved):

denis-trapeznikov (approved):

PR #1939, opened 2019-11-11
graph every op serialise test - wip

baykaner:
- fixed matrix multiply transpose serialisation
- fixed avgpool and maxpool
- implemented test for serialising every op in one graph


MissingNO57 (approved):
LGTM

PR #1947, opened 2019-11-12
Feature/persist consensus storage

n-hutton:
This PR adds a method to persist DKG keys even if the node crashes, plus a few bugfixes found during deployment.


libs/constellation/src/constellation.cpp
```diff
@@ -368,11 +372,13 @@ Constellation::Constellation(CertificatePtr certificate, Config config)
   }
 
   // Enable experimental features
-  assert(dag_);
-  dag_service_ = std::make_shared<ledger::DAGService>(muddle_->GetEndpoint(), dag_);
-  reactor_.Attach(dag_service_->GetWeakRunnable());
+  if(dag_)
```
n-hutton: Do not commit until dag PR is in

libs/beacon/include/beacon/beacon_manager.hpp
```diff
@@ -108,6 +108,17 @@ class BeaconManager
...
   std::string                    group_public_key() const;
   ///}
+  //
+
+private:
+  template <typename T, typename D>
+  friend struct serializers::MapSerializer;
+
+  // What the DKG should return
+  PrivateKey              secret_share_;       ///< Share of group private key (x_i)
```
ejfitzgerald, line 118: Why the move in the file?

libs/beacon/src/beacon_service.cpp
```diff
@@ -47,15 +48,44 @@ void TrimToSize(T &container, std::size_t max_size)
   }
 }
 
+template <typename T>
+T ChooseRandomlyFrom(T &container, std::size_t items)
```
ejfitzgerald, line 52: Worth making a common header? I am sure that I have written almost exactly the same function at some point

libs/beacon/include/beacon/beacon_manager.hpp
n-hutton: I wanted to demark the variables that are not metadata, thus if you want to ser/deser the class you must have at least these for it to fn correctly

libs/beacon/src/beacon_service.cpp
n-hutton: Makes sense

libs/beacon/include/beacon/aeon.hpp
```diff
@@ -62,6 +63,79 @@ struct AeonExecutionUnit
 
   Aeon aeon;
 };
-
 }  // namespace beacon
+
```
n-hutton, line 67: Make the aeon serializable

libs/beacon/include/beacon/beacon_manager.hpp
```diff
@@ -158,6 +163,51 @@ class BeaconManager
 
 namespace serializers {
 
+template <typename D>
```
n-hutton, line 166: serialize the absolute neccessary stuff to recover the beacon manager (not the certificate though which is set after). I don't think we want to serialize the certificate ever.

libs/beacon/src/beacon_service.cpp
```diff
@@ -98,6 +128,8 @@ BeaconService::BeaconService(MuddleInterface &muddle, const CertificatePtr &cert
   state_machine_->RegisterHandler(State::COMPLETE, this, &BeaconService::OnCompleteState);
   // clang-format on
 
+  ReloadState();
```
n-hutton, line 131: On construction, attempt to load the most recent aeon

libs/beacon/src/beacon_service.cpp
```diff
@@ -149,6 +227,8 @@ BeaconService::State BeaconService::OnWaitForSetupCompletionState()
         std::make_shared<BlockEntropy>(active_exe_unit_->aeon.block_entropy_previous);
     block_entropy_being_created_ = std::make_shared<BlockEntropy>(active_exe_unit_->block_entropy);
 
+    SaveState();
```
n-hutton, line 230: Save aeons every time DKG is successful

libs/beacon/src/beacon_service.cpp
```diff
@@ -159,11 +239,29 @@ BeaconService::State BeaconService::OnWaitForSetupCompletionState()
   return State::WAIT_FOR_SETUP_COMPLETION;
 }
 
+// Attempt to determine whether the node is attempting to generate entropy that is far in the past
+// (possible on start up when syncing)
+bool BeaconService::OutOfSync()
```
n-hutton, line 244: Now that we might push an old aeon into the beacon service, it needs a way to realise it has fallen behind and leave the state machine

libs/beacon/src/beacon_service.cpp
```diff
@@ -234,7 +338,7 @@ BeaconService::State BeaconService::OnCollectSignaturesState()
                                       BeaconServiceProtocol::GET_SIGNATURE_SHARES, index);
 
   // Timer to wait maximally for network events
-  timer_to_proceed_.Restart(std::chrono::milliseconds{50});
+  timer_to_proceed_.Restart(std::chrono::milliseconds{200});
```
n-hutton, line 341: This delay for network events was too small I think, causing failures in beta net

libs/beacon/src/trusted_dealer_beacon_service.cpp
```diff
@@ -36,7 +36,7 @@ void TrustedDealerSetupService::StartNewCabinet(
     BlockEntropy const &prev_entropy, DkgOutput const &output,
     std::pair<SharedNotarisationManager, CabinetNotarisationKeys> notarisation_keys)
 {
-  uint64_t round_end = round_start + aeon_period_ - 1;
+  uint64_t round_end = round_start + aeon_period_;
```
n-hutton: This caused the beacon to trigger the new out of sync functionality erronerously

libs/core/src/reactor.cpp
```diff
@@ -109,6 +109,15 @@ void Reactor::StartWorker()
   worker_ = std::make_unique<ProtectedThread>(&Reactor::Monitor, this);
 }
 
+Reactor::~Reactor()
+{
+  if (worker_)
```
n-hutton, line 114: the reactor needs this or it will std::terminate if shut down unexpectedly

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -457,9 +459,22 @@ MainChainRpcService::State MainChainRpcService::OnSynchronised(State current, St
...
 
-  if (chain_.HasMissingBlocks())
+  MainChain::BlockPtr head_of_chain = chain_.GetHeaviestBlock();
+  uint64_t const      seconds_since_last_block =
+      GetTime(fetch::moment::GetClock("default", fetch::moment::ClockType::SYSTEM)) -
+      head_of_chain->timestamp;
+
+  // Assume if the chain is quite old that there are peers who have a heavier chain we haven't
+  // heard about
+  if (seconds_since_last_block > 100 && head_of_chain->block_number != 0)
```
n-hutton, line 469: this avoids the case where over half the nodes crash, and start to re-sync but stall since they think they are synced as there are no loose blocks. There are no loose blocks because the miners at the head of the chain are waiting for them to catch up so they can generate entropy.

libs/version/src/cli_header.cpp
```diff
@@ -28,11 +28,13 @@ void DisplayCLIHeader(std::string const &name, std::string const &years,
                       std::string const &additional)
 {
 
-  std::cout << "\u25A0 \u25A0 \u25A0 .    " << name << ' ' << version::FULL << '\n';
-  std::cout << "\u25A0 \u25A0 . .   Copyright " << years << " (c) Fetch AI Ltd." << '\n';
-  std::cout << "\u25A0 . . .    " << additional << '\n';
+  std::cout << "\u25A0 \u25A0 \u25A0 .\n";
```
n-hutton, line 31: Seems cleaner this way

libs/beacon/src/trusted_dealer_beacon_service.cpp
jinmannwong: From what I understood of the set up the round start and round end variables are inclusive, i.e. the cabinet is in charge of the end points as well. If this is the case then the -1 one is consistent with the total number of rounds in the aeon being aeon_period. 

See for example how the values are set in consensus.cpp:
```
 cabinet_creator_->StartNewCabinet(cabinet_member_list, threshold,
                                        current_block_.block_number + 1,
                                        current_block_.block_number + aeon_period_,
                                        last_block_time + block_interval, current.block_entropy);
```

libs/beacon/include/beacon/beacon_manager.hpp
jinmannwong: I am confused about this not saving of the certificate at all. This certificate is tied to them being a member of the cabinet since the cabinet consists of the public keys of all people who won the auction. All messages sent during the DKG and entropy signature shares need to be signed with this certificate so that other DKG members know it is genuinely from them and worth processing.

jinmannwong: Also, cabinet size and cabinet index are deducible from identity_to_index if you know your own public key from certificate (which you will need as mentioned above)

jinmannwong (commented):
Left some comments to be resolved but no explicit changes as of yet - depends on comment resolution.

libs/beacon/src/trusted_dealer_beacon_service.cpp
n-hutton: I think there is slight incongruity with its actual use which led to my confusion. The variable is indeed inclusive, here is its use in consensus.cpp:

```
      // Safe to call this multiple times
      cabinet_creator_->StartNewCabinet(cabinet_member_list, threshold,
                                        current_block_.block_number + 1,
                                        current_block_.block_number + aeon_period_,
                                        last_block_time + block_interval, current.block_entropy);
```

the plus 1 is for the start since conceptually you already have genesis, whereas in the test it starts at 0. I have reverted my 'fix' and we can see if the CI passes now :)

libs/ledger/src/protocols/main_chain_rpc_service.cpp
ejfitzgerald: This seems like something that needs making a variable / constant

ejfitzgerald (approved):

PR #1955, opened 2019-11-13
TX malleability End 2 End test

steevi:


ejfitzgerald:
@steevi can you update this PR. We have landed the transaction fixes so the diff should be much clearer now

ejfitzgerald (changes requested):
See original comment 😁

scripts/end_to_end_test/smart_contract_tests/malleability_test.py
```diff
@@ -0,0 +1,55 @@
...
+    ENDPOINT = 'transfer'
+    # create the APIs
+    api = LedgerApi(options['host'], options['port'])
+
+    # Hardcoded private keys
+    id1PrivateKey = "d222867ac019aa7e5b5946ee23100b4437d54ec08db8e89f6e459d497b579a03"
+    id2PrivateKey = "499f40c1bf13e7716e62431b17ab668fa2688b7c94a011a3aab595477bc68347"
+
+    # Create identities from the private keys
+    id1 = Entity().from_hex(id1PrivateKey)
```
ejfitzgerald:
```suggestion
    id1 = Entity.from_hex(id1PrivateKey)
```

scripts/end_to_end_test/smart_contract_tests/malleability_test.py
```diff
@@ -0,0 +1,55 @@
...
+    id2 = Entity().from_hex(id2PrivateKey)
```
ejfitzgerald:
```suggestion
    id2 = Entity.from_hex(id2PrivateKey)
```

scripts/end_to_end_test/smart_contract_tests/malleability_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # Load 1000 tokens to id1
+    api.sync(api.tokens.wealth(id1, 1000))
+
+    # transaction payload that transfers 250 FET from id1 to id2. Signed with (r,s)
+    orig_tx = b'\xa1D\x00\x00\xc5\xab \xe3\xab\x84\\\xb4\xa1\xd2\xc3\xe4\xc3\xb0\x8f_\xf4*o\xf2\xa7\x1dv\x97\xba\x8f2\xc4\x15\xb7|\x7f\x8d\x85\x0b>\xf0%\xb1\x89\xa2\xd9\xbbJQZ\x84\xc3g=\xb6\xd3\xef%8],\x8d\x1e4\xb0n-\xe1\xc0\xfa\xc0\x85\x01\x14\x00\x00\x00\x00\x00\x00\x00\x00\x04\x0b\xaf\xbca\xa0\x85$7/I]\x9d\xee\x08\xad\xbc9\x82N\x98\x05\x06\x94p\x919\\\xec\xe1f6\xdd\xc0\x94\xb9\xd4\t\xd5\xb3N\xf0\xbb\xd9\xc9\x9c\\\xaf!\xfc78\x02G,\xf9j\x8a(\x0f\x84\xe83\xf9\x92@!\x92\xa0\xa1%\xd5Q\xb6\x08\x00\xd4A\xcb4\x83\xcd6W<"\xa7?"V=m\xd7\xb2~g{\x98\xbaN/wYh\x88\x83\x9a<o$9\xc9yI\xea(\x92?\x16\x8d6\nm\x15\\+\xe7\x95p\xaf'
```
ejfitzgerald: For readability I think that I would encode this pre-encoded transactions as hex instead of byte literals

scripts/end_to_end_test/smart_contract_tests/malleability_test.py
```diff
@@ -0,0 +1,55 @@
...
+    # Malicious transaction payload of the previous transaction. Signed with (r,n-s)
+    mal_tx = b"\xa1D\x00\x00\xc5\xab \xe3\xab\x84\\\xb4\xa1\xd2\xc3\xe4\xc3\xb0\x8f_\xf4*o\xf2\xa7\x1dv\x97\xba\x8f2\xc4\x15\xb7|\x7f\x8d\x85\x0b>\xf0%\xb1\x89\xa2\xd9\xbbJQZ\x84\xc3g=\xb6\xd3\xef%8],\x8d\x1e4\xb0n-\xe1\xc0\xfa\xc0\x85\x01\x14\x00\x00\x00\x00\x00\x00\x00\x00\x04\x0b\xaf\xbca\xa0\x85$7/I]\x9d\xee\x08\xad\xbc9\x82N\x98\x05\x06\x94p\x919\\\xec\xe1f6\xdd\xc0\x94\xb9\xd4\t\xd5\xb3N\xf0\xbb\xd9\xc9\x9c\\\xaf!\xfc78\x02G,\xf9j\x8a(\x0f\x84\xe83\xf9\x92@\xf1\xa01\x1c\xc9\xb4.\xda\xa4\x1dS~=\x1f\x9dT\xac\x15\x95\x04'l'\xbe)s\xce\xa8\\\x8e\x8a\x17\x14\x8a+\x1d\xc3\xe1\x88\xf9\x11\xc6\xb0_\x16=\x95\xb9d\xaeYJ4xp\x97>Y\xb3\xc9=5\xb8\x95"
+
+    # Creating jsons for the above mentioned transactions
+    legit_trans = api.tokens._post_tx_json(orig_tx, ENDPOINT)
```
ejfitzgerald: `_post_tx_json` is not a private method on this token api object. So this would be a little brittle. There is a free function that you can use `submit_json_transaction` that is part of the API directly which might be a better option

scripts/end_to_end_test/smart_contract_tests/malleability_test.py
```diff
@@ -0,0 +1,55 @@
...
+    mal_trans = api.tokens._post_tx_json(mal_tx, ENDPOINT)
+
+    # Sending the transactions to the ledger
+    api.sync(legit_trans)
+    api.sync(mal_trans)
```
ejfitzgerald:
```suggestion
api.sync([legit_trans, mal_trans])
```
Also worth making sure that `legit_trans` and `mal_trans` are the same value

steevi:
Waiting for the following modifications of the submit_json_function to verify that the test is working.

1. submit_json_function should return the transaction hash and not True or False
2. submit_json_function should accept an endpoint as input
3. the endpoint should be passed to the format_contract_url function. If I pass `endpoint='transfer'`, the url should be `protocol://host:port/api/contract/fetch/token/transfer`

Could you please also update the fetchai-ledger-api on the pypi?

ejfitzgerald (approved):

PR #1962, opened 2019-11-14
Update entropy benchmark with most recent seen entropy

jinmannwong:
Minor bug fix to entropy benchmark and threshold for DKG completion in beacon_setup_service.


libs/beacon/tests/unit/trusted_dealer_test.cpp
```diff
@@ -186,7 +186,6 @@ void RunTrustedDealer(uint16_t total_renewals = 4, uint32_t cabinet_size = 4,
       // Note, to avoid limiting the 'look ahead' entropy gen, set the block to ahead of numbers per
       // aeon
       member->beacon_service.MostRecentSeen((i * aeon_period) + aeon_period - 1);
-      member->setup_service.Abort(aeon_period);
     }
```
jinmannwong, line 189: Not needed as use of trusted beacon setup service means that the state where abort_below_ is check is never entered.

libs/beacon/benchmark/entropy_gen_bench.cpp
```diff
@@ -187,6 +187,7 @@ void EntropyGen(benchmark::State &state)
       nodes[m]->beacon_setup.StartNewCabinet(
           cabinet, test_attempt * entropy_rounds, start_time, prev_entropy,
           dealer.GetDkgKeys(nodes[m]->muddle_certificate->identity().identifier()));
+      nodes[m]->beacon_service.MostRecentSeen(test_attempt * entropy_rounds + entropy_rounds - 1);
```
n-hutton: I know everyone knows BODMAS, but could you bracket: (test_attempt*entropy_rounds)+...

n-hutton (changes requested):
Very minor change then good to go

libs/beacon/benchmark/entropy_gen_bench.cpp
jinmannwong: Ah, sure.

n-hutton (approved):

PR #1964, opened 2019-11-14
Bug/fixed point conversion basic vm engine [COLE-57]

JuanBesa:
Fixes fixed_point conversion to use the new conversion constructor that is available.

Fixes some minor issues when Float32 and Bool where passed as parameters or return values.


libs/dmlf/tests/execution/basic_engine_tests.cpp
```diff
@@ -1905,13 +1873,30 @@ TEST(DISABLED_BasicVmEngineDmlfTests, DisabledArrayArrayOpTests)
 {
   RunArrayTest("doFloat32", std::vector<float>{1.0f, 2.0f, 3.0f, 4.0f});
 
-  RunArrayTest("doFixed32",
```
markos, line 1908: One question, why are the doFixed32/doFixed64 tests removed?

libs/dmlf/src/basic_vm_engine.cpp
```diff
@@ -372,8 +377,7 @@ void ToFixed32(BasicVmEngine::LedgerVariant &var)
   {
     fp32_t val32;
     val32.Data() = var.As<int>();
-    fp64_t val64{fp64_t::FromBase(static_cast<int64_t>(val32.Data()) << 16)};
-    var = BasicVmEngine::LedgerVariant{val64};
+    var          = BasicVmEngine::LedgerVariant{fp64_t{val32}};
```
markos: it's valid, but I would suggest to use an explicit static_cast<fp64_t>(val32).

libs/dmlf/tests/execution/basic_engine_tests.cpp
JuanBesa: Oh, because they are working! I have this disabled test to show which parameters/return values don't work when using arrays. I had fixed the issue but incorrectly hadn't deleted them from here. They are in a test a little bit up.

engineerkatie (approved):
Super. LGTM.

PR #1967, opened 2019-11-15
VM ML bindings execution charge estimators

lrahmani:
Reference: 
- https://docs.google.com/document/d/1xIEYt8B-yPo0W4igpMe8Rel9hd2eU0GaCtqk6aywJBs/edit
- https://docs.google.com/document/d/16ORxnBv0BsIgdNgiGZx1UktejbENkNPg474CD-IpuJQ/edit

### DIFF
--- a/libs/meta/include/meta/tuple.hpp
+++ b/libs/meta/include/meta/tuple.hpp
@@ -74,6 +74,21 @@ struct TupleOperations<std::tuple<Ts...>>
   template <std::size_t N>
   using DropTerminal =
       TupleOperations<typename SplitTuple<std::tuple_size<type>::value - N, type>::Initial>;
+
+  template <typename T>
+  struct Prepend;
+  template <typename... Ts2>
+  struct Prepend<std::tuple<Ts2...>>
+  {
+    using type = std::tuple<Ts2..., Ts...>;
+  };
+  template <typename T>
+  struct Append;
+  template <typename... Ts2>
+  struct Append<std::tuple<Ts2...>>
+  {
+    using type = std::tuple<Ts..., Ts2...>;
+  };
 };
 
 template <typename>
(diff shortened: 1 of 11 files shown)


libs/meta/include/meta/tuple.hpp
```diff
@@ -74,6 +74,21 @@ struct TupleOperations<std::tuple<Ts...>>
...
+  {
+    using type = std::tuple<Ts2..., Ts...>;
+  };
+  template <typename T>
+  struct Append;
+  template <typename... Ts2>
+  struct Append<std::tuple<Ts2...>>
+  {
+    using type = std::tuple<Ts..., Ts2...>;
+  };
```
denis-trapeznikov, line 91: `template<class... Ts> using TupleCat = decltype(std::tuple_cat(std::declval<Ts>()...));` ?

libs/vm/include/vm/module/binding_interfaces.hpp
```diff
@@ -37,6 +37,46 @@ namespace vm {
...
+struct PrependedTupleImpl;
+
+template <typename T, typename... Ts>
+struct PrependedTupleImpl<T, std::tuple<Ts...>>
+{
+  using type = std::tuple<T, Ts...>;
+};
+
+template <typename T, typename Tuple>
+using PrependedTuple = typename PrependedTupleImpl<T, Tuple>::type;
```
denis-trapeznikov, line 50: `Cons` would be a more traditional name here. Besides, if we take `TupleCat` from my comment above, then it is trivially `template <class Car, class Cdr> using Cons = TupleCat<std::tuple<Car>, Cdr>;`.

denis-trapeznikov (changes requested):
Nitpicked few places to simplify.

wkaluza (changes requested):
VMTensor estimator tests were rejected in other reviews.

PR #1969, opened 2019-11-15
COLE-58 Easier muddle-colearn setup

engineerkatie:


libs/dmlf/src/colearn/muddle_learner_networker_impl.cpp
```diff
@@ -45,6 +52,35 @@ MuddleLearnerNetworkerImpl::MuddleLearnerNetworkerImpl(MuddlePtr mud, StorePtr u
...
+{
+  auto signer = std::make_shared<Signer>();
+  signer->Load(fetch::byte_array::FromBase64(priv));
+  CertificatePtr ident = signer;
+
+  netm_ = std::make_shared<NetMan>("LrnrNet", 4);
+  netm_->Start();
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+  auto mud = fetch::muddle::CreateMuddle("Test", ident, *netm_, "127.0.0.1");
```
JuanBesa, line 67: Is this correct?

libs/dmlf/tests/colearn_muddle/colearn_muddle.cpp
```diff
@@ -76,26 +76,16 @@ class LearnerTypedUpdates
   LearnerTypedUpdates(const std::string &pub, const std::string &priv, unsigned short int port,
```
JuanBesa, line 76: Const wrong way around

Also you could pass in the pub (and maybe the priv) as a copy and move it into pub_

JuanBesa (approved):
LGTM

PR #1970, opened 2019-11-15
Remove unnecessary call to UnInitialise

wkaluza:
The compiler already calls the `Analyser`'s `UnInitialise` method in its destructor.


robertdickson (approved):
lgtm

PR #1971, opened 2019-11-15
Fix multidigit FixedPoint tokenisation bug

wkaluza:
Enables support for multidigit integral fixed point literals, e.g. `123fp32`. Until now, Etch code containing such literals would fail to compile.


robertdickson (approved):
Looks good to me. Good find.

PR #1975, opened 2019-11-15
Validation Checks for Multi-Sig Wallets

ejfitzgerald:


scripts/generate_coverage.py
```diff
@@ -24,6 +24,19 @@
 build_directory = 'unknown'
 
 
+def find_tool(name):
+    tool_path = shutil.which('name')
+    if tool_path is not None:
+        tool_cmd = [tool_path]
+    elif os.path.exists('/usr/bin/xcrun'):
```
ejfitzgerald, line 31: Bonus: This is needed to get coverage to work with Apple Clang 11

libs/ledger/src/transaction_validator.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include "ledger/chaincode/deed.hpp"
+#include "ledger/chaincode/token_contract.hpp"
+#include "ledger/identifier.hpp"
+#include "ledger/transaction_validator.hpp"
+
+namespace fetch {
+namespace ledger {
+namespace {
+
+bool IsCreateWealth(chain::Transaction const &tx)
```
wkaluza, line 32: Are we committed to killing off wealth txs this sprint? If not, then it would be good to deduplicate this function and the one in `executor.cpp`

libs/ledger/tests/unit/transaction_validator_tests.cpp
```diff
@@ -0,0 +1,486 @@
...
+#include "ledger/chaincode/token_contract.hpp"
+#include "ledger/identifier.hpp"
+#include "ledger/state_sentinel_adapter.hpp"
+#include "ledger/storage_unit/fake_storage_unit.hpp"
+#include "ledger/transaction_validator.hpp"
+
+#include "gtest/gtest.h"
+
+#include <initializer_list>
+#include <ledger/chaincode/deed.hpp>
```
wkaluza: This is one of our headers - let's move it up and use quotes.

libs/ledger/tests/unit/transaction_validator_tests.cpp
```diff
@@ -0,0 +1,486 @@
...
+
+void TransactionValidatorTests::SetDeed(Deed const &deed)
+{
+  // create shard mask
+  BitVector shards{1};
+  shards.SetAllOne();
+
+  // create storage infrastructure
+  StateSentinelAdapter    storage_adapter{storage_, Identifier{"fetch.token"}, shards};
+  ContractContext         ctx{&token_contract_, Address{}, &storage_adapter, 0};
```
wkaluza:
```suggestion
  ContractContext         ctx{nullptr, Address{}, &storage_adapter, 0};
```

libs/ledger/tests/unit/transaction_validator_tests.cpp
```diff
@@ -0,0 +1,486 @@
...
+                .Seal()
+                .Sign(signer_)
+                .Build();
+
+  EXPECT_EQ(ContractExecutionStatus::TX_NOT_VALID_FOR_BLOCK, validator_(*tx, 1000));
+}
+
+TEST_F(TransactionValidatorTests, CheckDefaultCaseOnValidityBoundary)
+{
+  // build up a wealth transaction
```
wkaluza: Not true; ditto elsewhere in this file.

wkaluza (changes requested):

libs/ledger/src/transaction_validator.cpp
```diff
@@ -0,0 +1,134 @@
...
+  if (IsCreateWealth(tx))
+  {
+    return ContractExecutionStatus::SUCCESS;
+  }
+
+  // attach the token contract to the storage engine
+  StateAdapter storage_adapter{storage_, Identifier{"fetch.token"}};
+
+  {
+    ContractContext ctx{&token_contract_, tx.contract_address(), &storage_adapter, block_index};
```
wkaluza, line 75: This context does not need anything apart from the storage adapter

libs/ledger/src/transaction_validator.cpp
ejfitzgerald: Haha - I thought you would spot that - yes I duplicated only because we are going to remove it later this week

libs/ledger/tests/unit/transaction_validator_tests.cpp
ejfitzgerald: That Clion has a lot to answer for... done

libs/ledger/tests/unit/transaction_validator_tests.cpp
ejfitzgerald: Removed

wkaluza (approved):

PR #1977, opened 2019-11-16
Catch promise timeout to obj store protocol

n-hutton:


libs/ledger/src/storage_unit/storage_unit_client.cpp
```diff
@@ -394,9 +394,14 @@ StorageUnitClient::TxLayouts StorageUnitClient::PollRecentTx(uint32_t max_to_pol
       layouts.insert(layouts.end(), std::make_move_iterator(txs.begin()),
                      std::make_move_iterator(txs.end()));
     }
-    catch (std::runtime_error &e)
+    catch (service::PromiseError &e)
```
ejfitzgerald: I thought we have got rid of all of these. This should be std:: exception (so it will catch all of them)

ejfitzgerald (changes requested):

ejfitzgerald (approved):

PR #1981, opened 2019-11-18
Bugfix: ml-224, UInt256 issues and testing

vhavrylov-ncube:
Added a unit test skeleton for UInt256Wrapper class and UInt256 etch VM type.
Fixed wrong logarithm results (conversion tolerance vs conversion speed is still TBD)


markos (approved):
LGTM

PR #1982, opened 2019-11-18
Implemented StridedSlice

MissingNO57:


libs/ml/include/ml/ops/strided_slice.hpp
```diff
@@ -0,0 +1,352 @@
...
+  assert(output.shape() == this->ComputeOutputShape(inputs));
+
+  switch (inputs.at(0)->shape().size())
+  {
+    // 1D Tensor
+  case 1:
+  {
+    for (SizeType i{0}; i < output.shape().at(0); i++)
+    {
+      output.At(i) = inputs.at(0)->At(begins_.at(0) + i * strides_.at(0));
```
baykaner: I don't understand why you use .At(i), .At(i, j) etc. here.
Pretty sure you can do all this a lot better with iterators.
It looks like TensorSliceIterator has a constructor that takes stride already, so if you add another Slice method to tensor I think the strided slice implementation can be handled there.

libs/ml/include/ml/serializers/ml_types.hpp
```diff
@@ -2421,6 +2433,49 @@ struct MapSerializer<ml::OpSliceSaveableParams<TensorType>, D>
...
+  static uint8_t const OP_CODE  = 2;
+
+  static uint8_t const BEGINS  = 3;
+  static uint8_t const ENDS    = 4;
+  static uint8_t const STRIDES = 5;
+
+  template <typename Constructor>
+  static void Serialize(Constructor &map_constructor, Type const &sp)
+  {
+    auto map = map_constructor(6);
```
baykaner: you mean 5

libs/ml/tests/unit/ops/strided_slice.cpp
```diff
@@ -0,0 +1,507 @@
...
+      op.Backward({std::make_shared<TypeParam>(input)}, error);
+
+  // test correct shapes
+  ASSERT_EQ(backpropagated_signals.size(), 1);
+  ASSERT_EQ(backpropagated_signals[0].shape(), input.shape());
+
+  // test correct values
+  EXPECT_TRUE(backpropagated_signals[0].AllClose(gt));
+}
+
```
baykaner, line 505: can you also test for correct serialisation?

baykaner (changes requested):

libs/math/include/math/tensor.hpp
```diff
@@ -289,10 +289,14 @@ class Tensor
   ConstSliceType Slice() const;
   ConstSliceType Slice(SizeType index, SizeType axis = 0) const;
   ConstSliceType Slice(SizeVector indices, SizeVector axes) const;
+  ConstSliceType StridedSlice(SizeVector const &begins, SizeVector const &ends,
```
baykaner: I think it would make more sense to just call this Slice as well

libs/math/include/math/tensor.hpp
```diff
@@ -2200,6 +2204,23 @@ typename Tensor<T, C>::ConstSliceType Tensor<T, C>::Slice(std::vector<SizeType>
...
+typename Tensor<T, C>::ConstSliceType Tensor<T, C>::StridedSlice(
+    std::vector<SizeType> const &begins, std::vector<SizeType> const &ends,
+    std::vector<SizeType> const &strides) const
+{
+  std::vector<std::vector<SizeType>> range;
+  std::vector<SizeType>              axis;
+
+  for (SizeType j = 0; j < shape().size(); ++j)
+  {
+    range.push_back({begins.at(j), ends.at(j) + 1, strides.at(j)});
```
baykaner: shouldn't it be ends.at(j) without the + 1
And then you remove the -1 in the iterator?

libs/math/include/math/tensor_slice_iterator.hpp
```diff
@@ -387,16 +387,10 @@ class TensorSliceIterator
         volume = stride[i];
 
         s.volume      = volume;
-        SizeType diff = (s.to - s.from);
-        s.total_steps = diff / s.step;
-        if (s.total_steps * s.step < diff)
-        {
-          ++s.total_steps;
-        }
+        s.total_steps = ((s.to - s.from - 1) / s.step) + 1;
```
baykaner, line 390: if you remove the +1 from the Tensor::Slice implementation you can remove the -1 from this equation i think?

libs/ml/include/ml/ops/strided_slice.hpp
```diff
@@ -0,0 +1,156 @@
...
+  using TensorType    = T;
+  using SizeType      = fetch::math::SizeType;
+  using SizeVector    = fetch::math::SizeVector;
+  using ArrayPtrType  = std::shared_ptr<TensorType>;
+  using VecTensorType = typename Ops<T>::VecTensorType;
+  using SPType        = OpStridedSliceSaveableParams<T>;
+  using MyType        = StridedSlice<TensorType>;
+
+  explicit StridedSlice(SizeVector const &begins, SizeVector const &ends,
+                        SizeVector const &strides = {})
```
baykaner, line 42: it actually seems a little strange that this is a separate op from Slice() at all - it's calling the same implementation. I know that's how it is in tensorflow though so nothing to change here

baykaner (changes requested):

baykaner (approved):

PR #1984, opened 2019-11-18
Fix for segfault in beacon tests

### LINKED ISSUE #1957: Segfault noticed during beacon test
Noticed a segfault running the beacon test, but couldn't immediately replicate

jinmannwong:
Fix issue raised in #1957 - need to make sure no more messages are being sent to the RBC while the object is being destroyed at the end of the test.

Closes #1957


ejfitzgerald (approved):
Nice catch @jinmannwong

PR #1985, opened 2019-11-19
Implement limit for DKG transaction time

n-hutton:
Slight refactoring of DKG time increases, also sets an upper bound for how long the DKG can take. This makes it much less painful if the network goes down and everyone who comes back thinks there have been 100s of DKG failures.


libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -1525,145 +1527,160 @@ std::vector<std::weak_ptr<core::Runnable>> BeaconSetupService::GetWeakRunnables(
...
+  });
+
+  assert(expected_dkg_timespan_ != 0 && time_slots_in_dkg_ != 0);
+  assert(time_slots_to_end <= time_slots_in_dkg_);
+
+  // Note: fine to do floor arithmetic here, it might cause the deadline to happen in the past, but
+  // there is resilience to this.
+  uint64_t time_until_deadline_s =
+      (time_slots_to_end * expected_dkg_timespan_) / time_slots_in_dkg_;
+
```
jinmannwong, line 1586: Your changes have reverted the time allowed per slot back to something that depends on the number of states. The reason behind adding timer_per_slot_ is so that we can add more states in without reducing the time for each state, because the expected_dkg_timespan_ is fixed. I think this is important as the empirical values were calculated without states like the dry run and notarisations and without a fixed amount of time per slot you are eating into the time allowed for each state when you add more states.

jinmannwong (changes requested):

libs/beacon/src/beacon_setup_service.cpp
n-hutton: Oh right, I see we were coming from different directions here. I didn't get that just there and so refactored it when I noticed a possible issue with the code, which is that using doubles can result in divergent behaviour across machine architectures. I also thought it was incorrect that the DKG time was / 100 whereas the slots totalled 110, but seems I was not correct here.

Would you agree with the following change:
Change the map to be conceptually the base time a state should take
Have a lookup of a state modifer (1x usually)
?

libs/beacon/src/beacon_setup_service.cpp
```diff
@@ -1628,22 +1626,22 @@ void BeaconSetupService::SetTimeToProceed(BeaconSetupService::State state)
...
     uint16_t failures      = 0;
 
     while (reference_timepoint_ < current_time)
     {
       failures++;
       reference_timepoint_ += expected_dkg_timespan_;
       expected_dkg_timespan_ += std::min(expected_dkg_timespan_ + (expected_dkg_timespan_ / 2),
-                                         time_per_iteration * MAX_DKG_BOUND_MULTIPLE);
+                                         time_per_state * MAX_DKG_BOUND_MULTIPLE);
     }
```
jinmannwong, line 1645: In order for the time per slot to increase in each iteration you need to change time_per_state here so that it gets multiplied by 1.5 with each failure, no?

jinmannwong (changes requested):

libs/beacon/src/beacon_setup_service.cpp
n-hutton: OK, I have tested it and the logic is (now) correct:

```
    while (reference_timepoint_ < current_time)
    {
      failures++;
      reference_timepoint_ += expected_dkg_timespan_;
      expected_dkg_timespan_ = std::min(
          (expected_dkg_timespan_ + (expected_dkg_timespan_ / 2)),
          static_cast<uint64_t>(time_per_state * time_slots_in_dkg_ * MAX_DKG_BOUND_MULTIPLE));
    }

```

Each failure adds itself + 1/2, so 1.5* each time to a multiple of 4.

jinmannwong (approved):
Looks good.

PR #1987, opened 2019-11-19
fix SetLogLevel

n-hutton:


libs/logging/src/logging.cpp
```diff
@@ -166,11 +166,23 @@ void LogRegistry::Log(LogLevel level, char const *name, std::string &&message)
...
+
+  auto     logger_ref      = GetLogger(name);
+  LogLevel converted_level = ConvertToLevel(logger_ref.level());
+
+  if (level < converted_level)
+  {
+    return;
+  }
+
+  logger_ref.log(ConvertFromLevel(level), message);
```
ejfitzgerald: Erm this should be handled internally by speed log no?

n-hutton: Seems like it does not behave as expected

ejfitzgerald (approved):

PR #1988, opened 2019-11-19
UInt64 etch type initialization fixed

vhavrylov-ncube:
Fixed a wrong behavior when creating a UInt64 etch type with big values (>= uint64 max / 2)

### DIFF
--- a/libs/vm/src/generator.cpp
+++ b/libs/vm/src/generator.cpp
@@ -1,4 +1,3 @@
-
 //------------------------------------------------------------------------------
 //
 //   Copyright 2018-2019 Fetch.AI Limited
@@ -1308,9 +1307,9 @@ void Generator::HandleInteger64(IRExpressionNodePtr const &node)
 void Generator::HandleUnsignedInteger64(IRExpressionNodePtr const &node)
 {
   Executable::Instruction instruction(Opcodes::PushConstant);
-  auto value        = static_cast<uint64_t>(std::stoull(node->text));
-  instruction.index = AddConstant(Variant(value, TypeIds::UInt64));
-  uint16_t pc       = function_->AddInstruction(instruction);
+  auto                    value = static_cast<uint64_t>(std::stoull(node->text));
+  instruction.index             = AddConstant(Variant(value, TypeIds::UInt64));
+  uint16_t pc                   = function_->AddInstruction(instruction);
   AddLineNumber(node->line, pc);
 }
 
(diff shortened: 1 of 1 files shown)


markos (dismissed):
Looks Good.

baykaner (dismissed):

wkaluza:
This badly needs exception handling, either here or as a follow-up. We probably don't have to worry about `std::invalid_argument`, as the tokeniser should prevent offending inputs from reaching this code, but an unhandled `std::out_of_range` could break us. One way to do this could be to add an error string to `errors_` and then check if that's empty in `GenerateExecutable` after the call to `HandleBlock`. Similar checks already happen earlier in `GenerateExecutable`, so it's just more of the same. I'd let @robertdickson weigh in before merging.

To be clear, I think this change is a step in the right direction, as current behaviour is incorrect for large numbers. It's just a question of fixing the bug in a way that does not itself cause problems.

wkaluza (changes requested):
See comment above.

robertdickson (commented):
The problem here is that the Generator's job is to take a fully checked, error-free IR and generate an Executable from it -- the Generator itself is not the place to be doing checks on numbers or anything else, because that would suggest that the incoming IR has been allowed to be compiled with bad data in it. Maybe let this change go through for now, but with a view to adding superior checks to the HandleLiteral() call of the Parser, say.

vhavrylov-ncube:
From my point of view, a non-throwing std::strtoull could be used here just as std::atoll for signed integers. We still have undefined behavior for invalid inputs, but may be this is a subject for improving checking while parsing?
As far as I understand, the Generator is called rarely, and its maximum speed is not the most important consideration here, but rather maximum safety. So, in general, a try-catch block in "void Generator::HandleExpression(IRExpressionNodePtr const &node){}" over the switch with all node handler calls could do the job; in this case every conversion method in node handlers (atoi, atof, ...) can be replaced with exception-throwing (std::stof, stold and so on). Because if have a wrong input value despite all previous checks it is definitely an _exception_ :) and we must stop here before proceeding with undefined values (as it is now for uint64 with a big value). Of course this would be a big change and a bunch of work, but I believe we can plan it for future.

robertdickson:
Yes, agreed. Let's remove the exception danger here, and aim to validate literals earlier in the compilation process at the parsing stage.

wkaluza:
This week I'll be looking at perking up the tokeniser - if I choose the right values in tests, I may be able to get some coverage around this problem. At least this will establish which literals need tighter checks and what failure modes we are facing.

wkaluza (approved):

PR #1992, opened 2019-11-20
Adding fees to synergetic contracts

qati:
This PR refactors current fee structure in executor and introduces fees for synergetic contracts.


libs/ledger/include/ledger/fees/chargeable.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace fetch {
+namespace ledger {
+  
+class Chargeable
+{
+public:
+  // Construction / Destruction
+  Chargeable()                 = default;
+  Chargeable(Chargeable const &) = delete;
```
ejfitzgerald: I wouldn't over-constrain this interface by disabling copy and move

libs/ledger/include/ledger/fees/chargeable.hpp
```diff
@@ -0,0 +1,42 @@
...
+  Chargeable(Chargeable &&)      = delete;
+  virtual ~Chargeable()        = default;
+
+  virtual uint64_t CalculateFee() const = 0;
```
ejfitzgerald, line 41: It is also nice to document these interfaces. See: https://github.com/fetchai/ledger/blob/ce16976495a7e66f3dec2e0a94f76ef73a68e32e/libs/muddle/include/muddle/muddle_endpoint.hpp#L71

libs/ledger/include/ledger/fees/storage_fee.hpp
```diff
@@ -0,0 +1,50 @@
...
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "ledger/fees/chargeable.hpp"
+
+namespace fetch {
+namespace ledger {
+
+class StorageFee : public Chargeable
```
ejfitzgerald, line 27: While I appreciate this is a fairly simple class, I would still split this into `.hpp` and `.cpp` files

libs/ledger/src/executor.cpp
```diff
@@ -170,8 +170,18 @@ Executor::Result Executor::Execute(Digest const &digest, BlockIndex block, Slice
       storage_cache_->Clear();
     }
 
+    FeeManager::TransactionDetails tx_details{
```
ejfitzgerald: Is it worth making the constructor here to take the transaction object itself? This might make is easier in the future to extend?

libs/ledger/src/upow/naive_synergetic_miner.cpp
```diff
@@ -165,6 +168,23 @@ WorkPtr NaiveSynergeticMiner::MineSolution(Digest const &        contract_digest
                                            chain::Address const &contract_address,
                                            ProblemData const &   problem_data)
 {
+  // TODO(AB): State sharding needs to be added here
+  BitVector shards{num_lanes_};
+  shards.SetAllOne();
+  StateSentinelAdapter storage_adapter{storage_, Identifier{"fetch.token"}, shards};
```
ejfitzgerald: This should be just a state adapter. This will ensure that any operations from this will be read only

libs/ledger/src/upow/synergetic_execution_manager.cpp
```diff
@@ -52,6 +52,14 @@ SynergeticExecutionManager::SynergeticExecutionManager(DAGPtr dag, std::size_t n
         "The number of executors must be 1 because state concurrency not implemented");
   }
 
+  telemetry::Registry::Instance().CreateHistogram(
+      {0.000001, 0.000002, 0.000003, 0.000004, 0.000005, 0.000006, 0.000007, 0.000008, 0.000009,
+       0.00001,  0.00002,  0.00003,  0.00004,  0.00005,  0.00006,  0.00007,  0.00008,  0.00009,
+       0.0001,   0.0002,   0.0003,   0.0004,   0.0005,   0.0006,   0.0007,   0.0008,   0.0009,
+       0.001,    0.01,     0.1,      1,        10.,      100.},
+      "ledger_synergetic_executor_deduct_fees_duration",
```
ejfitzgerald, line 62: 👍 👍 👍 - Can we also add metrics for the other synergetic execution stages?

ejfitzgerald (changes requested):
Looks really good. Think that we need to hook up the restriction on the VM i.e. making sure we call `SetChargeLimit` on the VM instance

libs/ledger/src/upow/synergetic_contract.cpp
```diff
@@ -220,6 +220,11 @@ Status SynergeticContract::DefineProblem(ProblemData const &problem_data)
   auto vm  = std::make_unique<vm::VM>(module_.get());
   problem_ = std::make_shared<vm::Variant>();
 
+  if (charge_limit_ > 0)
+  {
+    vm->SetChargeLimit(charge_limit_);
```
ejfitzgerald, line 225: 👍

scripts/end_to_end_test/end_to_end_test.yaml
```diff
@@ -239,3 +239,26 @@ steps:
   - sleep: 5
   - run_python_test: { "script": ".smart_contract_tests.malleability_test" }
 ...
+
+---
+test_description: >
+  DISABLED - Test synergetic contract fees
```
ejfitzgerald: I guess you are waiting for the new version to be published?

scripts/end_to_end_test/end_to_end_test.yaml
```diff
@@ -239,3 +239,26 @@ steps:
...
+
+setup_conditions:
```
ejfitzgerald, line 248: Would also be good it we can think of a way to catch the negative case here in and end 2 end test. I.e. when the work is not completed because there are no funds available

libs/ledger/src/upow/synergetic_executor.cpp
```diff
@@ -87,15 +111,34 @@ void SynergeticExecutor::Verify(WorkQueue &solutions, ProblemData const &problem
...
-      status = contract->Complete(solution->address(), shard_mask);
-      contract->Detach();
+
+      // TODO(AB): charge limit
+      FeeManager::TransactionDetails tx_details{
+          solution->address(), solution->address(), shard_mask, solution->contract_digest(),
+          CHARGE_RATE,         CHARGE_LIMIT,        false};
+
+      ContractExecutionResult result;
+      FETCH_LOG_INFO(LOGGING_NAME, "Verify: tx created");
```
ejfitzgerald: I guess I would remove this debug

libs/ledger/src/upow/synergetic_executor.cpp
```diff
@@ -87,15 +111,34 @@ void SynergeticExecutor::Verify(WorkQueue &solutions, ProblemData const &problem
...
+            });
+      }
 
       if (SynergeticContract::Status::SUCCESS != status)
       {
         FETCH_LOG_WARN(LOGGING_NAME, "Failed to complete contract: 0x", contract->digest().ToHex(),
                        " Reason: ", ToString(status));
         return;
       }
+      FETCH_LOG_INFO(LOGGING_NAME, "Calculated fee: ", result.charge);
```
ejfitzgerald: This also seems too verbose a logging message I would downgrade to DEBUG

libs/ledger/src/upow/synergetic_executor.cpp
```diff
@@ -87,15 +111,34 @@ void SynergeticExecutor::Verify(WorkQueue &solutions, ProblemData const &problem
       contract->Attach(storage_);
       ContractContext ctx(&token_contract_, solution->address(), &storage_adapter, 0);
       contract->UpdateContractContext(ctx);
-      status = contract->Complete(solution->address(), shard_mask);
-      contract->Detach();
+
+      // TODO(AB): charge limit
+      FeeManager::TransactionDetails tx_details{
+          solution->address(), solution->address(), shard_mask, solution->contract_digest(),
+          CHARGE_RATE,         CHARGE_LIMIT,        false};
```
ejfitzgerald, line 120: This is punishing for the solution provider if the complete part of the contract turns out to be error prone

libs/ledger/src/upow/synergetic_executor.cpp
```diff
@@ -87,15 +111,34 @@ void SynergeticExecutor::Verify(WorkQueue &solutions, ProblemData const &problem
...
+      }
 
       if (SynergeticContract::Status::SUCCESS != status)
       {
         FETCH_LOG_WARN(LOGGING_NAME, "Failed to complete contract: 0x", contract->digest().ToHex(),
                        " Reason: ", ToString(status));
         return;
       }
+      FETCH_LOG_INFO(LOGGING_NAME, "Calculated fee: ", result.charge);
+      fee_manager_.Execute(tx_details, result, solution->block_index(), storage_);
```
ejfitzgerald, line 139: Maybe I am missing something but this deducts the fees from contract but I can't find where the fees are credited back

ejfitzgerald (changes requested):

scripts/end_to_end_test/end_to_end_test.yaml
qati: yes

libs/ledger/src/executor.cpp
```diff
@@ -355,28 +334,12 @@ bool Executor::ExecuteTransactionContract(Result &result)
...
-      FETCH_LOG_DEBUG(LOGGING_NAME, "Calculated charge for 0x", current_tx_->digest().ToHex(), ": ",
-                      scaled_charge, " (base: ", base_charge, " storage: ", storage_charge,
-                      " compute: ", compute_charge, " shards: ", allowed_shards_.PopCount(), ")");
-
-      // create wealth transactions are always free
-      if (!IsCreateWealth(*current_tx_))
-      {
-        result.charge += scaled_charge;
-      }
+      StorageFee storage_fee_{storage_adapter};
```
wkaluza: Style police: we normally reserve a traiing underscore for private members.

libs/ledger/src/upow/synergetic_execution_manager.cpp
```diff
@@ -79,6 +122,7 @@ ExecStatus SynergeticExecutionManager::PrepareWorkQueue(Block const &current, Bl
       FETCH_LOG_WARN(LOGGING_NAME, "Failed to get work from DAG Node: 0x", digest.ToHex());
       continue;
     }
+    work->UpdateBlockIndex(current.block_number);
```
wkaluza: You don't need this setter - wouldn't it be simpler inject the block number as an argument to `Work`'s constructor?

libs/ledger/src/upow/synergetic_contract.cpp
```diff
@@ -306,15 +329,37 @@ Status SynergeticContract::Complete(chain::Address const &address, BitVector con
   if (!vm->Execute(*executable_, clear_function_, error, output, *problem_, *solution_))
   {
     FETCH_LOG_WARN(LOGGING_NAME, "Solution execution failure: ", error);
+    charge_ += vm->GetChargeTotal();
```
wkaluza, line 332: :+1:

libs/ledger/src/upow/synergetic_contract.cpp
```diff
@@ -380,6 +425,7 @@ void SynergeticContract::Detach()
   storage_ = nullptr;
   problem_.reset();
   solution_.reset();
+  charge_ = 0;
```
wkaluza: Set `charge_limit_` to zero here for completeness?

libs/ledger/src/upow/synergetic_contract.cpp
```diff
@@ -400,6 +446,9 @@ char const *ToString(SynergeticContract::Status status)
   case SynergeticContract::Status::GENERAL_ERROR:
     text = "General Error";
     break;
+  case SynergeticContract::Status ::VALIDATION_ERROR:
```
wkaluza: Rogue space.
```suggestion
  case SynergeticContract::Status::VALIDATION_ERROR:
```

libs/ledger/src/upow/synergetic_executor.cpp
```diff
@@ -69,9 +91,13 @@ void SynergeticExecutor::Verify(WorkQueue &solutions, ProblemData const &problem
...
-    WorkScore calculated_score{0};
-    auto      status = contract->Work(solution->CreateHashedNonce(), calculated_score);
-
+    WorkScore                  calculated_score{0};
+    SynergeticContract::Status status;
+    {
+      telemetry::FunctionTimer const timer{*work_duration_};
+      status = contract->Work(solution->CreateHashedNonce(), calculated_score);
+    }
+    // TODO(AB): fee for invalid solution?
```
wkaluza: What do you mean here?

libs/ledger/src/upow/synergetic_executor.cpp
```diff
@@ -87,15 +113,37 @@ void SynergeticExecutor::Verify(WorkQueue &solutions, ProblemData const &problem
       contract->Attach(storage_);
       ContractContext ctx(&token_contract_, solution->address(), &storage_adapter, 0);
       contract->UpdateContractContext(ctx);
-      status = contract->Complete(solution->address(), shard_mask);
-      contract->Detach();
+
+      // TODO(AB): charge limit
```
wkaluza: TODO?

scripts/end_to_end_test/end_to_end_test.yaml
wkaluza: Do we have a new release ready? We should enable this prior to merging.

scripts/end_to_end_test/end_to_end_test.yaml
```diff
@@ -239,3 +239,46 @@ steps:
...
+  - run_contract: {"contract_name": "contract001", "nodes": [0], "wait_for_blocks": 15}
+  - wait_for_blocks: {"nodes": [0], "num": 5}
+  - query_balance: {"nodes": [0], "save_as": "balance2"}
+  - execute_expression: {"nodes": [0], "expression": "balance2 <= balance1"}
+  - stop_nodes: {"nodes": [0], "remove_dbs": True}
+...
+
+---
+test_description: >
+  DISABLED - Test synergetic contract fees in case of no founds
```
wkaluza: founds -> funds. Also we should enable this prior to merging.

libs/ledger/include/ledger/fees/chargeable.hpp
```diff
@@ -0,0 +1,45 @@
...
+//
+//------------------------------------------------------------------------------
+
+#include <cstdint>
+
+namespace fetch {
+namespace ledger {
+
+/**
+ * Chargeable is an abstract interface that exposes fees occurred during transaction, contract, etc.
```
wkaluza: occurred -> incurred

scripts/end_to_end_test/end_to_end_test.yaml
```diff
@@ -210,7 +210,7 @@ steps:
 
 ---
 test_description: >
-  Transactions sync and are executed across nodes
+  DISABLED - Transactions sync and are executed across nodes
```
wkaluza: Is this test unstable?

scripts/end_to_end_test/run_end_to_end_test.py
```diff
@@ -550,7 +570,93 @@ def wait_network_ready(parameters, test_instance):
...
+        b = api.tokens.balance(entity)
+        address = Address(entity)
+        output(
+            f"Requested balance ({b}) {address.to_hex()}, saved as {variable}")
+        if not hasattr(node_instance, "_variables"):
+            setattr(node_instance, "_variables", {})
+        node_instance._variables[variable] = b
+
+
+def execute_expression(parameters, test_instance):
```
wkaluza, line 594: This function might be a good candidate to use `eval()`: https://docs.python.org/3/library/functions.html#eval

scripts/end_to_end_test/run_end_to_end_test.py
```diff
@@ -610,6 +686,30 @@ def run_steps(test_yaml, test_instance):
 
 
 def run_test(build_directory, yaml_file, node_exe, name_filter=None):
+    global command_map
+
+    command_map = {
```
wkaluza: We never write to `command_map`, so I would make it a constant at file scope (renamed to COMMAND_MAP for style). That way you would not need all the `global` declarations either.

scripts/end_to_end_test/smart_contract_tests/balance.py
```diff
@@ -101,19 +101,24 @@ def run(options):
...
 
-    api.sync(contract.action(api, 'action_test', 10000, [entity1]))
-    assert contract.query(api, 'query_action_test') == 0
+    api.sync(contract.action(api, 'action_test', 200, [entity1]))
+    v = contract.query(api, 'query_action_test')
+    assert 800 <= contract.query(api, 'query_action_test') <= 1000
 
     submit_synergetic_data(api, contract, [100, 20, 3], entity1)
-    assert contract.query(api, 'query_clear_test') == 0
+    v = contract.query(api, 'query_clear_test')
```
wkaluza, line 115: This value of v is never used.

scripts/end_to_end_test/smart_contract_tests/balance.py
```diff
@@ -101,19 +101,24 @@ def run(options):
...
     api.sync(api.contracts.create(entity1, contract, 10000))
 
+    api.sync(api.tokens.transfer(entity1, contract.address, 1000, 500))
+
     assert contract.query(api, 'query_init_test') == 0
 
-    api.sync(contract.action(api, 'action_test', 10000, [entity1]))
-    assert contract.query(api, 'query_action_test') == 0
+    api.sync(contract.action(api, 'action_test', 200, [entity1]))
+    v = contract.query(api, 'query_action_test')
```
wkaluza, line 109: This value of v is never used.

scripts/end_to_end_test/smart_contract_tests/transfer.py
```diff
@@ -71,6 +71,9 @@ def setup(api):
     api.sync(api.contracts.create(entity1, contract1, 2000))
     api.sync(api.contracts.create(entity1, contract2, 2000))
 
+    api.sync(api.tokens.transfer(entity1, contract1.address, 10000, 500))
+    api.sync(api.tokens.transfer(entity1, contract2.address, 10000, 500))
```
wkaluza: Try to remove this as discussed. If that causes this test to fail, then something fishy is going on.

scripts/end_to_end_test/smart_contract_tests/synergetic_mining.py
```diff
@@ -118,6 +118,8 @@ def run(options):
     print('Create contract...')
     api.sync(api.contracts.create(entity1, contract, 10000))
 
+    api.sync(api.tokens.transfer(entity1, contract.address, 10000, 500))
```
wkaluza, line 121: Is this necessary for the test to pass?

scripts/end_to_end_test/smart_contract_tests/ledger_context.py
```diff
@@ -80,6 +80,8 @@ def run(options):
     # deploy the contract to the network
     status = api.sync(api.contracts.create(entity1, contract, 2000))[0]
 
+    api.sync(api.tokens.transfer(entity1, contract.address, 10000, 500))
```
wkaluza: Is this necessary for the test to pass?

scripts/end_to_end_test/smart_contract_tests/init_owner_address.py
```diff
@@ -53,6 +53,8 @@ def run(options):
     # deploy the contract to the network
     status = api.sync(api.contracts.create(entity1, contract, 2000))[0]
 
+    api.sync(api.tokens.transfer(entity1, contract.address, 10000, 500))
```
wkaluza: Is this necessary for the test to pass?

scripts/end_to_end_test/smart_contract_tests/balance.py
```diff
@@ -101,19 +101,24 @@ def run(options):
...
+    assert 800 <= contract.query(api, 'query_action_test') <= 1000
```
wkaluza: Please give all assertions meaningful failure messages. Makes debugging easier.
```suggestion
    v = contract.query(api, 'query_action_test')
    assert 800 <= v <= 1000, \
        'Expected query_action_test result to be between 800 and 1000, found {}'.format(v)
```

wkaluza (changes requested):

libs/ledger/src/upow/synergetic_executor.cpp
qati: right now the limit and rate is a hard coded constant. Would be nice if the user can specify the values

scripts/end_to_end_test/end_to_end_test.yaml
qati: 🙈

qati: It was disabled because CI used older ledger-api

libs/ledger/include/ledger/fees/chargeable.hpp
qati: 🙈

libs/ledger/src/upow/synergetic_executor.cpp
wkaluza: @ejfitzgerald Should we punish people for invalid solutions here?

scripts/end_to_end_test/end_to_end_test.yaml
qati: I think it's ok now, with extra sleep. I reenabled.

scripts/end_to_end_test/run_end_to_end_test.py
qati: I'm aware of eval, it was a choose not using it, to only allow the predefined expressions. And also the variable names are simple names in the expression but in reality they are attached to test instance.

scripts/end_to_end_test/smart_contract_tests/synergetic_mining.py
qati: yes, otherwise no syn function will be executed

scripts/end_to_end_test/smart_contract_tests/ledger_context.py
qati: no

scripts/end_to_end_test/end_to_end_test.yaml
qati: 0.11.0-a1?

ejfitzgerald (approved):
LGTM

wkaluza (approved):

PR #1993, opened 2019-11-20
Extending StructuredData with support for Address, UInt256 and Buffer

pbukva:


libs/core/include/core/byte_array/const_byte_array.hpp
```diff
@@ -129,7 +129,18 @@ class ConstByteArray
...
   {
-    return {char_pointer(), length_};
+    // Avoiding undefined behaviour (C++ standard defines passing nullptr as
+    // input to `std::basic_string(char const*, ...)` as undefined behaviour):
+    if (pointer() != nullptr)
+    {
+      return {char_pointer(), length_};
+    }
+
+    // This is to keep original behaviour always returning a string instance
```
cyenyxe: This documentation is contextual to the PR, I would recommend to drop the part `This is to keep the original behaviour`.

libs/core/include/core/byte_array/const_byte_array.hpp
```diff
@@ -129,7 +129,18 @@ class ConstByteArray
...
+    // even if ConstByteArray::pointer() is nullptr.
+    return {};
+    // Alternative approach would be to throw (commented-out for now)
```
cyenyxe: Since the rest of the code base already works assuming that null pointer -> empty string, I think we can leave it as is, but adding some method documentation like the following:

> Creates a string using the bytes in the array, or an empty string if there are none.

libs/core/src/byte_array/decoders.cpp
```diff
@@ -34,6 +34,11 @@ ConstByteArray FromBase64(ConstByteArray const &str)
   // After
   // https://en.wikibooks.org/wiki/Algorithm_Implementation/Miscellaneous/Base64
 
+  if (nullptr == str.pointer())
+  {
+    return {};
```
cyenyxe: The style of this `return` and the next should be the same, instead of `{}` and `""`.

libs/core/src/byte_array/encoders.cpp
```diff
@@ -29,11 +29,6 @@ namespace byte_array {
 
 ConstByteArray ToBase64(ConstByteArray const &str)
 {
-  if (str.empty())
```
cyenyxe: If this check is removed so the return value is always `ret`, it would be nice to initialize that variable, check its size and return if empty first thing, before executing casts and a myriad of other operations.

libs/vm-modules/tests/unit/state_tests.cpp
```diff
@@ -370,14 +376,18 @@ TEST_F(StateTests, test_serialisation_of_structured_data)
     endfunction
   )";
 
+  ConstByteArray expetced_buffer{"QWERTYUIOPasdfghjkl"};
```
cyenyxe: Typo: `expected_buffer`

libs/vm/include/vm/object.hpp
```diff
@@ -42,42 +42,44 @@ class Address;
...
 {
 };
 template <typename T>
-struct IsPtr<Ptr<T>> : std::true_type
+struct IsPtrImpl<Ptr<T>> : std::true_type
 {
 };
 
+template <typename T>
+static constexpr bool IsPtr = IsPtrImpl<T>::value;
```
cyenyxe, line 65: I had been wondering why had all the `::value` disappeared from the previous files in the review 😂 I like how this change makes other calls less cluttered.

libs/vm/include/vm/object.hpp
```diff
@@ -42,42 +42,44 @@ class Address;
...
 template <typename T>
-using IsString = std::is_base_of<String, std::decay_t<T>>;
+static constexpr bool IsString = std::is_base_of<String, std::decay_t<T>>::value;
 
 template <typename T, typename R = void>
 using IfIsExternal = std::enable_if_t<
-    (!IsPtr<std::decay_t<T>>::value) && (!IsObject<std::decay_t<T>>::value) &&
-        (!IsVariant<std::decay_t<T>>::value) && (!IsPrimitive<std::decay_t<T>>::value) &&
-        (!IsString<std::decay_t<T>>::value) && (!IsAddress<std::decay_t<T>>::value),
+    (!IsPtr<std::decay_t<T>>)&&(!IsObject<std::decay_t<T>>)&&(!IsVariant<std::decay_t<T>>)&&(
```
cyenyxe: Restore the previous spacing style, it was easier to read.

cyenyxe (changes requested):
Minor comments, the extensions seem well integrated with existing code / behavior.

libs/vm/include/vm/object.hpp
pbukva: The formatting is fully under control of our styling script (`apply_style.py`) which is executed automatically as the part of our CI. Any deviation from that styling will result in failure in CI.

libs/vm-modules/src/core/structured_data.cpp
```diff
@@ -39,8 +39,173 @@ namespace vm_modules {
...
+using fetch::vm_modules::ByteArrayWrapper;
+using fetch::vm_modules::math::UInt256Wrapper;
+
+template <typename T>
+constexpr uint8_t getTypeId() noexcept;
+
+template <>
+constexpr uint8_t getTypeId<vm::String>() noexcept
+{
+  return 's';
```
ejfitzgerald, line 52: I am not really sure that I like the idea of having yet another internal serialisations format. I would have preferred that we keep structured data as a loose wrapper around our MsgPack format and that we come up with a concrete ABI design that can be stared for actions, queries state etc.

libs/vm-modules/tests/unit/state_tests.cpp
```diff
@@ -403,14 +416,23 @@ TEST_F(StateTests, test_serialisation_of_structured_data)
       assert(arr_u64.count() == 1);
       assert(arr_u64[0] == 200u64);
 
+      return data.getBuffer("buffer");
     endfunction
   )";
 
   EXPECT_CALL(toolkit.observer(), Exists(state_name));
-  EXPECT_CALL(toolkit.observer(), Read(state_name, _, _));
+  EXPECT_CALL(toolkit.observer(), Read(state_name, _, _)).Times(AnyNumber());
```
ejfitzgerald: Presumably the number of times this is called is either 1 or 2. Might be better to stipulate this.

libs/vm/include/vm/object.hpp
ejfitzgerald: While true, you can always put in the exception guards if you feel like it is really making a bad job of the formatting

libs/vm-modules/src/core/structured_data.cpp
pbukva: Yes I fully agree, I knew from the very beginning that this is going to be contentious point.
I have chosen this idiom only as the first step in order to unlock the feature in smart contracts (considering that the serialisation format is just implementation detail, invisible to public api).
This implementation is intended to be changed to more suitable one.

I have been considering other idioms, including going directly to MsgPack instead of ledger Variant, however any of such different idioms would need additional effort to implement it comparing to currently chosen one.

libs/vm-modules/tests/unit/state_tests.cpp
pbukva: Well, this is because of stroring binary data (e.g. String, Buffer, UInt256), where the length is dynamic (hence `AnyNumber()` usage). I can change it to specific number necessary for the current test (I believe it would be `2` in this particular test data configuration).

pbukva:
Failing SHA256 etch test unearthed a bit bigger issues:
1) Endianess ignorance while converting `ByteArray -> UInt<T>`
2) Custom implementation of `UInt256Wrapper::ToString(...)` method which was originally ignoring to string conversion of underlying `UInt<T>` type (`UInt<T>::operator std::string()` method). When I corrected that in this PR, the issue popped.
3) The `SHA256Wrapper::Final()` function returns `UInt256Wrapper` rather than better suited `ByteArrayWrapper` exposing resulting hash to endianess issues mentioned in above two points.

libs/core/src/byte_array/encoders.cpp
pbukva: This peiece of the code will go back, it has been commited by mistake (was originally done only for debugging purposes).

markos:
fyi, @pbukva the UInt endianness issue is something that @vhavrylov-ncube has also discovered while fixing issues with UInt class in #1999 .

libs/vm/include/vm/object.hpp
pbukva: I can change this using the guards.
But, to be honest, code formatting/style is quite personal preference endeavour and can vary from person to person.

However, if these kind of comments are going to be a thing in long run, it is a bit contra productive to do this manually  (either after `apply_style.py` run, or during PR review where making&handling them is quite expensive). We should rather make a change to `apply_style.py` which shall be point of reference for code style.

libs/math/tests/unit/basic_math/big_number_tests.cpp
```diff
@@ -177,26 +177,24 @@ TEST(big_number_gtest, to_double_tests)
 }
 
 template <typename LongUInt>
-bool IsTrimmedSizeValid()
+void TestTrimmedWideSize()
 {
-  bool     isValid = LongUInt(uint64_t(0)).TrimmedSize() == 1;
+  EXPECT_EQ(0, LongUInt(uint64_t(0)).TrimmedWideSize());
```
vhavrylov-ncube, line 182: I guess TrimmedSize() or TrimmedWideSize() of a zero are still 1, because "0" is still a valid number, just occupying the least wide element? If no, it would be great to have some description in TrimmedWideSize() about what does it return.

libs/vm-modules/tests/unit/state_tests.cpp
pbukva: The point I wanted to make with using the `AnyNumber()` was that this aspect is *not* relevant for this test objective (it has been already tested by many unit tests that `Read(...)` method works properly in terms of how many times it is necessary to be called when retrieving data).

libs/math/tests/unit/basic_math/big_number_tests.cpp
pbukva: The Trimmed*Size() methods follow the size idiom used in containers (e.g. native C++ array, or std containers like std::array<...>, std::list<...>, etc. ...), so that methods can be used in cycles - for example `for (std::size_t i =0; i<TrimmedSize(); ++i) {...}` where cycle won't execute if `TrimmedSize()` returns `0`.

scripts/etch-language-tests/uint256.etch
```diff
@@ -2,6 +2,7 @@ function main()
 
    var uint256 = UInt256(100u64);
 
-   assert(toString(uint256) == "6400000000000000000000000000000000000000000000000000000000000000");
+   var expected_hex_str_in_big_endian_repr = "0000000000000000000000000000000000000000000000000000000000000064";
```
markos, line 5: that's actually little endian representation

pbukva: That should be big endian - the least significant byte is on the most right side of the array.
For example *big* endian representation of the `uint64(100<DEC>)` in memory will be `{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x64}`, and *little* endian will be `{0x64, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}`

markos: yes, you're right, I was confused, ignore this comment :)

pbukva:
This PR has been split in to two PRs - #2010 and #2011, in order to separate independent features and allow them to be considered & reviewed separatelly.

pbukva:
This PR has been split in to two PRs - #2010 and #2011, in order to separate independent features and allow them to be considered & reviewed separately.

PR #1994, opened 2019-11-20
Improvement/forward sync

denis-trapeznikov:
A basic implementation of sync-forward on startup of MainChain service.


libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -202,12 +214,16 @@ class MainChain
   /// @name Block Lookup
   /// @{
   BlockStatus InsertBlock(IntBlockPtr const &block, bool evaluate_loose_blocks = true);
-  bool LookupBlock(BlockHash const &hash, IntBlockPtr &block, bool add_to_cache = false) const;
```
denis-trapeznikov, line 203: At all callsitest where `add_to_cache` was not defaulted, it was explicitly `false`. I've assumed it's not needed at all.

libs/core/include/core/serializers/base_types.hpp
```diff
@@ -984,5 +984,45 @@ struct ArraySerializer<vectorise::UInt<S>, D>
   }
 };
 
+template <class T, class D, class... Fields>
```
n-hutton: This seems like an unrelated change and so I would put it in a different PR. However, I think that @ejfitzgerald already gave some reasons not to abstract this away and so I would not put it in the codebase.

libs/logging/include/logging/logging.hpp
```diff
@@ -58,7 +40,8 @@ std::string Format(Args &&... args)
 {
   // unroll all the arguments and generate the formatted output
   std::ostringstream oss;
-  Unroll<Args...>::Apply(oss, std::forward<Args>(args)...);
+  value_util::ForEach([&oss](auto &&arg) { oss << std::forward<decltype(arg)>(arg); },
```
n-hutton: This seems unrelated and so I would include it in another PR if this is what you are proposing. However, you might want to check with @ejfitzgerald and @troelsfr whether this is worth it. Seems like to me that code readability is reduced here

libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -176,12 +182,18 @@ class MainChain
...
-    uint64_t block_number{0};
+    using TipStats = std::tuple<Block::Weight, uint64_t, Block::Weight, BlockHash const &>;
+
+    Block::Weight total_weight{0};
+    Block::Weight weight{0};
+    uint64_t      block_number{0};
     // assuming every chain has a proper genesis
     BlockHash hash{chain::GENESIS_DIGEST};
 
+    constexpr TipStats Stats() const
```
n-hutton: It seems like this is here to make comparing tips become 
```
           return HeaviestTip::TipStats{a_total_weight, a_height, a_weight, a_hash} <
                 HeaviestTip::TipStats{b_total_weight, b_height, b_weight, b_hash};
```

When this could instead be resolved by defining < operator on the Tip?

libs/ledger/include/ledger/chain/time_travelogue.hpp
```diff
@@ -0,0 +1,53 @@
...
+ * Packet format used to convey heaviest chain for node sync.
+ */
+template <class B>
+struct TimeTravelogue
+{
+  using BlockType = B;
+  using Blocks    = std::vector<BlockType>;
+
+  Blocks  blocks;
+  int64_t next_direction;
```
n-hutton: It is not clear why next direction is 64 bits long here? Also, I would put the information about next_direction at the point it is defined (here). Templating on Block here seems unneccessary since it is only ever templated on a single type?

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -453,71 +557,126 @@ MainChain::Blocks MainChain::GetChainPreceding(BlockHash start, uint64_t limit)
...
 
   return result;
 }
 
 /**
- * Walk the block history collecting blocks until either genesis or the block limit is reached.
- * Unlike in GetChainPreceding, positive value in limit indicates forward-travel.
+ * Walk the block history collecting blocks until either genesis or the block_number limit reached.
+ * limit specifies destination block_number.
+ * If current_hash is empty, travel starts from tip with zero limit, and from genesis otherwise.
```
n-hutton: "zero limit" - It seems like this will allow an attack on nodes where you force them to walk the whole chain from tip to genesis with a single RPC call

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -453,71 +557,126 @@ MainChain::Blocks MainChain::GetChainPreceding(BlockHash start, uint64_t limit)
...
+  IntBlockPtr block;
+  BlockHash   next_hash;
 
   FETCH_LOCK(lock_);
+  if (!current_hash.empty())
+  {
+    if (!LookupBlock(current_hash, block, direction > 0 ? &next_hash : nullptr))
+    {
+      FETCH_LOG_ERROR(LOGGING_NAME, "Block lookup failure for block: ", ToBase64(current_hash));
+      throw std::runtime_error("Failed to lookup block");
```
n-hutton: I think we should avoid throws in the code if possible. In this instance you can cause a node to throw (and probably crash with an RPC call containing an invalid hash

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -326,6 +364,15 @@ MainChainRpcService::State MainChainRpcService::OnRequestHeaviestChain()
   return next_state;
 }
 
+inline bool Match(MainChainRpcService::Block const &left, MainChainRpcService::Block const &right)
+{
+  return left.hash == right.hash && left.previous_hash == right.previous_hash &&
```
n-hutton: Block has an == operator, seems this should be used rather than this function

libs/ledger/include/ledger/protocols/main_chain_rpc_service.hpp
```diff
@@ -162,6 +167,13 @@ class MainChainRpcService : public muddle::rpc::Server,
   Promise         current_request_;
   /// @}
 
+  /// @name Transient metadata for synchronisation — gap edges
+  /// @{
+  BlockPtr left_edge_;
```
n-hutton: It is not clear what left and right edges could mean here from first inspection

libs/logging/include/logging/logging.hpp
n-hutton: If I try to compile it, it produces:

```
In file included from /Users/nhutton/repos/ledger/libs/logging/include/logging/logging.hpp:20:
In file included from /Users/nhutton/repos/ledger/libs/logging/include/logging/backtrace.hpp:40:
In file included from /Users/nhutton/repos/ledger/vendor/backward-cpp/backward.hpp:68:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/algorithm:641:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/type_traits:2319:18: fatal error: recursive template instantiation exceeded maximum depth of 1024
forward(typename remove_reference<_Tp>::type& __t) _NOEXCEPT
                 ^
/Users/nhutton/repos/ledger/libs/logging/include/logging/logging.hpp:44:23: note: while substituting explicitly-specified template arguments into function template 'forward'
                      std::forward<Args>(args)...);
                      ^
/Users/nhutton/repos/ledger/libs/logging/include/logging/logging.hpp:43:15: note: in instantiation of function template specialization 'fetch::detail::Format<(lambda at /Users/nhutton/repos/ledger/libs/logging/include/logging/logging.hpp:43:23), (lambda at /Users/nhutton/repos/ledger/libs/logging/include/logging/logging.hpp:43:23), (lambda at /Users/nhutton/repos/ledger/libs/logging/include/logging/logging.hpp:43:23), (lambda at /Users/nhutton/repos/ledger/libs/logging/include/logging/logging.hpp:43:23), (lambda at /Users/nhutton/repos/ledger/libs/logging/include/logging/logging.hpp:43:23),
```

Looks like this style of sfinae recursion is more likely to max out the compiler's template recursion depth than the current implementation.

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -346,11 +393,128 @@ MainChainRpcService::State MainChainRpcService::OnWaitForHeaviestChain()
...
+            HandleChainResponse(current_peer_address_, earliest_block_it, latest_block_it);
+            if (response.next_direction != 0)
+            {
+              auto const &earliest_hash = blocks.back().hash;
+              right_edge_               = chain_.GetBlock(earliest_hash);
+              assert(right_edge_);
+            }
+          }
+        }
+        direction_ = response.next_direction;
```
n-hutton: It seems like you allow peers to tell you whether to sync forward or back. This could be turned in to an attack I believe

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -346,11 +393,128 @@ MainChainRpcService::State MainChainRpcService::OnWaitForHeaviestChain()
...
+              ++earliest_block_it;
+            }
+            else if (response.next_direction == 0)
+            {
+              FETCH_LOG_WARN(LOGGING_NAME,
+                             "Error: remote peer suggested sync stop with gap not yet closed");
+              return State::REQUEST_HEAVIEST_CHAIN;
+            }
+
+            HandleChainResponse(current_peer_address_, earliest_block_it, latest_block_it);
```
n-hutton: It looks here like you handle a chain response from a peer that does not connect to a left edge neccessarily - this won't work in practice

n-hutton (changes requested):
I think we need to catch up IRL to lay out how this should look. As it currently is, it looks too fragile and succeptible to other nodes providing you correct information. It seems from my interpretation of the code like another node could crash your node by:

- I respond with time travel indicating you need to go backwards
- You make an RPC call with start hash empty
- I respond with a chain of heavy invaid blocks that are loose (right edge does not meet left edge)
- You continually attempt to find blocks for this right edge from peers but can't, and never clear it

This situation would proably also happen if you had a fork so big your response differed to everyone else's (but was still valid)

libs/ledger/include/ledger/chain/main_chain.hpp
denis-trapeznikov: > When this could instead be resolved by defining < operator on the Tip?

Defining `operator<` would still leave a disjunction here as tips don't contain hashes. 
Isn't a _single_ comapison operator more _readable_?

libs/ledger/include/ledger/chain/time_travelogue.hpp
denis-trapeznikov: > It is not clear why next direction is 64 bits long here?

No idea, borrowed the  type from the original implementation where 64-bit uint was used to convey numbers of 10K at most.

In fact, next_dir is better defined as a separate enum.

> Templating on Block here seems unneccessary since it is only ever templated on a single type?

In fact,
it is `TimeTravelogue<BlockPtr>` in `MainChain` and `TimeTravelogue<Block>` in Protocol. Can be made the same type (need to check how `shared_ptr`s are deserialized).

libs/ledger/src/chain/main_chain.cpp
denis-trapeznikov: Below in code, answer's size is limited to `UPPER_LIMIT`, i.e. 10K, at most.

libs/ledger/src/chain/main_chain.cpp
denis-trapeznikov: Do you suggest `GetChainPreceding` be fixed as well?

Might be deserving a separate jira for exception handling.

libs/ledger/src/protocols/main_chain_rpc_service.cpp
denis-trapeznikov: `Block::operator==` only compares `hash`es but thanks, I've forgot to remove this proc.

libs/ledger/src/protocols/main_chain_rpc_service.cpp
denis-trapeznikov: This is always a fixed order: first sync forward, next maybe sync backwards, then stop. Above in this method unanticipated changes in direction are used as checks to discard a chunk.

libs/core/include/core/serializers/base_types.hpp
denis-trapeznikov: > However, I think that @ejfitzgerald already gave some reasons not to abstract this away

Not exactly.

In the meantime, I've run into tests failing last week, when added a serialized fields but forgot to change map size. Current implementation contains a triple boilerplate per serialized type. Such things are always better automated.

libs/logging/include/logging/logging.hpp
denis-trapeznikov: > Seems like to me that code readability is reduced here

Do you mean, recursive templates with explicit last specializations are more readable than a simple fold expression?

libs/ledger/src/protocols/main_chain_rpc_service.cpp
denis-trapeznikov: Hm?

```
              if (earliest_block.previous_hash != left_edge_->hash)
              {
                FETCH_LOG_WARN(LOGGING_NAME, "The earliest block received (0x",
                               earliest_block.hash.ToHex(), ", #", earliest_block.block_number,
                               ") does not stem from the current left edge (0x",
                               left_edge_->hash.ToHex(), ", #", left_edge_->block_number, ')');
                return State::REQUEST_HEAVIEST_CHAIN;
              }
```
or what is it?

libs/ledger/CMakeLists.txt
```diff
@@ -16,20 +16,20 @@ setup_compiler()
 
 setup_library(fetch-ledger)
 target_link_libraries(fetch-ledger
-                      PUBLIC fetch-core
-                             fetch-http
-                             fetch-network
+                      PUBLIC fetch-beacon
```
n-hutton, line 19: Why is this shuffled?

libs/ledger/include/ledger/chain/main_chain.hpp
n-hutton: You could put the hash in the tip to resolve this. Personally, I wouldn't say this is more readable since you can define whatever operations you like at the Tip struct/class level and just have a < b wherever you want a comparison in the code

libs/ledger/include/ledger/chain/main_chain.hpp
n-hutton: probably not, no

libs/ledger/include/ledger/chain/time_travelogue.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include "core/serializers/base_types.hpp"
+#include "ledger/chain/block.hpp"
+
+namespace fetch {
+namespace ledger {
+
+/**
+ * Packet format used to convey heaviest chain for node sync.
+ */
+template <class B>
```
n-hutton, line 30: Can we remove the template on this/make it concrete?

libs/ledger/include/ledger/protocols/main_chain_rpc_protocol.hpp
```diff
@@ -63,9 +65,10 @@ class MainChainProtocol : public service::Protocol
     return Copy(blocks);
   }
 
-  Blocks TimeTravel(Digest start, int64_t limit)
+  Travelogue TimeTravel(Digest start)
```
n-hutton, line 68: I think this should have a function comment to the effect of: Given a digest, provide this block and a number of blocks preceeding this on the heaviest chain. Peers will return nothing if the digest is not on the heaviest chain, and choose how many blocks to provide.

libs/ledger/include/ledger/protocols/main_chain_rpc_service.hpp
```diff
@@ -127,7 +132,10 @@ class MainChainRpcService : public muddle::rpc::Server,
   /// @{
   static constexpr char const *ToString(State state) noexcept;
   Address                      GetRandomTrustedPeer() const;
-  void                         HandleChainResponse(Address const &address, BlockList block_list);
+
+  void HandleChainResponse(Address const &address, BlockList blocks);
```
n-hutton, line 136: It seems like the OnWaitForHeaviestChain code is still there, which is why you are templating on HandleChainResponse - can we not delete the COMMON_SUB_CHAIN call in OnSyncronising? It might be possible that this would hide any errors in the forward sync by effectively doing the correct thing when miners are all close to HEAD

libs/ledger/include/ledger/chain/block.hpp
```diff
@@ -69,6 +70,7 @@ class Block
   uint64_t       timestamp{0u};      ///< The number of seconds elapsed since the Unix epoch
   BlockEntropy   block_entropy;      ///< Entropy that determines miner priority for the next block
   Weight         weight = 1;         ///< Block weight
+  uint64_t       chain_label{0};     ///< The label of a heaviest chain this block once belonged to
```
n-hutton, line 73: It is not entirely clear to me what the chain label could mean in this context?

scripts/ci-tool.py
```diff
@@ -20,6 +20,7 @@
 import fetchai_code_quality
 
 BUILD_TYPES = ('Debug', 'Release', 'RelWithDebInfo', 'MinSizeRel')
+LOG_LEVELS = ('trace', 'debug', 'info', 'warn', 'error', 'critical', 'none')
```
n-hutton: This seems unrelated, I would suggest you put it in a sperate PR along with the suggested changes for MapSerializer

n-hutton (changes requested):
So, there are some changes I would reccomend - although perhaps it is not neccessary for now to get rid of the HEAVIEST_SUBCHAIN for now in the interest of getting this PR into master

libs/ledger/src/chain/main_chain.cpp
n-hutton: ah yes, it makes sense to fix GetChainPreceding in a seperate PR

libs/ledger/CMakeLists.txt
denis-trapeznikov: Simply sorted lexicographically.

libs/logging/include/logging/logging.hpp
n-hutton: Personally, I do find this less readable - I suggest you present it to @ejfitzgerald who is the arbiter of such things.

libs/ledger/include/ledger/protocols/main_chain_rpc_protocol.hpp
denis-trapeznikov: Except for it's not an accurate description how it currently works, they can be off the heaviest chain provided they are proof.

libs/ledger/include/ledger/chain/block.hpp
denis-trapeznikov: It's a number that blocks on the heaviest chain share, from tip down to some block, purportedly the first fork.

scripts/ci-tool.py
denis-trapeznikov: Sure.
One problem with separate PRs is that often you need these changes straight away, not once a separate PR is merged.

libs/ledger/include/ledger/chain/time_travelogue.hpp
denis-trapeznikov: > Can we remove the template on this/make it concrete?

Then one minor issue arises. On the receiving end, when you cast `MainChainRpcService` to `Travelogue`, the blocks it's comprised of are to be inserted into the main chain and that's it. Current `MainChain`'s interface inserts blocks by reference, so, IMO, to deserialize them as vector of `shared_ptr`s would induce an unneeded overhead. OTOH, when they are collected in `MainChain::TimeTravel`, they are collected as pointers of `block_chain_`, and not copied (because they only need to be collected and serialized, pointers are good for that ... perhaps).

One obvious change (besides ignoring the fact of this overhead existing at all) would be to allow `MainChain` to insert blocks by shared pointer; then they can be deserialized as ready objects on the heap and inserted as such, without any extra copies or allocations; but this looks like a subject for a different, much less PR, where this templating could be eliminated successfully.

scripts/ci-tool.py
denis-trapeznikov: https://github.com/fetchai/ledger/pull/2057

libs/ledger/include/ledger/chain/block.hpp
denis-trapeznikov: Commented in block.hpp and main_chain.hpp.

libs/ledger/include/ledger/chain/main_chain.hpp
denis-trapeznikov: Largely reworked recently.

libs/ledger/include/ledger/chain/time_travelogue.hpp
denis-trapeznikov: Sync being only done forward, this is no longer a field to the struct.

libs/ledger/src/chain/main_chain.cpp
denis-trapeznikov: Removed at all—`TimeTravel()` now only has a single explicit argument.

libs/ledger/include/ledger/protocols/main_chain_rpc_service.hpp
denis-trapeznikov: Sync only done forward, gap edges are no longer used.

libs/ledger/src/protocols/main_chain_rpc_service.cpp
denis-trapeznikov: This is always a fixed order: first sync forward, then stop.

libs/ledger/src/protocols/main_chain_rpc_service.cpp
denis-trapeznikov: Currently, consistency checks are only performed with `ConsensusInterface::ValidBlock()`.

libs/ledger/include/ledger/chain/time_travelogue.hpp
n-hutton: Let's just open a jira ticket to do this refactoring in subsequent PRs. It seems that it should be fine to use shared pointers to blocks outside of the main chain (in block coordinator and main_chain_rpc), and use a std::move to add blocks to the main chain.

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -403,47 +529,82 @@ bool MainChain::RemoveBlock(BlockHash const &hash)
...
 
   FETCH_LOCK(lock_);
 
+  // asserting genesis block has a number of 0, and everything else is above
+  assert(GetBlock(chain::GENESIS_DIGEST));
+  assert(GetBlock(chain::GENESIS_DIGEST)->block_number == 0);
+#ifndef NDEBUG
+  for (auto const &block_entry : block_chain_)
+  {
+    assert(block_entry.second->block_number > 0 || block_entry.first == chain::GENESIS_DIGEST);
```
n-hutton: Is this NDEBUG neccessary? Assertions depend on NDEBUG anyway - I believe in Release mode we do not define it, whereas in Debug mode we do

n-hutton (changes requested):
OK, the proposed alternative MapSerializer is still in this PR and should be moved to a different PR. There are some other minor comments. I think in subsequent PRs to this we should:

Rework main chain to use block pointers instead of blocks
De-template TimeTravelogue and HandleChainResponse

It seems to me that TimeTravelogue would have been fine using pointers and having a copy at the point you call AddBlock. Optimising this using templates seems to be a premature optimisation; the cost of a memory copy is nothing in comparison to how inefficient our block sync protocol is algorithmically, even ignoring the fact that network events are on the scale of 100s of ms.

libs/ledger/src/chain/main_chain.cpp
denis-trapeznikov: > Is this NDEBUG neccessary?

Not at all, provided we're sure that the empty loop will be optimized out in Release mode.

n-hutton (approved):

PR #1997, opened 2019-11-20
Genesis state file consistent storage location

ejfitzgerald:


n-hutton (changes requested):
This isn't right - the genesis file location has the JSON configuration you will see at t=0. The db_prefix prepends the file which contains the serialized genesis block, a seperate thing.

ejfitzgerald:
> This isn't right - the genesis file location has the JSON configuration you will see at t=0. The db_prefix prepends the file which contains the serialized genesis block, a seperate thing.

I understand. The thing that has changed is the construction parameter which is only used for the db prefix.

Concretely as it is in `master` this only works if a genesis file location is **NOT** specified. In this case the value of `cfg_.genesis_file_location` is an empty string, then everything is golden. However, if the user specifies the location of the genesis file, as we do in the deployment configuration. Then things start going a little wonky.

In the case of the deployment the genesis file location is in fact the following path: `/app/config/genesis-file`. With the existing logic the DB path would therefore be translated into the following: `/app/config/genesis-filegenesis_block.db`. In the case of the deployment `/app/config` is actually a read only directory. Which prompted the discovery of this issue.

More generally, the premise that we have always worked off historically, was the the working directory for the process was implicitly the output directory for all the files for the ledger. This is why in the deployment the process is started in the `/app/data` directory and this is also the place where we would mount block storage.

This change brings the `GenesisFileCreator` in line with all the lane services by using the db prefix for this prefix - something I believe that has been used in the past mostly for running multiple instances of constellation in the same directory. This update means that regardless of the path for the genesis json file, this genesis state file will be a file relative to the working directory of the process. By default this means that output file would be called `nodestoragegenesis_block.db` - not really a super great name but more consistent with our other database files and importantly for deployment, not in the `/app/config/` directory.

The existing behaviour is kept, since the genesis JSON file (or the default file location if the setting is empty) is passed to the `LoadFile` method on the `GenesisFileCreator` object.

I apologise, I should have made this explicit in the description. Hopefully this makes sense, or I have completely miss understood something 😁

n-hutton (approved):
You're right - a howler for sure. My apologies

PR #1999, opened 2019-11-21
Bugfix/ml 224 uint256 issues

vhavrylov-ncube:
- imporved ToDouble conversion accuracy
- updated unit tests

### DIFF
--- a/libs/vectorise/include/vectorise/uint/uint.hpp
+++ b/libs/vectorise/include/vectorise/uint/uint.hpp
@@ -1041,7 +1041,7 @@ template <uint16_t S>
 constexpr uint64_t UInt<S>::TrimmedSize() const
 {
   uint64_t ret = WIDE_ELEMENTS;
-  while ((ret != 0) && (wide_[ret - 1] == 0))
+  while ((ret > 1) && (wide_[ret - 1] == 0))
   {
     --ret;
   }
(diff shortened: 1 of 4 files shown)


libs/vm-modules/tests/unit/big_number_wrapper_tests.cpp
```diff
@@ -360,4 +348,28 @@ TEST_F(UInt256Tests, DISABLED_uint256_type_casts)
   EXPECT_TRUE(result_is_ok);
 }
 
+TEST_F(UInt256Tests, DISABLED_uint256_sha256_assignment)
```
markos, line 352: why is this test disabled?

vhavrylov-ncube: It will fail now, as the constructor-from-bytes of the UInt256   or   a serialization of SHA256().final() needs to be changed, but I believe such rework should be planned future.

markos: ok, could you add this justification as a comment to the test? and also create a ticket to deal with the constructor.

libs/vectorise/include/vectorise/uint/uint.hpp
```diff
@@ -1089,17 +1089,26 @@ inline std::ostream &operator<<(std::ostream &s, UInt<S> const &x)
...
+    return static_cast<double>(x.ElementAt(0));
   }
-  const uint64_t most_significant_word = x.ElementAt(msw_index);
+  const size_t ELEMENTS_PER_WIDE = x.WIDE_ELEMENT_SIZE / x.ELEMENT_SIZE;
+  uint64_t     magnitude         = x.ELEMENTS * x.ELEMENT_SIZE - x.msb();
 
-  return pow(2, static_cast<double>(msw_index * BITS_IN_UINT64)) *
-         static_cast<double>(most_significant_word);
+  const size_t most_sign_byte  = (magnitude - 1) / x.ELEMENT_SIZE;
+  const size_t least_sign_byte = most_sign_byte - (ELEMENTS_PER_WIDE - 1);
```
markos: nitpicking, but most_significant_byte/least_significant_byte sound better, sign reminds of the sign bit and is confusing at a first glance.

libs/vm-modules/tests/unit/big_number_wrapper_tests.cpp
vhavrylov-ncube: There is a detailed comment inside the test code block :) Would it be better to move it outside?
Regarding the ticket - sure, I'll do it.

libs/vectorise/include/vectorise/uint/uint.hpp
vhavrylov-ncube: Absolutely agree! I just left a temporary naming here :)

markos (approved):
LGTM

PR #2000, opened 2019-11-21
Deprecate old learner networkers implementations - Add AbstractMessageController API [COLE-72][COLE-73]

lrahmani:


libs/dmlf/include/dmlf/colearn/abstract_message_controller.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include "dmlf/colearn/colearn_update.hpp"
+
+namespace fetch {
+namespace dmlf {
+namespace colearn {
+
+class AbstractMessageController
+{
+public:
+  using Bytes          = ColearnUpdate::Data;
```
JuanBesa, line 32: Why use Bytes and not Data?

libs/dmlf/include/dmlf/colearn/abstract_message_controller.hpp
```diff
@@ -0,0 +1,55 @@
...
+  using AlgorithmClass = ColearnUpdate::Algorithm;
```
JuanBesa, line 33: I think Algorithm Class is not very clear, neither is Algorithm for that matter but I think it is slightly better. Regardless we should agree on one name and use it everywhere. Maybe we should ask the ML people.

libs/dmlf/include/dmlf/colearn/abstract_message_controller.hpp
```diff
@@ -0,0 +1,55 @@
...
+  using UpdateClass    = ColearnUpdate::UpdateType;
```
JuanBesa, line 34: Same thing here as above. Why the name change?

libs/dmlf/include/dmlf/colearn/abstract_message_controller.hpp
```diff
@@ -0,0 +1,55 @@
...
+  using UpdatePtr      = std::shared_ptr<ColearnUpdate>;
+  using ConstUpdatePtr = std::shared_ptr<const ColearnUpdate>;
+
+  AbstractMessageController()          = default;
+  virtual ~AbstractMessageController() = default;
+
+  // To implement
+  virtual void           PushUpdate(UpdatePtr const &update, AlgorithmClass const &algorithm,
```
JuanBesa, line 42: I think this needs a bit more discussion. An Update seems to have two different stages in life. When the update is created by the algorithm controller it is lacking some parts (for example the fingerprint and time stamp) which will be added by the Message Controller. So, does it make sense for the Message Controller to receive an Update? Who would it receive an Update from? Is this really a Partial Update?

libs/dmlf/include/dmlf/colearn/abstract_message_controller.hpp
```diff
@@ -0,0 +1,55 @@
...
+                                    UpdateClass const &upd_class)           = 0;
+  virtual void           PushUpdate(Bytes const &update, AlgorithmClass const &algorithm,
```
JuanBesa, line 44: This is missing the Update Metadata

libs/dmlf/include/dmlf/colearn/muddle_learner_networker_impl.hpp
```diff
@@ -36,28 +37,34 @@ namespace fetch {
...
+  using RpcServerPtr                  = std::shared_ptr<RpcServer>;
+  using Bytes                         = byte_array::ByteArray;
+  using Store                         = UpdateStore;
+  using StorePtr                      = std::shared_ptr<Store>;
+  using NetMan                        = fetch::network::NetworkManager;
+  using NetManP                       = std::shared_ptr<NetMan>;
+  using Signer                        = fetch::crypto::ECDSASigner;
+  using Randomiser                    = RandomDouble;
+  using SignerPtr                     = std::shared_ptr<Signer>;
+  using Data                          = AbstractMessageController::Bytes;
```
JuanBesa, line 67: Why is it Data again?

libs/dmlf/include/dmlf/colearn/abstract_message_controller.hpp
```diff
@@ -0,0 +1,55 @@
...
+                                    UpdateClass const &upd_class)           = 0;
+  virtual std::size_t    GetUpdateCount(AlgorithmClass const &algorithm,
+                                        UpdateClass const &   upd_class) const = 0;
+  virtual std::size_t    GetUpdateTotalCount() const                        = 0;
+  virtual ConstUpdatePtr GetUpdate(AlgorithmClass const &algorithm,
```
JuanBesa, line 49: I understand we should have moved towards having the Criteria as a the main interface. Bit I guess that will break too much code right now.

libs/dmlf/include/dmlf/colearn/muddle_learner_networker_impl.hpp
```diff
@@ -36,28 +37,34 @@ namespace fetch {
...
+  using UpdatePtr                     = AbstractMessageController::UpdatePtr;
+  using ConstUpdatePtr                = AbstractMessageController::ConstUpdatePtr;
+  using Algorithm                     = UpdateStore::Algorithm;
+  using UpdateType                    = UpdateStore::UpdateType;
```
JuanBesa, line 71: Shouldn't these definitions come from the Abstract Message Controller? If we add Criteria to the GetUpdate, the it should also be defined from there.

libs/dmlf/include/dmlf/colearn/muddle_learner_networker_impl.hpp
```diff
@@ -74,25 +81,42 @@ class MuddleLearnerNetworkerImpl : public dmlf::AbstractLearnerNetworker
 
   void addTarget(const std::string &peer);
 
-  void PushUpdate(UpdateInterfacePtr const &update) override;
-  void PushUpdateType(const std::string &type_name, UpdateInterfacePtr const &update) override;
-  void PushUpdateBytes(const std::string &type_name, Bytes const &update);
+  void PushUpdate(UpdatePtr const &update, Algorithm const &algo = "algo0",
```
JuanBesa, line 87: "algo0"

libs/dmlf/include/dmlf/collective_learning/utilities/typed_update_adaptor.hpp
```diff
@@ -0,0 +1,125 @@
...
+#include "core/mutex.hpp"
+
+#include "dmlf/colearn/abstract_message_controller.hpp"
+#include "dmlf/collective_learning/utilities/type_map.hpp"
+
+#include "core/byte_array/byte_array.hpp"
+
+namespace fetch {
+namespace dmlf {
+namespace collective_learning {
```
JuanBesa, line 31: What is the difference between namespace "colearn" and "collective_learning" do we need both?

JuanBesa: Or maybe collective_learning should be outside of dmlf?

libs/dmlf/include/dmlf/collective_learning/utilities/typed_update_adaptor.hpp
```diff
@@ -0,0 +1,125 @@
...
+  using Lock  = std::unique_lock<Mutex>;
+
+  MessageControllerPtr msg_ctrl_;
+  TypeMap<>            update_types_;
+  mutable Mutex        update_types_m_;
+
+  template <typename T>
+  Bytes Serialize(std::shared_ptr<T> const &update)
+  {
+    fetch::serializers::MsgPackSerializer serializer;
```
JuanBesa, line 107: Since this one and the one in line 116 have to be the same I think it would be worth it to declare an alias for this typename.

libs/dmlf/include/dmlf/deprecated/type_map.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <iostream>
+#include <typeindex>
+#include <typeinfo>
+#include <unordered_map>
+
+namespace fetch {
+namespace dmlf {
+
+template <typename ValueType = std::string>
+class TypeMap_
```
JuanBesa, line 31: Should this be deprecated_TypeMap_  ?
Why does it have the underscore at the end?

libs/dmlf/tests/CMakeLists.txt
```diff
@@ -10,10 +10,9 @@ include(${FETCH_ROOT_CMAKE_DIR}/BuildTools.cmake)
...
 
-fetch_add_slow_test(dmlf_local_gtest fetch-dmlf local)
-fetch_add_slow_test(dmlf_muddle_gtest fetch-dmlf muddle)
+fetch_add_slow_test(dmlf_deprecated_gtest fetch-dmlf deprecated)
 fetch_add_test(dmlf_execution_gtest fetch-dmlf execution)
 fetch_add_test(dmlf_colearn_gtest fetch-dmlf colearn)
 fetch_add_slow_test(dmlf_dmlfmuddle_gtest fetch-dmlf dmlf_muddle)
 fetch_add_test(dmlf_var_converter_test fetch-dmlf var_converter)
-fetch_add_slow_test(colearn_muddle_gtest fetch-dmlf colearn_muddle)
+fetch_add_slow_test(dmlf_colearn_muddle_gtest fetch-dmlf colearn_muddle)
```
JuanBesa, line 18: Make sure to give a heads up to Katie

JuanBesa (commented):
I think the majority of it looks good. The part I am not sure about I have commented on. I think we are not fixing the main issue which is the interaction between all the pieces that occur in the AbstractMessageController.

JuanBesa (approved):
LGTMTNAF

PR #2004, opened 2019-11-21
VMModel AddLayer refactoring

vhavrylov-ncube:
- AddLayer.. methods preparation for templatizing
- added a necessary include to Graph


libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -35,9 +34,48 @@ namespace vm_modules {
 namespace ml {
 namespace model {
 
-using SizeType    = fetch::math::SizeType;
+using std::string;
```
baykaner: i know it's more verbose, but please don't use these declarations; instead specify the namespace as you need them in the code.

libs/vm-modules/include/vm_modules/ml/model/model.hpp
```diff
@@ -66,20 +83,26 @@ class VMModel : public fetch::vm::Object
...
-                              math::SizeType const &                   output_channels,
-                              math::SizeType const &                   input_channels,
-                              math::SizeType const &kernel_size, math::SizeType const &stride_size,
-                              fetch::vm::Ptr<fetch::vm::String> const &activation);
+  template <typename... LayerArgs>
+  void AddLayer(fetch::vm::Ptr<fetch::vm::String> const &layer, LayerArgs... args);
+  template <typename... LayerArgs>
+  void AddLayerSpecificImpl(SupportedLayerType layer, SequentialModelPtr me,
+                            math::SizeType const &inputs, math::SizeType const &hidden_nodes);
+  template <typename... LayerArgs>
```
baykaner: not obvious why you need the parameter packs for the AddLayerSpecificImpl methods?

libs/vm-modules/include/vm_modules/ml/model/model.hpp
```diff
@@ -66,20 +83,26 @@ class VMModel : public fetch::vm::Object
...
-                    math::SizeType const &kernel_size, math::SizeType const &stride_size);
-  void LayerAddConvActivation(fetch::vm::Ptr<fetch::vm::String> const &layer,
-                              math::SizeType const &                   output_channels,
-                              math::SizeType const &                   input_channels,
-                              math::SizeType const &kernel_size, math::SizeType const &stride_size,
-                              fetch::vm::Ptr<fetch::vm::String> const &activation);
+  template <typename... LayerArgs>
+  void AddLayer(fetch::vm::Ptr<fetch::vm::String> const &layer, LayerArgs... args);
+  template <typename... LayerArgs>
+  void AddLayerSpecificImpl(SupportedLayerType layer, SequentialModelPtr me,
```
baykaner: All these AddLayerSpecificImpl methods should be private right?

baykaner (changes requested):

libs/vm-modules/include/vm_modules/ml/model/model.hpp
vhavrylov-ncube: Oh, sure!

vhavrylov-ncube: done

libs/vm-modules/include/vm_modules/ml/model/model.hpp
vhavrylov-ncube: oops, an artifact :)

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: done

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -35,9 +34,60 @@ namespace vm_modules {
...
+};
+
+std::map<std::string, uint8_t> const VMModel::model_categories_{
+    {"none", static_cast<uint8_t>(ModelCategory::NONE)},
+    {"sequential", static_cast<uint8_t>(ModelCategory::SEQUENTIAL)},
+    {"regressor", static_cast<uint8_t>(ModelCategory::REGRESSOR)},
+    {"classifier", static_cast<uint8_t>(ModelCategory::CLASSIFIER)},
+};
+
+template <typename T>
```
baykaner, line 86: add doxygen style function comment and explain that this function:
"converts between user specified string and output type (e.g. activation, layer etc.)"

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -35,9 +34,60 @@ namespace vm_modules {
...
 namespace model {
 
-using SizeType    = fetch::math::SizeType;
+using fetch::math::SizeType;
+using fetch::ml::ops::LossType;
+using fetch::ml::OptimiserType;
+using fetch::ml::details::ActivationType;
 using VMPtrString = Ptr<String>;
 
+std::map<std::string, SupportedLayerType> const VMModel::layer_types_{
```
baykaner, line 43: wouldn't it be better if these were unordered maps?

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -35,9 +34,60 @@ namespace vm_modules {
...
+    {"sequential", static_cast<uint8_t>(ModelCategory::SEQUENTIAL)},
+    {"regressor", static_cast<uint8_t>(ModelCategory::REGRESSOR)},
+    {"classifier", static_cast<uint8_t>(ModelCategory::CLASSIFIER)},
+};
+
+template <typename T>
+inline T VMModel::ParseName(std::string const &name, std::map<std::string, T> const &dict,
+                            std::string const &errmsg) const
+{
+  auto found = dict.find(name);
```
baykaner: in general the style is to use auto to avoid typename repetitions; here, however, it's not obvious what type of thing auto is by looking at the right hand side.
T found = dict.find(name);
is preferable

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -35,9 +34,60 @@ namespace vm_modules {
...
+  if (found == dict.end())
+  {
+    throw std::runtime_error("Unknown " + errmsg + " name : " + name);
```
baykaner: vm runtime error would be better wouldn't it?

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -58,28 +108,12 @@ VMModel::VMModel(VM *vm, TypeId type_id, std::string const &model_category)
 
 void VMModel::Init(std::string const &model_category)
 {
-  model_config_ = std::make_shared<ModelConfigType>();
+  model_category_ = ModelCategory(ParseName(model_category, model_categories_, "model category"));
```
baykaner: split this into two lines for clarity

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -89,233 +123,65 @@ Ptr<VMModel> VMModel::Constructor(VM *vm, TypeId type_id,
...
-  {
-    optimiser_type = fetch::ml::OptimiserType::SGD;
-  }
-  else
-  {
-    throw std::runtime_error("invalid optimiser");
-  }
-
+  auto const loss_type      = ParseName(loss->string(), losses_, "loss function");
+  auto const optimiser_type = DeduceOptimizer(optimiser->string());
```
baykaner: don't use auto for these two cases please.

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -89,233 +123,65 @@ Ptr<VMModel> VMModel::Constructor(VM *vm, TypeId type_id,
...
 
 void VMModel::CompileSimple(fetch::vm::Ptr<fetch::vm::String> const &        optimiser,
                             fetch::vm::Ptr<vm::Array<math::SizeType>> const &in_layers)
 {
-  // construct the model with the specified layers
-  auto                        n_elements = in_layers->elements.size();
-  std::vector<math::SizeType> layers(n_elements);
-  for (std::size_t i = 0; i < n_elements; ++i)
+  OptimiserType optimiser_type = DeduceOptimizer(optimiser->string());
+  if (optimiser_type != OptimiserType::ADAM)
```
baykaner, line 165: why is this if condition here?
SGD should work fine too shouldn't it?

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -89,233 +123,65 @@ Ptr<VMModel> VMModel::Constructor(VM *vm, TypeId type_id,
...
   {
-    layers.at(i) = in_layers->elements.at(i);
+    throw std::runtime_error(
+        R"(Wrong optimiser, a "Simple" model can use only "adam", while given : )" +
+        optimiser->string());
   }
 
+  auto BuildLayersFromInput = [&in_layers]() {
```
baykaner: this is really unclear.
setting up a lambda function to return to an auto and passing to the DNNRegressor constructor hides a lot about what is going on.

If you want to functionalise this piece of code then make a static member function called BuildLayersFromInput that returns a vector<SizeType>. Invoke that function, and pass the results to the constructor when you need to.

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -89,233 +123,65 @@ Ptr<VMModel> VMModel::Constructor(VM *vm, TypeId type_id,
...
-  fetch::ml::OptimiserType optimiser_type;
-  if (optimiser->string() == "adam")
-  {
-    optimiser_type = fetch::ml::OptimiserType::ADAM;
-  }
-  else
-  {
-    throw std::runtime_error("invalid optimiser");
+  default:
+    throw std::runtime_error(
```
baykaner: vm runtime error would work fine?

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -89,233 +123,65 @@ Ptr<VMModel> VMModel::Constructor(VM *vm, TypeId type_id,
...
+        "Only REGRESSOR and CLASSIFIER model types take layers on compilation!");
   }
 
   model_->Compile(optimiser_type);
 }
+
+// TODO(VH): Why the method is called "Fit" while it does only 1 epoch training?
```
baykaner: delete comment. TODO's must have issue numbers.
To answer your question it matches the Keras api (except we haven't implemented a way to specify number of epochs to fit for yet - but keras defaults to 1)

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -89,233 +123,65 @@ Ptr<VMModel> VMModel::Constructor(VM *vm, TypeId type_id,
...
 void VMModel::Fit(vm::Ptr<VMTensor> const &data, vm::Ptr<VMTensor> const &labels,
                   fetch::math::SizeType const &batch_size)
 {
-  // prepare dataloader
-  auto dl = std::make_unique<TensorDataloader>();
-  dl->SetRandomMode(true);
-  dl->AddData({data->GetTensor()}, labels->GetTensor());
-  model_->SetDataloader(std::move(dl));
+  auto data_loader = std::make_unique<TensorDataloader>();
```
baykaner, line 215: here is an example where auto is preferred because the Type is spelled out on the right hand side. so this one is fine.

baykaner, line 215: why delete the comments in the function though?

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -453,6 +297,129 @@ fetch::vm::Ptr<VMModel> VMModel::DeserializeFromString(
...
+                                     std::vector<SupportedLayerType> &&valids) const
+{
+  static const std::map<SupportedLayerType, std::string> LAYER_NAMES_{
+      {SupportedLayerType::DENSE, "dense"},
+      {SupportedLayerType::CONV1D, "conv1d"},
+      {SupportedLayerType::CONV2D, "conv2d"},
+  };
+  if (std::find(valids.begin(), valids.end(), layer) == valids.end())
+  {
+    throw std::runtime_error("Invalid params specified for \"" + LAYER_NAMES_.at(layer) +
```
baykaner: vm runtime error again?

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -453,6 +297,129 @@ fetch::vm::Ptr<VMModel> VMModel::DeserializeFromString(
...
+                             "\" layer.");
+  }
+}
+
+VMModel::SequentialModelPtr VMModel::GetMeAsSequentialIfPossible()
+{
+  if (model_category_ != ModelCategory::SEQUENTIAL)
+  {
+    throw std::runtime_error("No \"add\" method exists for non-sequential models!");
```
baykaner: more vm runtime error

libs/vm-modules/include/vm_modules/ml/model/model.hpp
```diff
@@ -112,7 +117,46 @@ class VMModel : public fetch::vm::Object
...
+                            fetch::ml::details::ActivationType activation);
+
+  inline void AssertLayerTypeMatches(SupportedLayerType                layer,
+                                     std::vector<SupportedLayerType> &&valids) const;
+
+  template <typename T>
+  inline T ParseName(std::string const &name, std::map<std::string, T> const &dict,
+                     std::string const &errmsg) const;
+
+  inline fetch::ml::details::ActivationType DeduceActivation(std::string const &name) const;
```
baykaner: remove the DeduceX functions; the implementations are one line only, and you only use each one twice, and it's extremely clear what the inputs to Parse should be in context

baykaner (changes requested):

libs/vm-modules/include/vm_modules/ml/model/model.hpp
vhavrylov-ncube: Done

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: I prefer maps for cases when there are up to 10-15 elements. Hash computing could be a bit more expensive then short string comparisons, and O(log(n)) search is quite as fast as O(1) when N is small.

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: "auto" is a perfect way to avoid explicit usage of huge typenames, that are necessary to work with STL containers. In this particular case, it would be a 
typename std::map<std::string, T>::const_iterator found = ...
which from my point of view is not so much more clear than a plain "auto" :)

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: Oh sure! I forgot about it :)

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: done

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: done

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: let's keep them indeed

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: UPD: our static analyzer also thinks "auto" is suitable here.

vhavrylov-ncube: refactored without a need for "auto" usage :)

libs/vm-modules/src/ml/model/model.cpp
baykaner: doesn't look done

libs/vm-modules/tests/unit/ml.cpp
```diff
@@ -460,7 +459,7 @@ TEST_F(MLTests, serialisation_several_components_test)
 /////
 /// MODEL SERIALISATION TESTS
 /////
-TEST_F(MLTests, serialisation_model)
+TEST_F(MLTests, serialization_model)
```
baykaner: why switch to american spelling?

libs/vm-modules/tests/unit/ml.cpp
```diff
@@ -683,7 +682,7 @@ TEST_F(MLTests, model_string_serialisation_test)
   ASSERT_TRUE(toolkit.Run());
 }
 
-TEST_F(MLTests, non_permitted_serialisation_model_sequential_test)
+TEST_F(MLTests, non_permitted_serialization_model_sequential_test)
```
baykaner: again, why rename?

libs/vm-modules/tests/unit/ml.cpp
```diff
@@ -743,7 +742,28 @@ TEST_F(MLTests, non_permitted_serialisation_model_classifier_test)
     )";
 
   ASSERT_TRUE(toolkit.Compile(model_classifier_serialise_src));
-  EXPECT_FALSE(toolkit.Run());
+}
```
baykaner, line 746: 👍

baykaner (changes requested):

baykaner:
a few unresolved issues still

libs/vm-modules/src/ml/model/model.cpp
baykaner: okay we'll extend this in https://fetchai.atlassian.net/browse/ML-273

libs/vm-modules/tests/unit/ml.cpp
vhavrylov-ncube: oh no, it's a merge solving mistake

libs/vm-modules/tests/unit/ml.cpp
vhavrylov-ncube: done

libs/vm-modules/tests/unit/ml.cpp
vhavrylov-ncube: .SerializeTo(...) tempted me :) done.

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: It was done in hpp file :) moved here now.

baykaner (approved):
lgtm

PR #2005, opened 2019-11-22
Dev/pick more than one starting peer

engineerkatie:


libs/dmlf/src/networkers/muddle_learner_networker.cpp
```diff
@@ -75,10 +75,32 @@ MuddleLearnerNetworker::MuddleLearnerNetworker(const std::string &cloud_config,
...
+  auto                            config_peers = doc.root()["peers"];
+
+  auto config_peer_count = config_peers.size();
+  for (std::size_t peer_number = 0; peer_number < config_peer_count; peer_number++)
+  {
+    if (peer_number != instance_number)
+    {
+      peers_.emplace_back(config_peers[peer_number]["pub"].As<std::string>());
+    }
+  }
```
evsmithx, line 89: This duplicates code that is below at line 112 - I think this is unintentional?

JuanBesa (dismissed):

libs/dmlf/src/networkers/muddle_learner_networker.cpp
```diff
@@ -75,10 +75,32 @@ MuddleLearnerNetworker::MuddleLearnerNetworker(const std::string &cloud_config,
...
+
+  if (config_peer_count < 10)
```
evsmithx: Could we make 10 a parameter so that it's clear what this is doing? (I.e. replace the 10 below as well with something like min_peers)

libs/dmlf/src/networkers/muddle_learner_networker.cpp
```diff
@@ -75,10 +75,32 @@ MuddleLearnerNetworker::MuddleLearnerNetworker(const std::string &cloud_config,
...
   {
     initial_peers.insert(doc.root()["peers"][0]["uri"].As<std::string>());
   }
+  else
+  {
+    for (unsigned int i = 0; i < 10; i++)
+    {
+      auto peer_number = (instance_number + 1 + i) % config_peer_count;
+      if (peer_number != instance_number)
```
evsmithx: If we change the loop condition to i <= 10 then we can remove this if statement

evsmithx (changes requested):

libs/dmlf/src/networkers/muddle_learner_networker.cpp
engineerkatie: Rabbits. deleted the other one.

libs/dmlf/src/networkers/muddle_learner_networker.cpp
engineerkatie: No, we can't there's a modulo in there to cause trouble.

JuanBesa (dismissed):

evsmithx (dismissed):

JuanBesa (approved):

PR #2006, opened 2019-11-22
VMTensor .At() and .setAt() methods refactoring

vhavrylov-ncube:


libs/vm-modules/src/math/tensor.cpp
```diff
@@ -57,19 +57,20 @@ Ptr<VMTensor> VMTensor::Constructor(VM *vm, TypeId type_id, Ptr<Array<SizeType>>
...
       })
-      .CreateMemberFunction("at", &VMTensor::AtOne)
-      .CreateMemberFunction("at", &VMTensor::AtTwo)
-      .CreateMemberFunction("at", &VMTensor::AtThree)
-      .CreateMemberFunction("at", &VMTensor::AtFour)
-      .CreateMemberFunction("setAt", &VMTensor::SetAtOne)
-      .CreateMemberFunction("setAt", &VMTensor::SetAtTwo)
-      .CreateMemberFunction("setAt", &VMTensor::SetAtThree)
-      .CreateMemberFunction("setAt", &VMTensor::SetAtFour)
+      .CreateMemberFunction("at", &VMTensor::At<Index>)
```
baykaner, line 66: it's annoying but there's no obvious way around this right now is there?
If so, would you mind please adding in the bindings for the 5 & 6 dimensional tensor accessor/setter as well. There are some cases with convolution where you might get such a tensor.

baykaner: you can do a separate pr for that if you like

baykaner (dismissed):

libs/vm-modules/src/math/tensor.cpp
vhavrylov-ncube: Done

baykaner (approved):
👍

PR #2008, opened 2019-11-24
Tokenisation: prevent leading zeros and dangling decimal points

wkaluza:


ejfitzgerald (commented):
Looks good to me - just check with @robertdickson  and we should be good to go

robertdickson (approved):
looks great

PR #2010, opened 2019-11-25
Endianess handling of UInt<T> <-> ByteArray conversion

### LINKED ISSUE #1383: UInt<BITS>: Fix handling of residual bits in last element of `wide_[...]` array (bits beyond BITS)
In the current implementation (commit: e8dcf6a45f07bd44484de5cb8e750e66b19e1949), in the case when `UInt<BITS>::wide_[WIDE_ELEMENTS]` array is bigger(bit-wise) than requested `BITS` size (when `0 != BITS % (sizeof(WideElement) * 8)`), the _last_ element of the `UInt<BITS>::wide_` array data member (beyond `BITS`) shall be cleared/masked out after _each_ operation to avoid unexpected effects/impact on result from sequence of operations, for example:
the `(UInt<72>{1} << 72) >> 72` sequence is expected to return `0` value, as analogy to the same sequence with native C++ integer type would - for example with `uint64_t` type as following code demonstrates:
```cpp
#include <iostream>
using namespace std;
int main() {
    uint64_t x{1};
    x <<= 63; x <<= 1;
    x >>= 63; x >>= 1;
    cout << "x = " << x << endl;
}
```

pbukva:
Draft of of the solution how to handle endianess during `UInt<T>`  from/to `ByteArray` conversion.

Exposure of the of the `Base58` encoding for `ByteArray` (`Buffer`) type in Etch is just proposal, since I think it can be useful/necessary for situations when contract needs to cooperate with source it can not control - e.g. with another contract (contract-to-contract calls).

Resolves #1383 

NOTE: This PR is extract from #1993 PR, to isolate it from extension of StructuredData feature.


libs/core/include/core/byte_array/const_byte_array.hpp
```diff
@@ -129,7 +129,14 @@ class ConstByteArray
 
   explicit operator std::string() const
   {
-    return {char_pointer(), length_};
+    // Avoiding undefined behaviour (C++ standard defines passing nullptr as
+    // input to `std::basic_string(char const*, ...)` as undefined behaviour):
+    if (pointer() != nullptr)
```
ejfitzgerald, line 150: I am not sure that this is undefined behaviour because the length would be zero in this case. In either case I think it makes sense to have this as an explicit check

libs/json/src/document.cpp
```diff
@@ -71,7 +71,7 @@ void JSONDocument::ExtractPrimitive(Variant &variant, JSONToken const &token,
   case NUMBER_INT:
   {
     std::string const str{document.SubArray(token.first, token.second)};
-
+    errno   = 0;
```
ejfitzgerald, line 74: Is this strictly necessary?

libs/vm-modules/tests/unit/big_number_wrapper_tests.cpp
```diff
@@ -131,23 +131,27 @@ class UInt256Tests : public ::testing::Test
   static constexpr std::size_t    SIZE_IN_BITS  = 256;
   static constexpr std::size_t    SIZE_IN_BYTES = SIZE_IN_BITS / 8;
 
-  UInt256Wrapper zero{dummy_vm_ptr, dummy_typeid, 0};
+  UInt256Wrapper zero{
```
ejfitzgerald: How come this got reformatted

libs/vm-modules/src/math/bignumber.cpp
```diff
@@ -105,11 +94,13 @@ UInt256Wrapper::UInt256Wrapper(VM *vm, TypeId type_id, uint64_t data)
 {}
 
 Ptr<UInt256Wrapper> UInt256Wrapper::ConstructorFromBytes(VM *vm, TypeId type_id,
-                                                         Ptr<ByteArrayWrapper> const &ba)
+                                                         Ptr<ByteArrayWrapper> const &ba,
+                                                         bool input_is_little_endian)
```
ejfitzgerald: Presumably it just makes sense to standardise this interface to always expecting big endian

libs/vm-modules/src/core/byte_array_wrapper.cpp
```diff
@@ -42,16 +60,160 @@ ByteArrayWrapper::ByteArrayWrapper(VM *vm, TypeId type_id, byte_array::ByteArray
...
+    byte_array_ = std::move(decoded_array);
+    return true;
+  }
+  catch (std::runtime_error const &ex)
+  {
+    return false;
+  }
+}
+
+fetch::vm::Ptr<vm::String> ByteArrayWrapper::ToBase58()
```
ejfitzgerald, line 131: I don't think byte array wrapper should have a base58 representation. I want to reserve this to mean addresses only

libs/vectorise/include/vectorise/uint/uint.hpp
```diff
@@ -1081,39 +1138,69 @@ UInt<S>::operator std::string() const
...
   }
-  const auto exponent = static_cast<double>(least_eligible_byte_idx * x.ELEMENT_SIZE);
 
-  return static_cast<double>(mantisse) * pow(2, exponent);
+  return retval;
 }
 
-inline double Log(UInt<256> const &x)
+template <typename T>
+constexpr meta::EnableIf<std::is_same<meta::Decay<T>, UInt<256>>::value, double> Log(T &&x)
```
ejfitzgerald: I am not sure if we should have this log function on the UInt256. Since we are converting to double. Does it makes sense to simply include a count leading zeros function or similar? (and then people can use it as a psuedo log or similar)

libs/ledger/src/upow/work.cpp
```diff
@@ -102,7 +102,10 @@ Work::UInt256 Work::CreateHashedNonce() const
   hasher.Reset();
   hasher.Update(digest1);
 
-  return UInt256{hasher.Final()};
+  // Forcing little endian representation of the hash (even if it is actually
+  // represented in big endian encoding) in order to keep compatibility with
+  // proofs recorded in previously generated blocks.
+  return UInt256{hasher.Final(), true};
```
ejfitzgerald: @n-hutton is in the process of removing all the proof of work stuff. It doesn't make sense to worry about backwards compat.

ejfitzgerald (changes requested):
There are a lot of changes here. In future I would recommend separating them into the strictly the updates that you are making and then have an alternative PR which includes all the updates to the VM SFINAE code. It is very difficult to get a sense of the change that is being made with all of this noise.

I have a few comments, basically around the interface design giving people the option to create a U256 from a little endian buffer. I think this is generally a bad idea and we should simply standardise for this class on BE.

libs/json/src/document.cpp
pbukva: Yes it is, and I discovered it on my linux installation while executing unit tests. C/C++ standard explicitly says for all `strtoXX(...)` / `std::strtoXX(...)` functions that the `errno` global variable is set when something goes wrong (errno value remains unchanged when conversion is successful).

pbukva: I would say so. I discovered the issue on one of my dev setups (either Linux or Mac, can't remember which one) while executing unit tests. C/C++ standard does not enforce setting `errno` global variable for any of `strtoXX(...)` / `std::strtoXX(...)` functions if conversion is successful, it only defines that the the `errno` global variable is set when something goes wrong.
[Here](https://stackoverflow.com/a/14176593) is a little bit of reading on the topic.

libs/vectorise/include/vectorise/uint/uint.hpp
pbukva: The method was there before.
Wanted to ditch it exactly due to reasons you mentioned (we have to double conversion), but decided to keep it just in case there is some Etch code which relies on it (e.g. some smart/synergetic contract tests, examples of code in the documentation, etc. ...).
I will be more than happy to ditch it.

libs/core/include/core/byte_array/const_byte_array.hpp
pbukva: C++ standard for [std::basic_string<...>](https://en.cppreference.com/w/cpp/string/basic_string/basic_string) says this for the `basic_string( const CharT* s, size_type count, const Allocator& alloc = Allocator() );`:

pbukva: By mistake I have looked up description of a little bit different method on [std::basic_string<...>](https://en.cppreference.com/w/cpp/string/basic_string/basic_string) - variant 5 (`basic_string( const CharT* s, const Allocator& alloc = Allocator() )`) instead of 4 (`basic_string( const CharT* s, size_type count, const Allocator& alloc = Allocator() );`).

It looks that for variant 4) standard just defines: `"The behaviour is undefined if [s, s + count) is not a valid range."`, where the range should be valid (mathematically speaking) if `s` is  `nullptr`.
However, that shall result in to undefined behaviour if `count > 1`, because constructor would start copying a data starting from `nullptr`, unless constructor does additional check for null-ness of the `s` (what is not required by C++ standard). This should not happen in our case because `ConstByteArray` shall ensure that if `pointer()` method returns `nullptr` then `length_` data member shall have zero value.

libs/ledger/src/upow/work.cpp
```diff
@@ -102,7 +102,10 @@ Work::UInt256 Work::CreateHashedNonce() const
   hasher.Reset();
   hasher.Update(digest1);
 
-  return UInt256{hasher.Final()};
+  // Forcing little endian representation of the hash (even if it is actually
+  // represented in big endian encoding) in order to keep compatibility with
+  // proofs recorded in previously generated blocks.
+  return UInt256{hasher.Final(), memory::Endian::BIG};
```
vhavrylov-ncube: Much nicer than bool!

libs/vm-modules/src/math/bignumber.cpp
```diff
@@ -352,12 +315,6 @@ void UInt256Wrapper::Divide(Ptr<Object> &lhso, Ptr<Object> &rhso)
     lhs->number_ /= rhs->number_;
     return;
   }
-  if (rhs->IsTemporary())
-  {
-    rhs->number_ /= lhs->number_;
```
pbukva, line 346: Removed this block, because it division is not commutative operation => rhs & lhs can NOT be swapped

libs/vm-modules/src/math/bignumber.cpp
```diff
@@ -288,12 +256,7 @@ void UInt256Wrapper::Subtract(Ptr<Object> &lhso, Ptr<Object> &rhso)
     lhs->number_ -= rhs->number_;
     return;
   }
-  if (rhs->IsTemporary())
-  {
-    rhs->number_ -= lhs->number_;
```
pbukva, line 282: Removed this block, because it substration is not commutative operation => rhs & lhs can NOT be swapped

libs/vm-modules/tests/unit/big_number_wrapper_tests.cpp
pbukva: I don't know, I don't bother/waste time with formating anymore and fully rely on `apply_style.py`.

libs/vectorise/include/vectorise/memory/endian.hpp
```diff
@@ -0,0 +1,35 @@
+#pragma once
```
markos: imho, this should go to platform.hpp, as it already has helper functions for big/little endian

libs/vectorise/include/vectorise/uint/uint.hpp
```diff
@@ -980,33 +1051,36 @@ constexpr UInt<S> &UInt<S>::operator>>=(std::size_t bits)
...
   std::size_t lsb = 0;
-  for (std::size_t i = 0; i < WIDE_ELEMENTS; i++)
+  std::size_t i   = 0;
+
+  for (; i < WIDE_ELEMENTS - 1; i++, lsb += WIDE_ELEMENT_SIZE)
   {
-    std::size_t lsbi = platform::CountTrailingZeroes64(wide_[i]);
-    lsb += lsbi;
-    if (lsbi < 64)
+    auto const val{wide_[i]};
```
ejfitzgerald, line 1077: Out of interest extract copy for readability?

libs/vm-modules/include/vm_modules/core/byte_array_wrapper.hpp
```diff
@@ -37,12 +38,27 @@ class ByteArrayWrapper : public fetch::vm::Object
...
-  static fetch::vm::Ptr<ByteArrayWrapper> Constructor(fetch::vm::VM *vm, fetch::vm::TypeId type_id,
-                                                      int32_t n);
+  static fetch::vm::Ptr<ByteArrayWrapper> Constructor(vm::VM *vm, vm::TypeId type_id, int32_t n);
 
   fetch::vm::Ptr<ByteArrayWrapper> Copy();
+  fetch::vm::Ptr<vm::String>       ToBase64();
+  bool                             FromBase64(fetch::vm::Ptr<vm::String> const &value_b64);
+  fetch::vm::Ptr<vm::String>       ToHex();
+  bool                             FromHex(fetch::vm::Ptr<vm::String> const &value_hex);
+  fetch::vm::Ptr<vm::String>       ToBase58();
```
ejfitzgerald, line 48: I would prefer not to have base58 encoding exposed on the byte array so that this is reserved for the address type. It also ensures that all base58 numbers on the system will have a checksum

markos (dismissed):
LGTM

vhavrylov-ncube (dismissed):
LGTM

ejfitzgerald (approved):

libs/vectorise/include/vectorise/uint/uint.hpp
pbukva: That was supposed to be reference (`auto const &`).

PR #2013, opened 2019-11-25
Dev/pick more than one starting peer

engineerkatie:
In order to give us more chance of building a connected swarm, pick >1 starting peer.


engineerkatie:
Code included in larger PR.

PR #2015, opened 2019-11-25
Add VMTensor member function estimators - Add basic tests [COLE-70]

lrahmani:
Provides estimators as lambda functions.

### DIFF
--- a/libs/vm-modules/src/math/tensor.cpp
+++ b/libs/vm-modules/src/math/tensor.cpp
@@ -28,11 +28,14 @@
 
 using namespace fetch::vm;
 
-namespace fetch {
-namespace vm {
-static ChargeAmount CHARGE_UNIT = 1u;
+namespace {
+fetch::vm::ChargeAmount charge_func_of_tensor_size(
+    fetch::vm::Ptr<fetch::vm_modules::math::VMTensor> const &ctx, size_t factor = 1)
+{
+  return static_cast<fetch::vm::ChargeAmount>(fetch::vm::CHARGE_UNIT * ctx->size() * factor);
 }
-}  // namespace fetch
+
+}  // namespace
 
 namespace fetch {
 namespace vm_modules {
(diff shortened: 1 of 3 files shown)


libs/vm-modules/src/math/tensor.cpp
```diff
@@ -28,6 +28,12 @@
 
 using namespace fetch::vm;
 
+namespace fetch {
+namespace vm {
+static ChargeAmount CHARGE_UNIT = 1u;
```
wkaluza: make `constexpr`. Also, this isn't a tensor-specific quantity, so it does not belong in this file. Move it to where ChargeAmount is defined - remove duplicate definitions from other files.

libs/vm-modules/src/math/tensor.cpp
```diff
@@ -57,28 +63,85 @@ Ptr<VMTensor> VMTensor::Constructor(VM *vm, TypeId type_id, Ptr<Array<SizeType>>
...
+
+  ///////////////////////////////
+  // Prepare charge estimators //
+  ///////////////////////////////
+
+  // scalar estimation for cheap operations
+  ChargeAmount low_charge{vm::CHARGE_UNIT};
+
+  // object state-dependent estimators
+  auto charge_func_of_tensor_size = [](Ptr<VMTensor> const &this_,
```
wkaluza: move this to a free function in an anonymous namespace in this file - then your estimators won't have to drag a copy around and you won't have to capture it everywhere.

libs/vm-modules/tests/unit/math.cpp
```diff
@@ -383,4 +392,111 @@ TEST_F(MathTests, tensor_set_from_string)
...
+      tensor_shape[1] = 2u64;
+      var x = Tensor(tensor_shape);
+    endfunction
+  )";
+
+  ASSERT_TRUE(toolkit.Compile(tensor_setup_src));
+  ASSERT_TRUE(toolkit.Run(nullptr, tensor_setup_charge_min)) << stdout.str();
+}
+
+TEST_F(MathTests, tensor_setup_charge_estimator_min_charge_fails)
```
wkaluza, line 404: It's good that you are using descriptive titles, but try to convey the point of each test. The reason this fails is because the charge limit is exceeded; the test above succeeds, because the charge limit is sufficient. Otherwise whoever reads this in the future needs to hunt for that `- 1` to understand what you are testing, which defeats the point of having titles.

libs/vm-modules/tests/unit/math.cpp
```diff
@@ -383,4 +392,111 @@ TEST_F(MathTests, tensor_set_from_string)
...
+TEST_F(MathTests, tensor_fill_estimator_fails)
+{
+  size_t       tensor_size = 10;
+  ChargeAmount charge_max  = tensor_setup_charge_min + 2 * fetch::vm::CHARGE_UNIT * tensor_size;
+
+  static char const *tensor_to_string_src = R"(
+    function main()
+      var tensor_shape = Array<UInt64>(2);
+      tensor_shape[0] = 5u64;
+      tensor_shape[1] = 5u64;
```
wkaluza, line 447: Looking at this code, I question if this is the right level to test this. What you are checking is that the fill estimator returns a higher value if you give it a tensor with more work to do. Why not test that directly by calling the estimator in the unit test? The only reason you need magic numbers like

```
  size_t       tensor_size = 10;
  ChargeAmount charge_max  = tensor_setup_charge_min + 2 * fetch::vm::CHARGE_UNIT * tensor_size;
```

is that this test is running more code than you are interested in, which obscures the pass/fail criteria. Try extracting the estimators to a new file and testing them directly. Also, consider that some estimators may be so simple that they don't need testing.

wkaluza (changes requested):

libs/vm-modules/src/math/tensor.cpp
baykaner: Presumably this ChargeAmount should be set somewhere centrally in the vm for re-use in all modules?

Other than this point, this all looks great 👍

PR #2016, opened 2019-11-25
Bugfix/ml 224 uint256 issues

vhavrylov-ncube:
Added a typeid for UInt256
Implemented standalone and in-place +-*/ for UInt256
added tests for +-*/


libs/vm-modules/src/math/bignumber.cpp
```diff
@@ -250,6 +258,116 @@ bool UInt256Wrapper::FromJSON(JSONVariant const &variant)
...
+    return;
+  }
+  if (rhs->IsTemporary())
+  {
+    rhs->number_ += lhs->number_;
+    lhso = std::move(rhs);
+    return;
+  }
+  Ptr<UInt256Wrapper> n(
+      new UInt256Wrapper(vm_, fetch::vm::TypeIds::UInt256, lhs->number_ + rhs->number_));
```
robertdickson: Constructor doesn't need to be passed fetch::vm::TypeIds::UInt256 here. The constructor should pass TypeIds::UInt256 to the Object constructor itself.

libs/vm-modules/src/math/bignumber.cpp
```diff
@@ -250,6 +258,116 @@ bool UInt256Wrapper::FromJSON(JSONVariant const &variant)
...
+  lhs->number_ *= rhs->number_;
+}
+
+void UInt256Wrapper::Divide(Ptr<Object> &lhso, Ptr<Object> &rhso)
+{
+  Ptr<UInt256Wrapper> lhs = lhso;
+  Ptr<UInt256Wrapper> rhs = rhso;
+  if (lhs->IsTemporary())
+  {
+    lhs->number_ /= rhs->number_;
```
robertdickson, line 352: This should check for division by zero, and do a RuntimeError if so. I guess the other divide and inplace divisions need this check too.

vhavrylov-ncube: The zero division check is implemented inside UInt<> underlying container, and an exception is thrown in case of zero division there. 
ledger/libs/vectorise/include/vectorise/uint/uint.hpp:651
This class is basically just a wrapper above a concrete UInt<256> container. So I think an additional check here could be a redundant one.

robertdickson: But RutimeError() has to be called here so that the VM can halt execution -- do you need to wrap the operation in try/catch perhaps?

vhavrylov-ncube: Ah I see, the vm_->RuntimeError("") must be called, not just an exception. Ok, I'll do the check here then of course.

libs/vm-modules/src/math/bignumber.cpp
vhavrylov-ncube: There is no constructor with the proper typeId in UInt256Wrapper, all constructors take it as a parameter. I guess it is better to pass here a type id instead of creation of a specialized constructor only for using it in 4 math operations inside this file.

libs/vm-modules/tests/unit/big_number_wrapper_tests.cpp
```diff
@@ -189,66 +189,211 @@ TEST_F(UInt256Tests, uint256_raw_increase)
...
-        bigger.increase();
+  ASSERT_TRUE(toolkit.Compile(TEXT));
+  EXPECT_TRUE(toolkit.Run());
+}
 
-        var gt : Bool = smaller > bigger;
-        ok = ok && !gt;
+TEST_F(UInt256Tests, uint256_assignment)
+{
+  // NOTE! The current behavior for assignment of UInt256 is a reference-assignment,
```
baykaner: if this is still true you need to add TODO(issue number) here

libs/vm-modules/tests/unit/big_number_wrapper_tests.cpp
```diff
@@ -306,46 +451,42 @@ TEST_F(UInt256Tests, uint256_logValue)
...
-          ok = ok && (test_uint64 == correct_uint64);
 
           var test_str : String = toString(test);
-          var correct_str : String =
-          "000084e2506ce67c000000000000000000000000000000000000000000000000";
-          ok = ok && (test_str == correct_str);
-          return ok;
+          var expected_str_in_big_endian_enc : String =
+                "0000000000000000000000000000000000000000000000007ce66c50e2840000";
+          // Assertion disabled until UInt256 buffer init rework
```
baykaner: please uncomment this code. If it can't be merged until another PR then this will have to wait for that. Commenting out code like this tends to lead to forgetting to un-disable tests!

libs/vm-modules/tests/unit/big_number_wrapper_tests.cpp
```diff
@@ -306,46 +451,42 @@ TEST_F(UInt256Tests, uint256_logValue)
   EXPECT_NEAR(result, expected, expected * LOGARITHM_TOLERANCE);
 }
 
+// Warning: the test contains 1 temporary disabled assertion.
```
baykaner: ditch the comment - handle the tests correctly

baykaner (changes requested):

libs/vm-modules/src/math/bignumber.cpp
```diff
@@ -71,6 +72,14 @@ void UInt256Wrapper::Bind(Module &module)
...
       .EnableOperator(Operator::LessThan)
+      .EnableOperator(Operator::Add)
+      .EnableOperator(Operator::Subtract)
+      .EnableOperator(Operator::InplaceAdd)
+      .EnableOperator(Operator::InplaceSubtract)
+      .EnableOperator(Operator::Multiply)
+      .EnableOperator(Operator::Divide)
+      .EnableOperator(Operator::InplaceMultiply)
+      .EnableOperator(Operator::InplaceDivide)
       //        .EnableOperator(Operator::LessThanOrEqual)
```
markos: better remove those commented operators altogether or add a comment that they are tackled in another PR (and reference to that PR).

libs/vm-modules/src/math/bignumber.cpp
```diff
@@ -71,6 +72,14 @@ void UInt256Wrapper::Bind(Module &module)
...
       .EnableOperator(Operator::GreaterThan)
       //        .EnableOperator(Operator::GreaterThanOrEqual)
```
markos: same

libs/vm-modules/src/math/bignumber.cpp
markos: I found that a constructor that takes Id is needed in order to work with the recent changes in StructuredData, that way GetObject/SetObject can call a generic constructor. I had to do that for Fixed128 also, so it's ok.

markos: Having said that, for within the class itself, there is no point to carry TypeId everywhere, the Constructor with TypeId as argument is only useful within StructureData serialization/deserialization. Within the class itself, you could just call a plain constructor.

libs/vm-modules/src/math/bignumber.cpp
```diff
@@ -250,6 +259,127 @@ bool UInt256Wrapper::FromJSON(JSONVariant const &variant)
...
+    return;
+  }
+  if (rhs->IsTemporary())
+  {
+    rhs->number_ -= lhs->number_;
+    lhso = std::move(rhs);
+    return;
+  }
+  Ptr<UInt256Wrapper> n(
+      new UInt256Wrapper(vm_, fetch::vm::TypeIds::UInt256, lhs->number_ - rhs->number_));
```
markos: same comment as above for typeid in the constructor.

libs/vm-modules/tests/unit/big_number_wrapper_tests.cpp
vhavrylov-ncube: There is no issue for this rework, afaik. @robertdickson told me that what I describe for UInt256 is an expected assignment behavior for Object value types, which are now not implemented.

libs/vm-modules/tests/unit/big_number_wrapper_tests.cpp
vhavrylov-ncube: moved to a separate disabled test

libs/vm-modules/tests/unit/big_number_wrapper_tests.cpp
vhavrylov-ncube: done

libs/vm-modules/tests/unit/big_number_wrapper_tests.cpp
vhavrylov-ncube: I have created https://fetchai.atlassian.net/browse/ML-258 for this rework, and for now just removed this comment.

libs/vm-modules/src/math/bignumber.cpp
vhavrylov-ncube: done

libs/vm-modules/src/math/bignumber.cpp
vhavrylov-ncube: done

libs/vm-modules/src/math/bignumber.cpp
vhavrylov-ncube: Refactored the constructor from underlying container.

libs/vm-modules/src/math/bignumber.cpp
vhavrylov-ncube: done

libs/vm-modules/src/math/bignumber.cpp
vhavrylov-ncube: VM uses a constructor with type_id as an argument
libs/vm/include/vm/vm.hpp:353:23: 
    return Ptr<T>{new T(this, GetTypeId<T>(), std::forward<Ts>(args)...)};
so I have just added another utility constructor just for such use-cases.

baykaner (approved):
LGTM
did you raise a task to come back and re-enable that one disabled test?

PR #2017, opened 2019-11-25
Remove POW code and refactor

n-hutton:
Remove POW code from the codebase, leaving behind a consensus implementation that emulates POW behaviour.

It does this by randomly choosing a time to wait on top of the last block time - the spread of times around the block interval by multiple miners create the forking effect. Block weights are also random.

This means that the block coordinator logic is much cleaner, it just asks consensus if it can generate a block for both the POW and POS cases.

The behaviour externally should be the same.


libs/network/include/network/p2pservice/p2p_http_interface.hpp
```diff
@@ -156,7 +156,7 @@ class P2PHttpInterface : public http::HTTPModule
       block["hash"]         = "0x" + b->hash.ToHex();
       block["previousHash"] = "0x" + b->previous_hash.ToHex();
       block["merkleHash"]   = "0x" + b->merkle_hash.ToHex();
-      block["proof"]        = "0x" + b->proof.header().ToHex();
+      block["proof"]        = "0x" + b->merkle_hash.ToHex();  // TODO(HUT): check non breaking here.
```
n-hutton: @ejfitzgerald is this going to break block explorer?

n-hutton: (if I remove this field that is)

n-hutton: I guess I could switch it to entropy hex instead

ejfitzgerald: Just remove it, it is a simple update to block explorer anyway. Entropy hash makes more sense anyway

libs/ledger/include/ledger/chain/block.hpp
```diff
@@ -107,64 +101,58 @@ struct MapSerializer<ledger::Block, D>
...
     map.ExpectKeyGetValue(MERKLE_HASH, block.merkle_hash);
     map.ExpectKeyGetValue(BLOCK_NUMBER, block.block_number);
-    map.ExpectKeyGetValue(MINER, block.miner);
     map.ExpectKeyGetValue(MINER_ID, block.miner_id);
     map.ExpectKeyGetValue(LOG2_NUM_LANES, block.log2_num_lanes);
     map.ExpectKeyGetValue(SLICES, block.slices);
     map.ExpectKeyGetValue(DAG_EPOCH, block.dag_epoch);
     map.ExpectKeyGetValue(TIMESTAMP, block.timestamp);
     map.ExpectKeyGetValue(ENTROPY, block.block_entropy);
+    map.ExpectKeyGetValue(MINER, block.miner);
```
n-hutton, line 153: So I would like to remove this field (clashes with miner_id), but since it seems to be used for fee collecting I will leave it in for now

libs/ledger/include/ledger/chain/block_coordinator.hpp
```diff
@@ -121,7 +118,7 @@ class BlockSinkInterface;
  *                  │─ ─ ─ ─ ─ ─              │─ ─ ─ ─ ─ ─ ┘         │                  │
  *                  ▼                         ▼                      │                  │
  *        ┌──────────────────┐      ┌──────────────────┐             │                  │
- *        │ Post Exec. Block │      │   Proof Search   │             │                  │
+ *        │ Post Exec. Block │      │                  │             │                  │
```
n-hutton, line 121: Need to update this sometime :)

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -549,21 +515,16 @@ BlockCoordinator::State BlockCoordinator::OnPreExecBlockValidation()
       return State::RESET;
     }
 
-    if (consensus_)
-    {
-      consensus_->UpdateCurrentBlock(*previous);  // Only update with valid blocks
-      auto result = consensus_->ValidBlock(*current_block_);
+    consensus_->UpdateCurrentBlock(*previous);  // Only update with valid blocks
+    auto result = consensus_->ValidBlock(*current_block_);
```
n-hutton, line 518: Once @denis-trapeznikov stuff is done, the consensus valididity checks will not be in the block coordinator (instead guarantee main cahin provides valid blocks)

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -250,7 +251,7 @@ void MainChainRpcService::HandleChainResponse(Address const &address, BlockList
     it->UpdateDigest();
 
     // add the block
-    if (it->proof())
+    // TODO(HUT): put consensus check here.
```
n-hutton, line 254: @denis-trapeznikov

libs/network/include/network/p2pservice/p2p_http_interface.hpp
n-hutton: OK, I will remove this completely

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -198,13 +192,15 @@ BlockCoordinator::BlockCoordinator(MainChain &chain, DAGPtr dag,
...
-  state_machine_->RegisterHandler(State::PROOF_SEARCH,                 this, &BlockCoordinator::OnProofSearch);
 
   state_machine_->RegisterHandler(State::TRANSMIT_BLOCK,               this, &BlockCoordinator::OnTransmitBlock);
   state_machine_->RegisterHandler(State::RESET,                        this, &BlockCoordinator::OnReset);
   // clang-format on
 
+  assert(consensus_);
+
   state_machine_->OnStateChange([this](State current, State previous) {
+    FETCH_UNUSED(this);
```
ejfitzgerald: Just remove from the capture? I am probably missing something?

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -1036,7 +990,7 @@ BlockCoordinator::State BlockCoordinator::OnWaitForNewBlockExecution()
     // update the current block with the desired hash
     next_block_->merkle_hash = storage_unit_.CurrentHash();
 
-    FETCH_LOG_DEBUG(LOGGING_NAME, "Merkle Hash: ", ToBase64(next_block_->merkle_hash));
+    FETCH_LOG_INFO(LOGGING_NAME, "Merkle Hash: ", ToBase64(next_block_->merkle_hash));
```
ejfitzgerald: Revert back to debug?

ejfitzgerald (changes requested):
Looks good. I would not have gone as overboard with the simulated PoW stuff but it looks good to me.

I just have a couple of niggles that I think that make sense before landing it but nice one

libs/ledger/src/chain/block_coordinator.cpp
n-hutton: You do need this for periodic_print. Reverted this.

libs/ledger/src/chain/block_coordinator.cpp
n-hutton: reverted, good catch 👍

ejfitzgerald (approved):

libs/ledger/src/chain/block_coordinator.cpp
ejfitzgerald: Yes of course - I am blind 🤦‍♂

PR #2018, opened 2019-11-25
Extend option to run tests selectively to slow, unit, integration

n-hutton:
This should now be possible:

```
./scripts/ci-tool.py -T Debug --test-names chain-unit-tests,ledger-chain-tests
```

### DIFF
--- a/scripts/ci-tool.py
+++ b/scripts/ci-tool.py
@@ -249,11 +249,9 @@ def test_project(build_root, include_regex=None, exclude_regex=None, test_names=
         '-T', TEST_NAME
     ]
 
-    # Need to convert test names from set to a regex of ORs
-    names_as_list = list(test_names)
-    names_as_regex = "|".join(str(x) for x in names_as_list)
-
     if test_names is not None:
+        # Need to convert test names from set to a regex of ORs
+        names_as_regex = "|".join(str(x) for x in test_names)
         cmd = cmd + ['-R', names_as_regex]
     if include_regex is not None:
         cmd = cmd + ['-L', str(include_regex)]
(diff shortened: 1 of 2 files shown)


scripts/ci-tool.py
```diff
@@ -249,6 +249,12 @@ def test_project(build_root, include_regex=None, exclude_regex=None):
         '-T', TEST_NAME
     ]
 
+    # Need to convert test names from set to a regex of ORs
+    names_as_list = list(test_names)
+    names_as_regex = "|".join(str(x) for x in names_as_list)
```
wkaluza: You don't need the list here, do you? You can iterate over `test_names` directly.

scripts/end_to_end_test/run_end_to_end_test.py
```diff
@@ -620,20 +620,23 @@ def run_test(build_directory, yaml_file, node_exe, name_filter=None):
             for test in all_yaml:
                 # Get test setup conditions
                 setup_conditions = yaml_extract(test, 'setup_conditions')
+                description = yaml_extract(test, 'test_description')
+
+                test_disabled = "DISABLED" in description
 
-                # Check if name is not filtered out
+                # Check if name is not filtered out (overrides disabled specifier)
```
wkaluza: I don't think this is a good idea. If someone is working on a test, they might as well re-enable it in the YAML. This creates a risk where someone will locally fix a disabled test and commit changes without removing the `DISABLED` tag. This mistake would not be obvious from the GitHub diff and thus would be easy to miss in review.

wkaluza (changes requested):
Good change, but I wouldn't alter the disablement logic in e2e tests.

scripts/ci-tool.py
n-hutton: you're right

scripts/end_to_end_test/run_end_to_end_test.py
n-hutton: I can see this happening and I guess it is not much effort to enable the test

wkaluza (approved):

PR #2020, opened 2019-11-25
Peer Selector Override

ejfitzgerald:


libs/muddle/src/direct_message_service.cpp
```diff
@@ -420,16 +425,10 @@ DirectMessageService::UpdateStatus DirectMessageService::UpdateReservation(Addre
...
-        bool const is_outgoing = conn->Type() == network::AbstractConnection::TYPE_OUTGOING;
-
         // we have a duplicate entry. This means that the two nodes are trying to connect to each
         // other. In this case define the rule that the connection with the lowest address continues
         // the connection
-        bool const should_replace =
-            ((address_ < address) && is_outgoing) || ((address_ > address) && !is_outgoing);
-
-        if (should_replace)
+        if (ShouldReplace(address, IsOutgoing(conn)))
```
denis-trapeznikov: `IsOutgoing()` could be called from inside `ShouldReplace()`, if there's no other possible combination.

libs/muddle/src/peer_selector.cpp
```diff
@@ -358,10 +359,20 @@ void PeerSelector::OnAnnouncement(Address const &from, byte_array::ConstByteArra
...
+  }
+  catch (std::exception const &ex)
+  {
+    FETCH_LOG_ERROR(logging_name_, "Unable to deserialise announcement packet: ", ex.what());
+  }
 
-  serializers::MsgPackSerializer serializer{payload};
-  serializer >> peers;
+  // convert to a set
+  std::unordered_set<network::Peer> const peers{peer_list.begin(), peer_list.end()};
```
denis-trapeznikov, line 375: `ArraySerializer` can deserialize `unordered_set`s.

libs/muddle/src/peer_selector.cpp
```diff
@@ -403,36 +414,71 @@ void PeerSelector::OnAnnouncement(Address const &from, byte_array::ConstByteArra
...
   }
 
-  // update the peers info cache
+  // remove all internal references to this objects address and peer
+  std::unordered_set<Address> removed_addresses{};
   {
-    auto it = peers_info_.find(from);
-    if (it != peers_info_.end())
+    auto it = peers_info_.begin();
+    while (it != peers_info_.end())
```
denis-trapeznikov: `for(it = peers_info_.begin(); it != peers_info_.end();)`?

libs/muddle/src/peer_selector.cpp
```diff
@@ -403,36 +414,71 @@ void PeerSelector::OnAnnouncement(Address const &from, byte_array::ConstByteArra
...
         {
-          metadata.peer_data.emplace_back(PeerMetadata{peer});
+          peer_it       = peer_data.erase(peer_it);
+          peers_removed = true;
+        }
+        else
+        {
+          ++peer_it;
         }
       }
```
denis-trapeznikov, line 438: `peer_data` is an `std::vector`, which means this code works in O(N²). `std::remove_if` would probably better fit in here.

libs/muddle/src/peer_selector.cpp
```diff
@@ -403,36 +414,71 @@ void PeerSelector::OnAnnouncement(Address const &from, byte_array::ConstByteArra
...
-    }
-    else  // new entry
-    {
-      Metadata &metadata = peers_info_[from];
 
-      for (auto const &peer : peers)
+      // remove the entry completely if we have removed all previous peers
+      if (peers_removed && peer_data.empty())
```
denis-trapeznikov, line 441: Shouldn't `peer_data.empty()` be sufficient here?

libs/muddle/src/peer_selector.cpp
```diff
@@ -403,36 +414,71 @@ void PeerSelector::OnAnnouncement(Address const &from, byte_array::ConstByteArra
...
+    // lookup existing entry
+    metadata = &(it->second);
+  }
+  else
+  {
+    // create a new entry
+    metadata = &peers_info_[from];
+  }
+
+  if (metadata != nullptr)
```
denis-trapeznikov, line 473: Can't be false.

denis-trapeznikov (changes requested):

libs/muddle/src/peer_selector.cpp
ejfitzgerald: true with the code as it is at the moment, this is just a belt and braces check for the future 😁

libs/muddle/src/peer_selector.cpp
ejfitzgerald: No. In the case of the unresolvable address `peer_data` would still be empty. In this change I only want to remove resolved peers who have been updated

libs/muddle/src/peer_selector.cpp
ejfitzgerald: You are correct this is a N^2 operation, however, the N in this case is expected to be relatively small. I did consider using `std::remove_if` here but I think this is clearer especially as I want to preserve thte `peers_removed` flag.

libs/muddle/src/peer_selector.cpp
ejfitzgerald: Sure - I will update to be consistent with the other loop

libs/muddle/src/peer_selector.cpp
ejfitzgerald: You are correct, the issue here is that the serialisation object is a `std::vector` here. While it is true that the current implementation of the serialisation for `std::vector` and `std::unordered_set` is the same (it is just a message pack array). There is a chance that this will not be the case in the future. This is why I elected to make this copy.

To be clear in all use cases at the moment this is a vector of size 1, so while the addition copy is not great, the overhead is marginal.

@troelsfr is planning a big update in this area, I suspect that this code will be completely replaced. It would be therefore my preference to keep this inefficiency for the time being.

libs/muddle/src/peer_selector.cpp
denis-trapeznikov: Well, technically peers_removed can be calculated as `bool peers_removed = std::remove_if() != peers_data.end();`.

libs/muddle/src/direct_message_service.cpp
ejfitzgerald: Yes that is true, but would require passing the connection object into `ShouldReplace`. I don't really see much advantage in either case to be honest.

denis-trapeznikov: Well, if `is_outgoing` was computed in `ShouldReplace()`, the call itself would be shorter, and with less parentheses; and if `IsOutgoing()` is not used anywhere, it could be eliminated at all.

ejfitzgerald: I am a little surprised at the level of discussion around this point. The changes were originally made  to aid in some additional changes in this file which proved unnecessary.

For the time being I have reverted all these changes back to the original

ejfitzgerald:
@denis-trapeznikov if you are happy with the changes so far, can you approve the PR

denis-trapeznikov (approved):

PR #2021, opened 2019-11-26
Dev/update store smart [Cole-50][Cole-43]

JuanBesa:
This PR add several capabilities to Update Store. 

- No duplicate updates are stored
- Support for multiple learners
- Criteria now allows returning a Nan which means that the update is not interesting right now

Some defaults are included so that the new version mimics the current update-store queue like functionality.


lrahmani (dismissed):
LGTM.

- The criteria management is promising
- Avoiding pushing and getting duplicate updates is useful, we just need to make sure what is considered 'unique' about an update (i.e., proper semantic of ComputeFingerprint member function)
- Timestamping is ok

lrahmani (approved):
satisfied static analyser => happy life

PR #2022, opened 2019-11-26
bugfixes to model workflow & serialisation

baykaner:
- fixes that prevent calls to serialise model without dataloader, optimiser, and model being set


libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -364,8 +374,34 @@ typename VMModel::ModelPtrType &VMModel::GetModel()
 
 bool VMModel::SerializeTo(serializers::MsgPackSerializer &buffer)
 {
+
+  // can't serialise uncompiled model
+  if (!compiled_)
+  {
+    throw std::runtime_error("cannot set state with uncompiled model");
+  }
```
vhavrylov-ncube, line 383: Could we use vm->RuntimeError(...) here instead of c++ exception? This error (and consequtive ones in this file) looks like a VM-related, not a core problem.

evsmithx (approved):

PR #2023, opened 2019-11-26
fix bug with contract function invocation

robertdickson:
This fixes a bug in contract invocation in the etch compiler.


wkaluza (approved):
Nice one, lgtm

PR #2024, opened 2019-11-26
Add metrics for model

evsmithx:
Model can now have metrics specified at compile time, and the result of these is reported when Evaluate is called.


libs/ml/include/ml/model/model.hpp
```diff
@@ -167,6 +166,46 @@ void Model<TensorType>::Compile(OptimiserType optimiser_type, ops::LossType loss
     }
   }
 
+  for (auto const &met : metrics)
```
baykaner, line 178: since this takes more than a second to figure out what's going on, please add a comment here pointing out that this bit of code bode adds all the necessary metric nodes to the graph and stores their names in 'metrics_' for later access

libs/ml/include/ml/model/model.hpp
```diff
@@ -259,11 +298,33 @@ void Model<TensorType>::Predict(TensorType &input, TensorType &output)
...
+  if (batch_size == 0)
+  {
+    batch_size = dataloader_ptr_->Size();
+  }
+  auto evaluate_pair = dataloader_ptr_->PrepareBatch(batch_size, is_done_set);
+
+  this->graph_ptr_->SetInput(label_, evaluate_pair.first);
+  this->graph_ptr_->SetInput(input_, evaluate_pair.second.at(0));
+
+  DataVectorType ret;
```
baykaner, line 318: a comment would make these subsequent lines less cryptic, i.e.:
"call evaluate on the graph and store the loss in ret", then
"push further metrics into ret - due to graph caching these subsequent evaluate calls are cheap"

libs/ml/include/ml/model/model.hpp
```diff
@@ -259,11 +298,33 @@ void Model<TensorType>::Predict(TensorType &input, TensorType &output)
...
+  ret.emplace_back(*(this->graph_ptr_->Evaluate(error_).cbegin()));
+
+  for (auto const &met : metrics_)
+  {
+    ret.emplace_back(*(this->graph_ptr_->Evaluate(met).cbegin()));
```
baykaner, line 325: have you confirmed (via debugging or benchmarking) that the caching is doing it's job as we are hoping here - i.e. that these evaluate calls are very cheap. if not, please step through in debugging at least and just check it works how we expect.

libs/ml/include/ml/model/model.hpp
```diff
@@ -320,8 +381,8 @@ void Model<TensorType>::TrainImplementation(DataType &loss, SizeType n_rounds)
 
   dataloader_ptr_->SetMode(dataloaders::DataLoaderMode::TRAIN);
 
-  loss_              = DataType{0};
-  DataType min_loss  = fetch::math::numeric_max<DataType>();
+  loss_ = DataType{0};
+  DataType min_loss;
```
baykaner: instead of declaring an uninstantiated value here, move the declaration to the first point of use a few lines later, i.e.:
DataType min_loss = loss_;

libs/ml/include/ml/serializers/ml_types.hpp
```diff
@@ -391,6 +391,19 @@ void SerializeAnyOp(MapType &map, uint8_t code, fetch::ml::OpType const &op_type
...
+  case ml::OpType::METRIC_CATEGORICAL_ACCURACY:
+  {
+    SerializeImplementation<TensorType, D, ml::OpCategoricalAccuracySaveableParams<TensorType>>(
+        map, code, op);
+    break;
+  }
+  case ml::OpType::GRAPH:
+  case ml::OpType::SUBGRAPH:
+  {
+    throw ml::exceptions::InvalidMode(
```
baykaner, line 403: nice catch 👍

libs/ml/include/ml/utilities/graph_builder.hpp
```diff
@@ -551,6 +552,21 @@ void BuildNodeAndInsertTrainables(NodeSaveableParams<T> const &nsp, std::string
     g->AddTrainable(node, name);
     break;
   }
+  case OpType::METRIC_CATEGORICAL_ACCURACY:
```
baykaner, line 555: Since you've implemented a new op you'll need to update the graph test which adds one of every op type as well

baykaner (changes requested):
isn't this missing a necessary change to vm_modules/ml/model as well?
I would have thought some changes to evaluate calls are necessary there also?

libs/ml/include/ml/model/model.hpp
```diff
@@ -554,7 +616,7 @@ struct MapSerializer<ml::model::Model<TensorType>, D>
   template <typename Constructor>
   static void Serialize(Constructor &map_constructor, Type const &sp)
   {
-    auto map = map_constructor(13);
+    auto map = map_constructor(14);
```
vhavrylov-ncube: Would you consider naming this constant as 
static constexpr auto DESIRED_MAP_SIZE = 14;
or similar? Just to make a hint what does this value mean. The exact name of the constant is absolutely up to you, of course.

libs/ml/tests/unit/layers/PRelu.cpp
```diff
@@ -97,9 +97,10 @@ TYPED_TEST(PReluTest, node_forward_test)  // Use the class as a Node
   placeholder_op_ptr->SetData(data);
 
   fetch::math::SizeType in_size = 50u;
```
vhavrylov-ncube: this assignment
in_size = 50u;
is used 2 times in this file, so it would be great to also have a named constant for the value, m.b. smth like
MIN_ELIGIBLE_INPUT_SIZE or DEFAULT_INPUT_SIZE, you know it better :)

libs/ml/include/ml/serializers/ml_types.hpp
evsmithx: Thank you!

baykaner (dismissed):

baykaner (approved):

libs/ml/include/ml/model/model.hpp
evsmithx: Yep, checked in the debugger and cache status is CACHE_VALID

PR #2028, opened 2019-11-27
avoid accidentally stuttering during crash recovery

n-hutton:


ejfitzgerald:
Seems good from my side but worth a run past @jinmannwong

libs/beacon/src/beacon_service.cpp
```diff
@@ -283,21 +301,19 @@ BeaconService::State BeaconService::OnCollectSignaturesState()
...
-  if (state_machine_->previous_state() == State::PREPARE_ENTROPY_GENERATION)
+  // Only populate our signature information when we want to share it with peers (when less
+  // than lead blocks from the HEAD). Entropy generation can however sync ahead of the chain.
+  if (signatures_being_built_.find(index) == signatures_being_built_.end() &&
+      index <= (most_recent_round_seen_ + entropy_lead_blocks_))
   {
     SignatureInformation this_round;
     this_round.round                                        = index;
     this_round.threshold_signatures[identity_.identifier()] = active_exe_unit_->member_share;
     signatures_being_built_[index]                          = this_round;
```
n-hutton, line 315: Note: todo is done, signatures are trimmed now. Also, there can be no infinite loop due to the OutOfSync function

jinmannwong (dismissed):
Looks good to me.

PR #2019, opened 2019-11-25
 ML use an Estimator Object member functions when binding member functions [COLE-74]

lrahmani:

PR #2029, opened 2019-11-27
Feature/ml add estimator for vmmodel

lrahmani:


wkaluza (changes requested):
See my comments about changes related to VMTensor in #2015 and #2019.

PR #2030, opened 2019-11-27
guard against loose blocks when verifying

n-hutton:
When verifying blocks, it was assumed that there was a continuous chain preceeding the block to be verified. This is not neccessarily true, as:

A) The block coordinator will remove blocks sometimes
B) You might try to verify a loose block

This PR guards against that case. This will mean that the block validation on the blocks coming in externally (main_chain_rpc_something) will have to try...catch.


libs/ledger/src/consensus/consensus.cpp
```diff
@@ -161,7 +162,16 @@ Block GetBeginningOfAeon(Block const &current, MainChain const &chain)
...
   while (!ret.block_entropy.IsAeonBeginning() && current.block_number != 0)
   {
-    ret = GetBlockPriorTo(ret, chain);
+    auto prior = GetBlockPriorTo(ret, chain);
+
+    if (!prior)
+    {
+      FETCH_LOG_ERROR(LOGGING_NAME,
+                      "Failed to find the beginning of the aeon when traversing the chain!");
+      throw std::runtime_error("Failed to traverse main chain");
```
ejfitzgerald, line 171: Are these handled correctly at the block coordinator level

n-hutton: the BC catches all state exceptions, so it will not std::terminate if this were to throw. However, I believe it can only throw if you either try to update or verify a loose block from the BC, or remove a block from the chain elsewhere, which does not happen. So, I am  sure it will not throw this in the BC.

ejfitzgerald: Okay

ejfitzgerald (approved):

PR #2032, opened 2019-11-27
Single line comment fix

robertdickson:
Fixes a silly bug where // comments on the very last line were causing a parsing error


cyenyxe:
Can we have a test for this, to guarantee no regressions to the previous behavior?

wkaluza (approved):

baykaner (approved):
seems to work 👍

cyenyxe:
Would it be possible to have a test to prevent any future regressions?

PR #2034, opened 2019-11-28
fix timing in pow test

n-hutton:
I noticed that the simulated pow test failed in master here : https://jenkins.economicagents.com/blue/organizations/jenkins/ledger/detail/master/308/pipeline - this should hopefully resolve that.

To test this theory I will push a few empty commits to this PR to check the flakiness of the tests.

### DIFF
--- a/libs/ledger/src/consensus/simulated_pow_consensus.cpp
+++ b/libs/ledger/src/consensus/simulated_pow_consensus.cpp
@@ -124,7 +124,7 @@ NextBlockPtr SimulatedPowConsensus::GenerateNextBlock()
                   " deadline: ", decided_next_timestamp_ms_);
 
   // Round up to the nearest second for block timestamp
-  auto current_time_s_ceiled =
+  auto const current_time_s_ceiled =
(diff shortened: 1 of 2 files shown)


libs/ledger/src/consensus/simulated_pow_consensus.cpp
```diff
@@ -123,6 +123,10 @@ NextBlockPtr SimulatedPowConsensus::GenerateNextBlock()
   FETCH_LOG_DEBUG(LOGGING_NAME, "Generating block. Current time: ", current_time_ms,
                   " deadline: ", decided_next_timestamp_ms_);
 
+  // Round up to the nearest second for block timestamp
+  auto current_time_s_ceiled =
```
vhavrylov-ncube: Looks like it is not changed below, so m.b. "auto const" ?

n-hutton: m.b = maybe?
Sure, I have made the change now 👍

baykaner (approved):

qati (approved):

PR #2035, opened 2019-11-28
bugfix disabled test fixtures

baykaner:


n-hutton (approved):
Note to anyone reading this: it appears if you do not disable the fixture name, when running ctest locally it will run this test anyway, causing a failure.

evsmithx (approved):

PR #2038, opened 2019-11-28
bugfixed + character handling by tensor

baykaner:


evsmithx (approved):

libs/math/include/math/tensor.hpp
```diff
@@ -703,6 +703,7 @@ Tensor<T, C> Tensor<T, C>::FromString(byte_array::ConstByteArray const &c)
       }
       ++i;
       break;
+    case '+':
```
vhavrylov-ncube, line 706: Is there any unit test behind this? :)

PR #2041, opened 2019-11-29
VMModel error handling improved and tests added

vhavrylov-ncube:


libs/vm-modules/include/vm_modules/ml/model/model.hpp
```diff
@@ -118,6 +118,8 @@ class VMModel : public fetch::vm::Object
   ModelCategory      model_category_ = ModelCategory::NONE;
   bool               compiled_       = false;
 
+  static constexpr size_t min_allowed_hidden_layers_ = 2;
```
baykaner: this value should be 1; and you should fix the code that doesn't correctly handle that case

baykaner: also use std::size_t (code style)

libs/vm-modules/include/vm_modules/ml/model/model.hpp
```diff
@@ -101,7 +101,7 @@ class VMModel : public fetch::vm::Object
 
   static void Bind(fetch::vm::Module &module);
 
-  ModelPtrType &GetModel();
+  void SetUnderlyingModelInstance(ModelPtrType const &instance);
```
baykaner: just SetModel is fine:

also why do you remove the getter?

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -161,19 +157,17 @@ void VMModel::CompileSequential(fetch::vm::Ptr<fetch::vm::String> const &loss,
 void VMModel::CompileSimple(fetch::vm::Ptr<fetch::vm::String> const &        optimiser,
                             fetch::vm::Ptr<vm::Array<math::SizeType>> const &in_layers)
 {
-  OptimiserType const optimiser_type = ParseName(optimiser->string(), optimisers_, "optimiser");
-  if (optimiser_type != OptimiserType::ADAM)
+  size_t const total_hidden_layers = in_layers->elements.size();
```
baykaner: std::size_t

baykaner: also calling elements.size() is very cheap - you don't need to store it to a variable. but it's fine if you think it's clearer to read that way

libs/vm-modules/tests/unit/ml_model_tests.cpp
```diff
@@ -347,6 +347,140 @@ TEST_F(VMModelTests, model_init_with_wrong_name)
   EXPECT_THROW(toolkit.Run(), std::runtime_error);
 }
 
+TEST_F(VMModelTests, model_add_invalid_layers)
```
baykaner: this test has nice coverage, but it would be preferable to wrap up the re-usable code in the test fixture, and then have multiple small tests; that way when one fails you immediately know which invalid case is at fault

baykaner (changes requested):

libs/vm-modules/include/vm_modules/ml/model/model.hpp
vhavrylov-ncube: The only usage of this getter was in this file, to make assignments to its return value. This approach

a.getSomething() = b;

discards constness of the getter, and I guess it would be better to have explicitly non-constant setter.

libs/vm-modules/include/vm_modules/ml/model/model.hpp
vhavrylov-ncube: I would keep it 2 for now as it protects user from a runtime crash. Indeed we need to fix Graph code, but right now I can not estimate if that fix is small or huge (and what risk does it add), so I guess it might be better to create a separate task for this investigation, and set VMModel::min_allowed_hidden_layers_ to 1 in that scope.

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: Yeap, it's more about naming than cacheing :)

libs/vm-modules/tests/unit/ml_model_tests.cpp
vhavrylov-ncube: Agreed

libs/vm-modules/include/vm_modules/ml/model/model.hpp
vhavrylov-ncube: I leave a comment above this constant to explain the current value

vhavrylov-ncube: After clarification with @MissingNO57 : no changes are required to existing code, it works fine, just a naming confusion. I have renamed necessary constants and variables.

libs/ml/include/ml/model/sequential.hpp
```diff
@@ -45,12 +45,14 @@ class Sequential : public Model<TensorType>
...
 
+  SizeType LayerCount() const;
+
   template <typename X, typename D>
   friend struct serializers::MapSerializer;
 
 private:
-  SizeType    layer_count_ = 0;
   std::string prev_layer_;
+  SizeType    layer_count_ = 0;
```
baykaner: if you add any new member variables to an object, this will need to be reflected in the object serialisation and deserialistion. Please speak to me, Jiri, or Emma if you're not sure exactly how to do that.

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -137,72 +135,96 @@ Ptr<VMModel> VMModel::Constructor(VM *vm, TypeId type_id,
...
   }
 
-  size_t const n_elements = in_layers->elements.size();
-
-  std::vector<math::SizeType> layers;
-  layers.reserve(n_elements);
-  for (size_t i = 0; i < n_elements; ++i)
+  std::vector<math::SizeType> shapes;
+  shapes.reserve(total_layer_shapes);
+  for (size_t i = 0; i < total_layer_shapes; ++i)
```
baykaner: std::size_t (style)

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -395,29 +417,34 @@ void VMModel::AssertLayerTypeMatches(SupportedLayerType                layer,
   };
   if (std::find(valids.begin(), valids.end(), layer) == valids.end())
   {
-    std::string const message{"Invalid params specified for \"" + LAYER_NAMES_.at(layer) +
-                              "\" layer."};
-    vm_->RuntimeError(message);
-    throw std::runtime_error(message);
+    throw std::runtime_error("Invalid params specified for \"" + LAYER_NAMES_.at(layer) +
```
baykaner, line 420: I think this was better before. The reason is that if we try to run a smart contract that uses this model function and fails here, the vm->RuntimeError(message) should route that useful message to std::out for the developer in a sensible way - whereas merely throwing will not give this information to the developer.

baykaner: if you like we can discuss this next week and I can show you how to run a simple smart contract and try it out for yourself. That might make this clearer.

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -395,29 +417,34 @@ void VMModel::AssertLayerTypeMatches(SupportedLayerType                layer,
...
 }
 
 VMModel::SequentialModelPtr VMModel::GetMeAsSequentialIfPossible()
 {
   if (model_category_ != ModelCategory::SEQUENTIAL)
   {
-    std::string const message{"No \"add\" method exists for non-sequential models!"};
-    vm_->RuntimeError(message);
-    throw std::runtime_error(message);
+    throw std::runtime_error("Layer adding is allowed only for sequential models!");
```
baykaner, line 429: same as above

libs/vm-modules/tests/unit/ml_model_tests.cpp
```diff
@@ -32,11 +32,52 @@ namespace {
...
+
 class VMModelTests : public ::testing::Test
 {
 public:
   std::stringstream stdout;
   VmTestToolkit     toolkit{&stdout};
+
+  void TestInvalidLayerCompilation(std::vector<std::string> sources)
+  {
+    for (const auto &test_case : sources)
```
baykaner: still you could ditch this for-loop, take in just a single source, and separate all the tests for the most clarity

libs/vm-modules/tests/unit/ml_model_tests.cpp
```diff
@@ -32,11 +32,52 @@ namespace {
...
+    {
+      std::cout << "Testing invalid layer: " << test_case << std::endl;
```
baykaner: then you wouldn't need to print this too

baykaner (changes requested):
tests look much cleaner, but still I would separate them as much as possible. Also you must implement serialisation/deserialisation for the layer count.

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: Runtime error will be called in a caller function, the top level .AddLayer<> contains a try-catch block that now covers all possible usages of ParseName, Assert.. and AddLayerSpecificImpl, so any exception will be caught and turned into VM-runtime error. So we still have VM logs with this error if occurs.

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: oh sorry for that :)

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: I have replaced RuntimeErrors with exceptions wherever an immediate stop of the machine does not happen, e.g. where the control flow does not return to VM handlers. And vice-versa, in AddLayer and other etch-object-member-functions I replaced c++ exceptions with RuntimeErrors - because after return from this method control flow goes to VM hadler, the machine can be safely stopped and error log appears. Mixing both things in one code block did not provide expected result for me.

libs/vm-modules/src/ml/model/model.cpp
vhavrylov-ncube: This function is also always covered by catch blocks, either AddLayer<> or Compile... one, so the error message will be shown neatly.

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -395,29 +417,34 @@ void VMModel::AssertLayerTypeMatches(SupportedLayerType                layer,
...
 template <typename... LayerArgs>
 void VMModel::AddLayer(fetch::vm::Ptr<fetch::vm::String> const &layer, LayerArgs... args)
 {
-  auto const layer_type = ParseName(layer->string(), layer_types_, "layer type");
-  AddLayerSpecificImpl(layer_type, args...);
+  try
+  {
+    SupportedLayerType const layer_type = ParseName(layer->string(), layer_types_, "layer type");
+    AddLayerSpecificImpl(layer_type, args...);
+    compiled_ = false;
```
vhavrylov-ncube, line 441: This try block covers all Assertions and ParseName's, that could be invoked in AddLayerSpecificImpl(), so if assignment [compiled_ = false] takes place, it is guaranteed that a layer was indeed added.

libs/vm-modules/tests/unit/ml_model_tests.cpp
vhavrylov-ncube: Ok, I'll do it

libs/vm-modules/tests/unit/ml_model_tests.cpp
vhavrylov-ncube: done

libs/ml/include/ml/model/sequential.hpp
vhavrylov-ncube: Ok, but this variable actually already exists in
https://github.com/fetchai/ledger/blob/master/libs/ml/include/ml/model/sequential.hpp
and there are several lines of code using it.
I don't know why Github highlighted this line as a change, becase I've added a getter method, not a member.

vhavrylov-ncube: Serialisation/deserialisation of this member is already implemented in sequential.hpp, so no action required.

baykaner (approved):

PR #2042, opened 2019-11-29
Memleak fix in Nodes by using weak_ptr instead of shared_ptr

MissingNO57:


baykaner (dismissed):
very nice catch

baykaner (dismissed):

wkaluza (changes requested):
I don't know what memory leak you are fixing, but you are misusing the `weak_ptr` API in doing so. A call to `weak_ptr::lock()` can return a default constructed (i.e. empty) `shared_ptr` if the pointee had already been destroyed. Therefore code such as

```
auto ptr         = i.lock();
auto ret_err_sig = ptr->BackPropagate(*bp_it);
```

risks dereferencing a null pointer.

Please consult docs and fix.
https://en.cppreference.com/w/cpp/memory/weak_ptr/lock

wkaluza (approved):

baykaner (approved):

PR #2053, opened 2019-12-02
Deprecate create wealth in ledger

n-hutton:
This PR removes the ability for transactions to arbitrarily create wealth. Nodes must be started with a genesis file that specifies wealth at one or more addresses. For the end to end tests, this is a 'benefactor account', which is transferred from, where create wealth was previously used.


libs/ledger/src/executor.cpp
```diff
@@ -365,11 +359,7 @@ bool Executor::ExecuteTransactionContract(Result &result)
                       scaled_charge, " (base: ", base_charge, " storage: ", storage_charge,
                       " compute: ", compute_charge, " shards: ", allowed_shards_.PopCount(), ")");
 
-      // create wealth transactions are always free
-      if (!IsCreateWealth(*current_tx_))
-      {
-        result.charge += scaled_charge;
-      }
+      result.charge += scaled_charge;
```
wkaluza, line 362: Shouldn't this be capped at the charge limit?

wkaluza (changes requested):
Fix typo in the PR title and leftover comments.

libs/ledger/src/executor.cpp
n-hutton: I think that in this instance the result.charge is what it has cost, so it can and will exceed the charge limit (the amount the TX is willing to pay), in which case your TX fails. Second opinion @ejfitzgerald, since I don't really want to get this wrong.

wkaluza: My thinking was that the information of insufficient charge is already conveyed by `result.status = Status::INSUFFICIENT_CHARGE;` below. Obviously you'd need to rearrange this function a bit, but overall setting `result.charge` above the charge limit runs a needless risk of overcharging someone accidentally IMO.

PR #2055, opened 2019-12-02
Dev/colearn uri [COLE-63]

JuanBesa:
Introduces the ColearnURI

This add the class and adds it to the interface of UpdateStore. 
The goal of the URI is to allow easy logging of messages related to Updates as will as eventually simplify the interface for pushing/getting updates.


engineerkatie (approved):
lgtm

PR #2056, opened 2019-12-02
Test recent comment-related changes

wkaluza:


n-hutton (approved):
Looks great. I would generally edit the description though for anyone that missed standup or was daydreaming

PR #2060, opened 2019-12-03
Remove Identifier and contract digest

qati:
This PR removed the Identifier construct which ties chain code addresses to contract addresses. It also changes storage keys for contracts from hash to contract address and removes the need for the digest in calls.


libs/ledger/benchmark/state_sentinel_bench.cpp
```diff
@@ -40,7 +38,7 @@ void StateSentinelAdapter_BasicBenchmark(benchmark::State &state)
   BitVector shards{1};
   shards.SetAllOne();
 
-  StateSentinelAdapter adapter{storage, Identifier{"foo.bar"}, shards};
+  StateSentinelAdapter adapter{storage, {"foo.bar"}, shards};
```
ejfitzgerald: `ConstByteArray` has implicit constructor I would therefore drop the braces so that it is at least consistent with other parts of the code base.

```suggestion
  StateSentinelAdapter adapter{storage, "foo.bar", shards};
```

libs/ledger/src/chaincode/chain_code_cache.cpp
```diff
@@ -30,30 +30,66 @@
...
+                                                   StorageInterface &    storage)
 {
   // attempt to locate the contract in the cache
   ContractPtr contract = FindInCache(contract_id);
 
   // if this fails create the contract
   if (!contract)
   {
-    if (contract_id.type() == Identifier::Type::SMART_OR_SYNERGETIC_CONTRACT)
+    if (IsIdentity(contract_id))
```
ejfitzgerald: I would recommend using the `chain::Address::Parse` function here. This will do a better job with the parse and will also validate the addresses built in checksum.

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -120,11 +121,12 @@ Contract::Result SmartContractManager::OnCreate(chain::Transaction const &tx)
   nonce = FromBase64(nonce);
   chain::Address const payable_address{crypto::Hash<crypto::SHA256>(tx.from().address() + nonce)};
```
ejfitzgerald: I know that it is not part of this PR but can we rename this to contract address because I think that this will make this code simpler to reason about

libs/ledger/src/chaincode/smart_contract_wrapper.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include "ledger/chaincode/smart_contract_wrapper.hpp"
+#include "moment/clocks.hpp"
+
+namespace fetch {
+namespace ledger {
+
+SmartContractWrapper::SmartContractWrapper(ConstByteArray source)
+  : source{std::move(source)}
+  , creation_timestamp{
```
ejfitzgerald: We can't have a timestamp in something that will change state. Can we make this a block timestamp please

libs/ledger/src/resource_mapper.cpp
```diff
@@ -27,13 +26,38 @@
...
+}  // namespace
+
 uint32_t MapResourceToLane(byte_array::ConstByteArray const &resource,
                            byte_array::ConstByteArray const &contract, uint32_t log2_num_lanes)
 {
-  ledger::Identifier identifier(contract);
-
   return storage::ResourceAddress{
-      byte_array::ByteArray{}.Append(identifier.name_space(), ".state.", resource)}
+      byte_array::ByteArray{}.Append(GetNameSpace(contract), ".state.", resource)}
```
ejfitzgerald: How come this isn't just `contract`?

libs/ledger/src/state_adapter.cpp
```diff
@@ -120,7 +121,7 @@ StateAdapter::Status StateAdapter::Write(std::string const &key, void const *dat
  * Checks to see if the specified key exists in the database
  *
  * @param key The key to be checked
- * @return OK if the key exists, PERMISSION_DENIED if the key is incorrect, ERROR is the key does
+ * @return OK if the key exists, PERMISSION_DENIED if the key is incorrect, ERROR is thSe key does
```
ejfitzgerald: Typo?

ejfitzgerald (changes requested):

libs/ledger/src/chaincode/smart_contract_wrapper.cpp
qati: done

libs/chain/src/address.cpp
```diff
@@ -60,24 +60,31 @@ bool Address::Parse(ConstByteArray const &input, Address &output)
...
+      {
+        output.address_ = address;
+        output.display_ = input;
+        success         = true;
+      }
     }
   }
+  catch (...)
+  {
+    success = false;
```
wkaluza, line 87: log out an error here

libs/ledger/include/ledger/dag/dag_node.hpp
```diff
@@ -175,13 +173,12 @@ struct MapSerializer<ledger::DAGNode, D>
...
-  static uint8_t const HASH                    = 7;
-  static uint8_t const SIGNATURE               = 8;
-  static uint8_t const OLDEST_EPOCH_REFERENCED = 9;
-  static uint8_t const WEIGHT                  = 10;
+  static uint8_t const CONTRACT_ADDRESS        = 4;
+  static uint8_t const IDENTITY                = 5;
+  static uint8_t const HASH                    = 6;
+  static uint8_t const SIGNATURE               = 7;
+  static uint8_t const OLDEST_EPOCH_REFERENCED = 8;
+  static uint8_t const WEIGHT                  = 9;
```
wkaluza, line 181: There's a call to `map_constructor` below which should now receive 9 instead of 10.

libs/ledger/src/chaincode/chain_code_cache.cpp
```diff
@@ -30,30 +30,29 @@
...
   // if this fails create the contract
   if (!contract)
   {
-    if (contract_id.type() == Identifier::Type::SMART_OR_SYNERGETIC_CONTRACT)
+    chain::Address address;
+    if (chain::Address::Parse(contract_id, address))
     {
-      auto const contract_digest = contract_id.qualifier().FromHex();
-      contract                   = CreateSmartContract<SmartContract>(contract_digest, storage);
+      contract = CreateSmartContract<SmartContract>(address.display(), storage);
```
wkaluza: pass address.address() instead of display, which returns a base58 string. Let's keep things as raw bytes as much as possible.

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -118,13 +119,14 @@ Contract::Result SmartContractManager::OnCreate(chain::Transaction const &tx)
   }
 
   nonce = FromBase64(nonce);
-  chain::Address const payable_address{crypto::Hash<crypto::SHA256>(tx.from().address() + nonce)};
+  chain::Address const contract_address{crypto::Hash<crypto::SHA256>(tx.from().address() + nonce)};
 
-  Identifier scope;
-  if (!scope.Parse(calculated_hash + "." + payable_address.display()))
+  SmartContractWrapper contract;
+  if (GetStateRecord(contract, contract_address.display()))
```
wkaluza: leave address unencoded here

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -169,7 +171,7 @@ Contract::Result SmartContractManager::OnCreate(chain::Transaction const &tx)
   Result init_status;
   if (!on_init_function.empty())
   {
-    state().PushContext(scope.full_name());
+    state().PushContext(contract_address.display());
```
wkaluza, line 174: do not encode

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -184,8 +186,8 @@ Contract::Result SmartContractManager::OnCreate(chain::Transaction const &tx)
       return init_status;
     }
   }
-
-  auto const status = SetStateRecord(contract_source, calculated_hash);
+  auto const status = SetStateRecord(SmartContractWrapper{contract_source, init_status.block_index},
+                                     contract_address.display());
```
wkaluza: do not encode

libs/ledger/src/chaincode/smart_contract_manager.cpp
```diff
@@ -206,7 +208,7 @@ Contract::Result SmartContractManager::OnCreate(chain::Transaction const &tx)
 storage::ResourceAddress SmartContractManager::CreateAddressForContract(Digest const &digest)
 {
   // create the resource address in the form fetch.contract.state.<digest of contract>
-  return StateAdapter::CreateAddress(Identifier{NAME}, digest.ToHex());
+  return StateAdapter::CreateAddress(NAME, digest);
```
wkaluza: we no longer pass in a digest to this function, so let's rename the argument

libs/ledger/src/executor.cpp
```diff
@@ -49,36 +49,26 @@ namespace fetch {
...
   using ContractMode = chain::Transaction::ContractMode;
 
-  ConstByteArray contract_name{};
   switch (tx.contract_mode())
   {
   case ContractMode::NOT_PRESENT:
     break;
   case ContractMode::PRESENT:
-    contract_name = tx.contract_digest().address().ToHex() + "." + tx.contract_address().display();
+    identifier = tx.contract_address().display();
```
wkaluza:
```suggestion
    identifier = tx.contract_address().address();
```

libs/ledger/src/executor.cpp
```diff
@@ -49,36 +49,26 @@ namespace fetch {
 namespace ledger {
 namespace {
 
-bool GenerateContractName(chain::Transaction const &tx, Identifier &identifier)
+bool GenerateContractName(chain::Transaction const &tx, ConstByteArray &identifier)
```
wkaluza: This function can no longer return false, so arguably it's no longer pulling its weight - try to get rid of it?

libs/ledger/src/upow/synergetic_execution_manager.cpp
```diff
@@ -104,7 +104,7 @@ ExecStatus SynergeticExecutionManager::PrepareWorkQueue(Block const &current, Bl
 {
   telemetry::FunctionTimer const timer{*prepare_queue_duration_};
 
-  using WorkMap = std::unordered_map<ProblemId, WorkItemPtr>;
+  using WorkMap = std::unordered_map<chain::Address, WorkItemPtr>;
```
wkaluza, line 106: `ProblemId` can be deleted now, I think

libs/ledger/src/upow/synergetic_executor.cpp
```diff
@@ -75,10 +75,16 @@ void SynergeticExecutor::Verify(WorkQueue &solutions, ProblemData const &problem
     // in the case of the first iteration we need to create the contract and define the problem
     if (!contract)
     {
-      auto const &digest = solution->contract_digest();
+      auto const &address = solution->address().display();
```
wkaluza: premature encoding

libs/ledger/src/state_adapter.cpp
```diff
@@ -143,11 +144,11 @@ StateAdapter::Status StateAdapter::Exists(std::string const &key)
...
  * @return The generated resource address for this key
  */
-ResourceAddress StateAdapter::CreateAddress(Identifier const &scope, ConstByteArray const &key)
+ResourceAddress StateAdapter::CreateAddress(ConstByteArray const &scope, ConstByteArray const &key)
 {
-  FETCH_LOG_DEBUG("StateAdapter", "Creating address for key: ", key, " scope: ", scope.full_name());
+  FETCH_LOG_DEBUG("StateAdapter", "Creating address for key: ", key, " scope: ", scope);
 
-  return ResourceAddress{scope.full_name() + ".state." + key};
+  return ResourceAddress{scope + ".state." + key};
```
wkaluza, line 151: This is a good place to encode, I think. I'm not sure if scope can be equal to fetch.token here, so you may need to check for that

wkaluza (changes requested):

libs/ledger/src/state_adapter.cpp
qati: scope can be token as well. So I don't think this is a good place for encoding.

libs/ledger/src/chaincode/smart_contract_manager.cpp
wkaluza: Scratch that, encoding is necessary here for now

libs/ledger/src/chaincode/smart_contract_manager.cpp
wkaluza: As discussed, change Get- and SetStateRecord to accept Address argument, call display() inside

libs/ledger/src/chaincode/chain_code_cache.cpp
wkaluza: As discussed, pass in Address, call .display() inside CreateSmartContract

libs/ledger/src/chaincode/smart_contract_manager.cpp
wkaluza: see above

wkaluza (approved):

libs/ledger/include/ledger/chaincode/contract.hpp
```diff
@@ -138,9 +140,9 @@ class Contract : public Chargeable
   bool ParseAsJson(chain::Transaction const &tx, variant::Variant &output);
 
   template <typename T>
-  bool GetStateRecord(T &record, ConstByteArray const &key);
+  bool GetStateRecord(T &record, chain::Address const &address);
```
ejfitzgerald, line 143: I was not expecting this change why is the key always an address?

libs/ledger/include/ledger/dag/dag_node.hpp
```diff
@@ -183,7 +183,7 @@ struct MapSerializer<ledger::DAGNode, D>
   template <typename Constructor>
   static void Serialize(Constructor &map_constructor, Type const &node)
   {
-    auto map = map_constructor(10);
+    auto map = map_constructor(9);
```
ejfitzgerald, line 186: 👍 👍 👍 👍 👍

libs/ledger/src/chaincode/smart_contract_wrapper.cpp
```diff
@@ -17,15 +17,13 @@
...
 
 namespace fetch {
 namespace ledger {
 
-SmartContractWrapper::SmartContractWrapper(ConstByteArray source)
+SmartContractWrapper::SmartContractWrapper(ConstByteArray source, uint64_t timestamp)
   : source{std::move(source)}
-  , creation_timestamp{
-        GetTime(fetch::moment::GetClock("default", fetch::moment::ClockType::SYSTEM))}
+  , creation_timestamp{timestamp}
```
ejfitzgerald, line 26: 👍

scripts/end_to_end_test/end_to_end_test.yaml
```diff
@@ -133,7 +133,7 @@ steps:
 
 ---
 test_description: >
-  Test synergetic mining
+  DISABLED - Test synergetic mining
```
ejfitzgerald: Is this meant to be disabled at this time?

ejfitzgerald (changes requested):

ejfitzgerald (approved):

scripts/end_to_end_test/end_to_end_test.yaml
wkaluza: It's disabled on master

libs/ledger/include/ledger/chaincode/contract.hpp
wkaluza: We agreed that we'll leave it as is for now and introduce a generic overload as a follow up

PR #2063, opened 2019-12-03
Deployment Updates

ejfitzgerald:
Fixes from the deployment 

- Update promises interface so that it is less throwy
- Add cli tools for chain and tx syncing support
- Correct issue with DKG which allows duplicate aeon starts
- Correct main chain sync issue preventing full chain sync issues


apps/storage-tools/blk_ctl.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "block_storage_tool.hpp"
+
+#include "logging/logging.hpp"
+
+#include <chrono>
+#include <cstdlib>
+#include <fstream>
+#include <thread>
+
+namespace {
```
n-hutton: I understand that namespacing in cpp files like this is no longer relevant in modern cpp - is that right?

apps/storage-tools/blk_ctl.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+using namespace std::chrono_literals;
+
+static constexpr char const *LOGGING_NAME = "BlkCtl";
+
+}  // namespace
+
+int main()
+{
+  fetch::crypto::mcl::details::MCLInitialiser();
```
n-hutton: It makes me sad every time I see this in our codebase :'(

n-hutton: (ofc it is needed here)

apps/storage-tools/block_storage_client.cpp
```diff
@@ -0,0 +1,36 @@
...
+using fetch::ledger::MainChainProtocol;
+
+BlockStorageClient::BlockStorageClient(MuddleAddress peer, MuddleEndpoint &endpoint)
+  : peer_{std::move(peer)}
+  , client_{"BlockClient", endpoint, fetch::SERVICE_MAIN_CHAIN, fetch::CHANNEL_RPC}
+{}
+
+bool BlockStorageClient::DownloadCompleteChain(Blocks &blocks)
+{
+  auto promise = client_.CallSpecificAddress(peer_, fetch::RPC_MAIN_CHAIN,
```
n-hutton: Naturally @denis-trapeznikov will break this with his forward changes

apps/storage-tools/block_storage_tool.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+  // wait until we have established the connection to the remote peer
+  if (!WaitForPeerConnections(*net_, 1))
+  {
+    throw std::runtime_error{"Unable to connect to peers requested"};
+  }
+
+  reactor_.Start();
+
+  //  // create the storage client
```
n-hutton: dead code

libs/beacon/include/beacon/aeon.hpp
```diff
@@ -47,6 +47,13 @@ struct Aeon
 
   // Timeouts for waiting for other members
   uint64_t start_reference_timepoint{uint64_t(-1)};
+
+  bool operator==(Aeon const &other) const
```
n-hutton, line 51: 👍

libs/beacon/include/beacon/block_entropy.hpp
```diff
@@ -75,6 +75,16 @@ struct BlockEntropy : public BlockEntropyInterface
   uint64_t EntropyAsU64() const override;
   void     HashSelf();
   bool     IsAeonBeginning() const;
+
+  bool operator==(BlockEntropy const &other) const
+  {
+    return ((qualified == other.qualified) &&
```
n-hutton: @jinmannwong there are no gotchas here are there?

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -341,8 +341,13 @@ MainChainRpcService::State MainChainRpcService::OnWaitForHeaviestChain()
     {
       if (PromiseState::SUCCESS == status)
       {
-        // the request was successful, simply hand off the blocks to be added to the chain
-        HandleChainResponse(current_peer_address_, current_request_->As<BlockList>());
+        BlockList block_list{};
```
n-hutton: This will probably merge conflict with @denis-trapeznikov  stuff

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -455,19 +465,8 @@ MainChainRpcService::State MainChainRpcService::OnSynchronised(State current, St
...
-  // Assume if the chain is quite old that there are peers who have a heavier chain we haven't
-  // heard about
-  if (seconds_since_last_block > 100 && head_of_chain->block_number != 0)
-  {
-    FETCH_LOG_INFO(LOGGING_NAME, "Synchronisation appears to be lost - chain is old.");
-    state_machine_->Delay(std::chrono::milliseconds{1000});
-    next_state = State::REQUEST_HEAVIEST_CHAIN;
-  }
-  else if (chain_.HasMissingBlocks())
+  if (chain_.HasMissingBlocks())
```
n-hutton, line 511: Also, with @denis-trapeznikov  stuff, this will never happen since we do not allow loose blocks into the main chain.

libs/ledger/src/protocols/notarisation_service.cpp
```diff
@@ -159,18 +159,10 @@ NotarisationService::State NotarisationService::OnVerifyNotarisations()
 {
   BlockNotarisationShares ret;
 
-  try
+  // Attempt to resolve the promise and add it
+  if (!notarisation_promise_->IsSuccessful() || !notarisation_promise_->GetResult(ret))
```
n-hutton, line 163: Unrelated: I wonder what happens here if the promise were to fail

libs/ledger/src/storage_unit/storage_unit_client.cpp
```diff
@@ -556,7 +539,7 @@ bool StorageUnitClient::Lock(ShardIndex index)
                                                     RevertibleDocumentStoreProtocol::LOCK);
 
     // wait for the promise
-    success = promise->As<bool>();
+    success = promise->GetResult(success) && success;
```
n-hutton, line 542: This looks a bit suspect - if success were false and GetResult set it to true, would it still work as intended? In any case switching to success && promise->... would possibly be a minor time saver due to short circuit logic

libs/network/include/network/service/promise.hpp
```diff
@@ -215,29 +212,26 @@ class PromiseError final : public std::exception
...
-bool details::PromiseImplementation::As(T &ret) const
-{
-  if (!Wait())
+      success = true;
+    }
+  }
+  catch (std::exception const &ex)
   {
-    return false;
+    FETCH_LOG_WARN("Promise", "Error getting promise result: ", ex.what());
```
n-hutton: So now we won't see the service id etc in the reactor catch as a side effect of this right?

libs/semanticsearch/include/semanticsearch/schema/vocabulary_instance.hpp
```diff
@@ -19,7 +19,7 @@
 
 #include <functional>
 #include <map>
-#include <string>
+#include <memory> #include <string>
```
n-hutton: newline

libs/semanticsearch/include/semanticsearch/schema/vocabulary_instance.hpp
```diff
@@ -37,7 +37,7 @@ class VocabularyInstance : public std::enable_shared_from_this<VocabularyInstanc
   {
     // TODO(private issue 143): add destructor
     Vocabulary ret;
-    ret.reset(new VocabularyInstance(std::type_index(typeid(T)), new T(data)));
+    ret = std::make_shared<VocabularyInstance>(std::type_index(typeid(T)), new T(data));
```
n-hutton: Isn't this unrelated?

n-hutton (changes requested):
minor changes

libs/beacon/include/beacon/block_entropy.hpp
jinmannwong: I would say for using in the beacon setup service to not stack the same aeon twice we should only be checking equality of fields that are unique for an aeon. So things which are block dependent like block number, digest (which hashed the block number), confirmations, group signature and block notarisation should probably not be included. In this sense, I am not sure the entropy should be in the comparator of aeon because I don't think you would want to queue two aeons which only differ on their previous block entropy.

jinmannwong: This messiness arises because the previous entropy should not be in an aeon, since it is not an aeon quantity and the beacon setup service does not care about previous entropy. The previous entropy is something that needs to be set for the beacon service.

apps/storage-tools/block_storage_client.cpp
ejfitzgerald: Yes that is fine

apps/storage-tools/block_storage_tool.cpp
ejfitzgerald: Done

libs/ledger/src/protocols/main_chain_rpc_service.cpp
ejfitzgerald: Sure this is just minor promise stuff though the change is relatively simple

libs/ledger/src/protocols/main_chain_rpc_service.cpp
ejfitzgerald: Sure

libs/ledger/src/storage_unit/storage_unit_client.cpp
ejfitzgerald: The key thing is that `GetResult()` populated `success` so it needs to be this way around. I agree a little confusing, but accurate

libs/semanticsearch/include/semanticsearch/schema/vocabulary_instance.hpp
ejfitzgerald: Weird note sure where that came from, but fixed

libs/network/include/network/service/promise.hpp
ejfitzgerald: Good spot this is fixed in the latest updates

jinmannwong (dismissed):

libs/beacon/include/beacon/block_entropy.hpp
n-hutton: Great, thanks

n-hutton (dismissed):
LGTM

PR #2065, opened 2019-12-03
Added throw after allocating SharedArray fails

MissingNO57:


baykaner (commented):
let's have a test that demonstrates the behaviour

baykaner (dismissed):

baykaner (approved):

PR #2066, opened 2019-12-03
Feature/uint256 additional tests

baykaner:
This PR implements further tests for basic uint256 arithmetic


pbukva:
These tests are covered in the #2010 + that PR carries multiple fixes (e.g. fixes for 2 bugs related to special handling of temporary VM variables for non-commutative operators [/ and /=](https://github.com/fetchai/ledger/pull/2010#discussion_r353083258) and [- and -=](https://github.com/fetchai/ledger/pull/2010#discussion_r353083379).

Of course, tests in etch can stay, but it would be better to test the mathematical behaviour in unit tests for underlying `UInt<T>` class rather than in `UInt256Wrapper` VM class, where we should just test the ability to access/call exposed API of types in Etch.

vhavrylov-ncube (approved):
Great job!

PR #2067, opened 2019-12-03
Feature/ml 283 replace double conversions

markos:
replace double constants with exact FixedPoint integer values and avoid conversion from floats/double. This way make calculations fully deterministic.


baykaner (approved):
lgtm

PR #2069, opened 2019-12-04
Etch compile-time annotation checks

wkaluza:
Annotations are now checked at compile time.

Function definitions now only admit no annotations other than
* `@init`
* `@action`
* `@query`
* `@problem`
* `@objective`
* `@work`
* `@clear`

Contract function prototypes now have to be annotated with `@action`. Unannotated contract functions are disabled until we decide to enable them for c2c calls.

Attempting to put multiple `@init`-functions in one contract is now a compile-time error.

Attempting to put multiple annotations on one function definition or prototype is now a compile-time error.

`@init`-functions must either take no arguments or one `Address`.

`@init`-functions and `@action`s must now either be void or return `Int64`.

@ejfitzgerald FYI, especially the Int64 returns.

Edit: Queries are now non-void

Synergetic contracts are now validated at compile time:
* the four synergetic functions marked with `@clear`, `@problem`, `@objective` and `@work` need to all be present or absent
* attempting to have more than one of any of the synergetic functions in one contract is an error
* The prototypes and interdependencies of the synergetic functions are now enforced:
    * `@problem`: `function f(Array<StructuredData>) : NonVoidProblemType`
    * `@work`: `function f(NonVoidProblemType, UInt256) : NonVoidSolutionType`
    * `@objective`: `function f(NonVoidProblemType, NonVoidSolutionType) : Int64`
    * `@clear`: `function f(NonVoidProblemType, NonVoidSolutionType)`

### DIFF
--- a/libs/vm/src/analyser.cpp
+++ b/libs/vm/src/analyser.cpp
@@ -394,6 +394,415 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
   return true;
 }
 
+void Analyser::EnforceLedgerRestrictions(BlockNodePtr const &block_node)
+{
+  LedgerRestrictionMetadata metadata;
+
+  ValidateBlock(block_node, metadata);
+
+  CheckInitFunctionUnique(metadata);
+  if (CheckSynergeticFunctionsPresentAndUnique(metadata))
+  {
+    CheckSynergeticContract(metadata);
+  }
+}
+
+void Analyser::ValidateBlock(BlockNodePtr const &block_node, LedgerRestrictionMetadata &metadata)
+{
+  for (NodePtr const &child : block_node->block_children)
+  {
+    switch (child->node_kind)
+    {
+    case NodeKind::File:
+    {
+      BlockNodePtr file_node = ConvertToBlockNodePtr(child);
+      filename_              = file_node->text;
+      ValidateBlock(file_node, metadata);
+      break;
+    }
+    case NodeKind::ContractDefinition:
+    case NodeKind::StructDefinition:
+    {
+      ExpressionNodePtr name_node = ConvertToExpressionNodePtr(child->children[0]);
+      TypePtr           type      = name_node->type;
+      if (type)
+      {
+        ValidateBlock(ConvertToBlockNodePtr(child), metadata);
+      }
+      break;
+    }
+    case NodeKind::ContractFunction:
+    {
+      ValidateFunctionAnnotations(child);
+      ValidateFunctionPrototype(child, metadata);
+      break;
+    }
+    case NodeKind::FreeFunctionDefinition:
+    case NodeKind::MemberFunctionDefinition:
+    {
+      ValidateFunctionAnnotations(ConvertToBlockNodePtr(child));
+      ValidateFunctionPrototype(ConvertToBlockNodePtr(child), metadata);
+      break;
+    }
+    default:
+    {
+      break;
+    }
+    }  // switch
+  }
+}
+
+void Analyser::ValidateFunctionAnnotations(NodePtr const &function_node)
+{
+  auto const &annotations_node = function_node->children[0];
+  if (annotations_node == nullptr)
+  {
+    return;
+  }
+  assert(annotations_node->node_kind == NodeKind::Annotations);
+
+  if (annotations_node->children.size() != 1u)
+  {
+    AddError(annotations_node->line, "Functions are only allowed to bear one annotation");
+  }
+
+  for (auto const &annotation : annotations_node->children)
+  {
+    auto const &text = annotation->text;
+    if (text != INIT_ANNOTATION && text != ACTION_ANNOTATION && text != QUERY_ANNOTATION &&
+        text != PROBLEM_ANNOTATION && text != OBJECTIVE_ANNOTATION && text != WORK_ANNOTATION &&
+        text != CLEAR_ANNOTATION)
+    {
+      AddError(annotation->line, "Invalid annotation");
+    }
+  }
+}
+
+void Analyser::ValidateFunctionPrototype(NodePtr const &            function_node,
+                                         LedgerRestrictionMetadata &metadata)
+{
+  auto const &function_name_node = ConvertToExpressionNodePtr(function_node->children[1]);
+  auto const &annotations        = function_node->children[0];
+
+  bool is_action = false;
+  bool is_init   = false;
+  bool is_query  = false;
+
+  assert(!annotations || annotations->node_kind == NodeKind::Annotations);
+
+  switch (function_node->node_kind)
+  {
+  case NodeKind::ContractFunction:
+  {
+    if (HasAnnotation(annotations, ACTION_ANNOTATION))
+    {
+      is_action = true;
+    }
+    else
+    {
+      AddError(function_name_node->line, "Contract functions must carry the @action annotation");
+    }
+  }
+  break;
+  case NodeKind::FreeFunctionDefinition:
+  {
+    if (HasAnnotation(annotations, ACTION_ANNOTATION))
+    {
+      is_action = true;
+    }
+    if (HasAnnotation(annotations, INIT_ANNOTATION))
+    {
+      is_init = true;
+      metadata.init_functions.emplace_back(function_node, filename_);
+    }
+    if (HasAnnotation(annotations, QUERY_ANNOTATION))
+    {
+      is_query = true;
+    }
+    if (HasAnnotation(annotations, CLEAR_ANNOTATION))
+    {
+      metadata.clear_functions.emplace_back(function_node, filename_);
+    }
+    if (HasAnnotation(annotations, OBJECTIVE_ANNOTATION))
+    {
+      metadata.objective_functions.emplace_back(function_node, filename_);
+    }
+    if (HasAnnotation(annotations, PROBLEM_ANNOTATION))
+    {
+      metadata.problem_functions.emplace_back(function_node, filename_);
+    }
+    if (HasAnnotation(annotations, WORK_ANNOTATION))
+    {
+      metadata.work_functions.emplace_back(function_node, filename_);
+    }
+  }
+  break;
+  case NodeKind::MemberFunctionDefinition:
+  {
+    if (annotations)
+    {
+      AddError(annotations->line, "Annotations on member functions are not permitted");
+    }
+  }
+  break;
+  default:
+    assert(false);
+    break;
+  }
+
+  auto const &return_type = function_name_node->function->return_type != nullptr
+                                ? function_name_node->function->return_type
+                                : void_type_;
+
+  if (is_action)
+  {
+    if (return_type != int64_type_ && return_type != void_type_)
+    {
+      AddError(function_name_node->line, "@action functions must either be void or return Int64");
+    }
+  }
+
+  if (is_init)
+  {
+    if (return_type != int64_type_ && return_type != void_type_)
+    {
+      AddError(function_name_node->line, "@init functions must either be void or return Int64");
+    }
+
+    auto const param_types = function_name_node->function->parameter_types;
+    if (!param_types.empty() && !(param_types.size() == 1 && param_types.back() == address_type_))
+    {
+      AddError(function_name_node->line,
+               "@init functions must accept eithoer no parameters or exactly one parameter of type "
+               "Address");
+    }
+  }
+
+  if (is_query)
+  {
+    if (return_type == void_type_)
+    {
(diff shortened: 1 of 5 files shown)


ejfitzgerald (commented):
Outstanding @wkaluza  - This is perfect. I approve, just haven't pressed the green button so that Robert can voice an opinion

robertdickson (commented):
I agree, this is outstanding. @wkaluza and I spoke this morning and agreed that we would wait until my structs PR goes in before adding this, then I'll be more than happy to approve. That is due to happen tomorrow all being well, so not long to wait.

libs/vm/src/analyser.cpp
```diff
@@ -37,9 +36,40 @@
...
+std::string const WORK_ANNOTATION      = "@work";
+std::string const CLEAR_ANNOTATION     = "@clear";
+
+bool HasAnnotation(NodePtr const &annotations_node, std::string const &annotation_name)
+{
+  if (annotations_node)
+  {
+    assert(annotations_node->node_kind == NodeKind::Annotations);
+
+    for (auto const &annotation_ptr : annotations_node->children)
```
robertdickson: Niggle, but I don't use _ptr in names. Just use "annotation" here?

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+    {
+    case NodeKind::File:
+    {
+      BlockNodePtr file_node = ConvertToBlockNodePtr(child);
+      filename_              = file_node->text;
+      ValidateBlock(file_node, metadata);
+      break;
+    }
+    case NodeKind::ContractDefinition:
+    case NodeKind::StructDefinition:
```
robertdickson: Struct member functions can't have annotations, so NodeKind::StructDefinition isn't needed here.

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+      break;
+    }
+    case NodeKind::ContractFunction:
+    {
+      ValidateFunctionAnnotations(child);
+      ValidateFunctionPrototype(child, metadata);
+      break;
+    }
+    case NodeKind::FreeFunctionDefinition:
+    case NodeKind::MemberFunctionDefinition:
```
robertdickson: Struct member functions can't have annotations, so NodeKind::MemberFunctionDefinition isn't needed here.

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+      TypePtr           type      = name_node->type;
+      if (type)
+      {
+        ValidateBlock(ConvertToBlockNodePtr(child), metadata);
+      }
+      break;
+    }
+    case NodeKind::ContractFunction:
+    {
+      ValidateFunctionAnnotations(child);
```
robertdickson: Need to check that the function name node ( child->children[1] ) has a valid function, and ignore the node if not

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+      ValidateFunctionPrototype(child, metadata);
+      break;
+    }
+    case NodeKind::FreeFunctionDefinition:
+    case NodeKind::MemberFunctionDefinition:
+    {
+      ValidateFunctionAnnotations(ConvertToBlockNodePtr(child));
```
robertdickson: Need to check that the function name node ( child->children[1] ) has a valid function, and ignore the node if not

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+    {
+      metadata.problem_functions.emplace_back(function_node, filename_);
+    }
+    if (HasAnnotation(annotations, WORK_ANNOTATION))
+    {
+      metadata.work_functions.emplace_back(function_node, filename_);
+    }
+  }
+  break;
+  case NodeKind::MemberFunctionDefinition:
```
robertdickson: I don't think this is needed, as MemberFunctionDefinition nodes won't have any annotation node children

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+      AddError(annotations->line, "Annotations on member functions are not permitted");
+    }
+  }
+  break;
+  default:
+    assert(false);
+    break;
+  }
+
+  auto const &return_type = function_name_node->function->return_type != nullptr
```
robertdickson: function always has a valid return_type

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+                                ? function_name_node->function->return_type
+                                : void_type_;
+
+  if (is_action)
+  {
+    if (return_type != int64_type_ && return_type != void_type_)
+    {
+      AddError(function_name_node->line, "@action functions must either be void or return Int64");
```
robertdickson: "Void" probably won't make sense to users -- can we change this error message to something about not having a return type?

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+    if (return_type != int64_type_ && return_type != void_type_)
+    {
+      AddError(function_name_node->line, "@init functions must either be void or return Int64");
+    }
+
+    auto const param_types = function_name_node->function->parameter_types;
+    if (!param_types.empty() && !(param_types.size() == 1 && param_types.back() == address_type_))
+    {
+      AddError(function_name_node->line,
+               "@init functions must accept eithoer no parameters or exactly one parameter of type "
```
robertdickson: typo on "eithoer"

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+               "Address");
+    }
+  }
+
+  if (is_query)
+  {
+    if (return_type == void_type_)
+    {
+      AddError(function_name_node->line, "@query functions must not be void");
```
robertdickson: "must return a value" ?

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+  bool const problem_params_correct = problem.parameter_types.size() == 1 && problem_param_type &&
+                                      problem_param_type->name == "Array<StructuredData>";
+  if (!problem_params_correct)
+  {
+    AddError(
+        problem_file, problem_node->line,
+        "The @problem function should accept a single parameter of type Array<StructuredData>");
+  }
+
+  if (problem.return_type == nullptr || problem.return_type == void_type_)
```
robertdickson: return_type is always valid

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+  }
+
+  for (auto const &annotation : annotations_node->children)
+  {
+    auto const &text = annotation->text;
+    if (text != INIT_ANNOTATION && text != ACTION_ANNOTATION && text != QUERY_ANNOTATION &&
+        text != PROBLEM_ANNOTATION && text != OBJECTIVE_ANNOTATION && text != WORK_ANNOTATION &&
+        text != CLEAR_ANNOTATION)
+    {
+      AddError(annotation->line, "Invalid annotation");
```
robertdickson: Maybe add the bad annotation to the error message?

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+    }
+  }
+}
+
+void Analyser::ValidateFunctionPrototype(NodePtr const &            function_node,
+                                         LedgerRestrictionMetadata &metadata)
+{
+  auto const &function_name_node = ConvertToExpressionNodePtr(function_node->children[1]);
```
robertdickson, line 494: Just for completeness, have an assert for the function name node having a valid function here, and in ValidateFunctionAnnotations()

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+  auto const &objective_file = metadata.objective_functions.back().filename;
+  auto const &problem_file   = metadata.problem_functions.back().filename;
+  auto const &work_file      = metadata.work_functions.back().filename;
+
+  auto const &clear_node     = metadata.clear_functions.back().node;
+  auto const &objective_node = metadata.objective_functions.back().node;
+  auto const &problem_node   = metadata.problem_functions.back().node;
+  auto const &work_node      = metadata.work_functions.back().node;
+
+  auto const &clear     = *ConvertToExpressionNodePtr(clear_node->children[1])->function;
```
robertdickson: Rename these clear_function etc just to make easier to read.

libs/vm/src/analyser.cpp
```diff
@@ -357,25 +390,438 @@ bool Analyser::Analyse(BlockNodePtr const &root, std::vector<std::string> &error
...
+    return false;
+  }
+
+  if (metadata.clear_functions.size() == 1 && metadata.objective_functions.size() == 1 &&
+      metadata.problem_functions.size() == 1 && metadata.work_functions.size() == 1)
+  {
+    return true;
+  }
+
+  bool const clear_missing     = metadata.clear_functions.empty();
```
robertdickson: Pedantic, but maybe put these bools at the top and reuse?

libs/vm/src/analyser.cpp
wkaluza: done.

libs/vm/src/analyser.cpp
wkaluza: done

libs/vm/src/analyser.cpp
wkaluza: done

libs/vm/src/analyser.cpp
wkaluza: done

libs/vm/src/analyser.cpp
wkaluza: done

libs/vm/src/analyser.cpp
wkaluza: done

libs/vm/src/analyser.cpp
wkaluza: Sweet! Fixed.

libs/vm/src/analyser.cpp
wkaluza: good call, done

libs/vm/src/analyser.cpp
wkaluza: D'oh! Fixed.

libs/vm/src/analyser.cpp
wkaluza: I've gone with "must have a return type". If they forget to return when the function's not void, they'll get a meaningful error on the next attempt

libs/vm/src/analyser.cpp
wkaluza: fixed

libs/vm/src/analyser.cpp
wkaluza: done

libs/vm/src/analyser.cpp
wkaluza: done

libs/vm/src/analyser.cpp
wkaluza: done

libs/vm/src/analyser.cpp
wkaluza: done

robertdickson (approved):
Looks great, let's get this in.

PR #2070, opened 2019-12-04
Initial implementation of c2c calls

wkaluza:
Adds ledger support for contracts like the following:

```
// caller contract
contract c1_interface
  @action
  function eleven() : Int64;
endcontract

@action
function c2c_call(contract_name : String)
  contract c = c1_interface(contract_name);

  assert(c.eleven() == 11i64);
endfunction
```
```
// callee contract
@action
function eleven() : Int64
  return 11i64;
endfunction
```


scripts/install-test-dependencies.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -e
 
-pip3 install  wheel pyyaml requests --force-reinstall
-pip3 install  fetchai-ledger-api==0.11.0a3 --force-reinstall
+pip3 install wheel --force-reinstall
+pip3 install pyyaml requests --force-reinstall
```
wkaluza, line 5: pyyaml needs wheel to install without errors, but pip is not clever enough to know that.

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -495,8 +498,119 @@ Contract::Result SmartContract::InvokeAction(std::string const &name, chain::Tra
...
+      ss << "Execution of function " << function.name << " from contract " << identity
+         << " failed with error \"" << error << "\"";
+      error = ss.str();
+    }
+
+    c.state_adapter->PopContext();
+    vm->IncreaseChargeTotal(vm2.GetChargeTotal() - reference_charge);
+
+    auto it = call_history.find(called_contract_address);
+    call_history.erase(it);
```
ejfitzgerald: Think that it is probably safer (and more concise) in the following form:

```suggestion
    call_history.erase(called_contract_address);
```

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -495,8 +498,119 @@ Contract::Result SmartContract::InvokeAction(std::string const &name, chain::Tra
...
+      error = "Contract-to-contract call cycle detected";
+
+      return false;
+    }
+
+    call_history.insert(called_contract_address);
+
+    decltype(auto) c = context();
+
+    // TODO(WK) charge for reading from storage
```
ejfitzgerald: Can we raise an issue for this one

ejfitzgerald (dismissed):
Really excellent work, just a couple of very minor comments

ejfitzgerald (dismissed):

ejfitzgerald (approved):

PR #2073, opened 2019-12-04
Bug/supplied peers not being read from json

engineerkatie:


libs/dmlf/src/colearn/muddle_learner_networker_impl.cpp
```diff
@@ -223,6 +240,11 @@ void MuddleLearnerNetworkerImpl::PushUpdateBytes(UpdateType const &type_name, By
     auto next_ones = alg_->GetNextOutputs();
 
     Peers peers;
+    if (next_ones.empty())
+    {
+      FETCH_LOG_WARN(LOGGING_NAME, "PushUpdateBytes(2) got no peers from alg to send to.");
```
JuanBesa: What is (2)?

Shouldn't it be "... **has** no peers...?

libs/dmlf/src/colearn/muddle_learner_networker_impl.cpp
```diff
@@ -74,13 +74,14 @@ void MuddleLearnerNetworkerImpl::Setup(MuddlePtr mud, StorePtr update_store)
...
     byte_array::ConstByteArray bytes;
     double                     proportion;
     double                     random_factor;
 
     auto source = std::string(fetch::byte_array::ToBase64(from));
 
-    buf >> type_name >> bytes >> proportion >> random_factor;
+    buf >> algo_name >> type_name >> bytes >> proportion >> random_factor;
 
     std::cout << "from:" << source << ", "
```
JuanBesa, line 86: Would it be useful to add algo_name and type_name to the cout message?

JuanBesa (dismissed):
Please see comments

libs/dmlf/src/colearn/muddle_learner_networker_impl.cpp
engineerkatie: Yes, I should do that.

libs/dmlf/src/colearn/muddle_learner_networker_impl.cpp
engineerkatie: (2) is the second message of that form. "All error messages should be unique in the source"

engineerkatie:
Doing this in the private repo instead.

PR #2076, opened 2019-12-05
LDGR-627: Fix null String crash and Bloom filter loading failure

qati:
This PR fixes two issues:
* the crash caused by returning null String from contract function
* if the node crashes and the db files are corrupted Bloom filter loading will cause node crash


libs/ledger/src/chain/main_chain.cpp
```diff
@@ -99,11 +99,18 @@ MainChain::~MainChain()
...
+    try
+    {
+      std::ofstream out(BLOOM_FILTER_STORE, std::ios::binary | std::ios::out | std::ios::trunc);
+      LargeObjectSerializeHelper buffer{};
+      buffer << bloom_filter_;
 
-    out << buffer.data();
+      out << buffer.data();
+    }
+    catch (std::exception &e)
```
wkaluza: catch exceptions by const ref

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -806,11 +813,20 @@ void MainChain::RecoverFromFile(Mode mode)
...
+        LargeObjectSerializeHelper buffer{bloom_filter_data};
 
-      buffer >> bloom_filter_;
+        buffer >> bloom_filter_;
+      }
+      catch (std::exception &e)
+      {
+        FETCH_LOG_ERROR(LOGGING_NAME,
+                        "Failed to load Bloom filter from storage! Reason: ", e.what());
+        bloom_filter_.Reset();
```
wkaluza: There's a problem here, because now the main chain won't have a Bloom filter, so any transaction that comes up will be treated as new even if it is already in the chain. At the moment I think this requires a reset of the chain, as well as the filter.

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -806,11 +813,20 @@ void MainChain::RecoverFromFile(Mode mode)
...
+      {
+        byte_array::ByteArray bloom_filter_data{in};
 
-      LargeObjectSerializeHelper buffer{bloom_filter_data};
+        LargeObjectSerializeHelper buffer{bloom_filter_data};
 
-      buffer >> bloom_filter_;
+        buffer >> bloom_filter_;
+      }
+      catch (std::exception &e)
```
wkaluza: catch by const ref

libs/ledger/src/chaincode/smart_contract.cpp
```diff
@@ -741,8 +741,14 @@ SmartContract::Status SmartContract::InvokeQuery(std::string const &name, Query
     response["result"] = output.Get<fixed_point::fp64_t>();
     break;
   case vm::TypeIds::String:
-    response["result"] = output.Get<vm::Ptr<vm::String>>()->string();
-    break;
+  {
+    auto res = output.Get<vm::Ptr<vm::String>>();
```
wkaluza: res may be const

wkaluza (changes requested):

wkaluza (dismissed):

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -806,11 +813,20 @@ void MainChain::RecoverFromFile(Mode mode)
...
+        LargeObjectSerializeHelper buffer{bloom_filter_data};
 
-      buffer >> bloom_filter_;
+        buffer >> bloom_filter_;
+      }
+      catch (std::exception const &e)
+      {
+        FETCH_LOG_ERROR(LOGGING_NAME,
+                        "Failed to load Bloom filter from storage! Reason: ", e.what());
+        Reset();
```
ejfitzgerald, line 1017: What happens with this in practise? Is there any easy point in the code that we can signal in the block coordinator that we need to restart from genesis again?

ejfitzgerald (changes requested):
I think that this will work in the general case, however, because the block coordinator has no knowledge of this event, I can foresee issues with starting from genesis again. Maybe we can have a chat about this between us three this morning?

libs/ledger/src/chain/main_chain.cpp
qati: Added the signal to block coordinator to reset

ejfitzgerald (changes requested):
Sorry @qati can we revert back to the original code I think that should be great as discussed. Then lets try and get this to work

ejfitzgerald (dismissed):

ejfitzgerald (approved):

libs/ledger/src/chain/main_chain.cpp
n-hutton: my testing has found that this happens every time - so this is not a fix?

n-hutton (changes requested):

n-hutton:
see comment

ejfitzgerald:
@n-hutton is this testing docker based, if so it is a bug in crash reporter. I have fixed this

n-hutton:
My experience is that the bloom filter, if you shut the node down and bring it up again, will always fail to deser. It won't crash the node but it will call reset on the main chain forcing the BC back to genesis when you start up.

Not sure how the crash reporter would interfere with this (?)

n-hutton (approved):

wkaluza (approved):

PR #2054, opened 2019-12-02
Feature/vm model benchmarks

MissingNO57:
combined effort to implement:
- estimator functions for VMTensor and VMModel
- benchmarking of Tensor & Model functions (to tune estimators)
- relevant performance & tuning tests
- implemented tensor Min, Max, Sum etch bindings, estimators, tests
- temporarily disabled tests for vm_bindings tagged experimental (to avoid charge overflow)

PR #2077, opened 2019-12-05
Feature/ml 309 tensor min estimation

baykaner:
Should be merged in after #2054 
- refactoring some tensor estimator code
- implement binding for Tensor min function
- implement estimator for min function
- implement estimator min function test


baykaner:
this work was merged into #2054

PR #2081, opened 2019-12-05
Dev/uris in abstract interfaces

engineerkatie:
Make the last step to move URIs into the base interface.


engineerkatie:
Doing this in the private repo instead.

PR #2082, opened 2019-12-05
Replacing exceptions with logging: Examples, VMOptimiser

vhavrylov-ncube:
- replaced std::runtime error wtih std::cout or vm.RuntimeError()
  wherever possible.

### DIFF
--- a/libs/vm-modules/tests/unit/ml.cpp
+++ b/libs/vm-modules/tests/unit/ml.cpp
@@ -23,7 +23,9 @@
 #include "vm_modules/ml/training_pair.hpp"
 #include "vm_test_toolkit.hpp"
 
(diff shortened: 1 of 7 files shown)


libs/vm-modules/tests/unit/ml.cpp
```diff
@@ -319,6 +347,34 @@ TEST_F(MLTests, graph_string_serialisation_test)
   EXPECT_TRUE(initial_loss->GetTensor().AllClose(loss->GetTensor()));
 }
 
+TEST_F(MLTests, optimiser_construction_adam)
```
baykaner, line 350: very nice!

baykaner (approved):
LGTM

PR #2083, opened 2019-12-05
VMTensor benchmarks and VMModel Estimator coeficients tuning [COLE-70][COLE-71][COLE-74][COLE-42]

MissingNO57:


libs/vm-modules/src/math/tensor/tensor_estimator.cpp
```diff
@@ -93,65 +95,153 @@ ChargeAmount TensorEstimator::SetAtFour(uint64_t /*idx1*/, uint64_t /*idx2*/, ui
...
+    {
+      if (cur_dim == 0)
+      {
+        break;
+      }
+      --cur_dim;
+    }
+  }
+
+  if (not_found)
```
baykaner, line 159: estimators dont need logic like this; if the behaviour is invalid it'll throw anyway.
Also check with volodymyr that the tensor behaviour for squeeze is going to stay like this or whether it'l being updated (i.e. numpy squeeze i think doesn't throw if there's nothing to squeeze, it just doesn't do anything)

baykaner (dismissed):

baykaner (approved):

PR #2084, opened 2019-12-05
bugfixes seen during deployment and E2E tests

n-hutton:


libs/constellation/src/constellation.cpp
```diff
@@ -549,6 +549,9 @@ bool Constellation::Run(UriSet const &initial_peers, core::WeakRunnable bootstra
     }
   }
 
+  // Need to set new genesis hash etc.
+  chain_.Reset();
```
n-hutton, line 553: The chain must be reset if it is possible the genesis hash has changed

libs/ledger/include/ledger/protocols/main_chain_rpc_protocol.hpp
```diff
@@ -67,8 +68,18 @@ class MainChainProtocol : public service::Protocol
 
   Travelogue TimeTravel(Digest start)
   {
-    auto ret_val = chain_.TimeTravel(std::move(start));
-    return {Copy(ret_val.blocks), ret_val.heaviest_hash};
+    try
+    {
+      auto ret_val = chain_.TimeTravel(std::move(start));
```
n-hutton, line 73: This allows peers to request a time travel for a missing hash which causes the main chain to throw

libs/ledger/include/ledger/protocols/main_chain_rpc_service.hpp
```diff
@@ -91,6 +91,8 @@ class MainChainRpcService : public muddle::rpc::Server,
   MainChainRpcService(MainChainRpcService &&)      = delete;
   ~MainChainRpcService() override                  = default;
 
+  void Start();
```
n-hutton, line 94: Only attach the gossip listener after setting up the main chain + genesis, it is pointless otherwise.

libs/ledger/include/ledger/protocols/main_chain_rpc_service.hpp
```diff
@@ -169,11 +171,12 @@ class MainChainRpcService : public muddle::rpc::Server,
...
-  StateMachinePtr state_machine_;
-  Address         current_peer_address_;
-  BlockHash       current_missing_block_;
-  Promise         current_request_;
+  RpcClient             rpc_client_;
+  StateMachinePtr       state_machine_;
+  Address               current_peer_address_;
+  BlockHash             current_missing_block_;
+  Promise               current_request_;
+  std::atomic<uint16_t> loose_blocks_seen_{0};
```
n-hutton, line 179: Count loose blocks seen and use this to guess that we are out of sync. There was a condition where the forward sync would request from a node, get Genesis or genesis + 1 block, and then stay in the syncronised state forever, dumping all loose blocks

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -393,11 +378,21 @@ MainChainRpcService::State MainChainRpcService::OnWaitForHeaviestChain()
...
+          next_state = GetInitialState(mode_);
+        }
       }
       else
       {
         FETCH_LOG_INFO(LOGGING_NAME, "Heaviest chain request to: ", ToBase64(current_peer_address_),
                        " failed. Reason: ", service::ToString(status));
+
+        state_machine_->Delay(std::chrono::seconds{1});
+        next_state = GetInitialState(mode_);
```
n-hutton, line 400: If this condition happened, it would spin in this state forever printing the log message

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -393,11 +378,21 @@ MainChainRpcService::State MainChainRpcService::OnWaitForHeaviestChain()
             next_state = State::SYNCHRONISING;  // we have reached the tip
           }
         }
+        else
+        {
+          FETCH_LOG_WARN(LOGGING_NAME,
+                         "Received empty block response when forward syncing the chain!");
+          state_machine_->Delay(std::chrono::seconds{8});
```
n-hutton, line 390: might be worth to do something like Consensus::BlockInterval

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -428,7 +423,7 @@ MainChainRpcService::State MainChainRpcService::OnSynchronising()
     }
 
     FETCH_LOG_INFO(LOGGING_NAME, "Requesting chain from muddle://", ToBase64(current_peer_address_),
-                   " for block ", ToBase64(current_missing_block_));
+                   " for block ", current_missing_block_.ToHex());
```
n-hutton: @wkaluza has specified block hashes/hashes in general should be in hex format

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -496,15 +491,13 @@ MainChainRpcService::State MainChainRpcService::OnSynchronised(State current, St
...
 
   MainChain::BlockPtr head_of_chain = chain_.GetHeaviestBlock();
-  uint64_t const      seconds_since_last_block =
-      GetTime(fetch::moment::GetClock("default", fetch::moment::ClockType::SYSTEM)) -
-      head_of_chain->timestamp;
 
   // Assume if the chain is quite old that there are peers who have a heavier chain we haven't
   // heard about
-  if (seconds_since_last_block > 100 && head_of_chain->block_number != 0)
+  if (loose_blocks_seen_ > 5)
```
n-hutton, line 508: Might want some rethinking but it is worth avoiding the case where a single loose block bumps you from this state

libs/ledger/src/protocols/main_chain_rpc_service.cpp
ejfitzgerald:
```suggestion
                   " for block 0x", current_missing_block_.ToHex());
```

libs/ledger/src/protocols/main_chain_rpc_service.cpp
ejfitzgerald, line 508: This will probably conflict with my changes on `master`

libs/ledger/include/ledger/protocols/main_chain_rpc_service.hpp
ejfitzgerald, line 94: Maybe I am going mad but where is this called from? Does it also make sense to have a stop the drops the subscription on the floor too?

libs/ledger/src/storage_unit/storage_unit_client.cpp
```diff
@@ -144,43 +144,31 @@ bool StorageUnitClient::RevertToHash(Hash const &hash, uint64_t index)
...
-    FETCH_LOG_INFO(LOGGING_NAME, "Reverting state to genesis.");
+    FETCH_LOG_INFO(LOGGING_NAME, "Reverting state to genesis!");
+  }
 
-    // fill the tree with empty leaf nodes
-    for (std::size_t i = 0; i < num_lanes(); ++i)
-    {
-      tree[i] = chain::GENESIS_MERKLE_ROOT;
-    }
+  uint64_t const merkle_stack_size = permanent_state_merkle_stack_.size();
```
n-hutton, line 155: The old logic here was wrong. If you were to revert to genesis it incorrectly assumed that the lanes would all be GENESIS_MERKLE_ROOT - not true.

libs/ledger/include/ledger/protocols/main_chain_rpc_protocol.hpp
```diff
@@ -67,8 +68,18 @@ class MainChainProtocol : public service::Protocol
...
+      return {Copy(ret_val.blocks), ret_val.heaviest_hash};
+    }
+    catch (std::exception const &ex)
+    {
+      FETCH_LOG_WARN(LOGGING_NAME,
+                     "Failed to respond to time travel request for block hash: ", start.ToHex(),
```
ejfitzgerald:
```suggestion
                     "Failed to respond to time travel request for block hash: 0x", start.ToHex(),
```

ejfitzgerald (changes requested):

libs/ledger/include/ledger/protocols/main_chain_rpc_service.hpp
n-hutton: Urgh, forgot to include that. thanks.

libs/ledger/src/protocols/main_chain_rpc_service.cpp
n-hutton: already resolved it but yes.

ejfitzgerald (dismissed):

PR #2085, opened 2019-12-05
Improvement/root page

denis-trapeznikov:
This commit introduces a glamorous >_> root page at 127.0.0.1:8000/; in addition, a simple tree-like structures generic rendering mechanism was quickly devised, in case one needs one.


libs/http/include/http/server.hpp
```diff
@@ -331,6 +342,79 @@ class HTTPServer : public AbstractHTTPServer
     });
   }
 
+  void AddDefaultRootModule()
```
ejfitzgerald, line 332: My feeling is that this should be not be contained here in the HTTP server. I can we not move this to an external module and then provide the accessors for the routes via the HTTP server interface.

ejfitzgerald (changes requested):

libs/http/include/http/server.hpp
denis-trapeznikov: Ok, possible but—Either way the server should be aware of this module. Even if the module's definition itself is contained in a separate file, the module is added via server's special method (well, unless we declare it server's friend — which still requires changes to server's code).

ejfitzgerald (approved):

PR #2088, opened 2019-12-06
Feature/ml 312 ml 310 tensor max and sum estimation

baykaner:
- implements Max for etch tensor & estimator
- implements Sum for etch tensor & estimator


baykaner:
work merged into #2054

PR #2092, opened 2019-12-06
Feature/ml 330 arithmetic operators for etch tensor

baykaner:
- implements arithmetic operators for etch tensor - but these cannot be bound until operators can take estimators
- implements member functions which substitute for operators until then


libs/vm-modules/tests/unit/math/tensor_tests.cpp
```diff
@@ -355,6 +355,498 @@ TEST_F(MathTensorTests, tensor_set_from_string)
...
+      var x = Tensor(tensor_shape);
+      var y = Tensor(tensor_shape);
+      x.fill(7.0fp64);
+      y.fill(7.0fp64);
+      var result : Bool = x.isEqual(y);
+      return result;
+    endfunction
+  )";
+
+  //  static char const *tensor_equal_true_src = R"(
```
markos, line 378: leftover comment?

wkaluza (approved):

PR #2093, opened 2019-12-06
slightly more intelligent block coord recovery

n-hutton:
Previously, if the block coordinator came up and there was a mismatch between the main chain and the merkle tree most recently written to the DB, it would revert to genesis.

Now, it should try to step backwards from the heaviest block seeing if it can revert to a known state. It will do this all the way to genesis, but I don't anticipate this to be a costly operation relative to start up times

### DIFF
--- a/libs/ledger/src/chain/block_coordinator.cpp
+++ b/libs/ledger/src/chain/block_coordinator.cpp
@@ -200,7 +200,8 @@ BlockCoordinator::BlockCoordinator(MainChain &chain, DAGPtr dag,
   assert(consensus_);
 
   state_machine_->OnStateChange([this](State current, State previous) {
-    if (periodic_print_.Poll())
+    FETCH_UNUSED(this);
+    // if (periodic_print_.Poll())
     {
       FETCH_LOG_INFO(LOGGING_NAME, "Current state: ", ToString(current),
                      " (previous: ", ToString(previous), ")");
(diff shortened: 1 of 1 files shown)


libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -1083,7 +1116,12 @@ BlockCoordinator::State BlockCoordinator::OnReset()
   }
   else
   {
-    FETCH_LOG_ERROR(LOGGING_NAME, "Unable to find a previously executed block!");
+    FETCH_LOG_ERROR(LOGGING_NAME,
+                    "Unable to find a previously executed block! Doing a hard reset.");
+    Reset();
+
+    auto genesis_block = chain_.CreateGenesisBlock();
```
n-hutton, line 1123: This should avoid a segv immediately after the error message

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -1274,6 +1312,7 @@ void BlockCoordinator::Reset()
   last_executed_block_.ApplyVoid([](auto &digest) { digest = chain::ZERO_HASH; });
   execution_manager_.SetLastProcessedBlock(chain::ZERO_HASH);
   chain_.Reset();
+  current_block_ = chain_.GetHeaviestBlock();
```
n-hutton, line 1315: We should make heavier use of this reset function when there is an extinction level event

PR #2094, opened 2019-12-06
Feature/ml 282 implement fixedpoint string parser

markos:
Parse FixedPoint directly from string and avoid conversion to double.


wkaluza:
Rather than using regular expressions, wouldn't it be better (not to mention easier) to use the Etch tokeniser? Probably will be faster too, given that the stdlib's `<regex>` implementation tends to be sluggish.

markos:
@wkaluza I actually removed regex and did it with plain string handling/conversion.

libs/vm-modules/tests/unit/state_tests.cpp
```diff
@@ -579,11 +579,11 @@ TEST_F(StateTests, test_serialisation_of_fixed_point32)
   ASSERT_TRUE(static_cast<bool>(retval));
   ASSERT_EQ(int32_t{3}, retval->Count());
 
-  EXPECT_EQ(fetch::fixed_point::fp32_t{1.0},
+  EXPECT_EQ(fetch::fixed_point::fp32_t("1.0"),
```
baykaner, line 582: please add tests to check scientific notation works as well

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
```diff
@@ -932,6 +950,104 @@ constexpr FixedPoint<16, 16>::FixedPoint(FixedPoint<32, 32> const &o)
...
+      power10 *= 10;
+    }
+    fractional_part = (fractional_part * ONE_MASK) / power10;
+  }
+  else
+  {
+    integer_match = s_copy;
+  }
+
+  if (integer_match.length() > DECIMAL_DIGITS)
```
baykaner, line 1014: what if the integer part is 00000000000001 ?

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
```diff
@@ -706,13 +707,19 @@ inline std::ostream &operator<<(std::ostream &s, int128_t const &x)
   return s;
 }
 
+inline std::ostream &operator<<(std::ostream &s, uint128_t const &x)
+{
+  s << static_cast<uint64_t>(x >> 64) << "|" << static_cast<uint64_t>(x);
```
baykaner, line 713: is this operator for printing? if so, presumably we don't want to print number as 123|456 but as 123.456?
what am i missing?

libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -236,6 +236,129 @@ TEST(FixedPointTest, Conversion_64_64)
...
+  fp128_t large2("-1442695040888963407359924681001892137");
+  EXPECT_TRUE(fp128_t::IsStateOverflow());
+  EXPECT_EQ(large1, fp128_t::FP_MAX);
+  EXPECT_EQ(large2, fp128_t::FP_MIN);
+
+  fp128_t e3("0E-12fp128");
+  EXPECT_EQ(e3, fp128_t::_0);
+}
+
 TEST(FixedPointTest, Constants_16_16)
```
baykaner, line 369: can you please add a few tests to check various parsing cases; e.g.:
- FixedPoint("000000000000001.0000000000000000000000000000000000000000000000001")
- FixedPoint("000000000000001.0000000000000000000000000000000000000000000000000")
etc.

baykaner (changes requested):

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
markos: no, this is only for printing the uint128_t value, the container, used only for debugging purposes.

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
markos: hm, I thought the parser in the vm already catches those, will check if that's the case. I can add an extra zeros trimming if not.

libs/vm-modules/tests/unit/state_tests.cpp
markos: ok

libs/math/tests/unit/fixed_point/fixed_point.cpp
markos: ok

libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -236,6 +236,129 @@ TEST(FixedPointTest, Conversion_64_64)
...
 
+TEST(FixedPointTest, FromString_16_16)
+{
+  // Get raw value
+  fp32_t one("1fp32");
+  fp32_t zero_point_five("0.5fp32");
+  fp32_t one_point_five("1.5fp32");
+  fp32_t two_point_five("2.5fp64");
+  fp32_t m_one_point_five("-1.5fp");
+  fp32_t m_one_point_five_em10("-1.5e-3fp32_t");
```
vhavrylov-ncube: em10, but e-3 (not e-10) is it ok?

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
wkaluza: The tokeniser won't let through any superfluous leading zeros. An integer part of a number is either exactly "0" or starts with a non zero digit

libs/math/tests/unit/fixed_point/fixed_point.cpp
markos: resolved.

libs/math/tests/unit/fixed_point/fixed_point.cpp
markos: scientific notation was removed, after discussion about fp problems.

libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
markos: added removal of leading and trailing zeros so this should be covered now.

libs/vm-modules/tests/unit/state_tests.cpp
markos: scientific notation was removed after discussion.

baykaner (approved):

PR #2096, opened 2019-12-06
implement guard for charge total overflow

baykaner:


libs/vm-modules/tests/unit/vm_charge_tests.cpp
```diff
@@ -65,6 +65,9 @@ class CustomType : public Object
   void TooExpensive(uint8_t /*unused*/, uint16_t /*unused*/)
   {}
 
+  void OverflowExpensive(uint8_t /*unused*/, uint16_t /*unused*/)
+  {}
+
```
wkaluza: we don't need this, see further comments

libs/vm-modules/tests/unit/vm_charge_tests.cpp
```diff
@@ -80,6 +83,8 @@ ChargeAmount const high_charge_limit = 1000;
 ChargeAmount const affordable_charge = 10;
 ChargeAmount const expensive_charge  = 1000;
 
+ChargeAmount const max_charge_amount = std::numeric_limits<ChargeAmount>::max();
```
wkaluza, line 83: This is good, but we also need this:
```
auto max_charge_estimator = [](uint8_t x, uint16_t y) -> ChargeAmount {
  return max_charge_amount;
};
```

libs/vm-modules/tests/unit/vm_charge_tests.cpp
```diff
@@ -463,6 +482,23 @@ TEST_F(
...
+  static char const *TEXT = R"(
+    function main()
+      CustomType.overflowExpensive(3u8, 4u16);
+    endfunction
+  )";
+
+  ASSERT_TRUE(toolkit.Compile(TEXT));
+  ASSERT_FALSE(toolkit.Run(nullptr, max_charge_amount));
+}
+
```
wkaluza: We don't need the complications of member functions, we're just writing an overflow check. Replace this second test with the following (nearly identical to the first test, but note use of `max_charge_estimator` instead of `max_charge_amount`).

```
TEST_F(VmChargeTests, functor_bind_with_charge_estimate_execution_does_not_overflow_charge_total_with_estimator)
{
  toolkit.module().CreateFreeFunction("overflowExpensive", handler, max_charge_estimator);

  static char const *TEXT = R"(
    function main()
      overflowExpensive(3u8, 4u16);
    endfunction
  )";

  ASSERT_TRUE(toolkit.Compile(TEXT));
  ASSERT_FALSE(toolkit.Run(nullptr, max_charge_amount));
}
```

wkaluza (changes requested):

wkaluza (dismissed):

wkaluza (approved):

PR #2100, opened 2019-12-06
Fix for ML-329 Tensor.At() on wrong indices

vhavrylov-ncube:
Fixed Tensor .At() accessor behavior when wrong index values are passed
Cleaned up existing unit tests
added tests for bad cases

### DIFF
--- a/libs/math/include/math/tensor.hpp
+++ b/libs/math/include/math/tensor.hpp
@@ -459,6 +459,12 @@ private:
   template <SizeType N, typename FirstIndex, typename... Indices>
   SizeType UnrollComputeColIndex(FirstIndex &&index, Indices &&... indices) const
   {
+    if (shape_[N] <= SizeType(index))
+    {
+      throw exceptions::WrongIndices(
+          "Tensor::At : index " + std::to_string(SizeType(index)) + " is out of bounds of axis " +
+          std::to_string(N) + " (max possible index is " + std::to_string(shape_[N] - 1) + ").");
+    }
     return static_cast<SizeType>(index) * stride_[N] +
            UnrollComputeColIndex<N + 1>(std::forward<Indices>(indices)...);
   }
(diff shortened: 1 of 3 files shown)


libs/math/include/math/tensor.hpp
```diff
@@ -459,13 +459,25 @@ class Tensor
   template <SizeType N, typename FirstIndex, typename... Indices>
   SizeType UnrollComputeColIndex(FirstIndex &&index, Indices &&... indices) const
   {
+    if (shape_[N] <= SizeType(index))
```
baykaner: pedantic but shape_.at(N) is safer

vhavrylov-ncube: sure!

libs/vm-modules/tests/unit/math/tensor_tests.cpp
```diff
@@ -135,42 +111,68 @@ TEST_F(MathTensorTests, tensor_4_dim_fixed64_fill)
...
 
-TEST_F(MathTensorTests, tensor_set_on_invalid_index)
+TEST_F(MathTensorTests, tensor_at_invalid_index_count_too_few)
 {
   static char const *SRC = R"(
-    function main() : Tensor
+    function main()
       var tensor_shape = Array<UInt64>(1);
       tensor_shape[0] = 2u64;
+      tensor_shape[1] = 2u64;
```
baykaner, line 154: i don't think this really tests anything about tensor though, you're just setting up an Array with one element and then trying to access its second element. That fails before you even try to construct a tensor

baykaner (changes requested):
looks like you have one irrelevant test (or at least a test that should live elsewhere), other than that it looks good

libs/vm-modules/tests/unit/math/tensor_tests.cpp
vhavrylov-ncube: Oh, no, my fault here: Array<UInt64>(2) was intended; 2-dim tensor with 1-dim accessor called.

baykaner (approved):

PR #2101, opened 2019-12-06
Feature/deprecate create wealth noremove

n-hutton:
A version of create wealth removal that still leaves the interface in for testing


n-hutton (commented):
Hold off on merge

PR #2102, opened 2019-12-06
Makes shutdown on segfault system parameter

troels-im:
This PR now allows constellation to be ran as 
```
./apps/constellation/constellation -block-interval 3000 -shutdown-on-fail
```
which will cause constellation to throw an exception when a segfault is discovered followed by a graceful shutdown.


troels-im:
Merged into #2104

PR #2103, opened 2019-12-06
Refactoring: Moving inline implementations to CPP files

troels-im:


troels-im:
Included in CI refactor.

PR #2113, opened 2019-12-09
VMModel::Fit, SerialiseTo and new Tensor operators estimations

MissingNO57:
Should be merged after #2083 and #2092


MissingNO57:
Separated Model Fit and SerialiseTo to new PR

PR #2114, opened 2019-12-09
Added metrics to etch VMModel::CompileSequential

evsmithx:
Added an optional list of metrics for etch model compilation, to match ML Model.


libs/ml/include/ml/model/model.hpp
```diff
@@ -490,6 +508,11 @@ void Model<TensorType>::TrainImplementation(DataType &loss, SizeType n_rounds)
   loss = loss_;
 }
 
+template <typename TensorType>
+bool Model<TensorType>::DataLoaderIsSet()
+{
+  return dataloader_ptr_->Size() != 0;
```
baykaner, line 518: is it possible for this to be called while dataloader_ptr_ is still a null_ptr?

libs/vm-modules/include/vm_modules/ml/model/model_estimator.hpp
```diff
@@ -239,6 +253,7 @@ class ModelEstimator
     // Model
     DataType forward_pass_cost{0.0};
     DataType backward_pass_cost{0.0};
+    DataType metrics_cost{0.0};
```
baykaner: all of these can now be string initialisations (maybe coordinate with Jiri for the nicest way to do this for t he static constexpr functions above)

libs/ml/include/ml/model/model.hpp
baykaner, line 518: can dataloader_ptr_ ever be a null ptr?

libs/vm-modules/include/vm_modules/ml/model/model_estimator.hpp
```diff
@@ -179,6 +185,14 @@ class ModelEstimator
   {
     return DataType(0.003333333333333);
   };
+  static constexpr DataType SCEL_FORWARD_IMPACT()
+  {
+    return DataType(0.003333333333333);
```
baykaner: Konstantinos' PR is in, so FixedPoint initialisations should all be strings

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -768,4 +768,44 @@ TEST_F(VMModelTests, DISABLED_regressor_model_test)
...
+          label_shape[1] = 1000u64;
+          var data = Tensor(data_shape);
+          var label = Tensor(label_shape);
+
+          // set up model
+          var model = Model("sequential");
+          model.add("dense", 10u64, 10u64, "relu");
+          model.add("dense", 10u64, 10u64, "relu");
+          model.add("dense", 10u64, 1u64);
+          model.compile("mse", "adam", {"mse"});
```
baykaner, line 791: nice 👍 

can you also demonstrate using accuracy?

libs/vm-modules/src/ml/model/model_estimator.cpp
```diff
@@ -216,20 +216,95 @@ ChargeAmount ModelEstimator::CompileSequential(Ptr<String> const &loss,
          COMPUTE_CHARGE_COST;
 }
 
+ChargeAmount ModelEstimator::CompileSequentialWithMetrics(
```
baykaner, line 259: you also need to add a couple of estimator tests (see vm-modules/test/unit/model/model_estimator); these should be sanity checks that just ensure the model estimator does what we expect it to over a range of input values

baykaner (commented):
very nice; a few extra bits needed and then we can merge this in

libs/ml/include/ml/model/model.hpp
evsmithx: Good catch!

libs/vm-modules/include/vm_modules/ml/model/model_estimator.hpp
evsmithx: Jiri has done these in his PR

vhavrylov-ncube (commented):
Great job!

baykaner (approved):

PR #2116, opened 2019-12-09
VMTensor Transpose and Reshape improvements, estimation and estimator tests

vhavrylov-ncube:
+ added unit tests for VMTensor Transpose for good and bad cases

### DIFF
--- a/libs/vm-modules/src/math/tensor/tensor.cpp
+++ b/libs/vm-modules/src/math/tensor/tensor.cpp
@@ -223,6 +223,19 @@ Ptr<VMTensor> VMTensor::Unsqueeze() const
 
 bool VMTensor::Reshape(Ptr<Array<SizeType>> const &new_shape)
 {
+  if (new_shape->elements.empty())
+  {
+    RuntimeError("Can not reshape a Tensor : new shape is empty!");
+    return false;
+  }
+  for (SizeType axis_size : new_shape->elements)
+  {
+    if (axis_size == 0)
+    {
+      RuntimeError("Can not reshape a Tensor : axis of size 0 found in new shape!");
+      return false;
+    }
+  }
   return tensor_.Reshape(new_shape->elements);
 }
 
--- a/libs/vm-modules/src/math/tensor/tensor.cpp
+++ b/libs/vm-modules/src/math/tensor/tensor.cpp
@@ -223,11 +223,7 @@ Ptr<VMTensor> VMTensor::Unsqueeze() const
 
 bool VMTensor::Reshape(Ptr<Array<SizeType>> const &new_shape)
 {
-  if (new_shape->elements.empty())
-  {
-    RuntimeError("Can not reshape a Tensor : new shape is empty!");
-    return false;
-  }
+  std::size_t total_new_elements = 0;
   for (SizeType axis_size : new_shape->elements)
   {
     if (axis_size == 0)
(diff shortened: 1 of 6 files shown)


libs/vm-modules/src/math/tensor/tensor.cpp
```diff
@@ -195,12 +195,31 @@ Ptr<VMTensor> VMTensor::Unsqueeze()
...
+    return false;
+  }
+  for (SizeType axis_size : new_shape->elements)
+  {
+    if (axis_size == 0)
+    {
+      RuntimeError("Can not reshape a Tensor : axis of size 0 found in new shape!");
+      return false;
+    }
+  }
```
baykaner, line 247: you don't check that number of elements in new_shape are the same as in old_shape. As Ira pointed out, this would be necessary to match numpy/tf behaviour - and we should probably match their behaviour if it's easy to do

libs/vm-modules/src/math/tensor/tensor_estimator.cpp
```diff
@@ -214,7 +214,7 @@ ChargeAmount TensorEstimator::Transpose()
 {
   if (tensor_.shape().size() != 2)
   {
-    return MaximumCharge("Cannot squeeze tensor, no dimensions of size 1");
+    return MaximumCharge("Cannot transpose tensor, only two-dimensional one can be transposed.");
```
baykaner:
```suggestion
    return MaximumCharge("Cannot transpose tensor, only two-dimensional tensor can be transposed.");
```

libs/vm-modules/tests/unit/math/tensor_tests.cpp
```diff
@@ -468,16 +465,195 @@ TEST_F(MathTensorTests, tensor_sum_etch_test)
...
+
+  auto const transposed = vm_tensor.Transpose()->GetTensor();
+
+  DataType const result = transposed.At<fetch::math::SizeType, fetch::math::SizeType>(1, 0);
+  DataType const expected{2.2};
+
+  EXPECT_EQ(tensor.shape().at(0), transposed.shape().at(1));
+  EXPECT_EQ(tensor.shape().at(1), transposed.shape().at(0));
+
+  EXPECT_TRUE(fetch::math::Abs(expected - result) < DataType::TOLERANCE);
```
baykaner: after konstantinos' parser gets merged in this test can be replaced with EXPECT_TRUE(expected == result)

baykaner (changes requested):

libs/vm-modules/src/math/tensor/tensor.cpp
vhavrylov-ncube: well, here I just checked an obviously bad cases, but did not change the existing "good" behaviour. I agree with Ira that silent dropping of a tensor content while reshaping is not the most clear and user-friendly approach; however, I don't consider myself as enough skilled ML engineer to decide what the correct behavior is :) I would suggest discussing that with ML team and/or requirement management.

vhavrylov-ncube: UPD: VMTensor behavior could safely be changed without affecting underlying math::Tensor implementation, so I added element count equality check here.

libs/vm-modules/tests/unit/math/tensor_tests.cpp
vhavrylov-ncube: I have replaced this assertion with EXPECT_EQ and it works in all cases but 14273.000000000, so it needs additional investigation.

vhavrylov-ncube: UPD: here is ok to use "==" but still we have cases where delta-based comparison is necessary.

baykaner (approved):

PR #2127, opened 2019-12-10
Added FP64 for LFG generator and refactored weight init

MissingNO57:
Should be merged after #2137


libs/core/include/core/random/lfg.hpp
```diff
@@ -142,6 +163,23 @@ class LaggedFibonacciGenerator
...
   static constexpr double inv_double_max_ =
       1. / static_cast<double>(std::numeric_limits<RandomType>::max());
+
+  static const fp128_t fp128_max_;
+  static const fp128_t FP128_ONE;
+  static const fp128_t FP128_TWO;
 };
+
+template <uint64_t P, uint64_t Q>
+const fetch::fixed_point::fp128_t LaggedFibonacciGenerator<P, Q>::fp128_max_ =
```
baykaner: since you use each of these constants only once i'm not sure it make much sense to define them like this: why not just directly use static_cast<fp128_t>(1)?

baykaner (dismissed):

libs/ml/include/ml/ops/weights.hpp
```diff
@@ -121,42 +121,56 @@ class Weights : public fetch::ml::ops::Variable<T>
     {
     case WeightsInitialisation::ZEROS:
     {
-      array.Fill(typename TensorType::Type(0));
+      array.Fill(static_cast<DataType>(0));
```
vhavrylov-ncube, line 124: Would "DataType{0}" be better here?

baykaner (approved):

PR #2129, opened 2019-12-10
Feature/ml 289 flatten

iryna-melnyk:


vhavrylov-ncube (commented):
Looks good. Please add a unit test to ml_model_tests.cpp that will use this function from Etch side: this way you can check that the script compiles and runs with expected results. I would suggest at least 3 tests: adding with proper parameter (must compile), adding with invalid parameter (must not compile), flattening a 2-dimensional previous layer and checking if the model provides correct output (must compile, run and return expected result)

PR #2131, opened 2019-12-10
tensor dot, argmax, and copy

baykaner:
implemented tensor dot, argmax, and copy + tests + initial estimators

### DIFF
--- a/libs/vm-modules/src/math/tensor/tensor.cpp
+++ b/libs/vm-modules/src/math/tensor/tensor.cpp
@@ -410,8 +410,8 @@ vm::Ptr<VMTensor> VMTensor::ArgMax(SizeType const &indices)
 
 vm::Ptr<VMTensor> VMTensor::Dot(vm::Ptr<VMTensor> const &other)
 {
-  auto ret_tensor = fetch::math::Dot(GetTensor(), other->GetTensor());
-  Ptr<VMTensor> ret = Ptr<VMTensor>{new VMTensor(this->vm_, this->type_id_, ret_tensor)};
+  auto          ret_tensor = fetch::math::Dot(GetTensor(), other->GetTensor());
+  Ptr<VMTensor> ret        = Ptr<VMTensor>{new VMTensor(this->vm_, this->type_id_, ret_tensor)};
   return ret;
 }
 
--- a/libs/vm-modules/src/math/tensor/tensor.cpp
+++ b/libs/vm-modules/src/math/tensor/tensor.cpp
@@ -168,8 +168,7 @@ vm::Ptr<vm::Array<SizeType>> VMTensor::VMShape() const
 
   for (std::size_t i = 0; i < tensor_.shape().size(); ++i)
   {
-    auto val              = tensor_.shape().at(i);
-    array->elements.at(i) = static_cast<SizeType>(val);
+    array->elements.at(i) = static_cast<SizeType>(tensor_.shape().at(i));
   }
 
   return array;
--- a/libs/vm-modules/src/math/tensor/tensor.cpp
+++ b/libs/vm-modules/src/math/tensor/tensor.cpp
@@ -88,6 +88,8 @@ void VMTensor::Bind(Module &module, bool const enable_experimental)
           .CreateMemberFunction("setAt", &VMTensor::SetAt<Index, Index, Index, Index, DataType>,
                                 UseEstimator(&TensorEstimator::SetAtFour))
           .CreateMemberFunction("size", &VMTensor::size, UseEstimator(&TensorEstimator::size))
+          .CreateMemberFunction("shape", &VMTensor::VMShape,
+                                UseEstimator(&TensorEstimator::VMShape))
           .CreateMemberFunction("fill", &VMTensor::Fill, UseEstimator(&TensorEstimator::Fill))
           .CreateMemberFunction("fillRandom", &VMTensor::FillRandom,
                                 UseEstimator(&TensorEstimator::FillRandom))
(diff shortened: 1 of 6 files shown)


libs/vm-modules/tests/unit/math/tensor_tests.cpp
```diff
@@ -355,6 +356,34 @@ TEST_F(MathTensorTests, tensor_set_from_string)
...
+  ASSERT_TRUE(toolkit.Compile(tensor_from_string_src));
+  Variant res;
+  ASSERT_TRUE(toolkit.Run(&res));
+
+  auto const                    tensor = res.Get<Ptr<fetch::vm_modules::math::VMTensor>>();
+  fetch::math::Tensor<DataType> gt({4});
+  gt.Fill(static_cast<DataType>(2.0));
+
+  EXPECT_TRUE(gt.AllClose(tensor->GetTensor()));
+}
```
evsmithx, line 421: I assume that x and y are deep copies and not references to the same object - it would be good to have a test for this.

libs/vm-modules/tests/unit/math/tensor_tests.cpp
```diff
@@ -972,6 +1001,92 @@ TEST_F(MathTensorTests, tensor_sum_etch_test)
...
+  ASSERT_TRUE(toolkit.Compile(tensor_dot_src));
+  Variant res;
+  ASSERT_TRUE(toolkit.Run(&res));
+
+  auto const                    tensor = res.Get<Ptr<fetch::vm_modules::math::VMTensor>>();
+  fetch::math::Tensor<DataType> gt({2, 2});
+  gt.Fill(static_cast<DataType>(2.0));
+
+  std::cout << "gt.ToString(): " << gt.ToString() << std::endl;
+  std::cout << "tensor->GetTensor().ToString(): " << tensor->GetTensor().ToString() << std::endl;
```
evsmithx: Debugging code

evsmithx: Also this test could be slightly more complex!

evsmithx (dismissed):

libs/vm-modules/src/math/tensor/tensor.cpp
```diff
@@ -153,6 +161,20 @@ SizeType VMTensor::size() const
...
 }
 
+vm::Ptr<vm::Array<SizeType>> VMTensor::VMShape() const
+{
+  auto array = this->vm_->CreateNewObject<Array<SizeType>>(
+      this->vm_->GetTypeId<SizeType>(), static_cast<int32_t>(tensor_.shape().size()));
+
+  for (std::size_t i = 0; i < tensor_.shape().size(); ++i)
+  {
+    auto val              = tensor_.shape().at(i);
```
vhavrylov-ncube: auto const?

libs/vm-modules/src/math/tensor/tensor_estimator.cpp
```diff
@@ -164,6 +179,36 @@ ChargeAmount TensorEstimator::Sum()
...
+
+ChargeAmount TensorEstimator::Dot(vm::Ptr<VMTensor> const & /*other*/)
+{
+  SizeType padded_size = fetch::math::Tensor<DataType>::PaddedSizeFromShape(tensor_.shape());
+  SizeType size        = fetch::math::Tensor<DataType>::SizeFromShape(tensor_.shape());
+
+  return static_cast<ChargeAmount>(DEFAULT_PADDED_SIZE_COEF * padded_size +
+                                   DEFAULT_SIZE_COEF * size + DEFAULT_CONST_COEF) *
+         COMPUTE_CHARGE_COST;
+}
```
vhavrylov-ncube, line 225: these three methods above and .Copy() share identical code, so it would be great to just have 1 and and call it.

libs/vm-modules/tests/unit/math/tensor_tests.cpp
```diff
@@ -355,6 +357,69 @@ TEST_F(MathTensorTests, tensor_set_from_string)
...
+  )";
+
+  ASSERT_TRUE(toolkit.Compile(tensor_from_string_src));
+  Variant res;
+  ASSERT_TRUE(toolkit.Run(&res));
+
+  auto tensor_shape{res.Get<Ptr<IArray>>()};
+
+  std::vector<SizeType> ret;
+  std::vector<SizeType> gt({2, 3, 4});
```
vhavrylov-ncube: std::vector<SizeType> const expected({2, 3, 4})   would be clearer

libs/vm-modules/tests/unit/math/tensor_tests.cpp
```diff
@@ -355,6 +357,69 @@ TEST_F(MathTensorTests, tensor_set_from_string)
...
+
+  while (tensor_shape->Count() > 0)
+  {
+    ret.emplace_back(tensor_shape->PopFrontOne().Get<uint64_t>());
+  }
+
+  EXPECT_TRUE(gt == ret);
```
vhavrylov-ncube: EXPECT_EQ (if fails) returns both arguments' names and their values, while EXPECT_TRUE just notifies that "it is false", so I would prefer  
 EXPECT_EQ(gt, ret);

vhavrylov-ncube (dismissed):

libs/vm-modules/src/math/tensor/tensor_estimator.cpp
baykaner: it probably won't stay that way

vhavrylov-ncube (dismissed):

vhavrylov-ncube (approved):
LGTM

PR #2132, opened 2019-12-10
Thread Sanity for new networking

troels-im:
This PR fixes a number of issues related to thread sanity in the new networking code and introduces a debug mutex to detect simple deadlocks.


libs/core/include/core/mutex.hpp
```diff
@@ -18,10 +18,93 @@
...
+void RegisterMutexAcquisition(DebugMutex *mutex, std::thread::id thread);
+void UnregisterMutexAcquisition(DebugMutex *mutex, std::thread::id thread);
+void FindDeadlock(DebugMutex *mutex, std::thread::id thread);
+
+class DebugMutex
+{
+public:
+  DebugMutex()
+    : filename_{"unknown"}
+    , line_{0}
```
denis-trapeznikov: Better as default initializers.

libs/core/include/core/mutex.hpp
```diff
@@ -18,10 +18,93 @@
...
+  {}
+
+  DebugMutex(std::string const &filename, int32_t line)
```
denis-trapeznikov: Nitpicking: `filename` by value.

libs/core/src/mutex.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+void FindDeadlock(DebugMutex *mutex, std::thread::id thread)
+{
+  mutex_register.FindDeadlock(mutex, thread);
+}
+
+void Mutexregister::RegisterMutexAcquisition(DebugMutex *mutex, std::thread::id thread)
+{
+  std::lock_guard<std::mutex> guard(mutex_);
```
denis-trapeznikov: `FETCH_LOCK`?

libs/core/include/core/mutex.hpp
```diff
@@ -18,10 +18,93 @@
...
+  };
+
+private:
+  std::mutex  mutex_;
+  std::string filename_{};
+  int32_t     line_{0};
+};
+
+#ifndef NDEBUG
+using Mutex = DebugMutex;
```
denis-trapeznikov, line 164: It is not immediately clear how could the two be possibly interchangeable. `DebugMutex`'s most useful ctor accepts filename and line, while `std::mutex` has none of such. Moreover, wouldn't it be more informative to store the location of lock acquistion, not the mutex construction itself, as is done in `FETCH_LOCK` (`FETCH_LOCK` itself could be modified to make good use of `DebugMutex`).

libs/muddle/src/kademlia/peer_tracker.cpp
```diff
@@ -407,41 +418,54 @@ void PeerTracker::PullPeerKnowledge()
...
     {
-      continue;
-    }
+      FETCH_LOCK(pull_mutex_);
+      address = peer_pull_queue_.front();
 
-    auto search_for = peer_pull_map_[address];
+      if (address.size() != Packet::ADDRESS_SIZE)
+      {
+        continue;
```
denis-trapeznikov, line 460: Is it a dead loop? If address is invalid, it is not popped, so considered over and over again.

libs/muddle/src/kademlia/peer_tracker.cpp
```diff
@@ -407,41 +418,54 @@ void PeerTracker::PullPeerKnowledge()
...
+    // It is important that no lock is held when this is called
     auto promise = rpc_client_.CallSpecificAddress(address, RPC_MUDDLE_KADEMLIA,
                                                    PeerTrackerProtocol::FIND_PEERS, search_for);
 
     // wrap the promise is a task
 
     auto call_task = std::make_shared<PromiseTask>(
-        promise, tracker_configuration_.promise_timeout,
+        promise, cfg.promise_timeout,
         [this, address, search_for, pull_id](service::Promise const &promise) {
```
denis-trapeznikov: NP: `address ` and `search_for` should rather be `move`d.

libs/muddle/src/kademlia/peer_tracker.cpp
```diff
@@ -587,27 +608,35 @@ void PeerTracker::OnResolvedPull(uint64_t pull_id, Address const &peer, Address
                     " code: ", int(promise->state()));
 
     // In case of failure, we stop following
-    if (peer_pull_map_.find(peer) != peer_pull_map_.end())
     {
-      peer_pull_map_.erase(peer);
+      FETCH_LOCK(pull_mutex_);
+      if (peer_pull_map_.find(peer) != peer_pull_map_.end())
```
denis-trapeznikov, line 631: An excess `if`.

libs/muddle/src/kademlia/peer_tracker.cpp
```diff
@@ -653,26 +682,28 @@ void PeerTracker::ConnectToDesiredPeers()
...
-          break;
-        }
+          // Breaking if we never did a pull before
+          if (it == last_pull_from_peer_.end())
+          {
+            break;
+          }
 
+          // Breaking if it is sufficiently long ago
+          if (ToSeconds(Clock::now() - it->second) > 300)  // TODO(tfr): Fix parameter
```
denis-trapeznikov, line 720: NP: `Clock::now()`  should probably be evaluated once at loop's start, should be enough for one second-resolution.

libs/muddle/src/kademlia/peer_tracker.cpp
```diff
@@ -753,50 +790,60 @@ void PeerTracker::Periodically()
...
     ConnectToDesiredPeers();
   }
+
   // Ensuring that we keep connections open which we are currently
   // pulling data from
-  for (auto const &item : uri_resolution_tasks_)
   {
-    if (item.first != own_address_)
+    FETCH_LOCK(uri_mutex_);
+    FETCH_LOCK(direct_mutex_);
```
denis-trapeznikov, line 832: We should introduce a form to lock several mutexes all at once in natural order, same as `std::scoped_lock` does.

libs/muddle/src/kademlia/table.cpp
```diff
@@ -51,16 +52,18 @@ KademliaTable::Peers KademliaTable::FindPeerInternal(KademliaAddress const &kam_
 
   // Creating initial list
   std::unordered_map<Address, PeerInfo> results;
-  for (auto &p : by_logarithm_[log_id].peers)
   {
-    results.insert({p->address, *p});
+    FETCH_LOCK(peer_info_mutex_);
+    for (auto &p : by_logarithm_[log_id].peers)
+    {
+      results.insert({p->address, *p});
```
denis-trapeznikov, line 59: `emplace`.

libs/muddle/src/kademlia/table.cpp
```diff
@@ -154,16 +155,19 @@ KademliaTable::Peers KademliaTable::FindPeerByHammingInternal(KademliaAddress co
...
   std::unordered_map<Address, PeerInfo> results;
-  for (auto &p : by_hamming_[hamming_id].peers)
   {
-    results.insert({p->address, *p});
-  }
+    FETCH_LOCK(peer_info_mutex_);
 
+    for (auto &p : by_hamming_[hamming_id].peers)
+    {
+      results.insert({p->address, *p});
```
denis-trapeznikov, line 163: `emplace`.

libs/muddle/src/kademlia/table.cpp
```diff
@@ -175,32 +179,33 @@ KademliaTable::Peers KademliaTable::FindPeerByHammingInternal(KademliaAddress co
     if (scan_left && (left != 0))
     {
       left -= 1;
+      FETCH_LOCK(peer_info_mutex_);
       for (auto const &p : by_hamming_[left].peers)
       {
         results.insert({p->address, *p});
```
denis-trapeznikov, line 185: Again `emplace`.

libs/muddle/src/kademlia/table.cpp
```diff
@@ -481,14 +504,12 @@ std::size_t KademliaTable::active_buckets() const
   return ret;
 }
 
-uint64_t KademliaTable::first_non_empty_bucket() const
-{
-  return first_non_empty_bucket_;
-}
-
 void KademliaTable::SetCacheFile(std::string const &filename, bool load)
```
denis-trapeznikov, line 507: NP: `filename` by value.

libs/muddle/src/router.cpp
```diff
@@ -590,9 +593,17 @@ void Router::SendToConnection(Handle handle, PacketPtr const &packet, bool exter
...
+  auto                  success = [this, wpacket]() {
+    auto packet = wpacket.lock();
+    if (packet)
+    {
+      ClearDeliveryAttempt(packet);
+    }
+  };
+  auto fail = [this, wpacket, external, reschedule_on_fail]() {
+    auto packet = wpacket.lock();
+    if (packet && reschedule_on_fail)
```
denis-trapeznikov, line 607: `if (reschedule_on_fail) { auto packet = wpacket.lock(); if (packet)` ?

denis-trapeznikov (changes requested):

libs/muddle/src/kademlia/peer_tracker.cpp
troels-im: Thanks. That is a really good observation.

libs/core/include/core/mutex.hpp
troels-im: Yes, you are right.

libs/core/include/core/mutex.hpp
troels-im: Ctor is gone now

libs/core/src/mutex.cpp
troels-im: `FETCH_LOCK` cannot be used here as that would lead to infinite recursion.

libs/core/include/core/mutex.hpp
troels-im: They are not interchangeable in the most general case. `DebugMutex` provides a subset of `std::mutex` capabilities, so as long as the code compiles and runs with `DebugMutex`, it is guaranteed that it will with `std::mutex`

troels-im:
Replaced by #2170

PR #2135, opened 2019-12-11
Feature/increased consensus testing

n-hutton:
Unit test consensus.cpp. Still needs to test the weight is correct, and everything meets threshold requirements.


libs/ledger/src/chain/block.cpp
```diff
@@ -87,7 +87,7 @@ void Block::UpdateDigest()
   // Generate hash stream
   serializers::MsgPackSerializer buf;
   buf << previous_hash << merkle_hash << block_number << miner << log2_num_lanes << timestamp
-      << tx_merkle_tree.root();
+      << tx_merkle_tree.root() << block_entropy;
```
n-hutton: Needed so that block entropy is part of hash + signature (avoid man in the middle attack on your blocks)

libs/ledger/tests/consensus/pos_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+TEST_F(ConsensusTests, test_not_an_aeon)
+{
+  BlockPtr block = ValidNthBlock(1);
+  block->block_entropy.confirmations.clear();
+
+  ASSERT_EQ(consensus_->ValidBlock(*block), ledger::ConsensusInterface::Status::NO);
+}
+
+TEST_F(ConsensusTests, DISABLED_test_zero_weight)
```
n-hutton: Will enable once I get Jenny's fixes in.

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -505,38 +504,21 @@ BlockCoordinator::State BlockCoordinator::OnPreExecBlockValidation()
...
-    if (expected_block_number != current_block_->block_number)
-    {
-      FETCH_LOG_WARN(LOGGING_NAME, "Block validation failed: Block number mismatch. Expected: ",
-                     expected_block_number, " Actual: ", current_block_->block_number, " (0x",
-                     current_block_->hash.ToHex(), ')');
-
-      RemoveBlock(current_block_->hash);
-      return State::RESET;
-    }
+    consensus_->UpdateCurrentBlock(*previous);
```
jinmannwong, line 559: No need to check anymore that previous is not null?

libs/ledger/tests/consensus/pos_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+  // return a valid first block after genesis (entropy not
+  // populated fully)
+  BlockPtr ValidNthBlock(uint64_t desired_block_number)
+  {
+    BlockPtr ret     = std::make_shared<Block>();
+    BlockPtr genesis = MainChain::CreateGenesisBlock();
+
+    ret->block_number               = desired_block_number;
+    ret->block_entropy.block_number = ret->block_number;
+    ret->block_entropy.block_number = ret->block_number;
```
jinmannwong: Block number in entropy duplicated.

libs/ledger/tests/consensus/pos_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+  ASSERT_EQ(consensus_->ValidBlock(*block), ledger::ConsensusInterface::Status::NO);
+}
+
+TEST_F(ConsensusTests, DISABLED_test_wrong_weight)
+{
+  BlockPtr block = ValidNthBlock(1);
+  block->weight++;
+
+  ASSERT_EQ(consensus_->ValidBlock(*block), ledger::ConsensusInterface::Status::NO);
+}
```
jinmannwong, line 235: I'm not sure weight is tested in ValidBlock.

libs/ledger/tests/consensus/pos_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+TEST_F(ConsensusTests, test_wrong_miner_identity)
+{
+  BlockPtr block  = ValidNthBlock(1);
+  block->miner_id = qual_[1];
+
+  ASSERT_EQ(consensus_->ValidBlock(*block), ledger::ConsensusInterface::Status::NO);
```
jinmannwong: Is this to test incorrect signature? Because I think it also fails in invalid weight.

libs/ledger/tests/consensus/pos_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+}
+
+TEST_F(ConsensusTests, test_timestamp_too_early)
+{
+  BlockPtr block = ValidNthBlock(1);
+  block          = ValidNthBlock(2);
+
+  ASSERT_EQ(consensus_->ValidBlock(*block), ledger::ConsensusInterface::Status::NO);
```
jinmannwong, line 254: I am confused here, is this not a loose block?

libs/ledger/tests/consensus/pos_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+}
+
+TEST_F(ConsensusTests, test_timestamp_before_previous)
+{
+  BlockPtr block = ValidNthBlock(1);
+  block          = ValidNthBlock(2);
+
+  block->timestamp = 0;
+
+  ASSERT_EQ(consensus_->ValidBlock(*block), ledger::ConsensusInterface::Status::NO);
```
jinmannwong, line 264: Same here, doesn't block at block number 1 have to be added to consensus first?

libs/ledger/tests/consensus/pos_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+{
+  BlockPtr block = ValidNthBlock(1);
+
+  block->block_entropy.confirmations[ConstByteArray("a")];
+  block->block_entropy.confirmations[ConstByteArray("b")];
+
+  block->UpdateDigest();
+
+  ASSERT_EQ(consensus_->ValidBlock(*block), ledger::ConsensusInterface::Status::NO);
+}
```
jinmannwong, line 376: Why is the digest mismatching if it is updated after the changes to confirmations?

libs/ledger/tests/consensus/pos_tests.cpp
jinmannwong: Also, I am not sure this requires the GetBlockGenerationWeight bug fixes. You are testing whether the miner identity has the weight given by GetBlockGenerationWeight, not whether GetBlockGenerationWeight is outputting the right thing.

jinmannwong (dismissed):
Some comments on the preciseness of the tests.

libs/ledger/src/chain/block_coordinator.cpp
n-hutton: I put this check in the ValidBlock check
```
  if (current.hash.size() != DIGEST_LENGTH_BYTES ||
      current.previous_hash.size() != DIGEST_LENGTH_BYTES)
  {
    return Status::NO;
  }
```

libs/ledger/tests/consensus/pos_tests.cpp
n-hutton: good catch

libs/ledger/tests/consensus/pos_tests.cpp
n-hutton: Ahh, you're right. Good catch

libs/ledger/tests/consensus/pos_tests.cpp
n-hutton: Indeed it is not -only tested implicitly via the timings. I disabled it to pass CI but will enable it before merging to master (after your bugfix).

libs/ledger/tests/consensus/pos_tests.cpp
n-hutton: The ValidNthBlock adds them to the chain when you call it, this is a little confusing I realise now

libs/ledger/tests/consensus/pos_tests.cpp
n-hutton: as comment above

libs/ledger/tests/consensus/pos_tests.cpp
n-hutton: the digest will now be correct for this block, but the signature should be invalid - so I should rename this test. Consensus does not check UpdateDigest is called since this is done when the block is first received

libs/ledger/src/chain/block.cpp
```diff
@@ -86,8 +86,20 @@ void Block::UpdateDigest()
...
-      << tx_merkle_tree.root();
+
+  // clang-format off
+  buf << previous_hash;
+  buf << merkle_hash;
+  buf << block_number;
+  buf << miner;
+  buf << miner_id;
+  buf << log2_num_lanes;
+  buf << tx_merkle_tree.root();
```
n-hutton, line 97: Couldn't we just sort and serialize all TXs instead of constucting a merkle tree here?

ejfitzgerald (approved):

PR #2141, opened 2019-12-11
Feature/tensor fromstring fixedpoint

markos:
FromString() specialized for FixedPoint to use the parser
Also bug fixes with the FixedPoint parser


libs/math/include/math/tensor.hpp
```diff
@@ -35,6 +35,7 @@
 #include "math/tensor_iterator.hpp"
 #include "math/tensor_slice_iterator.hpp"
 #include "math/tensor_view.hpp"
+#include "vectorise/fixed_point/fixed_point.hpp"
```
baykaner: dont think you need this anymore?

baykaner (dismissed):

libs/math/include/math/tensor.hpp
baykaner: ah right, you can only ditch it in the follow up PR

baykaner (dismissed):

baykaner (dismissed):

baykaner (dismissed):

baykaner (approved):

PR #2144, opened 2019-12-11
Refactor/ml 286 datatype init with strings

vhavrylov-ncube:


libs/ml/tests/unit/layers/scaled_dot_product_attention.cpp
```diff
@@ -50,13 +50,13 @@ TYPED_TEST(ScaledDotProductAttention, input_output_dimension_check)  // Use the
...
+  query_data.Fill(fetch::math::Type<DataType>("0.1"));
   TypeParam key_data = TypeParam({4, 5, 2});
-  key_data.Fill(static_cast<DataType>(0.2));
+  key_data.Fill(fetch::math::Type<DataType>("0.2"));
   TypeParam value_data = TypeParam({3, 5, 2});
-  value_data.Fill(static_cast<DataType>(0.3));
+  value_data.Fill(fetch::math::Type<DataType>("0.3"));
   TypeParam mask_data = TypeParam({1, 7, 2});
-  mask_data.Fill(static_cast<DataType>(1));
+  mask_data.Fill(fetch::math::Type<DataType>("1"));
```
markos, line 59: below you have Datatype{1}, I think for consistency all integers should just be like this and avoid using string.

libs/ml/tests/unit/layers/scaled_dot_product_attention.cpp
```diff
@@ -295,17 +293,17 @@ TYPED_TEST(ScaledDotProductAttention, saveparams_test)
...
+  value_data.Fill(fetch::math::Type<DataType>("0.1"));
+  mask_data.Fill(fetch::math::Type<DataType>("1"));
 
   // create labels
   TypeParam labels({12, 25, 4});
   labels.FillUniformRandom();
 
   // Create layer
-  LayerType layer(key_dim, static_cast<DataType>(1));
+  LayerType layer(key_dim, fetch::math::Type<DataType>("1"));
```
markos: same, Datatype{1}

vhavrylov-ncube: corrected

libs/ml/tests/unit/layers/scaled_dot_product_attention.cpp
vhavrylov-ncube: In this code block above there are 3 more usage of DataType init, so I did it to keep consistent code.

evsmithx (dismissed):

libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -241,6 +241,7 @@ TEST(FixedPointTest, FromString_16_16)
   // Get raw value
   fp32_t one("1fp32");
   fp32_t one2("1.0fp32");
+  fp32_t m_one("-1fp32");
```
baykaner: i think we should just replace these with this:
fp32_t m_one("-1");

baykaner (dismissed):

baykaner (approved):

PR #2137, opened 2019-12-11
implemented universal DataType constructor as a math free function

baykaner:
implements helper function for constructing floats/fixed data types from string

PR #2149, opened 2019-12-11
Removing floats from the math library and tests

evsmithx:
math_unit_tests now runs successfully with the fixed point from float constructor marked private, i.e. no construction of an fp from a float is allowed outside the fp class. I haven't done the fixed_point tests because I don't understand them! This pull request depends on #2137 and #2127.


libs/ml/tests/unit/model/dnn_regressor_test.cpp
```diff
@@ -106,7 +106,7 @@ TYPED_TEST(DNNRegressorModelTest, adagrad_dnnregressor)
 {
   using DataType = typename TypeParam::Type;
   ASSERT_TRUE(regressor_details::RunTest<TypeParam>(fetch::ml::OptimiserType::ADAGRAD,
-                                                    DataType{1e-4f}, DataType{0.03f}, 400));
+                                                    DataType{1e-4f}, DataType{0.04f}, 400));
```
vhavrylov-ncube: Ditch scientific notation, "0.0001"

libs/ml/tests/unit/model/sequential_test.cpp
```diff
@@ -112,7 +112,7 @@ TYPED_TEST(SequentialModelTest, adagrad_sequential)
 {
   using DataType = typename TypeParam::Type;
   ASSERT_TRUE(sequential_details::RunTest<TypeParam>(fetch::ml::OptimiserType::ADAGRAD,
-                                                     DataType{1e-4f}, DataType{0.03f}, 400));
+                                                     DataType{1e-4f}, DataType{0.04f}, 400));
```
vhavrylov-ncube: Scientific notation

libs/math/tests/unit/distance/entropy_tests.cpp
```diff
@@ -37,10 +37,10 @@ TYPED_TEST(EntropyTest, entropy)
...
-  A.Set(SizeType{0}, DataType(0.1));
-  A.Set(SizeType{1}, DataType(0.2));
-  A.Set(SizeType{2}, DataType(0.3));
-  A.Set(SizeType{3}, DataType(0.4));
+  A.Set(SizeType{0}, fetch::math::Type<DataType>("0.1"));
+  A.Set(SizeType{1}, fetch::math::Type<DataType>("0.2"));
+  A.Set(SizeType{2}, fetch::math::Type<DataType>("0.3"));
+  A.Set(SizeType{3}, fetch::math::Type<DataType>("0.4"));
 
   EXPECT_NEAR(double(statistics::Entropy(A)), 1.84643934467102, 1e-4);
```
vhavrylov-ncube: (optional) As far as we use fixed point here, may be it can be EXPECT_EQ?

vhavrylov-ncube (dismissed):
Scientific notation artifacts left

libs/ml/tests/unit/model/dnn_classifier_test.cpp
```diff
@@ -120,7 +120,7 @@ TYPED_TEST(DNNClassifierModelTest, adagrad_dnnclasifier)
 {
   using DataType = typename TypeParam::Type;
   ASSERT_TRUE(classifier_details::RunTest<TypeParam>(fetch::ml::OptimiserType::ADAGRAD,
-                                                     DataType{1e-2f}, DataType{0.01f}, 400));
+                                                     DataType{1e-2f}, DataType{0.03f}, 400));
 }
```
vhavrylov-ncube: scientific notation

libs/math/tests/unit/activation_functions/sigmoid.cpp
```diff
@@ -144,23 +149,23 @@ TYPED_TEST(SigmoidTest, exact_values)
...
-  numpy_output.Set(SizeType{0}, SizeType{1}, typename TypeParam::Type(0.76852478));
-  numpy_output.Set(SizeType{1}, SizeType{0}, typename TypeParam::Type(0.66818777));
-  numpy_output.Set(SizeType{1}, SizeType{1}, typename TypeParam::Type(1));
-
-  ASSERT_TRUE(output.AllClose(numpy_output));
+  numpy_output.Set(SizeType{0}, SizeType{0}, fetch::math::Type<DataType>("0.57444252"));
+  numpy_output.Set(SizeType{0}, SizeType{1}, fetch::math::Type<DataType>("0.76852478"));
+  numpy_output.Set(SizeType{1}, SizeType{0}, fetch::math::Type<DataType>("0.66818777"));
+  numpy_output.Set(SizeType{1}, SizeType{1}, DataType{1});
+  ASSERT_TRUE(output.AllClose(numpy_output, fetch::math::function_tolerance<DataType>()));
```
baykaner, line 167: you've added the tolerance here - what tolerance was it using before?

libs/math/tests/unit/basic_math/standard_functions_test.cpp
```diff
@@ -76,26 +76,27 @@ TYPED_TEST(StandardFunctionTests, clamp_array_1D_test)
...
+  A_clamp_expected(3)        = DataType{2};
+  A_clamp_expected(4)        = DataType{3};
+  A_clamp_expected(5)        = DataType{3};
 
   // Compare results with expected results
-  Clamp(DataType(2), DataType(3), A);
+  Clamp(DataType{2}, DataType{3}, A);
 
-  ASSERT_TRUE(A.AllClose(A_clamp_expected, DataType{1e-5f}, DataType{1e-5f}));
+  ASSERT_TRUE(A.AllClose(A_clamp_expected, fetch::math::Type<DataType>("0.00001"),
```
baykaner: I think there shouldn't be any tolerance in this test. It should just be ExpectEq(A, A_clamp_expected);

libs/math/tests/unit/basic_math/standard_functions_test.cpp
```diff
@@ -105,26 +106,27 @@ TYPED_TEST(StandardFunctionTests, clamp_array_2D_test)
...
+  A_clamp_expected(1, 0)     = DataType{2};
+  A_clamp_expected(1, 1)     = DataType{3};
+  A_clamp_expected(1, 2)     = DataType{3};
 
   // Compare results with expected results
-  Clamp(DataType(2), DataType(3), A);
+  Clamp(DataType{2}, DataType{3}, A);
 
-  ASSERT_TRUE(A.AllClose(A_clamp_expected, DataType{1e-5f}, DataType{1e-5f}));
+  ASSERT_TRUE(A.AllClose(A_clamp_expected, fetch::math::Type<DataType>("0.00001"),
```
baykaner: same here; should be no tolerance

libs/math/tests/unit/distance/entropy_tests.cpp
baykaner: i think this is tested for floats as well though.
we can replace 1e-4 with function tolerance though, right?

libs/math/tests/unit/distance/perplexity.cpp
```diff
@@ -38,10 +38,10 @@ TYPED_TEST(PerplexityTest, entropy)
...
-  A.Set(SizeType{0}, DataType(0.1));
-  A.Set(SizeType{1}, DataType(0.2));
-  A.Set(SizeType{2}, DataType(0.3));
-  A.Set(SizeType{3}, DataType(0.4));
+  A.Set(SizeType{0}, fetch::math::Type<DataType>("0.1"));
+  A.Set(SizeType{1}, fetch::math::Type<DataType>("0.2"));
+  A.Set(SizeType{2}, fetch::math::Type<DataType>("0.3"));
+  A.Set(SizeType{3}, fetch::math::Type<DataType>("0.4"));
 
   EXPECT_NEAR(double(statistics::Perplexity(A)), 3.59611546662432, 1e-3);
```
baykaner: same issue of tolerance

libs/math/tests/unit/tensor/basic_tests.cpp
```diff
@@ -38,15 +38,15 @@ Tensor<fetch::fixed_point::FixedPoint<I, F>,
...
+  static fetch::random::Random gen;
   Tensor<fetch::fixed_point::FixedPoint<I, F>,
          fetch::memory::SharedArray<fetch::fixed_point::FixedPoint<I, F>>>
       a1(n);
 
   fetch::fixed_point::FixedPoint<I, F> rn{0};
   for (std::size_t i = 0; i < n; ++i)
   {
-    rn       = fetch::fixed_point::FixedPoint<I, F>(gen.AsDouble());
+    rn       = fetch::fixed_point::FixedPoint<I, F>(gen.generator.AsFP64());
```
baykaner: in the future we should probably rework this 'generator' since it uses the mersenne twister from the std library - in other words, it's a source of indeterminacy - however since it's only in a test and doesn't seem to ever cause failures it's not a priority right now

libs/ml/tests/unit/model/dnn_regressor_test.cpp
baykaner: these DO need string constructors! Type<DataType>("0.0001")

libs/ml/tests/unit/model/dnn_classifier_test.cpp
baykaner: and string construction

libs/vectorise/include/vectorise/meta/math_type_traits.hpp
```diff
@@ -90,8 +89,11 @@ using IfIsNotVectorRegister = fetch::meta::EnableIf<IsNotVectorRegister<T>, R>;
 template <typename DataType, typename ReturnType>
 using IfIsNotImplemented = fetch::meta::IfIsNotImplemented<DataType, ReturnType>;
 
-template <typename DataType>
-using IfIsUnsignedInteger = fetch::meta::IfIsUnsignedInteger<DataType>;
+// template <typename DataType>
```
baykaner: I've removed this in my PR subsequently - merge the PR in again to ensure this doesn't get added back in via auto merge

baykaner (dismissed):

libs/ml/tests/unit/model/dnn_regressor_test.cpp
evsmithx: The ML stuff shouldn't be in there! It's been pulled in from another PR.

evsmithx: To avoid duplicating work I'm inclined to fix these in my defloating ml PR. This PR (for math) will go in first so it won't overwrite the ml one.

libs/math/tests/unit/activation_functions/sigmoid.cpp
evsmithx: The default tolerance, "0.00001", which is too strict for FP<16,16>. I tried it without the tolerance and the other types passed

libs/math/tests/unit/tensor/basic_tests.cpp
evsmithx: It's the fetch RNG, which is the lagged fibonacci generator - I don't think it uses the std library?

MissingNO57 (approved):
LGTM

libs/ml/tests/unit/model/sequential_test.cpp
evsmithx: ML tests will be defloated in a separate PR

libs/ml/tests/unit/model/dnn_classifier_test.cpp
evsmithx: ML tests will be defloated in a separate PR

PR #2153, opened 2019-12-12
Iimproved startup

ejfitzgerald:


libs/chain/src/constants.cpp
```diff
@@ -17,24 +17,101 @@
...
+
+}  // namespace
+
+Digest GetGenesisDigest()
+{
+  Digest digest = genesis_state.Apply([](GenesisState const &state) { return state.digest; });
+
+  if (digest.empty())
+  {
+    throw std::logic_error("Genesis has not been initialised");
```
n-hutton, line 47: 👍

libs/constellation/src/constellation.cpp
```diff
@@ -91,6 +90,35 @@ constexpr char const *LOGGING_NAME = "constellation";
 const std::size_t HTTP_THREADS{4};
 char const *      GENESIS_FILENAME = "genesis_file.json";
 
+class Defer
+{
+public:
+  using Callback = void (Constellation::*)();
+
+  explicit Defer(Constellation *instance, Callback callback)
```
n-hutton, line 98: no need for explicit here

libs/constellation/src/constellation.cpp
```diff
@@ -338,69 +360,225 @@ Constellation::Constellation(CertificatePtr const &certificate, Config config)
...
+  // - recover from previous genesis init
+  {
+    std::string const genesis_file_path =
+        cfg_.genesis_file_location.empty() ? GENESIS_FILENAME : cfg_.genesis_file_location;
+
+    GenesisFileCreator creator(*storage_, external_identity_, cfg_.db_prefix);
+
+    genesis_status = creator.LoadFile(genesis_file_path, cfg_.proof_of_stake, params);
+
+    if (genesis_status == GenesisFileCreator::Result::FAILURE)
```
n-hutton, line 478: 👍

libs/constellation/src/constellation.cpp
```diff
@@ -338,69 +360,225 @@ Constellation::Constellation(CertificatePtr const &certificate, Config config)
...
+    {
+      FETCH_LOG_ERROR(LOGGING_NAME, "Failed to restore or generate Genesis configuration");
+      return false;
+    }
+  }
+
+  // create the DAG
```
n-hutton, line 485: Probably need to clear up the settings.cpp in another PR

libs/constellation/src/constellation.cpp
```diff
@@ -338,69 +360,225 @@ Constellation::Constellation(CertificatePtr const &certificate, Config config)
...
+                               external_identity_->identity());
+
+  if (cfg_.proof_of_stake)
+  {
+    consensus_->SetDefaultStartTime(params.start_time);
+    consensus_->SetMaxCabinetSize(params.cabinet_size);
+
+    if (params.snapshot)
+    {
+      consensus_->Reset(*params.snapshot);
```
n-hutton, line 536: I will probably have to change this in my wealth PR - the flow should be basically the same for POS

libs/constellation/src/constellation.cpp
```diff
@@ -691,50 +739,225 @@ bool Constellation::Run(UriSet const &initial_peers, core::WeakRunnable bootstra
...
+  lane_services_.StopExternal();
+
+  ResetItem(synergetic_miner_);
+  ResetItem(dag_service_);
+
+  http_modules_.clear();
+  ResetItem(health_check_module_);
+  ResetItem(http_open_api_module_);
+
+  ResetItem(messenger_api_);
```
n-hutton, line 801: Won't this happen anyway? Is this so you can do it in a particular order, or for easier debugging?

libs/ledger/include/ledger/storage_unit/fake_storage_unit.hpp
```diff
@@ -92,8 +92,8 @@ class FakeStorageUnit final : public StorageUnitInterface
   TransactionStore             transaction_store_{};
   StatePtr                     state_{std::make_shared<State>()};
   StateHistory                 state_history_{};
-  StateHashStack               state_history_stack_{fetch::chain::GENESIS_MERKLE_ROOT};
-  Hash                         current_hash_{fetch::chain::GENESIS_MERKLE_ROOT};
+  StateHashStack               state_history_stack_{fetch::chain::GetGenesisDigest()};
```
n-hutton, line 95: 👍

scripts/end_to_end_test/end_to_end_test.yaml
```diff
@@ -110,6 +110,7 @@ setup_conditions:
 steps:
   - send_txs: { "name": "batch1", "amount": 10, "nodes": [0] }
   - verify_txs: { "name": "batch1", "nodes": [0] }
+  - sleep: 20
```
n-hutton, line 113: Is this due to inceased start up time?

n-hutton (approved):

scripts/end_to_end_test/end_to_end_test.yaml
ejfitzgerald: Yes that is correct

libs/constellation/src/constellation.cpp
ejfitzgerald: Yes this is over kill but I want to have the shutdown exactly when when this function is called

libs/constellation/src/constellation.cpp
ejfitzgerald: Good spot - it was a single parameter function at one point. I think I will fix this on delta if you don't mind

PR #2155, opened 2019-12-12
Feature/add more tensor from string tests

markos:
Make Tensor::FromString more robust, add more tests

### DIFF
--- a/libs/math/include/math/tensor.hpp
+++ b/libs/math/include/math/tensor.hpp
@@ -752,7 +752,7 @@ Tensor<T, C> Tensor<T, C>::FromString(byte_array::ConstByteArray const &c)
         break;
       }
       prev_backslash = false;
-      [[gnu::fallthrough]];  // explicit fallthrough to the next case
+      FETCH_FALLTHROUGH;  // explicit fallthrough to the next case
     case '\r':
     case '\n':
       if (reached_actual_data)
--- a/libs/math/include/math/tensor.hpp
+++ b/libs/math/include/math/tensor.hpp
@@ -752,7 +752,7 @@ Tensor<T, C> Tensor<T, C>::FromString(byte_array::ConstByteArray const &c)
         break;
       }
       prev_backslash = false;
-      [[fallthrough]];  // explicit fallthrough to the next case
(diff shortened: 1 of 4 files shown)


libs/math/include/math/tensor.hpp
```diff
@@ -707,8 +707,10 @@ Tensor<T, C> Tensor<T, C>::FromString(byte_array::ConstByteArray const &c)
     UNSET,
     COLON,
     NEWLINE
-  } new_row_marker         = UNSET;
-  bool reached_actual_data = false;
+  } new_row_marker           = UNSET;
+  bool   reached_actual_data = false;
+  size_t first_row_size      = 0;
```
vhavrylov-ncube: std::size_t

libs/math/tests/unit/tensor/constructor.cpp
```diff
@@ -37,6 +37,144 @@ TYPED_TEST(TensorConstructorTest, string_construction)
...
   ASSERT_EQ(tensor.At(0, 2), TypeParam(4));
+
+  auto tensor1 = fetch::math::Tensor<TypeParam>::FromString(R"(
+  	1 3 4 4.1
+  	5 6 7 7.999
+  	8 9 10 11.11111
+  )");
+  ASSERT_EQ(tensor1.shape().at(0), 3);
+  ASSERT_EQ(tensor1.shape().at(1), 4);
+  ASSERT_EQ(tensor1.At(0, 0), math::Type<TypeParam>("1"));
```
vhavrylov-ncube: (optional) ASSERT_EQ will immediately kill the test if failed, but each comparison here and in next line is actually not so critical - if it fails, we still could check the rest of elements; so I would suggest EXPECT_EQ for element comparisons.

libs/math/include/math/tensor.hpp
markos: fixed

libs/math/tests/unit/tensor/constructor.cpp
markos: replaced with EXPECT

markos:
Closing because it's dealt in #2162

PR #2156, opened 2019-12-12
Test/ml 353 misuse of etch ml model

iryna-melnyk:


libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -848,4 +848,140 @@ TEST_F(VMModelTests, model_with_accuracy_metric)
...
+          model.fit(data, label, 32u64);
+
+          // evaluate
+          var mets = model.evaluate();
+          return mets;
+        endfunction
+      )";
+
+  ASSERT_TRUE(toolkit.Compile(SRC_METRIC));
+  ASSERT_FALSE(toolkit.Run());
```
baykaner, line 879: great! 👍  this behaviour is consistent with tensorflow

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -848,4 +848,140 @@ TEST_F(VMModelTests, model_with_accuracy_metric)
...
+}
+
+TEST_F(VMModelTests, model_sequential_multiple_compile)
+{
+  static char const *SEQUENTIAL_SRC = R"(
+        function main()
+          var model = Model("sequential");
+          model.add("dense", 10u64, 10u64, "relu");
+          model.compile("mse", "adam");
+          model.compile("scel", "adam");
```
baykaner, line 902: it would be good if we could verify somehow that model.compile is actually changing the loss function for us each time.

Also, I would suspect that model will be adding a new op each time we call compile, but maybe it doesn't delete the old loss op from the previous compile call.

Can you investigate these two points please?

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -848,4 +848,140 @@ TEST_F(VMModelTests, model_with_accuracy_metric)
...
+}
+
+TEST_F(VMModelTests, model_sequential_add_after_compile)
+{
+  static char const *SEQUENTIAL_SRC = R"(
+      function main()
+         var model = Model("sequential");
+         model.add("dense", 10u64, 10u64, "relu");
+         model.compile("mse", "adam");
+         model.add("dense", 10u64, 1u64, "relu");
```
baykaner, line 918: again - let's verify that after the second call to compile we have a 2 layer network, and that we have a 1 layer network before that

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -848,4 +848,140 @@ TEST_F(VMModelTests, model_with_accuracy_metric)
...
+
+        return mets;
+      endfunction
+    )";
+
+  ASSERT_TRUE(toolkit.Compile(SEQUENTIAL_SRC));
+  ASSERT_TRUE(toolkit.Run());
+}
+
+TEST_F(VMModelTests, DISABLED_model_sequential_predict_bad_data)
```
baykaner, line 964: lets enable this test but EXPECT_THROW:

seems reasonable to throw when you are passing a tensor of size 0 in for prediction

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -848,4 +848,140 @@ TEST_F(VMModelTests, model_with_accuracy_metric)
...
+        model.compile("mse", "adam", {"categorical accuracy"});
+
+        var prediction = model.predict(data);
+      endfunction
+    )";
+
+  ASSERT_TRUE(toolkit.Compile(SEQUENTIAL_SRC));
+  ASSERT_FALSE(toolkit.Run());
+}
+
```
baykaner, line 986: please add another test to check EXPECT_THROW if we compile and then evaluate (this should not be permitted without training first)

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: I've found a valid statement for this case - EXPECT_DEATH(toolkit.Run(), "1000");

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: Done

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: So, each time when model.compile(...) is called we add to graph an error node:
graph_ptr_->template AddNode<ops::CrossEntropyLoss<TensorType>>("Error", {output_, label_});)
when we try to add to graph a node with existing name it checks for duplicates and returns an exception:
ml::exceptions::InvalidMode( "OperationType is not shareable. Cannot make duplicate of node named: " + node_name);

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: True. When compile is called second time we really have a 2-layered network because node names are unique for each dense layer. But again, it throws an InvalidMode exception because of ambiguous node names ("Error" node already exists in the graph)

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
baykaner: so when is this actually failing? On the second Compile call? (with a different loss op "scel"), or on the third compile call (which reuses "mse")?

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
baykaner: hmmm.... okay, well I feel like the ideal behaviour would be have .add trigger a 'decompile' - which handles removing the node from the graph (but we have no infrastructure to support that currently) - so I think it would be a tricky job

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: On the second Compile call? - yes

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: I can create a ticket for that

iryna-melnyk:
look at https://github.com/fetchai/ledger/pull/2171

PR #2143, opened 2019-12-11
Enable charge estimators for object operators

wkaluza:

PR #2158, opened 2019-12-12
Feature/ml 344 enable etch tensor operators

baykaner:
- to follow #2143 
- will remove tensor.add(), tensor.subtract() etc. but enable operators for tensors (e.g. x + y)


libs/vm-modules/src/math/bignumber.cpp
```diff
@@ -406,6 +406,67 @@ bool UInt256Wrapper::IsGreaterThan(Ptr<Object> const &lhso, Ptr<Object> const &r
   return rhs->number_ < lhs->number_;
 }
 
+vm::ChargeAmount UInt256Wrapper::AddChargeEstimator(fetch::vm::Ptr<Object> const & /*lhso*/,
+                                                    fetch::vm::Ptr<Object> const & /*rhso*/)
+{
+  return 1;
```
vhavrylov-ncube: I would suggest a named constant private field instead, just for a tuning purpose later.

libs/vm-modules/tests/unit/vm_charge_tests.cpp
```diff
@@ -549,4 +634,262 @@ TEST_F(VmChargeTests,
...
+      var obj1 = CustomType();
+      -obj1;
+    endfunction
+  )";
+
+  ASSERT_TRUE(toolkit.Compile(TEXT));
+  ASSERT_TRUE(toolkit.Run(nullptr, high_charge_limit));
+}
+
+TEST_F(VmChargeTests, negate_operator_bind_with_charge_estimate_execution_fails_when_limit_exceeded)
```
vhavrylov-ncube: "fail" in unit test naming could be a trouble for later bugfixing ;)

libs/vm/include/vm/address.hpp
```diff
@@ -175,6 +175,42 @@ class Address : public Object
     return lhs->address_ >= rhs->address_;
   }
 
+  ChargeAmount IsEqualChargeEstimator(Ptr<Object> const & /*lhso*/,
+                                      Ptr<Object> const & /*rhso*/) override
+  {
+    return 1;
+  }
```
vhavrylov-ncube: the same here, a named constant or constants would be better.

libs/vm/src/object.cpp
```diff
@@ -115,143 +115,331 @@ std::size_t Object::GetHashCode()
...
 {
-  RuntimeError(std::string(__func__) + "operator not implemented");
+  RuntimeError(std::string(__func__) + ": operator not implemented");
   return false;
 }
 
 bool Object::IsNotEqual(Ptr<Object> const & /* lhso */, Ptr<Object> const & /* rhso */)
 {
-  RuntimeError(std::string(__func__) + "operator not implemented");
+  RuntimeError(std::string(__func__) + ": operator not implemented");
```
vhavrylov-ncube: " :   .... ", a space right after " and before : would be neater.
And to simplify such corrections let the phrase be a 

static constexpr char * OP_NOT_IMPL = " .... " ;

used across all functions.

libs/vm/src/object.cpp
```diff
@@ -115,143 +115,331 @@ std::size_t Object::GetHashCode()
...
+}
+
+ChargeAmount Object::IsEqualChargeEstimator(Ptr<Object> const & /* lhso */,
+                                            Ptr<Object> const & /* rhso */)
+{
+  RuntimeError(std::string(__func__) + ": estimator not implemented");
+  return 1;
```
vhavrylov-ncube: static constexpr char* ESTIMATOR_NOT_IMPL = " : estimator not implemented."

ejfitzgerald (dismissed):

wkaluza (approved):

PR #2160, opened 2019-12-12
Feature/ml 355 fixedpoint string parser reject alpha

markos:
FixedPoint string parser should reject all alpha characters


libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
```diff
@@ -958,13 +958,16 @@ template <uint16_t I, uint16_t F>
...
   auto index  = s.find("fp");
   auto s_copy = std::string(s, 0, index);
 
+  bool contains_alpha = std::find_if(s_copy.begin(), s_copy.end(),
+                                     [](char c) { return !std::isalpha(c); }) != s_copy.end();
+  if (contains_alpha)
+  {
+    throw std::runtime_error("FixedPoint parsing from string does not allow alpha characters!");
+  }
+
```
vhavrylov-ncube, line 970: As far as I understood, this change will lead to runtime crash of all the tests that still could use scientific notation in FixedPoint initialization, right?

markos: unless they catch yes, exception should be handled outside of constructor

vhavrylov-ncube (dismissed):

baykaner (approved):

PR #2161, opened 2019-12-12
Feature/ml 359 deepcopy uint256 fixed128

vhavrylov-ncube:
+ added ".copy()" etch methods for deep copying of UInt256 and Fixed128 types.


libs/vm-modules/tests/unit/fixed_point.cpp
```diff
@@ -237,4 +237,43 @@ TEST_F(FixedPointTest, multidigit_integral_fixed_point128)
...
+      function main()
+        var a = 9876.54321000fp128;
+        var b = 9876.54321000fp128;
+
+        a = b.copy();
+        assert(a == b, "deep copy failed!");
+
+        a += 1.0fp128;
+
+        assert(a > b, "deep copy failed!");
```
cyenyxe: For consistency with the other test (whose message I understand more clearly):
```suggestion
        assert(a > b, "b is corrupted by increasing a!");
```

libs/vm/src/module.cpp
```diff
@@ -390,15 +390,11 @@ Module::Module()
...
       .CreateMemberFunction("set", &IShardedState::SetFromAddress);
 
-  GetClassInterface<Fixed128>().CreateSerializeDefaultConstructor(
-      [](VM *vm, TypeId) -> Ptr<Fixed128> {
+  GetClassInterface<Fixed128>()
+      .CreateSerializeDefaultConstructor([](VM *vm, TypeId) -> Ptr<Fixed128> {
         return Ptr<Fixed128>{new Fixed128(vm, fixed_point::fp128_t::_0)};
-      });
-
-  // GetClassInterface<UInt256Wrapper>().CreateSerializeDefaultConstructor(
```
cyenyxe, line 361: Why is there not one for UInt256?

vhavrylov-ncube: It is done in ledger/libs/vm-modules/src/math/bignumber.cpp:68

baykaner (approved):

PR #2165, opened 2019-12-12
Feature/deprecate wealth2

n-hutton:
This PR removes the ability for transactions to arbitrarily create wealth. Nodes must be started with a genesis file that specifies wealth at one or more addresses. For the end to end tests, this is a 'benefactor account', which is transferred from, where create wealth was previously used.


libs/ledger/src/fees/fee_manager.cpp
```diff
@@ -92,14 +84,12 @@ bool FeeManager::CalculateChargeAndValidate(TransactionDetails &             tx,
...
-  if (!tx.is_create_wealth)
-  {
-    result.charge += scaled_charge;
-  }
+  result.charge += scaled_charge;
 
   // determine if the chain code ran out of charge
   if (result.charge > tx.charge_limit)
   {
+    result.charge = scaled_charge;
```
ejfitzgerald: This is not needed

libs/ledger/include/ledger/protocols/main_chain_rpc_service.hpp
```diff
@@ -93,6 +93,8 @@ class MainChainRpcService : public muddle::rpc::Server,
   MainChainRpcService(MainChainRpcService &&)      = delete;
   ~MainChainRpcService() override                  = default;
 
+  void Start();
```
ejfitzgerald: Why do we need this again?

scripts/end_to_end_test/end_to_end_test.yaml
```diff
@@ -244,7 +244,7 @@ steps:
 
 ---
 test_description: >
-  Test synergetic contract fees
+  DISABLED - Test synergetic contract fees
```
ejfitzgerald, line 247: How come disabled?

scripts/end_to_end_test/end_to_end_test.yaml
```diff
@@ -267,7 +267,7 @@ steps:
 
 ---
 test_description: >
-  Test synergetic contract fees in case of no funds
+  DISABLED - Test synergetic contract fees in case of no funds
```
ejfitzgerald, line 270: How come disabled?

ejfitzgerald (changes requested):

scripts/end_to_end_test/end_to_end_test.yaml
n-hutton: Ran out of time, I mentioned I would ask atilla to fix them (rely on wealth creation)

scripts/end_to_end_test/end_to_end_test.yaml
n-hutton: as above

libs/ledger/include/ledger/protocols/main_chain_rpc_service.hpp
n-hutton: we do not.

ejfitzgerald (approved):

PR #2167, opened 2019-12-13
Improved precision of AsFP64 for LFG generator

MissingNO57:


evsmithx (dismissed):

libs/core/include/core/random/lfg.hpp
```diff
@@ -79,32 +81,41 @@ class LaggedFibonacciGenerator
...
+   * @return float or double uniformly distributed random value between 0.0 and 1.0
+   */
+  template <typename DataType>
+  meta::IfIsFloat<DataType, DataType> AsFP() noexcept
   {
-    auto fp128_u_max_ = static_cast<uint64_t>(fp128_t::FP_MAX);
-
-    auto fp_val = static_cast<fp128_t>(this->operator()() % fp128_u_max_);
-    return fp_val / fp128_t::FP_MAX;
+    return static_cast<DataType>(static_cast<double>(this->operator()()) * inv_double_max_);
```
baykaner, line 102: Please delete this method and move the implementation directly into AsDouble(). This will avoid 2 casts, and make the intention much clearer for others devs.

baykaner: Okay, so I think the reason you made this function is to allow for a unified interface; in this case, the name AsFP is wrong. You're not returning an FP - you're returning a DataType.
Please rename these Three functions AsType: Then it should be clear when it's called that AsType<Type>() is going to return a Type

libs/math/benchmark/trigonometry/trigonometry.cpp
```diff
@@ -32,7 +32,7 @@ void BM_Sin(benchmark::State &state)
   for (auto _ : state)
   {
     state.PauseTiming();
-    val = T(lfg.AsDouble());
+    val = T(lfg.AsFP<double>());
```
baykaner: These should be replaced with val = lfg.AsType<Type>();

I know that this is not the line being benchmarked, but since this provide a very bad example that other developers might see and copy thinking that it is safe. this is NOT safe!

libs/math/include/math/tensor.hpp
```diff
@@ -1576,7 +1576,7 @@ Tensor<T, C> &Tensor<T, C>::FillUniformRandom()
 {
   for (SizeType i = 0; i < this->size(); ++i)
   {
-    this->operator[](i) = Type(random::Random::generator.AsDouble());
+    this->operator[](i) = random::Random::generator.AsFP<Type>();
```
baykaner: 👍 
AsType<Type>() - but it's the same principle

baykaner (changes requested):

baykaner (dismissed):

baykaner (approved):

PR #2170, opened 2019-12-13
Fixing network race conditions

troels-im:
This PR fixes a many race conditions in the networking code and improves general stability.


libs/core/include/core/mutex.hpp
```diff
@@ -17,15 +17,166 @@
...
+  {
+    lockable_.unlock();
+  }
+
+private:
+  DebugMutex &lockable_;
+};
+
+//#ifndef NDEBUG
+using Mutex = DebugMutex;
```
ejfitzgerald, line 164: Can we control this via the existing cmake option instead of as default in debug mode please

libs/core/include/core/mutex.hpp
```diff
@@ -17,15 +17,166 @@
...
 namespace fetch {
+class DebugMutex;
+
+struct LockLocation
+{
+  std::string filename{};
+  int32_t     line{};
+};
+
+class Mutexregister
```
ejfitzgerald:
```suggestion
class MutexRegister
```

libs/http/src/https_client.cpp
```diff
@@ -33,7 +33,10 @@ namespace http {
  */
 HttpsClient::HttpsClient(std::string host, uint16_t port)
   : HttpClient{std::move(host), port}
-{}
+{
+  context_.set_default_verify_paths();
```
ejfitzgerald: Already in `master`

libs/muddle/internal/kademlia/table.hpp
```diff
@@ -218,7 +250,9 @@ struct MapSerializer<muddle::KademliaTable, D>
       item.ReportExistence(p, p.last_reporter);
     }
 
-    map.ExpectKeyGetValue(CONNECTION_EXPIRY, item.connection_expiry_);
+    FETCH_LOCK(item.desired_mutex_);
```
ejfitzgerald: Worth explicitly marking priority on in the class declaration of these mutexs?

ejfitzgerald (changes requested):

ejfitzgerald (dismissed):

ejfitzgerald (dismissed):

ejfitzgerald (approved):

PR #2172, opened 2019-12-13
Feature/ml 289 flatten temp

iryna-melnyk:


libs/vm-modules/src/ml/model/model_estimator.cpp
```diff
@@ -158,6 +158,12 @@ ChargeAmount ModelEstimator::LayerAddConvActivation(
   return MaximumCharge("Not yet implemented");
 }
 
+ChargeAmount ModelEstimator::LayerAddFlatten(Ptr<fetch::vm::String> const &layer)
+{
+  FETCH_UNUSED(layer);
+  return MaximumCharge("Not yet implement");
```
vhavrylov-ncube: Not yet implemented

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -848,4 +853,18 @@ TEST_F(VMModelTests, model_with_accuracy_metric)
   EXPECT_LE(metrics->elements.at(1), 1);
 }
 
+TEST_F(VMModelTests, DISABLED_model_sequential_flatten)
+{
```
vhavrylov-ncube, line 856: Why disabled? Would it fail now?

libs/vm-modules/include/vm_modules/ml/model/model.hpp
```diff
@@ -143,6 +144,7 @@ class VMModel : public fetch::vm::Object
                                            math::SizeType const &                   inputs,
                                            math::SizeType const &                   hidden_nodes,
                                            fetch::vm::Ptr<fetch::vm::String> const &activation);
+  void LayerAddFlatten(fetch::vm::Ptr<fetch::vm::String> const &layer);
```
baykaner, line 152: this is fine for now, but we should take note that as soon as we add implement another layer that takes no inputs (except the string) we're going to have to refactor these layers to a slightly less clear format; something like 'LayerAddOneParam(Ptr<String> const &layer)' which has an if/else to determine whether to call Flatten, or another layer type. @vhavrylov-ncube previously implemented a version of this

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
baykaner: yes - that is because the estimator will set maximum charge, which will cost too much for the VM to execute. That is expected behaviour for now, but see https://fetchai.atlassian.net/browse/ML-377

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -848,4 +853,18 @@ TEST_F(VMModelTests, model_with_accuracy_metric)
...
+  static char const *SRC_METRIC = R"(
+        function main()
+          var model = Model("sequential");
+          model.add("flatten");
+          model.compile("scel", "adam", {"categorical accuracy"});
+        endfunction
+      )";
+
+  ASSERT_TRUE(toolkit.Compile(SRC_METRIC));
+  ASSERT_TRUE(toolkit.Run());
```
baykaner, line 866: you should extend this test (or write another one if you prefer), so that you actually pass in some data to the model, call model.predict, return the output tensor from the etch function, and compare the shape and values of the tensor with another 'ground truth' tensor.

baykaner (changes requested):
looks good - just need to extend the testing slightly

baykaner (dismissed):

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: done

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -847,4 +852,60 @@ TEST_F(VMModelTests, model_with_accuracy_metric)
...
+TEST_F(VMModelTests, DISABLED_model_sequential_flatten_tensor_data)
+{
+  static char const *SRC_METRIC = R"(
+        function main() : Tensor
+          var tensor_shape = Array<UInt64>(2);
+          tensor_shape[0] = 3u64;
+          tensor_shape[1] = 3u64;
+
+          var x = Tensor(tensor_shape);
+          var str_vals = "0.5, 7.1, 9.1; 6.2, 7.1, 4.; -99.1, 14328.1, 10.0;";
```
vhavrylov-ncube, line 874: as we discussed, above creation could be reworked just to
var x = Tensor(Array\<UInt64\>(1));
because later it is resized and reshaped anyway.

libs/vm-modules/src/ml/model/model_estimator.cpp
iryna-melnyk: Done

evsmithx (approved):

vhavrylov-ncube (approved):
LGTM

iryna-melnyk:
#2196

PR #2175, opened 2019-12-16
remove unneccessary address in block

n-hutton:


libs/ledger/src/execution_manager.cpp
```diff
@@ -177,7 +177,7 @@ ExecutionManager::ScheduleStatus ExecutionManager::Execute(Block const &block)
   // update the last block hash
   state_.ApplyVoid([&block](Summary &summary) {
     summary.last_block_hash   = block.hash;
-    summary.last_block_miner  = block.miner;
+    summary.last_block_miner  = chain::Address(block.miner_id);
```
n-hutton, line 180: Is this neccessary?

ejfitzgerald (commented):
Can you double check that fees are not affected by this change, but otherwise looks good

libs/network/include/network/p2pservice/p2p_http_interface.hpp
```diff
@@ -155,7 +155,7 @@ class P2PHttpInterface : public http::HTTPModule
       block["hash"]         = "0x" + b->hash.ToHex();
       block["previousHash"] = "0x" + b->previous_hash.ToHex();
       block["merkleHash"]   = "0x" + b->merkle_hash.ToHex();
-      block["miner"]        = b->miner.display();
+      block["miner"]        = b->miner_id.identifier().ToBase64();
```
ejfitzgerald: This needs to be the address (block explorer needs it to be)

ejfitzgerald (changes requested):

libs/network/include/network/p2pservice/p2p_http_interface.hpp
n-hutton: Ok

ejfitzgerald (approved):

PR #2179, opened 2019-12-16
Set opcode charges based on benchmark times

jrriehl:
Opcode charges are based on benchmark times, as described [here](https://github.com/jrriehl/ledger/blob/benchmark/scripts/benchmark/OpcodePricing.md), in nanoseconds, raised to the nearest integer.

The convention adopted here is that 1 charge unit = 1 nanosecond of CPU time on the 4 GHz core used for testing. Since ChargeAmount is an integer and the basic opcodes run on the order of nanoseconds, we need to have at least enough resolution to distinguish between these.

We use the Fixed64 benchmarks for the primitive opcodes, since they are the most expensive of those tested, but in the future we could discriminate between the particular primitive used.

Opcodes associated with Object types and those that have not yet been benchmarked are assigned to the previous default value (1), but we should probably consider raising the default charge while we determine the appropriate charges.


wkaluza (approved):

wkaluza:
Let's get this in so we can build on it. Really thorough work. Some of the findings suggest that there's something funny going on behind the scenes - I wouldn't have expected the `PushFalse` opcode to be noticeably slower than `PushTrue`, for example.

jrriehl:
@wkaluza, thanks for the review! Just noticed that the numbers in the report are slightly out of date. I ran the benchmarks for more iterations and they settled to more reasonable values (PushTrue and PushFalse turn out to be roughly the same). These are already in the code, but I'll update the report to match these.

PR #2180, opened 2019-12-16
Etch defensive checks

robertdickson:
This PR adds various defensive bomb-proofing to prevent rogue scripts blowing up the etch compiler.


libs/vm/src/analyser.cpp
```diff
@@ -925,6 +963,11 @@ void Analyser::BuildFreeFunctionDefinition(BlockNodePtr const &function_definiti
   ExpressionNodePtr function_name_node =
       ConvertToExpressionNodePtr(function_definition_node->children[1]);
   std::string const &function_name = function_name_node->text;
+  if (num_free_functions_++ >= MAX_FREE_FUNCTIONS)
+  {
+    throw FatalErrorException(filename_, function_name_node->line,
```
ejfitzgerald, line 968: Are these exceptions caught by the top level compile function?

robertdickson: They are caught inside Analyse(), they can't escape that function.

robertdickson: It means fatal error in the sense of fatal semantic error, not fatal error for the compiler itself.

ejfitzgerald (dismissed):

ejfitzgerald (approved):

PR #2182, opened 2019-12-16
Bugfix/startup walk back

n-hutton:
This PR does two things:

- Fixes overly strict requirement on the merkle state matching the chain state
- Creates a whiltelist of initial miners

### DIFF
--- a/libs/constellation/src/constellation.cpp
+++ b/libs/constellation/src/constellation.cpp
@@ -549,7 +549,10 @@ bool Constellation::OnBringUpExternalNetwork(
     }
   }
 
-  consensus_->UpdateCurrentBlock(*chain_->CreateGenesisBlock());
+  if(recovery_block_)
+  {
+    consensus_->UpdateCurrentBlock(*recovery_block_);
+  }
 
   block_packer_ = std::make_unique<BlockPackingAlgorithm>(cfg_.log2_num_lanes);
 
@@ -992,6 +995,8 @@ bool Constellation::CheckStateIntegrity()
     // we need to update the execution manager state and also our locally cached state about the
     // 'last' block that has been executed
     execution_manager_->SetLastProcessedBlock(current_block->hash);
+
+    recovery_block_ = current_block;
   }
   else
   {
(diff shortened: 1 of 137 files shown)


libs/constellation/src/constellation.cpp
```diff
@@ -526,24 +531,21 @@ bool Constellation::OnBringUpExternalNetwork(
...
+  else
+  {
+    FETCH_LOG_INFO(LOGGING_NAME, "No snapshot to reset consensus with.");
   }
 
-  execution_manager_ = std::make_shared<ExecutionManager>(
-      cfg_.num_executors, cfg_.log2_num_lanes, storage_,
-      [this] { return std::make_shared<Executor>(storage_); }, tx_status_cache_);
+  // Update with genesis to trigger loading any saved state
+  consensus_->UpdateCurrentBlock(*chain_->CreateGenesisBlock());
```
ejfitzgerald, line 548: This doesn't seem right?

libs/constellation/src/constellation.cpp
```diff
@@ -927,39 +929,69 @@ bool Constellation::GenesisSanityChecks(GenesisFileCreator::Result genesis_statu
...
+      FETCH_LOG_WARN(LOGGING_NAME, "The revert operation failed!");
       return false;
     }
 
-    if (!storage_->RevertToHash(current_block->merkle_hash, current_block->block_number))
+    FETCH_LOG_INFO(LOGGING_NAME, "Reverted storage unit.");
+    FETCH_LOG_INFO(LOGGING_NAME, "Reverting DAG to: ", current_block->block_number);
+
+    // Need to revert the DAG too
+    if (dag_ && !dag_->RevertToEpoch(current_block->block_number))
```
ejfitzgerald, line 988: 👍

libs/constellation/src/constellation.cpp
ejfitzgerald: This should be the heaviest block on the chain presumably?

n-hutton: I think it is fine - you are allowed to update consensus with a current block that is not contigous in block number so while this looks weird I think it is safe. I did previously have a solution that updated it to the current block that we recovered to in constellation but I deemed it unneccessary code.

ejfitzgerald (approved):

PR #2184, opened 2019-12-17
ML-306 Added test for changing batch size on etch model

evsmithx:
Added test for changing the batch size on an etch model


libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -847,4 +847,48 @@ TEST_F(VMModelTests, model_with_accuracy_metric)
...
+          data_shape2[1] = 123u64;
+          var label_shape2 = Array<UInt64>(2);
+          label_shape2[0] = 7u64;
+          label_shape2[1] = 123u64;
+          var data2 = Tensor(data_shape2);
+          var label2 = Tensor(label_shape2);
+
+          // train the model again
+          model.fit(data2, label2, 16u64);
+
```
vhavrylov-ncube, line 1339: Assuming there is no one assert in etch code and neither one later in GTEST, does this test only check that the provided code could just run without crash?

evsmithx: Yep! I can't really think of anything else to check

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -847,4 +847,48 @@ TEST_F(VMModelTests, model_with_accuracy_metric)
...
+          var label = Tensor(label_shape);
+
+          // set up model
+          var model = Model("sequential");
+          model.add("dense", 10u64, 10u64, "relu");
+          model.add("dense", 10u64, 7u64);
+          model.compile("scel", "adam");
+
+          // train the model
+          model.fit(data, label, 32u64);
```
iryna-melnyk, line 1325: It would be useful to check the current batch size after each model.fit if possible

evsmithx: There's no way of doing it currently with the member functions of VMModel, and I think that adding methods for it would only be used here so there's not much point!

vhavrylov-ncube (approved):
LGTM

PR #2186, opened 2019-12-17
strip hex output from fixedpoint string

markos:
When hex debug output is enabled, FixedPoint string constructor throws an exception.

### DIFF
--- a/libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
+++ b/libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
@@ -995,6 +995,11 @@ FixedPoint<I, F>::FixedPoint(std::string const &s)
   auto index  = s.find("fp");
   auto s_copy = std::string(s, 0, index);
 
+#ifdef FETCH_FIXEDPOINT_DEBUG_HEX
+  index  = s_copy.find("(0x");
+  s_copy = std::string(s_copy, 0, index);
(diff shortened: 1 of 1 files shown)


libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
```diff
@@ -995,6 +995,11 @@ FixedPoint<I, F>::FixedPoint(std::string const &s)
   auto index  = s.find("fp");
   auto s_copy = std::string(s, 0, index);
 
+#ifdef FETCH_FIXEDPOINT_DEBUG_HEX
+  index  = s_copy.find("(0x");
+  s_copy = std::string(s_copy, 0, index);
+#endif
+
   bool contains_alpha = std::find_if(s_copy.begin(), s_copy.end(),
```
vhavrylov-ncube, line 1003: Is there a way to test it?

evsmithx (commented):
Could you make a test checking this behaviour?

vhavrylov-ncube (approved):

markos:
If FETCH_FIXEDPOINT_DEBUG_HEX is enabled during compilation, then without this patch, both math-unit-tests and vm-modules-unit-tests will fail. They both pass with this patch. Essentially the string constructor will see "(0x" and fail on 'x' as an alpha character.

PR #2192, opened 2019-12-17
Segfault on shutdown + more

troels-im:
This PR fixes three distinct issues that can happen, two of which are only relevant at shutdown.


ejfitzgerald (dismissed):
👍 👍 👍

vhavrylov-ncube (approved):

PR #2193, opened 2019-12-18
Correct Validation on blocks with missing confirmations

ejfitzgerald:


libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -187,8 +187,10 @@ void MainChainRpcService::OnNewBlock(Address const &from, Block &block, Address
 
   if (!ValidBlock(block, "new block"))
   {
-    FETCH_LOG_WARN(LOGGING_NAME,
-                   "Gossiped block did not prove valid. Loose blocks seen: ", loose_blocks_seen_);
+    FETCH_LOG_WARN(LOGGING_NAME, "New Block: #", block.block_number, " 0x", block.hash.ToHex(),
```
ejfitzgerald, line 190: Minor logging annoyances when looking through the logs

n-hutton: 👍

PR #2197, opened 2019-12-18
Test/ml 353 misuse of etch ml model

iryna-melnyk:
ML-353: tests for misuse of etch ml model
updated version


vhavrylov-ncube (approved):
LGTM

PR #2198, opened 2019-12-18
Dropout layer binding for VMModel

vhavrylov-ncube:
+ tests
+ corrected Optimizer console output


Jenkinsfile
```diff
@@ -1,268 +0,0 @@
-DOCKER_IMAGE_NAME = 'gcr.io/organic-storm-201412/fetch-ledger-develop:v0.4.2'
```
evsmithx: I assume this was removed by mistake!

libs/ml/include/ml/ops/activations/dropout.hpp
```diff
@@ -74,64 +80,103 @@ class Dropout : public fetch::ml::ops::Ops<T>
...
+    if (error_signal.shape() != inputs.front()->shape())
+    {
+      throw fetch::math::exceptions::WrongShape(
+          "Can not do a backward pass through Dropout layer: error signal shape mismatch inputs "
+          "shape.");
+    }
+    if (scaling_coefs_.shape() != inputs.front()->shape())
+    {
+      throw fetch::math::exceptions::WrongShape(
+          "Can not do a backward pass through Dropout layer: dropout scaling coeffitients shape "
```
evsmithx: coefficients

evsmithx (changes requested):

Jenkinsfile
vhavrylov-ncube: No, I removed this file temporary to prevent CI build of this PR until it is draft. It is a workaround )) I'll restore it for the final review.

libs/ml/include/ml/ops/activations/dropout.hpp
vhavrylov-ncube: agreed!

iryna-melnyk (dismissed):
Looks good to me. But have some doubts regarding hardcoded random_seed = 25102015 for this layer. I was told that random seed can not be changed from etch. @EVSmithFetchAI  is it so?

evsmithx:
> Looks good to me. But have some doubts regarding hardcoded random_seed = 25102015 for this layer. I was told that random seed can not be changed from etch. @EVSmithFetchAI is it so?

It's okay here - it's not necessary to set the random seed in that constructor but it's not going to give different values on different machines (which would be a problem)

vhavrylov-ncube:
Speed tests failed - investigating...

vhavrylov-ncube:
Reverted changes in ops::Dropout

evsmithx (approved):

PR #2202, opened 2019-12-18
Refactor: more defloating of fixedpoint initialisations

vhavrylov-ncube:
+ replaced fload/double initialisation with string constructors


libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -339,7 +339,7 @@ TEST(FixedPointTest, FromString_64_64)
   EXPECT_EQ(m_one_point_five, m_one_point_five_flt);
 
   fp128_t e1("2.718281828459045235360287471352662498");
-  fp128_t e2(2.718281828459045235360287471352662498);
+  fp128_t e2("2.718281828459045235360287471352662498");
   EXPECT_TRUE(e1.Near(e2));
   EXPECT_NE(e1, e2);
```
evsmithx: I would expect e1 == e2 now? I'm not sure this test tests anything meaningful any more!

libs/ml/tests/unit/layers/scaled_dot_product_attention.cpp
```diff
@@ -19,6 +19,7 @@
 #include "ml/layers/scaled_dot_product_attention.hpp"
 
 #include "gtest/gtest.h"
+#include "ml/layers/scaled_dot_product_attention.hpp"
```
evsmithx, line 20: duplicate include

evsmithx (changes requested):
Can you get @markos to check the fixed_point.cpp test to make sure they're still doing the right thing?

libs/math/tests/unit/fixed_point/fixed_point.cpp
vhavrylov-ncube: I guess you're right, it's a mistake; I wouldn't touch this test then, let @markos review it later; it is not critical change anyway.

libs/ml/tests/unit/layers/scaled_dot_product_attention.cpp
vhavrylov-ncube: removed

markos (approved):
Looks Good(TM)

evsmithx (approved):

PR #2204, opened 2019-12-18
Do not save history if it does nothing

n-hutton:
When writing to the storage at the bottom of the state database, do not record operations which have no effect (setting a value to it's current value). This results in a significant space saving.


ejfitzgerald:

Thanks!

PR #2205, opened 2019-12-18
optionally (during revert) allow longer promise timeouts

n-hutton:
Reverting could take a long time during the node start up - this was observed. Make a special case for this to allow a longer promise timeout.


ejfitzgerald:

Thanks!

PR #2207, opened 2019-12-18
Force consensus to be thread safe

n-hutton:
During block validation, consensus accesses the chain and the state database. This can happen async. when blocks arrive during gossip, so consensus should be protected with a mutex (avoids races on the stake manager).


ejfitzgerald:

Thanks!

PR #2215, opened 2019-12-18
Deterministic Serialisation

ejfitzgerald:
Backport of #2210

### DIFF
--- a/libs/shards/src/service_identifier.cpp
+++ b/libs/shards/src/service_identifier.cpp
@@ -75,5 +75,17 @@ bool ServiceIdentifier::operator==(ServiceIdentifier const &other) const
   return (type_ == other.type_) && (instance_ == other.instance_);
 }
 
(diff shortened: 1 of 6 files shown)


libs/shards/src/service_identifier.cpp
```diff
@@ -75,5 +75,17 @@ bool ServiceIdentifier::operator==(ServiceIdentifier const &other) const
...
+bool ServiceIdentifier::operator<(ServiceIdentifier const &other) const
+{
+  if (type_ == other.type_)
+  {
+    return instance_ < other.instance_;
+  }
+  else
+  {
+    return type_ < other.type_;
+  }
```
denis-trapeznikov, line 87: How about something like
```
using SrvcIdData = std::pair<Type, uint32_t>;
return SrvcIdData{type_, instance_} < SrvcIdData{other.type_, other.instance_};
```
or simply
```
return type_ < other.type_ || (type_ == other.type_ && instance_ < other.instance_);
```
?

PR #2217, opened 2019-12-19
add missing shift and associative operators

markos:
Add missing operators <<, >> and associative operators for +, -, *, / and bit operations &, |, ^, so that one can now do 2.0 * x or 1/x etc.


libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -785,6 +793,7 @@ TYPED_TEST(BasicArithmeticTest, Division)
   TypeParam zero(0);
   TypeParam one(1);
   TypeParam two(2);
+  TypeParam three(3);
```
vhavrylov-ncube, line 812: "TypeParam const" would be better (here and in other similar places) if you don't plan to change the variables

libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -810,6 +819,42 @@ TYPED_TEST(BasicArithmeticTest, Division)
...
+class BitOperationsTest : public ::testing::Test
+{
+};
+TYPED_TEST_CASE(BitOperationsTest, FixedPointTypes);
+TYPED_TEST(BitOperationsTest, ShiftLeft)
+{
+  // Positive
+  TypeParam one{1};
+  TypeParam two{2};
+  TypeParam x{1.6519711627625};
```
vhavrylov-ncube, line 853: initialisation with a float, not a string constructor - is it intended?

libs/math/tests/unit/fixed_point/fixed_point.cpp
markos: yes, I plan to do a massive change for that and the constructors -to use the {} construtor- in all these tests, but not in this PR

libs/math/tests/unit/fixed_point/fixed_point.cpp
markos: yes, for checking associativity and operations it's fine :)

vhavrylov-ncube (dismissed):
LGTM

libs/math/tests/unit/fixed_point/fixed_point.cpp
evsmithx: The float constructor is still there for now, but as it's not safe to use in library code I think in the long run we'll have to disable it. So I think it would be better to use the string constructor here.

evsmithx (dismissed):

vhavrylov-ncube (approved):
Only a whitespace change after previous approval. Reapproved.

PR #2218, opened 2019-12-19
add rand() support for fixedpoint types, with tests

markos:
Add rand() support for Fixed32, Fixed64 and Fixed128 types, with tests.


vhavrylov-ncube (dismissed):
Looks good; are charge estimates benchmarked?

markos:
not directly benchmarked but followed the general rule -which came from a benchmark- that if int is 1, then fp32 is 4, fp64 is 6 and fp128 is 12+.

vhavrylov-ncube (approved):

PR #2222, opened 2019-12-19
ML-383 VMTrainingPair removed and VMPair implemented

MissingNO57:


libs/core/include/core/serializers/base_types.hpp
```diff
@@ -792,6 +792,29 @@ struct MapSerializer<std::map<K, V>, D>
   }
 };
 
+template <typename K, typename V, typename D>
+struct PairSerializer<std::pair<K, V>, D>
+{
+public:
+  using Type       = std::map<K, V>;
```
evsmithx: Should this be map and not pair?

evsmithx (dismissed):

evsmithx (approved):

PR #2223, opened 2019-12-19
fix post inc/dec operators

markos:
postfix inc/dec operators did not increase the correct object.


vhavrylov-ncube (approved):
LGTM

iryna-melnyk (approved):
Looks good to me

PR #2224, opened 2019-12-19
Feature/increased telemetry and logging

n-hutton:
Add telemetry for some cases we suspect could be error prone


ejfitzgerald:

Thanks!

PR #2227, opened 2019-12-19
cherry pick for rel

n-hutton:


ejfitzgerald:

Thanks!

PR #2237, opened 2019-12-20
Activation layer experimental estimators and tests

vhavrylov-ncube:


libs/vm-modules/include/vm_modules/ml/model/model_estimator.hpp
```diff
@@ -218,6 +218,9 @@ class ModelEstimator
   static ChargeAmount MaximumCharge(std::string const &log_msg = "");
 
   static ChargeAmount ToChargeAmount(fixed_point::fp64_t const &val);
+
+  static ChargeAmount constexpr EXPERIMENTAL_LARGE_COST  = std::numeric_limits<uint32_t>::max();
```
iryna-melnyk: I would suggest to use int32_T MAX instead of uint 'cause we actually have a forward_pass_cost with signed 32 bit mantissa

iryna-melnyk (changes requested):

libs/vm-modules/include/vm_modules/ml/model/model_estimator.hpp
vhavrylov-ncube: Agreed

vhavrylov-ncube: Removed constatns completely. unnecessary.

iryna-melnyk (approved):

iryna-melnyk (commented):
...

iryna-melnyk (dismissed):
approved but need someone else to take a look

iryna-melnyk (dismissed):
looks good to me

libs/ml/include/ml/core/node.hpp
```diff
@@ -237,8 +237,10 @@ std::shared_ptr<TensorType> Node<TensorType>::Evaluate(bool is_training)
     {
       throw std::runtime_error("NaN encountered in Node::Evaluate");
     }
-
-    assert(!math::state_overflow<DataType>());
+    if (math::state_overflow<DataType>())
+    {
+      throw std::runtime_error("Overflow encountered in Node::Evaluate");
+    }
```
evsmithx: I left the overflow checks as assertions because overflows are sometimes recoverable, unlike NaNs where the result of all operations will become NaN. Do you think it's better to have a runtime_error?

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -78,6 +107,31 @@ class VMModelTests : public ::testing::Test
     // method for a model must end in model compilation error and safe stop.
     ASSERT_FALSE(toolkit.Compile(src));
   }
+
+  void TestActivation(std::string const &input, std::string const &activation,
+                      std::string const &expected)
+  {
+    std::string src =
+        std::regex_replace(ACTIVATION_LAYER_TEST_SOURCE, std::regex("<<ACTIVATION>>"), activation);
+    src = std::regex_replace(src, std::regex("<<INPUT>>"), input);
```
evsmithx, line 116: Regex is a good idea here!

evsmithx (changes requested):

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: the idea is: assert failure would anyway cause a runtime crash with unclear error message (because of messy core dump). In release build, no assertion would be made, but if overflows are not _always_ recoverable, I guess we need to stop VM here. However, it's just a guess, so if you are sure the assert is better here, I'll restore it.

evsmithx (approved):

PR #2241, opened 2019-12-23
build fixes + lower resolution on block hash

n-hutton:


libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -1142,10 +1194,26 @@ BlockCoordinator::State BlockCoordinator::OnReset()
 
   if (block != nullptr && !block->hash.empty())
   {
-    block_hash_->set(*reinterpret_cast<uint64_t const *>(block->hash.pointer()));
+    if ((block->block_number % 100) == 0)
```
ejfitzgerald, line 1201: This is not going to make the block hash less noisy but whatever not really an issue

libs/ledger/src/chain/block_coordinator.cpp
```diff
@@ -520,10 +522,24 @@ BlockCoordinator::State BlockCoordinator::OnSynchronised(State current, State pr
...
 
-  // Failure will set this to a nullptr
-  next_block_ = consensus_->GenerateNextBlock();
+    // Failure will set this to a nullptr
+    next_block_ = consensus_->GenerateNextBlock();
+  }
+  catch (std::exception const &ex)
+  {
+    FETCH_LOG_WARN(LOGGING_NAME, "Failed to update consensus with error: ", ex.what());
+    consensus_update_failure_total_->increment();
```
ejfitzgerald, line 535: I think that both of these cases need to result in a switch to the reset state

ejfitzgerald (approved):

PR #2243, opened 2019-12-23
ML-390: fixed values in conv tests

iryna-melnyk:
fixed conv tests


evsmithx (approved):

evsmithx (commented):
Can these tests now be re-enabled?

PR #2244, opened 2019-12-23
test/ML-377-enable-tests

iryna-melnyk:
tests in ml_model_tests were partially enabled
I hope this PR will be merged after  2237 & 2243

### DIFF
--- a/libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
+++ b/libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
@@ -453,27 +453,31 @@ TEST_F(VMModelTests, model_add_dense_relu)
 }
 
 // Disabled until implementation of AddLayerConv estimator
-TEST_F(VMModelTests, DISABLED_model_add_conv1d_noact)
+TEST_F(VMModelTests, model_add_conv1d_noact)
 {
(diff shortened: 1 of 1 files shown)


libs/ml/include/ml/core/node.hpp
```diff
@@ -237,8 +237,10 @@ std::shared_ptr<TensorType> Node<TensorType>::Evaluate(bool is_training)
     {
       throw std::runtime_error("NaN encountered in Node::Evaluate");
     }
-
-    assert(!math::state_overflow<DataType>());
+    if (math::state_overflow<DataType>())
```
baykaner: I'm not convinced that we want to be doing this. What is the justification for switching to runtime overflow checks?

baykaner (changes requested):
testing looks good. Not sure about the runtime node overflow checks though

libs/ml/include/ml/core/node.hpp
iryna-melnyk: My changes were based on  #2237 to avoid a lot of conflicts in ml_model_tests
I'll restore node.hpp

iryna-melnyk: this branch was rebased so all unneeded changes are gone

baykaner (approved):

PR #2246, opened 2019-12-23
Bugfix/ml 402 fix tensor reshape

baykaner:
- fixes reshape op to work as expected

### DIFF
--- a/libs/math/tests/unit/tensor/tensor_operations.cpp
+++ b/libs/math/tests/unit/tensor/tensor_operations.cpp
@@ -21,21 +21,85 @@
 #include "test_types.hpp"
 #include "vectorise/fixed_point/fixed_point.hpp"
 
-namespace fetch {
-namespace math {
-namespace test {
+namespace {
+
+using SizeType = fetch::math::SizeType;
 
 template <typename T>
 class TensorOperationsTest : public ::testing::Test
 {
 };
 
-TYPED_TEST_CASE(TensorOperationsTest, FloatIntAndUIntTypes);
+TYPED_TEST_CASE(TensorOperationsTest, fetch::math::test::FloatIntAndUIntTypes);
+
+template <typename TypeParam>
+void ReshapeTest(std::vector<SizeType> const &initial_shape,
+                 std::vector<SizeType> const &final_shape)
+{
+  fetch::math::Tensor<TypeParam> t1(initial_shape);
+  fetch::math::Tensor<TypeParam> t2(final_shape);
+  t1.Reshape(final_shape);
+  EXPECT_TRUE(t1.shape() == t2.shape());
+}
+
+TYPED_TEST(TensorOperationsTest, two_dim_reshape_test)
+{
+  ReshapeTest<TypeParam>({2, 3}, {6, 1});
+  ReshapeTest<TypeParam>({4, 4}, {2, 8});
+  ReshapeTest<TypeParam>({1, 2}, {2, 1});
+  ReshapeTest<TypeParam>({0, 0}, {0, 0});
+}
+
+TYPED_TEST(TensorOperationsTest, three_and_two_dim_reshape_test)
+{
+  ReshapeTest<TypeParam>({2, 3, 1}, {6, 1});
+  ReshapeTest<TypeParam>({2, 1, 3}, {6, 1});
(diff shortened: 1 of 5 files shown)


libs/math/tests/unit/tensor/tensor_operations.cpp
```diff
@@ -21,21 +21,85 @@
...
+  ReshapeTest<TypeParam>({2, 1, 3, 1}, {6, 1, 1});
+  ReshapeTest<TypeParam>({2, 1, 1, 3}, {2, 1, 3});
+  ReshapeTest<TypeParam>({2, 3, 4, 5}, {12, 2, 5});
+
+  ReshapeTest<TypeParam>({6, 1, 1}, {2, 3, 1, 1});
+  ReshapeTest<TypeParam>({2, 1, 3}, {2, 1, 3, 1});
+  ReshapeTest<TypeParam>({1, 1, 6}, {2, 1, 1, 3});
+  ReshapeTest<TypeParam>({3, 4, 10}, {2, 3, 4, 5});
+}
+
```
vhavrylov-ncube, line 100: What about negative cases, when reshape should fail/throw ?

libs/ml/include/ml/ops/reshape.hpp
```diff
@@ -40,12 +40,15 @@ class Reshape : public fetch::ml::ops::Ops<T>
 
   explicit Reshape(std::vector<SizeType> new_shape)
     : new_shape_(std::move(new_shape))
-  {}
+  {
+    new_size_ = fetch::math::Product(new_shape_);
```
vhavrylov-ncube: Wouldn't it be better to place this in initializer list?

libs/ml/include/ml/ops/reshape.hpp
```diff
@@ -40,12 +40,15 @@ class Reshape : public fetch::ml::ops::Ops<T>
...
+  }
 
   explicit Reshape(SPType const &sp)
     : Ops<T>(sp)
   {
     new_shape_ = sp.new_shape;
+    new_size_  = sp.new_size;
```
vhavrylov-ncube: initialiser list again

libs/ml/tests/unit/ops/reshape.cpp
```diff
@@ -0,0 +1,293 @@
...
+      new_op.Backward({std::make_shared<const TensorType>(data)}, error);
+
+  // test correct values
+  EXPECT_TRUE(error_signal.at(0).AllClose(
+      new_error_signal.at(0), fetch::math::function_tolerance<typename TypeParam::Type>(),
+      fetch::math::function_tolerance<typename TypeParam::Type>()));
+  fetch::math::state_clear<DataType>();
+}
+
+TYPED_TEST(ReshapeTest, Reshape_graph_serialization_test)
```
vhavrylov-ncube: serialisation

libs/ml/tests/unit/ops/reshape.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+  data.FillUniformRandom();
+  auto gt_error_it = gt_error.begin();
+  auto data_it     = data.begin();
+  for (std::size_t j = 0; j < gt_error.size(); ++j)
+  {
+    *gt_error_it = *data_it;
+    ++gt_error_it;
+    ++data_it;
+  }
```
evsmithx: I think you can use gt_error = data.Copy() here.

evsmithx (commented):
There doesn't seem to be a test here for the order that the elements are put into the new tensor, i.e. whether `[1, 2, 3, 4].reshape(2,2)` gives [[1, 2], [3, 4]] or [[1, 3], [2, 4]].

libs/math/tests/unit/tensor/tensor_operations.cpp
evsmithx: The parameters are vectors of SizeTypes, so presumably it would give a compilation error.

vhavrylov-ncube (dismissed):
Looks good to me

evsmithx (changes requested):
Batch dimension of tensor in reshape op needs to stay the same. Tests for this and also for ComputeOutputShape needed, as discussed on slack.

evsmithx (approved):

PR #2247, opened 2019-12-24
benchmarck for categorical accuracy

AlexTeslenko:
added benchmark for categorical accuracy forward pass


libs/ml/benchmark/metrics/metrics.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+  fetch::ml::ops::CategoricalAccuracy<TensorType> ca;
+
+  for (auto _ : state)
+  {
+    ca.Forward(inputs, output);
+  }
+}
+
+BENCHMARK_TEMPLATE(BM_CategoricalAccuracy, int, 2, 2)->Unit(benchmark::kMicrosecond);
```
baykaner: we don't want to benchmark for int particularly; but we should benchmark for:
float, double, fp32, fp64, fp128

baykaner (changes requested):

libs/ml/benchmark/metrics/metrics.cpp
AlexTeslenko: ok, will amend that

libs/ml/benchmark/metrics/metrics.cpp
```diff
@@ -0,0 +1,97 @@
...
+BENCHMARK_TEMPLATE(BM_CategoricalAccuracy, float, 1000, 1000)->Unit(benchmark::kMillisecond);
+BENCHMARK_TEMPLATE(BM_CategoricalAccuracy, float, 2000, 2000)->Unit(benchmark::kMillisecond);
+
+BENCHMARK_TEMPLATE(BM_CategoricalAccuracy, double, 2, 2)->Unit(benchmark::kMicrosecond);
+BENCHMARK_TEMPLATE(BM_CategoricalAccuracy, double, 10, 10)->Unit(benchmark::kMicrosecond);
+BENCHMARK_TEMPLATE(BM_CategoricalAccuracy, double, 100, 100)->Unit(benchmark::kMillisecond);
+BENCHMARK_TEMPLATE(BM_CategoricalAccuracy, double, 1000, 1000)->Unit(benchmark::kMillisecond);
+BENCHMARK_TEMPLATE(BM_CategoricalAccuracy, double, 2000, 2000)->Unit(benchmark::kMillisecond);
+
+BENCHMARK_TEMPLATE(BM_CategoricalAccuracy, fetch::fixed_point::FixedPoint<16, 16>, 2, 2)
```
baykaner, line 64: fyi: instead of writing out FixedPoint<16, 16> you can use fp32_t (I believe it's in either the fixed_point or math namespace)

baykaner (dismissed):

baykaner (approved):

PR #2249, opened 2019-12-26
Improvement/add block by pointer

denis-trapeznikov:
Moving BlockPtr definitions from MainChain class into the namespace; AddBlock() accepts pointers; TimeTravelogue is templated no more.


libs/ledger/src/upow/naive_synergetic_miner.cpp
```diff
@@ -135,7 +135,7 @@ void NaiveSynergeticMiner::Mine()
     std::ostringstream oss;
     for (auto const &element : problem_spaces)
     {
-      oss << "\n -> 0x" << element.first.contract_address.display();
+      oss << "\n -> 0x" << element.first.address().ToHex();
```
ejfitzgerald: If this is actually a contract address then I think that we should display it with `.display()` as this would be more consistent

libs/ledger/include/ledger/chain/main_chain.hpp
```diff
@@ -167,21 +164,22 @@ class MainChain
...
-  BlockHash  GetHeaviestBlockHash() const;
-  Blocks     GetHeaviestChain(uint64_t limit = UPPER_BOUND) const;
-  Blocks     GetChainPreceding(BlockHash start, uint64_t limit = UPPER_BOUND) const;
-  Travelogue TimeTravel(BlockHash current_hash) const;
-  bool       GetPathToCommonAncestor(
-            Blocks &blocks, BlockHash tip_hash, BlockHash node_hash, uint64_t limit = UPPER_BOUND,
-            BehaviourWhenLimit behaviour = BehaviourWhenLimit::RETURN_MOST_RECENT) const;
+  BlockPtr      GetHeaviestBlock() const;
+  BlockHash     GetHeaviestBlockHash() const;
+  MutableBlocks GetHeaviestChain(uint64_t limit = UPPER_BOUND) const;
```
ejfitzgerald: If we are basically dropping the internal block type now, then I guess this make sense to drop the Mutable part from the name

ejfitzgerald (commented):
Mostly minor comments

libs/ledger/src/upow/naive_synergetic_miner.cpp
```diff
@@ -135,7 +135,7 @@ void NaiveSynergeticMiner::Mine()
     std::ostringstream oss;
     for (auto const &element : problem_spaces)
     {
-      oss << "\n -> 0x" << element.first.address().ToHex();
+      oss << "\n -> 0x" << element.first.display().ToHex();
```
ejfitzgerald:
```suggestion
      oss << "\n -> 0x" << element.first.display();
```

ejfitzgerald (commented):
Just minor suggestions - lets get this merged. This PR is in danger of getting stale and we have some more updates from the release branch that will need to get merged on top of it

ejfitzgerald (dismissed):

ejfitzgerald (approved):

PR #2251, opened 2019-12-27
Ops-based charge estimations for Model.predict()

vhavrylov-ncube:
Attempt to implement charge estimation using recursive operational-wise Graph traversing, without using VMModelEstimator class helper.


libs/ml/include/ml/core/graph.hpp
```diff
@@ -401,6 +420,18 @@ TensorType Graph<TensorType>::ForwardPropagate(std::string const &node_name, boo
   return ForwardImplementation(node_name, is_training, false);
 }
 
+template <typename TensorType>
```
baykaner, line 433: this stuff is really not simple at all and **wont** be obvious when we look back at it in the future so please add a doxygen-style function comment explaining **which** node node_name is pointing to when you invoke this function and a description of what the function is for more generally

libs/ml/include/ml/core/graph.hpp
```diff
@@ -157,6 +157,22 @@ class Graph
 
   void ResetGradients();
 
+  virtual typename fetch::ml::ops::Ops<T>::ChargeAmount ForwardPassChargeCost(
```
baykaner: doxygen style function comment please

libs/ml/include/ml/core/node.hpp
```diff
@@ -52,6 +52,8 @@ class Node
   using VecTensorType    = typename fetch::ml::ops::Ops<TensorType>::VecTensorType;
   using SPType           = fetch::ml::NodeSaveableParams<TensorType>;
   using NodeErrorMapType = std::unordered_map<Node<TensorType> *, std::vector<TensorType>>;
+  using ShapeType =
```
baykaner: ShapeType is ALWAYS a math::SizeVector everywhere, and you should never explicitly set uint64_t for a SizeType or ShapeType since we should be allowed to change this to a uint32 or uint128 and everything should still work fine. Therefore just use SizeVector

libs/ml/include/ml/core/node.hpp
```diff
@@ -141,6 +143,105 @@ class Node
     return static_cast<bool>(cached_output_status_ == CachedOutputState::VALID_CACHE);
   }
 
+  virtual uint64_t ForwardPassChargeCost()
+  {
+    static constexpr uint64_t my_default_cost = 100;  // TODO(VH): Calculate/measure me
```
baykaner: even if we only use this in one function I would suggest moving this static constexpr down to the private member variables so it's easy to see what are the initialised/hard-coded parameters all in one place

libs/ml/include/ml/core/node.hpp
```diff
@@ -141,6 +143,105 @@ class Node
...
+    // TODO(VH): call OpForwardCost() of the Op - we need to know Data shape to do this.
+    if (input_nodes_.empty())
+    {
+      FETCH_LOG_INFO("Node", " Input node reached: default cost " +
+                                 std::to_string(my_default_cost) + " added.");
+      return my_default_cost;
```
baykaner: I'm not 100% sure about this logic. Seems a bit like we're indirectly saying 'is this node holding a placeholder op?'
It's probably fine, but if we ever have more than one op that doesn't need any input nodes we're assuming they should have the same cost. Should they?

libs/ml/include/ml/core/node.hpp
```diff
@@ -141,6 +143,105 @@ class Node
...
+    }
+    uint64_t                           total_cost = my_default_cost;
```
baykaner: these uint64_t's should probably be ChargeAmount's I guess?

libs/ml/include/ml/core/node.hpp
```diff
@@ -141,6 +143,105 @@ class Node
...
+    std::vector<std::vector<uint64_t>> input_shapes;
```
baykaner: std::vector<std::vector<uint64_t>> should be
std::vector<SizeVector>

libs/ml/include/ml/core/node.hpp
```diff
@@ -141,6 +143,105 @@ class Node
...
+    for (auto const &i : input_nodes_)
+    {
+      if (auto input_node_ptr = i.lock())
+      {
+        auto const input_node_forward_cost = input_node_ptr->ForwardPassChargeCost();
+        total_cost += input_node_forward_cost;
```
baykaner: why not just:
total_cost += input_node_ptr->ForwardPassChargeCost();

libs/ml/include/ml/core/node.hpp
```diff
@@ -141,6 +143,105 @@ class Node
...
+        }
+      }
+      else
+      {
+        throw std::runtime_error("Unable to lock weak pointer.");
+      }
+    }
+    auto const my_cost_of_input = op_ptr_->OpForwardCost(input_shapes);
+    total_cost += my_cost_of_input;
+
```
baykaner, line 209: Perhaps we should also check for overflow of total_cost?

libs/ml/include/ml/core/node.hpp
```diff
@@ -141,6 +143,105 @@ class Node
...
+    return total_cost;
+  }
+
+  void SetDefaultOutputShape(ShapeType const &new_shape)
+  {
+    default_output_shape_ = new_shape;
+    op_ptr_->SetDefaultOutputShape(new_shape);
+  }
+
+  ShapeType DefaultOutputShape()
```
baykaner: please add function comments explaining why these functions exist

libs/ml/include/ml/core/node.hpp
```diff
@@ -149,6 +250,7 @@ class Node
   TensorType        cached_output_;
   CachedOutputState cached_output_status_;
   OpType            operation_type_;
+  ShapeType         default_output_shape_{};
```
baykaner: since you've added a private member variable to Node you will need to update Node serialisation as well (+ add tests to check that this serialisation->deserialisation correctly reproduces this member variable)

libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -189,14 +197,44 @@ class FullyConnected : public SubGraph<T>
...
+    }
+
+    FETCH_LOG_INFO(DESCRIPTOR, "  Cost calculated : " + std::to_string(total_cost) + "\n");
+    return total_cost;
+  }
+
   static constexpr char const *DESCRIPTOR = "FullyConnected";
 
 private:
+  using Graph<T>::nodes_;
```
baykaner: please don't do this, just use this->nodes_ because it makes the intention clearer

libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -82,26 +82,34 @@ class FullyConnected : public SubGraph<T>
...
     std::string weights =
         this->template AddNode<fetch::ml::ops::Weights<TensorType>>(name + "_Weights", {});
+    nodes_.at(weights)->SetDefaultOutputShape({out_size_, in_size_});
     std::string weights_matmul = this->template AddNode<fetch::ml::ops::MatrixMultiply<TensorType>>(
         name + "_MatrixMultiply", {weights, flat_input});
+    nodes_.at(weights_matmul)->SetDefaultOutputShape({out_size_, 1});
     std::string bias =
         this->template AddNode<fetch::ml::ops::Weights<TensorType>>(name + "_Bias", {});
+    nodes_.at(bias)->SetDefaultOutputShape({out_size_, 1});
+
```
baykaner: i wonder if there's a way we can elegantly separate the layer construction work from the OutputShape setting work (without repeating code); having them interleaved like this seems suboptimal

libs/ml/include/ml/model/model.hpp
```diff
@@ -113,6 +113,12 @@ class Model
   friend struct serializers::MapSerializer;
   friend class fetch::vm_modules::ml::model::ModelEstimator;
 
+  using ChargeAmount = uint64_t;  // TODO(VH): Move me to an external declaration header.
```
baykaner: don't do this; get ChargeAmount from wherever it's genuinely declared!
If this means importing from vm we naturally don't want to do that from the ML library, so it would be better to move ChargeAmount from VM to meta/types.hpp or something like that, and then have vm/vm-modules/ml all include that

libs/ml/include/ml/ops/add.hpp
```diff
@@ -104,6 +104,25 @@ class Add : public fetch::ml::ops::Ops<T>
     return OpType::OP_ADD;
   }
 
+  virtual typename fetch::ml::ops::Ops<T>::ChargeAmount OpForwardCost(
```
baykaner: you don't need virtual here (unless you plan to have an op that inherits from add)

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -40,6 +40,8 @@ class Ops
   using SizeType      = fetch::math::SizeType;
   using ArrayPtrType  = std::shared_ptr<TensorType>;
   using VecTensorType = std::vector<std::shared_ptr<TensorType const>>;
+  using VecShapesType = std::vector<std::vector<SizeType>>;
```
baykaner: don't forget about SizeVector

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -51,6 +53,17 @@ class Ops
    * in ASSERT. On Forward you can use output.shape() and on Backward there is error_signal.shape()
    */
   virtual std::vector<SizeType> ComputeOutputShape(VecTensorType const &inputs) const = 0;
+  std::vector<SizeType>         ComputeDefaultOutputShape(VecShapesType const &input_shapes)
```
baykaner: i'm not sure about your naming - is 'default' really the right word for all these functions?

In any case please add function comment with description

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: Sure! I'll update type system for all use cases.

baykaner (commented):
good work so far

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: The thing here is: a placeholder is shapeless, and its shape can be deduced either from input data (which requires passing input data to all estimation chain), or from the shape of next Dense layer. For now I only traverse the Graph backwards, from output node to the input one; deducing shape of a placeholder Input requires one or more step(s) in opposite direction to find nearest Dense layer and use its shape as a reference. I am going to implement this as well but a bit later.

vhavrylov-ncube: Also there is a problem now with Weight Ops that are not Placeholders but do not have any Input nodes, so I guess here we have to check if the node Ops type is Placeholder or not (and if not, no additional charge would be added)

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: Sure, the type system is still work-in-progress ))

libs/ml/include/ml/layers/fully_connected.hpp
vhavrylov-ncube: the idea behind this is: I know the shapes for all internal sub-nodes of a fully-connected layer, so there is no need to recursively traverse them and calculate shapes.
I would do these assignments in a separate place, but the node names are only accessible in this function, so m.b. I just compose these setters in a block right after creation instructions.
Well, another approach could be just add a possibility to pass expected output shape right to Ops constructor (a separate special shape-aware constructor);

libs/ml/include/ml/model/model.hpp
vhavrylov-ncube: meta/types.hpp sounds just great!

libs/ml/include/ml/ops/add.hpp
vhavrylov-ncube: right, it's a copypasta artifact

libs/ml/include/ml/ops/ops.hpp
vhavrylov-ncube: Not sure "default" is the optimal one, the intention was to show it is not the "working" shape, e.g. "default" ignores batch-size of input data and always have 1 as a trailing dimension. It could be also smth like "single", "minimal", "primitive", so I am open to advice here :)

libs/ml/include/ml/model/model.hpp
vhavrylov-ncube: /libs/meta/include/meta/vm_types.hpp created for this purpose.

libs/ml/include/ml/layers/fully_connected.hpp
vhavrylov-ncube: removed

libs/ml/include/ml/core/graph.hpp
vhavrylov-ncube: added a comment

libs/ml/include/ml/core/graph.hpp
vhavrylov-ncube: done

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: reworked: now any non-placeholder inputless (leaf) Node returns the same charge as a Placeholder one, and the estiamtion depends on output shape.

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: Added operation type check.

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: std::vector<SizeType> is used wherever possible.

libs/ml/include/ml/layers/fully_connected.hpp
vhavrylov-ncube: Moved in a separate code block to prevent interleaving.

libs/ml/include/ml/ops/ops.hpp
vhavrylov-ncube: I've changed it to SliceOutputShape - where a "slice" is 1 data point ("slice") from an input batch.

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: done

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: fixed

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: done

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: Got it, ok. I would not do this in this particular PR because it is not going to be merged as is anyway, but yes in the product version I'll add serialisation for new fields in Node and Ops

libs/ml/include/ml/core/node.hpp
```diff
@@ -141,18 +154,184 @@ class Node
...
+      if (this->OpCode() == OpType::OP_PLACEHOLDER)
+      {
+        FETCH_LOG_INFO(name_.c_str(), "Input Placeholder layer cost added");
+      }
+      else
+      {
+        FETCH_LOG_ERROR(name_.c_str(),
+                        "Can not estimate a forward pass cost of a non-placeholder leaf Node!");
+      }
+      return op_ptr_->OpForwardCost(expected_in_shapes);
```
baykaner, line 181: I think static analysis wont permit this anyway; you'll need to do something like:
if (input_nodese_.empty())
{
  ...
  ret = op_ptr_->OpForwardCost(expected_in_shapes);
}
else
{
  ret = ...;
}
return ret;

libs/ml/include/ml/core/node.hpp
```diff
@@ -141,18 +154,184 @@ class Node
...
+      }
+      else
+      {
+        throw std::runtime_error("Unable to lock weak pointer.");
+      }
+    }
+    auto const my_cost_of_input = op_ptr_->OpForwardCost(input_shapes);
+    if (std::numeric_limits<fetch::vm::ChargeAmount>::max() - total_cost <= my_cost_of_input)
+    {
+      FETCH_LOG_ERROR(this->name_.c_str(),
```
baykaner, line 205: i'm not sure this is actually an error; maybe you just check if an overflow would have occurred and if so set total_cost to max()?

vhavrylov-ncube: I would appreciate a piece of advice on: how to check the overflow here? We have an overflow flag in VM, but this file is not a part of VM itself.

libs/ml/include/ml/core/node.hpp
```diff
@@ -141,18 +154,184 @@ class Node
...
 
   std::shared_ptr<ops::Ops<TensorType>> op_ptr_;
+
+  static const std::set<OpType> shaped_operations_;
 };
 
+/// A list of Ops that "knows" their output shape on construction, such as FullyConnected.
+template <typename TensorType>
+const std::set<OpType> Node<TensorType>::shaped_operations_{OpType::LAYER_FULLY_CONNECTED};
+
```
baykaner, line 344: quite possible i've missed something important, but it feels like we should just set up an input layer, and assume that there are no other ops that 'know' their output shape on construction. In this way we can try to cleanly separate what happens during graph.AddNode and graph.Compile (i.e. at AddNode time the shapes will always be unknown until we call Compile)

baykaner: If we do this, we don't need to include set and can ditch the shaped_operations_ set, i think

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -73,8 +97,93 @@ class Ops
...
+    }
+    ss << "[";
+    for (auto const &dim : out_shape)
+    {
+      ss << " " << dim;
+    }
+    ss << " ])";
+    return ss.str();
+  }
+
```
baykaner, line 169: I think we should consider making this a static free function utility that takes the op as input parameter? Or do we rely on private member access?

vhavrylov-ncube: we rely on private out_shape in this implementation, but yeap it could be extracted as a free function. However, I would separate it as a task "implement .summary() method for VMModel" or alike. This method is a debug/demo dummy to see what's happening while shapes are being linked and estimation is computed, and I believe in a production version we do not need it at all to prevent log flooding with unnecessary info.

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: Well, in general - yes, but then it require some rework of FullyConnected layer initialisation. 
Right now weight tensor (its shape depends on input count, and basically this tensor is the reason why Dense is fixed-shape) is inited in constructor of FullyConnected layer, therefore we need to pass input count as a parameter to constructor.
On the other hand, we can move weight initialisation from constructor to .Compile() call, and then, if we have a fixed-shape Input layer, all other shapes could be deduced and all Weight tensors initialized with proper sizes. I would choose this approach, but it is a big change that is a bit out of scope of this particular PR.

PR #2254, opened 2019-12-30
moved tensor implementations into tensor directory

baykaner:
- moved various math/tensor_* files together into math/tensor/


vhavrylov-ncube (approved):
Only include paths - safe enough.

PR #2255, opened 2019-12-30
Project build time tools

ejfitzgerald:
This PR adds the following changes to repository:

1. It adds a new compiler option which switches off all compiler caching and instead invokes a helper script (`build-profile-gen.py`) that monitors the execution time of each of the compile targets and then generates a `.prof-json` file for each of the target (a separate file was chosen so that the builds can be done in a multithreaded environment to more closely represent the default build case)

2. A companion helper script (`build-profile-collect.py` can then be used to scan through a target build folder, read all the `*.prof-json` files and generate a summary file: `build-profile-summary.csv`

3. A further script (`build-profile-view.py`) can optionally used to filter and display the results from the summary file.

@wkaluza we should catch up when you are back as to how much this conflicts with the flame graph idea you had previously been working on


scripts/build-profile-gen.py
```diff
@@ -0,0 +1,55 @@
...
+    else:
+        idx = 0
+
+    cmd = sys.argv[idx + 1:]
+
+    # attempt to determine output file
+    if '-o' in cmd:
+        output_path = cmd[cmd.index('-o') + 1]
+    else:
+        output_path = cmd[-1]
```
wkaluza, line 23: This is very implicit. The way all our tools behave is that compilers use `-o` and linkers use thelast argument to specify the output file. You could do this more cleanly by having CMake pass a CLI arg to this script to let it know whether it is running a linkage or a compilation.

scripts/build-profile-gen.py
```diff
@@ -0,0 +1,55 @@
...
+
+    profile_path = output_path + '.prof-json'
+    with open(profile_path, 'w') as profile_file:
+        profile = {
+            'cmd': cmd,
+            'outputPath': output_path,
+            'duration': delta,
+            'exitCode': exit_code,
+            'workingDir': os.getcwd(),
+        }
```
wkaluza: I'd store the start and end times here instead of duration.

scripts/build-profile-view.py
```diff
@@ -0,0 +1,87 @@
...
+    subparsers = parser.add_subparsers()
+
+    parser_summary = subparsers.add_parser('summary')
+    parser_summary.set_defaults(handler=run_summary)
+
+    parser_files = subparsers.add_parser('files')
+    parser_files.add_argument(
+        '--gt', type=float, help='Filter the set to files greater than a specified duration')
+    parser_files.add_argument(
+        '--lt', type=float, help='Filter the set to files less than a specified duration')
```
wkaluza: I do not understand these help strings.

scripts/build-profile-collect.py
```diff
@@ -0,0 +1,73 @@
...
+                    ouput_file_path, args.build_folder)
+
+                if profile['exitCode'] != 0:
+                    print('Skipping: {} (process failed)')
+                    continue
+
+                if ouput_file_path.endswith('.o'):
+                    operation = 'compile'
+                else:
+                    operation = 'link'
```
wkaluza, line 41: We do not need to infer this, see other comments.

wkaluza (changes requested):

scripts/build-profile-gen.py
ejfitzgerald: Done

scripts/build-profile-view.py
ejfitzgerald: Done

scripts/build-profile-view.py
```diff
@@ -14,9 +14,9 @@ def parse_commandline():
 
     parser_files = subparsers.add_parser('files')
     parser_files.add_argument(
-        '--gt', type=float, help='Filter the set to files greater than a specified duration')
+        '--gt', type=float, help='Filter output to results with a duration greater than the value specified')
```
wkaluza: `help='Filter output to results with a duration greater than the value specified (in seconds)'`

scripts/build-profile-view.py
```diff
@@ -14,9 +14,9 @@ def parse_commandline():
...
     parser_files.add_argument(
-        '--lt', type=float, help='Filter the set to files less than a specified duration')
+        '--lt', type=float, help='Filter output to results with a duration greater than the value specified')
```
wkaluza: `help='Filter output to results with a duration greater than the value specified (in seconds)'`

scripts/build-profile-view.py
ejfitzgerald: Done

scripts/build-profile-view.py
ejfitzgerald: Done

wkaluza (approved):

PR #2256, opened 2019-12-31
Feature/ml 366 remove float constructor

markos:
Remove FixedPoint float constructor, instead provide a FromFloat method.
This is to disallow implicit conversion from float/double numbers.
As a helper function to allow explicit conversion in typeless situations fetch::math::AsType was added, which does call FromFloat() in the case of FixedPoint.
Also remove all operators that handle non-FixedPoint types, so it will not be possible to do eg 1.5 * x anymore.


libs/math/include/math/base_types.hpp
```diff
@@ -242,5 +242,30 @@ static constexpr meta::IfIsFixedPoint<T, T> Type(std::string const &val)
...
+template <typename T, typename U>
+static constexpr meta::IfIsFloat<T, T> AsType(U val, meta::IfIsFloat<U> * /*unused*/ = nullptr)
+{
+  return static_cast<T>(val);
+}
+
+template <typename T, typename U>
+static constexpr meta::IfIsFixedPoint<T, T> AsType(U val, meta::IfIsFloat<U> * /*unused*/ = nullptr)
+{
+  return T::FromFloat(val);
```
baykaner: I think we need to rework this slightly. We want AsType to be the ONLY way to type convert a fixed point to a float, and privatise T::FromFloat(123.123). This is because if we allow multiple ways of doing the type conversion we make it easier to miss one later on when we're trying to find all cases where a fixed point is constructed via type conversion from a float. It's also more explicit that this is a type conversion this way.

Therefore, I suggest you have a struct here (perhaps wrapped in an anonymous namespace) that is a friend of FixedPoint that does the T::FromFloat for you. Then all other calls to T::FromFloat should be calls to AsType<float/double>

baykaner (changes requested):

libs/math/include/math/base_types.hpp
markos: I see your point, I'll do the necessary changes.

markos: made FromFloat() private, now constructionfrom float is only available through AsType<>, fixed and pushed

libs/dmlf/include/dmlf/collective_learning/utilities/utilities.hpp
```diff
@@ -240,7 +240,7 @@ fetch::dmlf::collective_learning::ClientParams<typename TensorType::Type> Client
   }
   if (!doc["learning_rate"].IsUndefined())
   {
-    client_params.learning_rate = static_cast<DataType>(doc["learning_rate"].As<float>());
+    client_params.learning_rate = fetch::math::AsType<DataType>(doc["learning_rate"].As<float>());
```
baykaner: @EVSmithFetchAI I would suggest that we probably don't want any indeterminacies anywhere in dmlf, right? (otherwise we would never be able to link to smart contracts). Can you think of an easy way to ditch the type casting in this case and just have doc["learning_rate"] be read directly as fixed point?

libs/dmlf/tests/execution/basic_engine_tests.cpp
```diff
@@ -1244,10 +1244,12 @@ TEST(BasicVmEngineDmlfTests, Add64)
 
 TEST(BasicVmEngineDmlfTests, AddFixed)
 {
-  ExecutionResult result = RunStatelessTest(
-      AddFixed, "add", Params{LedgerVariant(fp64_t(4.5)), LedgerVariant(fp32_t(5.5))});
+  ExecutionResult result =
+      RunStatelessTest(AddFixed, "add",
+                       Params{LedgerVariant(fetch::math::AsType<fp64_t>(4.5)),
```
baykaner: Let's just do fixed point construction from string in this test instead.

libs/math/benchmark/trigonometry/trigonometry.cpp
```diff
@@ -32,7 +33,7 @@ void BM_Sin(benchmark::State &state)
   for (auto _ : state)
   {
     state.PauseTiming();
-    val = T(lfg.AsType<double>());
+    val = fetch::math::AsType<T>(lfg.AsType<double>());
```
baykaner: replace with lfg.AsType<T>()

baykaner: same for all cases in this file

libs/ml/benchmark/embeddings/embeddings.cpp
```diff
@@ -76,7 +76,7 @@ void BM_Setup_And_Train_Embeddings(::benchmark::State &state)
   auto embedding_dimensions = static_cast<SizeType>(state.range(1));
   auto n_datapoints         = static_cast<SizeType>(state.range(2));
   auto n_epochs             = static_cast<SizeType>(state.range(3));
-  auto learning_rate        = DataType{0.1f};
+  auto learning_rate        = fetch::math::AsType<DataType>(0.1f);
```
baykaner: use string construction instead

baykaner: use string construction

libs/ml/include/ml/ops/abs.hpp
```diff
@@ -90,13 +90,14 @@ class Abs : public Ops<T>
     assert(error_signal.size() == inputs.at(0)->size());
 
     TensorType return_signal(inputs.at(0)->shape());
+    using DataType = typename TensorType::Type;
```
baykaner: move this to be with all the other using's as a public member of the class

libs/ml/tests/unit/ops/activations/sigmoid.cpp
```diff
@@ -50,7 +50,8 @@ TYPED_TEST(SigmoidTest, forward_test)
   op.Forward({std::make_shared<const TensorType>(data)}, prediction);
 
   // test correct values
-  ASSERT_TRUE(prediction.AllClose(gt, DataType{1e-5f}, DataType{1e-5f}));
+  ASSERT_TRUE(prediction.AllClose(gt, fetch::math::AsType<DataType>(1e-5f),
```
baykaner: just string construct these

baykaner: similarly for others in this file

libs/ml/tests/unit/ops/activations/softmax.cpp
```diff
@@ -50,7 +50,8 @@ TYPED_TEST(SoftmaxTest, forward_test)
   op.Forward({std::make_shared<const TensorType>(data)}, prediction);
 
   // test correct values
-  ASSERT_TRUE(prediction.AllClose(gt, DataType{1e-5f}, DataType{1e-5f}));
+  ASSERT_TRUE(prediction.AllClose(gt, fetch::math::AsType<DataType>(1e-5f),
```
baykaner: more string constructions? I presume it'll still pass; in any case shouldn't we be using function tolerance here rather than 1e-5?

baykaner: similarly throughout this file

libs/ml/tests/unit/ops/avg_pool_1d.cpp
```diff
@@ -52,11 +52,13 @@ TYPED_TEST(AvgPool1DTest, forward_test_3_2_2)
   {
     for (SizeType i{0}; i < 10; ++i)
     {
-      data(0, i, i_b) = static_cast<DataType>(data_input[i]) + static_cast<DataType>(i_b * 10);
+      data(0, i, i_b) = fetch::math::AsType<DataType>(data_input[i]) +
+                        fetch::math::AsType<DataType>(static_cast<double>(i_b * 10));
```
baykaner: all of these tests that fill a tensor indirectly via a std::vector<double> seem pretty wrong to me. I think they can all be re-written only using Tensor<DataType> (which avoids this bizarre double casting)

libs/ml/tests/unit/ops/convolution_1d.cpp
```diff
@@ -289,8 +290,10 @@ TYPED_TEST(Convolution1DTest, backward_3x3x2_5x3x3x2)
   ASSERT_EQ(prediction.at(1).shape(), kernels.shape());
 
   // Test correct values
-  ASSERT_TRUE(prediction[0].AllClose(gt1, DataType{1e-5f}, DataType{1e-5f}));
-  ASSERT_TRUE(prediction[1].AllClose(gt2, DataType{1e-5f}, DataType{1e-5f}));
+  ASSERT_TRUE(prediction[0].AllClose(gt1, fetch::math::AsType<DataType>(1e-5f),
+                                     fetch::math::AsType<DataType>(1e-5f)));
+  ASSERT_TRUE(prediction[1].AllClose(gt2, fetch::math::AsType<DataType>(1e-5f),
```
baykaner: function tolerance again?

libs/ml/tests/unit/ops/prelu_op.cpp
```diff
@@ -91,7 +91,8 @@ TYPED_TEST(PReluOpTest, backward_test)
       op.Backward({std::make_shared<TypeParam>(data), std::make_shared<TypeParam>(alpha)}, error);
 
   // test correct values
-  ASSERT_TRUE(prediction[0].AllClose(gt, DataType(1e-5), DataType(1e-5)));
+  ASSERT_TRUE(prediction[0].AllClose(gt, fetch::math::AsType<DataType>(1e-5),
```
baykaner: more function tolerance?

libs/ml/tests/unit/ops/tanh.cpp
```diff
@@ -51,22 +52,23 @@ TYPED_TEST(TanHTest, forward_all_positive_test)
...
   }
 
   fetch::ml::ops::TanH<TypeParam> op;
 
   TypeParam prediction(op.ComputeOutputShape({std::make_shared<const TypeParam>(data)}));
   op.Forward({std::make_shared<const TypeParam>(data)}, prediction);
 
-  ASSERT_TRUE(
-      prediction.AllClose(gt, typename TypeParam::Type(1e-4), typename TypeParam::Type(1e-4)));
+  ASSERT_TRUE(prediction.AllClose(gt, fetch::math::AsType<DataType>(1e-4),
```
baykaner: these can be string constructions on 0.0001

libs/ml/tests/unit/ops/tanh.cpp
```diff
@@ -196,8 +200,8 @@ TYPED_TEST(TanHTest, saveparams_backward_all_negative_test)
 
   for (uint64_t i(0); i < n; ++i)
   {
-    data.Set(i, typename TypeParam::Type(dataInput[i]));
-    error.Set(i, typename TypeParam::Type(errorInput[i]));
+    data.Set(i, fetch::math::AsType<DataType>(dataInput[i]));
+    error.Set(i, fetch::math::AsType<DataType>(errorInput[i]));
```
baykaner: errorInput and dataInput is wrong style (should be error_input and data_input)

libs/vm-modules/src/math/random.cpp
```diff
@@ -53,7 +53,7 @@ fetch::meta::IfIsFloat<T, T> Rand(VM *vm, T const &a = T{.0}, T const &b = T{1.0
   if (a >= b)
   {
     vm->RuntimeError("Invalid argument: rand(a, b) must satisfy a < b");
-    return T{.0};
+    return fetch::math::AsType<T>(0.0);
```
baykaner: wouldn't T{0} work for this anyway?

libs/vm-modules/src/math/random.cpp
```diff
@@ -68,15 +68,15 @@ fetch::math::meta::IfIsNotFixedPoint128<T, T> Rand(VM *vm, T const &a = T{.0}, T
...
   }
 
   std::random_device rd;
   std::mt19937_64    mt(rd());
 
   auto a_dbl = static_cast<double>(a);
   auto b_dbl = static_cast<double>(b);
-  return static_cast<T>(std::uniform_real_distribution<double>{a_dbl, b_dbl}(mt));
+  return fetch::math::AsType<T>(std::uniform_real_distribution<double>{a_dbl, b_dbl}(mt));
```
baykaner, line 79: This is under IFIsNotFixedPoint SFINAE so we KNOW that T(double) will work just fine. So we don't need AsType, right?

libs/vm-modules/src/math/random.cpp
```diff
@@ -96,8 +96,8 @@ IfIsPtrFixed128<T, Ptr<T>> Rand(VM *vm, Ptr<T> const &a, Ptr<T> const &b)
 
   auto a_dbl = static_cast<double>(a->data_);
   auto b_dbl = static_cast<double>(b->data_);
-  auto x =
-      static_cast<fixed_point::fp128_t>(std::uniform_real_distribution<double>{a_dbl, b_dbl}(mt));
+  auto x     = fetch::math::AsType<fixed_point::fp128_t>(
+      std::uniform_real_distribution<double>{a_dbl, b_dbl}(mt));
```
baykaner, line 100: this method is not safe anyway right? (i.e. AsFP128 will return a random FP128 constructed via double). Do we have a Jira task for fixing this? If not can you please raise one.

libs/vm-modules/include/vm_modules/math/tensor/tensor_estimator.hpp
```diff
@@ -275,6 +275,9 @@ class TensorEstimator
   static ChargeAmount ToChargeAmount(fixed_point::fp64_t const &val);
 
   VMObjectType &tensor_;
+
+  DataType padded_size_;
```
baykaner: I think you should ditch this.

I understand that this reduces the number of lines of code in the estimator implementation but it also means that the estimator is doing some of the actual work of the tensor, which is fundamentally not what estimators should be doing (i.e. you would need to update these values when you reshape/resize the tensor, and now you have to ensure estimator always tracks its tensor size/shape properly). You would also need to serialise and deserialise these two values (which I think you missed?).

libs/vm-modules/src/math/tensor/tensor_estimator.cpp
```diff
@@ -42,7 +42,11 @@ using SizeVector = ArrayType::SizeVector;
 
 TensorEstimator::TensorEstimator(VMTensor &tensor)
   : tensor_{tensor}
-{}
+{
```
baykaner: I think you should revert all of this.

If you don't like the repeated  code, feel free to add a helper function in an anonymous namespace for getting the padded_size and size since those lines are repeated a lot

libs/vm-modules/src/ml/model/model_estimator.cpp
```diff
@@ -65,8 +65,8 @@ ModelEstimator &ModelEstimator::operator=(ModelEstimator &&other) noexcept
 ChargeAmount ModelEstimator::LayerAddDense(Ptr<String> const &layer, SizeType const &inputs,
                                            SizeType const &hidden_nodes)
 {
-  SizeType size{0};
-  SizeType padded_size{0};
+  DataType size{0};
```
baykaner: Don't think there's any point making any changes to vmmodules/ml estimators. These will all be ditched in favour of a better mechanism that @vhavrylov-ncube is working on

baykaner (changes requested):
There are a lot of tests where a tensor is filled with data from a std::vector<double>
If you don't feel like fixing all of those in this PR please raise an issue for one of us to fix them

libs/dmlf/tests/execution/basic_engine_tests.cpp
markos: fixed.

libs/math/benchmark/trigonometry/trigonometry.cpp
markos: fixed.

libs/ml/benchmark/embeddings/embeddings.cpp
markos: fixed.

libs/ml/include/ml/ops/abs.hpp
markos: fixed.

libs/vm-modules/src/math/random.cpp
markos: fixed

libs/vm-modules/src/math/random.cpp
markos: we would have to replace the whole RNG used in this case instead of the std one. I'm fine with that and it would also be more deterministic -eg. in case C++ spec decide to use another RNG underneath. I will create a ticket for it.

libs/vm-modules/include/vm_modules/math/tensor/tensor_estimator.hpp
markos: ok, I'll put them back in the methods

libs/ml/tests/unit/ops/tanh.cpp
markos: fixed

libs/vm-modules/src/math/random.cpp
markos: this is actually IfIsNotFixedPoint128, so it's a FixedPoint, but not a FixedPoint128, which is handled below, a bit confusing name for a SFINAE I agree.

libs/ml/tests/unit/ops/tanh.cpp
markos: fixed

libs/vm-modules/include/vm_modules/math/tensor/tensor_estimator.hpp
markos: fixed

libs/vm-modules/src/ml/model/model_estimator.cpp
markos: done

libs/vm-modules/src/math/tensor/tensor_estimator.cpp
markos: done

libs/vm-modules/src/math/random.cpp
markos: Ml-436 added.

libs/dmlf/include/dmlf/collective_learning/utilities/utilities.hpp
markos: used doc["learning_rate"].As<DataType>() instead

libs/math/include/math/distance/absolute.hpp
```diff
@@ -30,31 +30,31 @@ typename ArrayType::Type AbsoluteDistance(ArrayType const &A, ArrayType const &B
...
   auto it2   = B.cbegin();
   assert(it1.size() == it2.size());
-  auto ret = Type(0);
+  auto ret = Type{0};
   Type d;
 
   while (it1.is_valid())
   {
-    if ((*it1 > 0) && (*it2 > 0))
+    if ((*it1 > Type{0}) && (*it2 > Type{0}))
```
vhavrylov-ncube, line 38: As Type{0} is used in many places here, may be 
static Type const ZERO = Type{0}; 
and use it everywhere?

libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -1146,58 +1174,56 @@ TYPED_TEST(TranscendentalTest, Exp)
...
   // TypeParam extreme{22741975.0640760175883770};
 
   double    r;
   double    delta, max_error = 0, avg_error = 0;
-  TypeParam x, y, scalex, scaley, margin, tolerance;
-  scalex    = 5.0;
-  margin    = 1.0;
-  tolerance = TypeParam::TOLERANCE;
+  TypeParam x, y;
+  double    scalex, scaley, margin;
```
vhavrylov-ncube, line 1216: "double const" with immediate init would be great if you do not change the values.

libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -1212,25 +1238,24 @@ TYPED_TEST(TranscendentalTest, Pow_positive_x_gt_1)
...
 
 TYPED_TEST(TranscendentalTest, Pow_positive_x_lt_1)
 {
   double    r;
-  double    delta, max_error = 0, avg_error = 0;
-  TypeParam x, y, scalex, scaley, margin, tolerance;
-  scalex    = 1.0;
-  margin    = 0.001;
-  tolerance = TypeParam::TOLERANCE;
+  double    delta, max_error = 0, avg_error = 0, scaley;
```
vhavrylov-ncube, line 1247: doube const if the value is not changed

libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -1251,47 +1276,45 @@ TYPED_TEST(TranscendentalTest, Pow_positive_x_lt_1)
...
+  EXPECT_NEAR(static_cast<double>(e1) / std::pow(static_cast<double>(a), 2), 1.0, tolerance);
+  EXPECT_NEAR(static_cast<double>(e2) / std::pow(static_cast<double>(a), 3), 1.0, tolerance);
   EXPECT_TRUE(TypeParam::IsNaN(TypeParam::Pow(a, a)));
   double    r;
-  double    delta, max_error = 0, avg_error = 0;
-  TypeParam x, y, scalex, scaley, margin, tolerance;
-  scalex    = 10.0;
-  margin    = 0.0001;
-  tolerance = TypeParam::TOLERANCE;
+  double    delta, max_error = 0, avg_error = 0, scaley;
```
vhavrylov-ncube, line 1290: doube const if the value is not changed

libs/math/tests/unit/fixed_point/fixed_point.cpp
```diff
@@ -1664,97 +1688,101 @@ TYPED_TEST(TrigonometryTest, Tan)
...
 TYPED_TEST(TrigonometryTest, ASin)
 {
   double    r;
   double    delta, max_error = 0, avg_error = 0;
   TypeParam x, scale, margin;
-  scale  = TypeParam::_1;
-  margin = TypeParam::TOLERANCE * 10;
+  scale          = TypeParam::_1;
+  margin         = TypeParam::TOLERANCE * TypeParam{10};
+  auto tolerance = static_cast<double>(TypeParam::TOLERANCE);
```
vhavrylov-ncube, line 1701: auto const

libs/ml/tests/unit/ops/activations/sigmoid.cpp
```diff
@@ -50,7 +50,8 @@ TYPED_TEST(SigmoidTest, forward_test)
   op.Forward({std::make_shared<const TensorType>(data)}, prediction);
 
   // test correct values
-  ASSERT_TRUE(prediction.AllClose(gt, DataType{1e-5f}, DataType{1e-5f}));
+  ASSERT_TRUE(prediction.AllClose(gt, fetch::math::Type<DataType>("0.00001"),
```
vhavrylov-ncube, line 53: good catch!

libs/ml/tests/unit/ops/avg_pool_1d.cpp
```diff
@@ -131,48 +137,48 @@ TYPED_TEST(AvgPool1DTest, backward_test_2_channels)
...
-  gt(1, 1, 1) = DataType{0.0f};
-  gt(1, 2, 0) = DataType{2.25f};
-  gt(1, 2, 1) = DataType{0.0f};
-  gt(1, 3, 0) = DataType{2.25f};
-  gt(1, 3, 1) = DataType{0.0f};
-  gt(1, 4, 0) = DataType{1.25f};
-  gt(1, 4, 1) = DataType{0.0f};
+  gt(0, 0, 0) = DataType{1} / DataType{2};
+  gt(0, 0, 1) = DataType{0};
+  gt(0, 1, 0) = DataType{5} / DataType{4};
```
vhavrylov-ncube, line 164: well, division of an integer by 2 or by 4 would result in xx.5 or xx.25 floats, which could be represented in fixed point with 100% accuracy, so DataType{5 / 4} is the just same (in this particular case!), right?

libs/ml/tests/unit/ops/loss_functions/softmax_cross_entropy_loss.cpp
```diff
@@ -47,7 +47,7 @@ TYPED_TEST(SoftmaxCrossEntropyTest, perfect_match_forward_test)
   // these are not logits - a softmax will get called on this
   data1.At(0, 0) = DataType{0};
   data1.At(1, 0) = DataType{0};
-  data1.At(2, 0) = DataType{999999.};
+  data1.At(2, 0) = DataType{999999};
```
vhavrylov-ncube, line 50: a great catch!

vhavrylov-ncube (commented):
looks good to me, but let somebody else go through this ))

libs/ml/tests/unit/ops/embeddings.cpp
```diff
@@ -203,8 +203,7 @@ TYPED_TEST(EmbeddingsTest, saveparams_test)
   new_op.Forward({std::make_shared<TensorType const>(input)}, new_prediction);
 
   // test correct values
-  EXPECT_TRUE(
-      new_prediction.AllClose(prediction, static_cast<DataType>(0), static_cast<DataType>(0)));
+  EXPECT_TRUE(new_prediction.AllClose(prediction, DataType{0}, DataType{0}));
```
baykaner, line 206: ideally all of these static_cast<DataType>(0) -> DataType{0} should be done in a separate PR.
This PR already has to be large and hence tricky to review, but it's made even larger with all these tidying up changes.

libs/vm-modules/src/math/tensor/tensor.cpp
```diff
@@ -75,7 +75,8 @@ void VMTensor::Bind(Module &module, bool const enable_experimental)
   using Index = fetch::math::SizeType;
 
   auto tensor_constructor_charge_estimate = [](Ptr<Array<SizeType>> const &shape) -> ChargeAmount {
-    SizeType padded_size = fetch::math::Tensor<DataType>::PaddedSizeFromShape(shape->elements);
+    DataType padded_size =
```
baykaner, line 78: i don't really understand why this change is made? It seems to result in additional casting in the tensor estimator and is anyway a confusing representation of padded_size (size has to be integral anyway)

libs/vm-modules/src/ml/model/model_estimator.cpp
```diff
@@ -65,8 +65,8 @@ ModelEstimator &ModelEstimator::operator=(ModelEstimator &&other) noexcept
 ChargeAmount ModelEstimator::LayerAddDense(Ptr<String> const &layer, SizeType const &inputs,
```
baykaner: there still seems to be changes in this file. I don't understand why you need to modify the estimators at all?

baykaner (changes requested):

baykaner (approved):

PR #2258, opened 2020-01-02
ML-290: added reshape layer binding

iryna-melnyk:
added reshape binding and some tests


libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -1137,6 +1165,46 @@ TEST_F(VMModelTests, model_sequential_flatten_2d_in_2d_out)
...
+                print(prediction.toString());
+
+                return prediction;
+              endfunction
+      )";
+
+  Variant res;
+  ASSERT_TRUE(toolkit.Compile(SRC_METRIC));
+  ASSERT_TRUE(toolkit.Run(&res, ChargeAmount{0}));
+  ASSERT_EQ(stdout.str(),
```
vhavrylov-ncube: I would return a Tensor from etch script and then validate its content rather then rely on console output

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -1137,6 +1165,46 @@ TEST_F(VMModelTests, model_sequential_flatten_2d_in_2d_out)
   EXPECT_TRUE(constructed_shape == expected.shape());
 }
 
+TEST_F(VMModelTests, model_sequential_reshape_3d_in_2d_out)
```
baykaner, line 1235: isn't this 1d in and 2d out?

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -1137,6 +1165,46 @@ TEST_F(VMModelTests, model_sequential_flatten_2d_in_2d_out)
...
+            "0.500000000;"
+            "6.199999999;"
+            "7.099999999;"
+            "7.099999999;"
+            "9.099999999;"
+            "4.000000000;");
+  auto const tensor            = res.Get<Ptr<fetch::vm_modules::math::VMTensor>>();
+  auto const constructed_shape = tensor->shape();
+  fetch::math::Tensor<fetch::fixed_point::fp64_t> expected({6, 1});
+  EXPECT_TRUE(constructed_shape == expected.shape());
```
baykaner, line 1413: test seems pretty reasonable. I would feel a bit more comfortable if there are one or two other tests that just check the shapes of tensor reshape (e.g. 1D -> 3D, 3D -> 2D, 10D->2D, and an invalid reshape option where the sizes are not equal? )

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: nope :) when you call tensor.fromString() method it reinits the tensor so it doesnt matter how to init an initial shape. it could be even {1,2,3,4,5,6,7,8,9} shape. BTW we have a task about binding fromString constructor to avoid these unpleasant initial array creation. In fact, a tensor with shape {2,3,1} is created in this test

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: As for me this form is more human-readable (with the same result) than tensor element-wise comparison. But if you insist...

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: I've added more tests. Take a look please

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -1137,6 +1165,254 @@ TEST_F(VMModelTests, model_sequential_flatten_2d_in_2d_out)
   EXPECT_TRUE(constructed_shape == expected.shape());
 }
 
+TEST_F(VMModelTests, model_sequential_reshape_2d_in_2d_out)
+{
+  static char const *SRC_METRIC = R"(
+          function main() : Tensor
+                          var shape = Array<UInt64>(1);
+                          shape[0] = 1u64;
+                          var x = Tensor(shape);
```
baykaner, line 1174: this part is almost entirely irrelevant; Perhaps we should modify the empty constructor to instantiate an empty tensor. Then you can just do:
var x = Tensor();
x.fromString(string);

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -1137,6 +1165,254 @@ TEST_F(VMModelTests, model_sequential_flatten_2d_in_2d_out)
...
+                          model.add("reshape", to_shape);
+                          model.compile("scel", "adam");
+                          var prediction = model.predict(x);
+
+                          return prediction;
+          endfunction
+      )";
+
+  ASSERT_TRUE(toolkit.Compile(SRC_METRIC));
+  ASSERT_FALSE(toolkit.Run(nullptr, ChargeAmount{0}));
```
baykaner, line 1232: why does this fail? 1, 4 has the same number of elements as the input string right?

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
baykaner: thanks that's great

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: {4,1} != {1,4}

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: Or even better: map tensor fromString constructor to create tensors in etch like so:
var x = Tensor(string);
As far as I remember only constructor from shape is mapped to etch so I need to write this ugly stuff. I think ml tests could be fixed after one of these bindings is implemented

baykaner: that's fine feel free to do it in another PR

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
baykaner: so why doesn't it just reshape your 4,1 tensor into a 1,4 tensor? I don't see why it fails to run?

baykaner (approved):

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: it is an attempt to reshape a tensor of shape 1 by batch_dim into a 4 by batch_dim and batch dim should'n be changed

PR #2259, opened 2020-01-02
Refactor/ml 418 move ml ops implementations to source files (A->C)

baykaner:
Moves implementations of some ops into source files.
Resulted in reduced ml build times on a local test (with 4 threads):
before:  ml  3995.735270
after:     ml  3122.697119


libs/ml/tests/unit/ops/activations/softmax.cpp
```diff
@@ -19,10 +19,9 @@
 #include "core/serializers/main_serializer_definition.hpp"
 #include "gtest/gtest.h"
 #include "math/base_types.hpp"
-#include "ml/ops/activation.hpp"
+#include "ml/ops/activations/softmax.hpp"
```
vhavrylov-ncube, line 22: Do I understand right that this change just removes unnecessary include? I mean, activation.hpp only includes softmax.hpp, and it is the softmax.hpp that we really need here? [good change!]

vhavrylov-ncube (dismissed):

libs/ml/tests/unit/ops/activations/softmax.cpp
baykaner: activation.cpp includes all the activations; here we only need softmax. Plus, the activation.hpp logic has been moved into source

vhavrylov-ncube (dismissed):
LGTM

vhavrylov-ncube (dismissed):
LGTM

vhavrylov-ncube (dismissed):
A good update!

iryna-melnyk:
LGTM

iryna-melnyk (approved):
LGTM

PR #2260, opened 2020-01-03
VMModel InputLayer skeleton added

vhavrylov-ncube:
- added a skeleton for InputLayer and etch binding to add it to the Model
- implemented smoke unit tests
- updated layer names map


vhavrylov-ncube:
closed because the background idea was changed.

PR #2261, opened 2020-01-03
Add noisec as vendor

jinmannwong:
Add noisec as vendor for implementing secure channels between two nodes.

LDGR-668


ejfitzgerald (approved):
Seems good to me. I think in the short term we need to bring all the vendor projects under the fetchai project banner, but we can do that at a later date

PR #2262, opened 2020-01-03
ML-442: empty constructor binding

iryna-melnyk:
A tiny change which helps to reduce a number of lines for tensor initialisation in etch:
var x = Tensor();
x.fromString(string);

### DIFF
--- a/libs/vm-modules/src/math/tensor/tensor.cpp
+++ b/libs/vm-modules/src/math/tensor/tensor.cpp
@@ -90,7 +90,6 @@ void VMTensor::Bind(Module &module, bool const enable_experimental)
 
   auto interface =
       module.CreateClassType<VMTensor>("Tensor")
-          .CreateConstructor(&VMTensor::EmptyConstructor)
           .CreateConstructor(&VMTensor::Constructor, tensor_constructor_charge_estimate)
           .CreateSerializeDefaultConstructor([](VM *vm, TypeId type_id) -> Ptr<VMTensor> {
             return Ptr<VMTensor>{new VMTensor(vm, type_id)};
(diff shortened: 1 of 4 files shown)


libs/vm-modules/src/math/tensor/tensor.cpp
```diff
@@ -85,6 +90,7 @@ void VMTensor::Bind(Module &module, bool const enable_experimental)
 
   auto interface =
       module.CreateClassType<VMTensor>("Tensor")
+          .CreateConstructor(&VMTensor::EmptyConstructor)
```
baykaner: you still need to charge for doing that!

Also you should make this an experimental binding

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -1096,9 +1093,7 @@ TEST_F(VMModelTests, model_sequential_flatten_tensor_data)
 {
```
baykaner, line 1093: add some tests for empty tensor construction and interaction (test how it fails when you assign values to it, check you can reshape it and then assign values etc.)

baykaner (changes requested):

libs/vm-modules/src/math/tensor/tensor.cpp
iryna-melnyk: ok, got it

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
iryna-melnyk: Done

baykaner (approved):

PR #2265, opened 2020-01-06
Input layer of a fixed shape for VMModel

vhavrylov-ncube:
- Model can now have an Input layer of a given (fixed) shape
- all other shapes for Dense layers and Ops-layers (Flatten, Dense, ...) are deduced from the single InputLayer shape
- ml node functionality partially transferred to .cpp file from header-only.
- Layer (Nodes) input and output shapes are automatically inferred/deduced using the Input layer shape
- refactored graph compilatioon process
- refactored Dense layer
- added tests


libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -75,43 +75,65 @@ class FullyConnected : public SubGraph<T>
...
+  {
+    assert(!this->expected_slice_input_shapes_.empty());
+    assert(!this->slice_output_shape_.empty());
+    this->nodes_.at(input_)->SetExpectedSliceInputShapes(this->expected_slice_input_shapes_);
+    this->nodes_.at(input_)->SetSliceOutputShape(this->expected_slice_input_shapes_.front());
+    FETCH_LOG_INFO(DESCRIPTOR, "-- Compiling sub-graph ... --");
+
+    // When expected input shape is known, it is possible to compute flattened input
+    // shape of this fully connected layer
+    this->nodes_.at(flat_input_)
```
baykaner: flat_input_ is a confusing naming choice given how it's now being used

vhavrylov-ncube: input_ and flat_input_ could be the same if the real input Tensor is 2-dimensional (e.g. data is plain 1 dim vector) and could be 2 different layers if this Dense is time_distributed (I don't know what does it mean, but the code says so :) 
I think we could name this node "flattened_input_" or alike.

libs/ml/include/ml/model/sequential.hpp
```diff
@@ -47,6 +47,8 @@ class Sequential : public Model<TensorType>
 
   SizeType LayerCount() const;
 
+  void SetExpectedInputShape(std::vector<math::SizeType> const &shape);
```
baykaner: naming: drop the 'expected'.

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -73,8 +95,128 @@ class Ops
     return is_training_;
   }
 
+  virtual void SetSliceOutputShape(Shape const &new_shape)
```
baykaner: I don't like all of these XSliceX names. I understand why you're saying slice, but in general a slice can be across any dimension; you're talking about a very particular way of slicing - batch slicing. Maybe these should all be called SetBatchOutputShape, ComputeBatchOutputShape etc.?

libs/ml/src/core/node.cpp
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/**
+ * @brief SliceOutputShape computes an output shape of the Node, if only 1 data slice (e.g.
+ * with batch size == 1) is provided to the Graph input. If there is no cached output shape,
+ * the method is recursively called until either a cached shape or input Node is encountered.
+ * @return vector of SizeType.
+ */
+template <typename TensorType>
+Shape Node<TensorType>::SliceOutputShape()
```
baykaner: BatchOutputShape

libs/ml/src/core/node.cpp
```diff
@@ -0,0 +1,147 @@
...
+                                        op_ptr_->OutputShapeAsString());
+    }
+
+    return candidate;
+  }
+
+  if (input_nodes_.empty())
+  {
+    FETCH_LOG_INFO(name_.c_str(), "Shape deduction reached a Graph leaf : " + this->name_);
+    return candidate;
```
baykaner: I suspect static analysis won't let you do this pattern anyway:
if (X)
{
  return ...
}

if (Y)
{
  return ...
}

in any case it would be clearer if there were else statements

libs/ml/include/ml/model/sequential.hpp
vhavrylov-ncube: done

libs/ml/include/ml/ops/ops.hpp
vhavrylov-ncube: Batch is a nice name too!

libs/ml/src/core/node.cpp
vhavrylov-ncube: I'll double check static analysis behavior in this case.

libs/ml/src/core/node.cpp
vhavrylov-ncube: done

evsmithx (dismissed):

libs/ml/include/ml/core/graph.hpp
```diff
@@ -811,7 +844,7 @@ void Graph<TensorType>::SetInputReference(std::string const &node_name, TensorTy
 template <typename TensorType>
 void Graph<TensorType>::SetInput(std::string const &node_name, TensorType const &data)
 {
-  SetInputReference(node_name, data.Copy());
+  SetInputReference(node_name, data);
```
vhavrylov-ncube: copying is redundant here, because SetInputReference takes its argument by const ref, and "data" is already a constant ref here.

libs/ml/src/core/node.cpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+    return candidate;
+  }
+
+  // If no input nodes exist - it is impossible to infer/deduce their shapes.
+  if (input_nodes_.empty())
+  {
+    FETCH_LOG_INFO(name_.c_str(), "Shape deduction reached a Graph leaf : " + this->name_);
+    return candidate;
+  }
```
iryna-melnyk: I would suggest to rearrange branching order and put check for emptiness first (because you actually called it more than once which is not necessary IMHO) it may improve the readability

vhavrylov-ncube: done!

libs/ml/include/ml/core/node.hpp
```diff
@@ -397,14 +396,17 @@ void Node<TensorType>::ResetCache(bool input_size_changed)
...
+void Node<TensorType>::SetNodeSaveableParams(NodeSaveableParams<TensorType> const &       nsp,
+                                             std::shared_ptr<ops::Ops<TensorType>> const &op_ptr)
 {
   name_                 = nsp.name;
   cached_output_status_ = CachedOutputState::CHANGED_SIZE;
   operation_type_       = nsp.operation_type;
   op_ptr_               = op_ptr;
 }
 
+std::string OutputShapeAsString(fetch::math::SizeVector const &out_shape);
```
baykaner: these two functions look like utilities.
Firstly, I believe they should be static functions.
Secondly, they should either live only in the cpp (if we're sure they will never be invoked from anywhere except node), or they should move to a more general utilities location (like ml/utilities).

libs/ml/include/ml/layers/PRelu.hpp
```diff
@@ -99,6 +99,15 @@ class PRelu : public SubGraph<T>
   }
 
   static constexpr char const *DESCRIPTOR = "ParametricRelu";
```
baykaner, line 60: presumably we can ditch this and just have the Descriptor function?

libs/ml/include/ml/layers/fully_connected.hpp
```diff
@@ -189,22 +259,66 @@ class FullyConnected : public SubGraph<T>
...
   SizeType in_size_          = fetch::math::numeric_max<SizeType>();
   SizeType out_size_         = fetch::math::numeric_max<SizeType>();
   bool     time_distributed_ = false;
 
-  void Initialise(TensorType &weights, WeightsInit init_mode)
-  {
-    fetch::ml::ops::Weights<TensorType>::Initialise(weights, in_size_, out_size_, init_mode);
-  }
+  // Non-saveable params
+  bool                          is_initialised_ = false;
```
baykaner: all these non-saveable params leave me slightly nervous about whether dense layer serialisation->deserialisation will ALWAYS result in the same behaviour.

For example. is_initialised_ can be true for a dense layer. Serialise and deserialise the layer and the new dense layer has is_initialised_ = false. Couldn't that be a cause of differing behaviour in the layers?

baykaner: the safe (but also less efficient) thing to do is serialised all the parameters - then behaviour is guaranteed; but you're right that there are definitely some parameters that are just effectively internal caching and don't need to be serialised.

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -52,14 +54,34 @@ class Ops
...
+  /**
+   * @brief ComputeBatchOutputShape is expensive function and should be used only for initialisation
+   * or in ASSERT. On Forward you can use output.shape() and on Backward there is
+   * error_signal.shape()
+   * @param input_shapes
+   * @return
+   */
+  virtual std::vector<SizeType> ComputeBatchOutputShape(ShapeVector const &input_shapes)
+  {
+    VecTensorType dummies;
```
baykaner, line 66: inefficient formulation. More efficient formulation is either to construct using input_shape.size() and then emplace back OR use vector.reserve() and then assign dummies[i] = ...
emplace back is probably neater

libs/ml/include/ml/saveparams/saveable_params.hpp
```diff
@@ -37,8 +37,10 @@ struct OpsSaveableParams
 
   virtual ~OpsSaveableParams() = default;
 
-  fetch::ml::OpType op_type     = OpType::NONE;
-  bool              is_training = true;
+  fetch::ml::OpType                    op_type     = OpType::NONE;
+  bool                                 is_training = true;
+  std::vector<fetch::math::SizeVector> batch_input_shapes{};
```
baykaner, line 42: if you're adding these two parameters to saveable params then you need to update the serialisers affected by this right?

libs/ml/src/core/node.cpp
```diff
@@ -0,0 +1,236 @@
...
+ * @brief BatchOutputShape computes an output shape of the Node, if only 1 data slice (e.g.
+ * with batch size == 1) is provided to the Graph input. If there is no cached output shape,
+ * the method is recursively called until either a cached shape or input Node is encountered.
+ * @return vector of SizeType.
+ */
+template <typename TensorType>
+Shape Node<TensorType>::BatchOutputShape()
+{
+  Shape const &candidate = op_ptr_->BatchOutputShape();
+  bool const   op_is_subgraph =
```
baykaner: this is an unnecessary step. you could just:

auto subgraph_ptr = dynamic_pointer_cast<Subgraph>(op_ptr_);

if (!candidate.empty() && subgraph_ptr)
{

}

baykaner: if it's a matter of clarity I would say that the comment on line 84 handles that for you - "if the op is not a subgraph ..."

libs/ml/src/core/node.cpp
```diff
@@ -0,0 +1,236 @@
...
+      if (input_node_ptr->OperationType() != OpType::OP_PLACEHOLDER)
+      {
+        FETCH_LOG_INFO(name_.c_str(),
+                       "Got an empty shape as return from non-placeholder layer! : " +
+                           input_node_ptr->GetNodeName());
+      }
+      continue;
+    }
+
+    input_shapes.emplace_back(in_shape);
```
baykaner: i think it's a little clearer to formulate it like this:

if (!in_shape.empty())
{
  input_shapes.emplace_back(in_shape);
}
else
{
  // error handling...
}

libs/ml/src/core/node.cpp
```diff
@@ -0,0 +1,236 @@
...
+///////////////////////////////
+
+template class Node<math::Tensor<int8_t>>;
+template class Node<math::Tensor<int16_t>>;
+template class Node<math::Tensor<int32_t>>;
+template class Node<math::Tensor<int64_t>>;
+template class Node<math::Tensor<uint8_t>>;
+template class Node<math::Tensor<uint16_t>>;
+template class Node<math::Tensor<uint32_t>>;
+template class Node<math::Tensor<uint64_t>>;
```
baykaner: don't need these

libs/ml/tests/unit/graph/graph_every_op.cpp
```diff
@@ -114,7 +114,83 @@ void ComparePrediction(GraphPtrType g, GraphPtrType g2, std::string node_name)
   EXPECT_TRUE(prediction.AllClose(prediction2, DataType{0}, DataType{0}));
 }
 
-TYPED_TEST(GraphRebuildTest, graph_rebuild_every_op)
+TYPED_TEST(GraphRebuildTest, graph_rebuild_time_distributed_ops)
```
baykaner: why do we separate out these tests? is this to do with time distributed not working with shape deduction currently? will they be merged back together after?

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -1579,4 +1591,88 @@ TEST_F(VMModelTests, model_fit_and_refit)
...
+
+         var data_shape = Array<UInt64>(4);
+         data_shape[0] = num_channels;
+         data_shape[1] = image_height;
+         data_shape[2] = image_width;
+         data_shape[3] = 25u64;
+
+         var model = Model("sequential");
+
+         model.addExperimental("input", data_shape);
```
baykaner, line 1657: so the input layer is working; but we don't yet have the right bindings for dense and conv to allow just output shape specification?
But presumably all the underlying code is there for that, we just need to add a couple of functions to vmmodules sequential, right?

baykaner (changes requested):

libs/ml/include/ml/layers/PRelu.hpp
vhavrylov-ncube: I believe yes, but for now ::DESCRIPTOR is used in a template code 

template \<typename OperationType\>
bool Graph\<TensorType\>::UpdateVariableName(
...
  std::string op_descriptor = (OperationType::DESCRIPTOR);

so I guess we will ditch both DESCRIPTOR and OpType fields and replace them by virtual polymorphic functions after refactoring of GraphBuilder and that code above.

libs/ml/src/core/node.cpp
vhavrylov-ncube: agreed

libs/ml/src/core/node.cpp
vhavrylov-ncube: removed

libs/ml/tests/unit/graph/graph_every_op.cpp
vhavrylov-ncube: well, the reason for separation was - time-distributed ones failed until I fixed them later. Now I just wonder if it would be better to split that "rebuild-all" approach to a smaller groups or even individual tests. But you're right it looks like a separate task, not related to current change, so I'll just merge it back for now.

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
vhavrylov-ncube: model.addExperimental("dense", 6u64);  -- this is already here for output-only Dense declaration; regarding Conv2D - it has only "input channels" parameter, it might be auto-deduced from some input shape of previous layer, but I am not ready to tell if this implementation is trivial.

libs/ml/src/core/node.cpp
vhavrylov-ncube: In this method I've tried to keep the next approach: if something went wrong, we return, else we continue. This way each next instruction is executed only if there were no errors before. Keeping an "else" block will result in less clearer flow here, as it is a bit harder to understand which of the blocks ("if" or "else" one) is for good case.

libs/ml/include/ml/ops/ops.hpp
vhavrylov-ncube: done

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
vhavrylov-ncube: Well, the first leyer in a Model could be also Dropout, Flatten, Reshape or another one, so it moght be a good idea to create interface for adding any type of layer to a Model with ptoviding input shape array as a parameter (as it is possible in Keras), however, it is a large change so I would prefer to keep only Dense as an example and an InputLayer as a default working solution for now.

libs/ml/include/ml/saveparams/saveable_params.hpp
vhavrylov-ncube: oh, I've missed it, sure!

libs/ml/include/ml/layers/fully_connected.hpp
vhavrylov-ncube: I agree. I have removed all fields possible, leaving only what are really necessary to keep, and serialised them.

baykaner (dismissed):

vhavrylov-ncube (commented):
approved from my side

baykaner (approved):

PR #2267, opened 2020-01-06
Improvement/noexcept mainchain

denis-trapeznikov:
`MainChain` class got stripped of the only two `throw` statements it did have.

(Simply rotating program state monad so that it lies in the realm of nullable references rather than in that of exceptionful execution.)


ejfitzgerald (approved):
LGTM

PR #2270, opened 2020-01-07
Added more VMPair tests

MissingNO57:


libs/vm-modules/tests/unit/pair_tests.cpp
```diff
@@ -73,4 +73,136 @@ TEST_F(PairTests, assign_string_u32_test)
...
+    endfunction
+  )";
+
+  ASSERT_TRUE(toolkit.Compile(TEXT));
+  ASSERT_TRUE(toolkit.Run());
+
+  ASSERT_EQ(stdout.str(), "-234.5669,-123.455999999");
+}
+
+TEST_F(PairTests, assign_fixed128_uint256_test)
```
baykaner: this name seems to be incorrect - you are assigning a fixed128 and a string

baykaner (changes requested):

baykaner (approved):

PR #2271, opened 2020-01-07
Refactor/ml 415 move ml ops activations implementations to source files

baykaner:


libs/math/include/math/activation_functions/sigmoid.hpp
```diff
@@ -36,23 +36,19 @@ void Sigmoid(ArrayType const &t, ArrayType &ret)
 
   auto array_it = t.cbegin();
   auto rit      = ret.begin();
-  Type zero{0};
-  Type one{1};
-  Type min_one{-1};
```
vhavrylov-ncube, line 41: as far as I understood, this min_one was not used at all right?

libs/math/tests/unit/activation_functions/sigmoid.cpp
```diff
@@ -158,6 +158,8 @@ TYPED_TEST(SigmoidTest, sigmoid_2x2)
 
   TypeParam output = fetch::math::Sigmoid(array1);
 
+  std::cout << "output.ToString(): " << output.ToString() << std::endl;
```
vhavrylov-ncube: are you sure you want to have this console output?

libs/ml/src/ops/activations/gelu.cpp
```diff
@@ -0,0 +1,146 @@
...
+{
+  FETCH_UNUSED(me);
+  assert(me.get() == this);
+
+  auto copyshare = std::make_shared<MyType>(*this);  // calls default copy constructor of MyType
+
+  return copyshare;
+}
+
+// 0.5x(1+tanh(0.797885x+0.035677x^3))
```
vhavrylov-ncube: I guess it would be great to have some more explanation about what does this formula mean and why is it used.

libs/ml/src/ops/activations/logsoftmax.cpp
```diff
@@ -0,0 +1,119 @@
...
+                                                         TensorType const &   error_signal)
+{
+  assert(inputs.size() == 1);
+  assert(inputs.front()->shape() == error_signal.shape());
+
+  TensorType return_signal = error_signal.Copy();
+  TensorType t(error_signal.shape());
+  fetch::math::Softmax((*inputs.front()), t, axis_);
+
+  // return_signal.InlineMultiply(t);
```
vhavrylov-ncube: looks like an artifact, is it still needed?

baykaner: interesting, i guess not. will ditch it.

libs/ml/src/ops/activations/gelu.cpp
baykaner: ill add a reference to the paper

libs/math/include/math/activation_functions/sigmoid.hpp
baykaner: no, it was used but i've refactored it out

vhavrylov-ncube (dismissed):
LGTM

vhavrylov-ncube (approved):
LGTM

PR #2274, opened 2020-01-07
Do not trim cabinet history after reversion

wkaluza:


libs/ledger/src/consensus/consensus.cpp
```diff
@@ -945,7 +951,10 @@ void Consensus::SetDefaultStartTime(uint64_t default_start_time)
 void Consensus::AddCabinetToHistory(uint64_t block_number, CabinetPtr const &cabinet)
 {
   cabinet_history_[block_number] = cabinet;
-  TrimToSize(cabinet_history_, HISTORY_LENGTH);
+  if (block_number > cabinet_history_.crbegin()->first)
```
ejfitzgerald, line 954: 👍

libs/ledger/src/consensus/consensus.cpp
```diff
@@ -443,7 +450,7 @@ void Consensus::UpdateCurrentBlock(Block const &current)
   }
 
   // Don't try to set previous when we see genesis!
-  if (current.block_number == 0)
+  if (MainChain::IsGenesisBlock(current))
```
ejfitzgerald:
```suggestion
  if (current.IsGenesis())
```

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1797,6 +1797,13 @@ MainChain::IntBlockPtr MainChain::CreateGenesisBlock()
   return genesis;
 }
 
+bool MainChain::IsGenesisBlock(Block const &block)
+{
+  return block.block_number == 0 && block.previous_hash == chain::ZERO_HASH &&
```
ejfitzgerald: This function actually exists in the block - see suggestion below. Would be worth checking with @denis-trapeznikov - currently I believe I think it only checks with a zero hash to signal genesis. But I think that this can be updated now I think to add at least the genesis digest check here.

I would not make this change (altering `IsGenesis()` method) in this PR (for the release branch) but would update when we port this fix to `master`. Thoughts?

ejfitzgerald (changes requested):

libs/ledger/src/chain/main_chain.cpp
wkaluza: Sounds good to me. In particular I think that we should err on the side of caution and check everything that is set in `MainChain::CreateGensisBlock()`, plus the block number being zero. Denis?

libs/ledger/src/consensus/consensus.cpp
wkaluza: Done.

ejfitzgerald (approved):

PR #2280, opened 2020-01-08
Improvement/expo backtrace

denis-trapeznikov:
This commit introduces very simple saturated exponential walkback in MainChain RPC service. Leaps are limited to be at most 16K blocks back, consequently, no logarithmic refinement attempted after we're back on a known chain.

It also now contains changes from #2249, to not add something only in order to later merge it (since that latter is now in a conflicted state, it can be discarded probably, with this PR to ~~rule~~ merge them all).


libs/core/include/core/containers/vector_util.hpp
```diff
@@ -0,0 +1,30 @@
...
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+namespace fetch {
+namespace core {
+
+template <class Left, class Right>
+void Append(Left &left, Right &&right)
```
ejfitzgerald: Personally I think I would call this extend, either way can we change the name of the file to reflect the function name

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -1998,40 +2006,31 @@ DigestSet MainChain::DetectDuplicateTransactions(BlockHash const &           sta
...
-      {
-        break;
-      }
-
       for (auto const &slice : block->slices)
       {
         for (auto const &tx : slice)
         {
-          if (transaction_digests.find(tx.digest()) != transaction_digests.end())
+          if (potential_duplicates.count(tx.digest()) > 0)
```
ejfitzgerald: I have been bitten with the performance of count in the past, worth double checking

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -463,6 +436,7 @@ State MainChainRpcService::OnWaitForBlocks()
 
     state_machine_->Delay(std::chrono::milliseconds{100 * consecutive_failures_});
     return State::REQUEST_NEXT_BLOCKS;
+  default:;
```
ejfitzgerald: I would remove default case, otherwise you won’t get compiler warning when the enum is changed

libs/ledger/src/upow/naive_synergetic_miner.cpp
```diff
@@ -135,7 +135,7 @@ void NaiveSynergeticMiner::Mine()
     std::ostringstream oss;
     for (auto const &element : problem_spaces)
     {
-      oss << "\n -> 0x" << element.first.contract_address.display();
+      oss << "\n -> 0x" << element.first.display().ToHex();
```
ejfitzgerald: I guess we talked about this in previous PR

libs/ledger/tests/chain/main_chain_service_tests.cpp
```diff
@@ -329,16 +462,15 @@ TEST_F(MainChainServiceTests, CheckWaitingToFullfilResponse)
   EXPECT_CALL(rpc_client_, TimeTravel(other1_, GetGenesisDigest()))
       .WillOnce(Return(TraveloguePromise{promise}));
 
-  Tick(State::SYNCHRONISING, State::START_SYNC_WITH_PEER);
```
ejfitzgerald, line 342: While I know you abhor duplication, being able to quickly comment stages in the pipeline is very easy with the old structure. Let’s discuss over tea in the office

ejfitzgerald (changes requested):

libs/ledger/tests/chain/main_chain_service_tests.cpp
```diff
@@ -474,14 +628,147 @@ TEST_F(MainChainServiceTests, CheckWhenGenesisAppearsToBeInvalid)
...
+MainChainProtocol::Travelogue TimeTravel(BlockGenerator::BlockPtr const &heaviest_block,
+                                         BlockGenerator::BlockPtrs       local_blocks)
+{
+  return {heaviest_block->hash, heaviest_block->block_number, TravelogueStatus::HEAVIEST_BRANCH,
+          std::move(local_blocks)};
+}
+
+}  // namespace
+
+TEST_F(MainChainServiceTests, CheckExponentialBackStep)
```
ejfitzgerald: How long does this test take to run, does it make sense to either move it or make it part of part of the integration test suite?

libs/ledger/tests/chain/mock_main_chain_rpc_client.hpp
```diff
@@ -29,5 +29,14 @@ class MockMainChainRpcClient : public fetch::ledger::MainChainRpcClientInterface
 
   MOCK_METHOD2(GetHeaviestChain, BlocksPromise(MuddleAddress, uint64_t));
   MOCK_METHOD4(GetCommonSubChain, BlocksPromise(MuddleAddress, Digest, Digest, uint64_t));
+  /*
+  virtual TraveloguePromise TimeTravel(MuddleAddress peer, Digest start) override
```
ejfitzgerald: ?

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -54,6 +54,8 @@ using PromiseState           = fetch::service::PromiseState;
 using State                  = MainChainRpcService::State;
 using Mode                   = MainChainRpcService::Mode;
 
+constexpr uint64_t max_sensible_step_back = 10000;
```
ejfitzgerald:
```suggestion
constexpr uint64_t MAX_SENSIBLE_STEP_BACK = 10000;
```

scripts/install-test-dependencies.sh
```diff
@@ -4,7 +4,7 @@ set -e
 pip3 install wheel --force-reinstall
 pip3 install pyyaml requests --force-reinstall
 pip3 install lark-parser==0.7.8 --force-reinstall
-pip3 install fetchai-ledger-api==0.11.0a4 --force-reinstall
+pip3 install fetchai-ledger-api==1.0.0rc1 --force-reinstall
```
ejfitzgerald, line 7: I think that we revert this because we now have solved the underlying issue

ejfitzgerald (approved):

PR #2281, opened 2020-01-08
ML-420: move templated ops...M-O

iryna-melnyk:
NOTE: there is a place in newly created one_hot.cpp file, where includes should be in specific order which differs from autoformatted one:

#include "ml/ops/one_hot.hpp"
#include "math/one_hot.hpp"
UPDATE: the issue was resolved


libs/ml/include/ml/ops/mask_fill.hpp
```diff
@@ -40,76 +38,32 @@ class MaskFill : public Ops<T>
...
-  {
-    FETCH_UNUSED(me);
-    assert(me.get() == this);
-
-    auto copyshare = std::make_shared<MyType>(*this);  // calls default copy constructor of MyType
-
-    return copyshare;
-  }
+      std::shared_ptr<fetch::ml::ops::Ops<TensorType>> me) override;
   /**
```
baykaner: generally i preferred to move these descriptions to the cpp as well; that way the hpp is very concise and clear, and if you want more details they're all in one place (the cpp)

libs/ml/include/ml/ops/matrix_multiply.hpp
```diff
@@ -46,22 +42,11 @@ class MatrixMultiply : public fetch::ml::ops::Ops<T>
...
   ~MatrixMultiply() override = default;
 
-  std::shared_ptr<OpsSaveableParams> GetOpSaveableParams() override
-  {
-    auto ret         = std::make_shared<SPType>();
-    ret->transpose_a = transpose_a_;
-    ret->transpose_b = transpose_b_;
-    return ret;
-  }
+  std::shared_ptr<OpsSaveableParams> GetOpSaveableParams() override;
```
baykaner, line 49: it looks like you forgot to move the implementation for MakeSharedCopy into source?

libs/ml/src/ops/matrix_multiply.cpp
```diff
@@ -0,0 +1,418 @@
...
+
+template <class T>
+MatrixMultiply<T>::MatrixMultiply(const SPType &sp)
+  : Ops<T>(sp)
+{
+  transpose_a_ = sp.transpose_a;
+  transpose_b_ = sp.transpose_b;
+}
+
+template <class T>
```
baykaner: generally I have been using template <typename TensorType> rather than <class T> but this is mostly a matter of style. In this context class vs. typename makes almost no difference (unless we do a template template, but we are not going to do that!!!)

libs/ml/src/ops/one_hot.cpp
```diff
@@ -0,0 +1,105 @@
...
+  }
+  else
+  {
+    shape.insert(shape.begin() + static_cast<math::PtrDiffType>(axis_), depth_);
+  }
+
+  return shape;
+}
+
+}  // namespace ops
```
baykaner, line 121: missing explicit instantiations for all cpp files

baykaner (changes requested):

libs/ml/include/ml/ops/matrix_multiply.hpp
iryna-melnyk: Possibly so

libs/ml/src/ops/matrix_multiply.cpp
iryna-melnyk: Did you mean typename?

iryna-melnyk: Ah, I see - was confused with first sentence

wkaluza:
> NOTE: there is a place in newly created one_hot.cpp file, where includes should be in specific order which differs from autoformatted one:
> 
> #include "ml/ops/one_hot.hpp"
> #include "math/one_hot.hpp"

I'm a little alarmed by this. What is the reason for the need to `#include` these headers in a specific order? The style checker can be appeased if there is a valid reason for this, but given that these are both headers we own, there may well not be.

Headers that depend on other headers being included are a maintenance nightmare. We should discuss and possibly fix this as tech debt.

iryna-melnyk:
> I'm a little alarmed by this. What is the reason for the need to `#include` these headers in a specific order? The style checker can be appeased if there is a valid reason for this, but given that these are both headers we own, there may well not be.
> 
> Headers that depend on other headers being included are a maintenance nightmare. We should discuss and possibly fix this as tech debt.

We already had a discussion with Khan, this issue could be fixed by including base_types in math/one_hot.hpp.
Basically, a separation of definitions and declarations of one_hot.hpp revealed this issue. It wasn't a problem at all before the refactoring. The reason was 'SizeType' is undefined in math/one_hot.hpp. On the other hand ops/one_hot.hpp has a definition of 'SizeType'. AFAIR the compiler resolves includes in order they are presented. And at the beginning math/one_hot.hpp was out of scope of my ticket.

libs/ml/include/ml/ops/matrix_multiply.hpp
iryna-melnyk: Done

libs/ml/src/ops/matrix_multiply.cpp
iryna-melnyk: Done

libs/ml/include/ml/ops/mask_fill.hpp
iryna-melnyk: Done

libs/ml/src/ops/one_hot.cpp
iryna-melnyk: Well, I did the same explicit instantiations for all except MaxPool1D. I will leave a separate comment there

libs/ml/src/ops/max_pool_1d.cpp
```diff
@@ -0,0 +1,200 @@
...
+///////////////////////////////
+/// EXPLICIT INSTANTIATIONS ///
+///////////////////////////////
+
+// template class MaxPool1D<math::Tensor<int8_t>>;
+// template class MaxPool1D<math::Tensor<int16_t>>;
+template class MaxPool1D<math::Tensor<int32_t>>;
+template class MaxPool1D<math::Tensor<int64_t>>;
+// template class MaxPool1D<math::Tensor<uint8_t>>;
+// template class MaxPool1D<math::Tensor<uint16_t>>;
```
iryna-melnyk: These were commented because of an error (implicit conversion loses integer precision)

iryna-melnyk: This was resolved by adding a static cast to tensor.cpp file

libs/math/include/math/tensor/tensor.hpp
```diff
@@ -820,7 +820,7 @@ void Tensor<T, C>::Set(Args... args)
   }
 
   uint64_t index = TensorSetter<0, Args...>::IndexOf(stride_, shape_, std::forward<Args>(args)...);
-  Type     value = TensorSetter<0, Args...>::ValueOf(std::forward<Args>(args)...);
+  Type value = static_cast<Type>(TensorSetter<0, Args...>::ValueOf(std::forward<Args>(args)...));
```
baykaner: this will fail static analysis i think:
auto value = static_cast<Type>(...)

iryna-melnyk: Yep, you're correct

libs/ml/tests/unit/ops/max_pool.cpp
```diff
@@ -19,6 +19,7 @@
 #include "core/serializers/main_serializer_definition.hpp"
 #include "math/base_types.hpp"
 #include "ml/ops/max_pool.hpp"
+#include "ml/ops/max_pool_2d.hpp"
```
iryna-melnyk, line 22: I don't want to move back ml/ops/max_pool_2d.hpp to ml/ops/max_pool.hpp but if it is needed please let me know

baykaner (dismissed):

baykaner (dismissed):

baykaner (dismissed):

baykaner (dismissed):

baykaner (approved):

PR #2284, opened 2020-01-09
Conv1D and 2D input size specification

MissingNO57:


MissingNO57:
Will be fixed in #2265, which made this PR redundant.

PR #2285, opened 2020-01-09
moved tsne implementation to source

baykaner:


libs/ml/src/clustering/tsne.cpp
```diff
@@ -0,0 +1,497 @@
...
+}
+
+///////////////////////////////
+/// EXPLICIT INSTANTIATIONS ///
+///////////////////////////////
+
+template class TSNE<math::Tensor<int8_t>>;
+template class TSNE<math::Tensor<int16_t>>;
+template class TSNE<math::Tensor<int32_t>>;
+template class TSNE<math::Tensor<int64_t>>;
```
iryna-melnyk: what about unsigned types?

baykaner: initially I ignored these because the code has a DataType{-2} in it; but I found a way to re-arrange the math to not require that, so I can include unsigned ints. Huzzah!

iryna-melnyk (approved):
looks good to me

PR #2287, opened 2020-01-10
Refactor/ml 429 move ml regularisers implementation to source files

baykaner:


libs/ml/src/ops/embeddings.cpp
```diff
@@ -17,7 +17,7 @@
 //------------------------------------------------------------------------------
 
 #include "ml/ops/embeddings.hpp"
-#include "ml/saveparams/saveable_params.hpp"
+//#include "ml/saveparams/saveable_params.hpp"
```
iryna-melnyk: it seems you forgot to remove it

iryna-melnyk (approved):
looks good

PR #2289, opened 2020-01-10
moved ml/core implementation to source

AlexTeslenko:
moved ml/core implementation to source


libs/ml/include/ml/core/node.hpp
```diff
@@ -153,258 +153,5 @@ class Node
   std::shared_ptr<ops::Ops<TensorType>> op_ptr_;
```
baykaner, line 157: There seems to be several methods you have not moved over to source file

libs/ml/include/ml/core/graph.hpp
```diff
@@ -275,747 +275,6 @@ std::string Graph<TensorType>::AddNode(std::string const &             node_name
   return updated_name;
```
baykaner, line 270: you should move as many #includes from hpp to cpp as possible. Use forward declarations where necessary

baykaner (changes requested):

libs/ml/include/ml/core/graph.hpp
```diff
@@ -275,747 +275,6 @@ std::string Graph<TensorType>::AddNode(std::string const &             node_name
...
 }
 
-/**
```
iryna-melnyk, line 279: comment to line 244: I guess AddNode should be also moved to cpp file, right?

baykaner: some of them are extremely difficult or impossible to move (e.g. functions that include a parameter pack will be in general not possible to move, and templated functions will be extremely verbose to move)

libs/ml/include/ml/core/node.hpp
AlexTeslenko: Maybe it is better to leave such one-line methods in header file and just make them inline?

libs/ml/include/ml/core/graph.hpp
```diff
@@ -18,20 +18,9 @@
 //------------------------------------------------------------------------------
 
 #include "ml/core/node.hpp"
-#include "ml/meta/ml_type_traits.hpp"
 #include "ml/ops/weights.hpp"
 #include "ml/state_dict.hpp"
```
baykaner, line 23: I think you can forward declare this and just include it in the source file?

libs/ml/include/ml/core/graph.hpp
```diff
@@ -18,20 +18,9 @@
 //------------------------------------------------------------------------------
 
 #include "ml/core/node.hpp"
-#include "ml/meta/ml_type_traits.hpp"
 #include "ml/ops/weights.hpp"
```
baykaner: I think weights are not used anywhere in the header - so can be deleted or moved to source

libs/ml/include/ml/core/node.hpp
```diff
@@ -20,16 +20,6 @@
 #include "logging/logging.hpp"
 #include "ml/ops/ops.hpp"
 #include "ml/ops/weights.hpp"
```
baykaner: i think weights include can be moved to source? I don't see any usage of weights in the header

libs/ml/src/core/graph.cpp
```diff
@@ -19,7 +19,18 @@
 #include "math/tensor/tensor.hpp"
 #include "math/tensor/tensor_slice_iterator.hpp"
 #include "ml/core/graph.hpp"
-
+#include "ml/layers/normalisation/layer_norm.hpp"
```
baykaner: this looks like it definitely should not be here

libs/ml/src/core/graph.cpp
```diff
@@ -19,7 +19,18 @@
...
+#include "ml/meta/ml_type_traits.hpp"
+#include "ml/ops/embeddings.hpp"
```
baykaner: also this should not be here

baykaner (changes requested):

libs/ml/include/ml/core/graph.hpp
AlexTeslenko: If I just forward declare StateDict, then I get this error:
error: implicit instantiation of undefined template 'fetch::ml::StateDict<fetch::math::Tensor<signed char, fetch::memory::SharedArray<signed char, 1> > >'
  virtual fetch::ml::StateDict<TensorType> StateDict();

Cannot figure out if it is possible to fix it without including "ml/state_dict.hpp"
                                           ^

libs/ml/include/ml/core/graph.hpp
AlexTeslenko: If I remove "ml/ops/weights" then I need include these headers:

 #include "ml/regularisers/regulariser.hpp"
#include "ml/exceptions/exceptions.hpp"

baykaner (approved):

PR #2291, opened 2020-01-10
implemented small integer types for used BLAS routines

baykaner:


vhavrylov-ncube (dismissed):
LGTM

libs/math/src/linalg/blas/gemm_nt_vector.cpp
```diff
@@ -117,6 +117,38 @@ void Blas<S, Signature(_C <= _alpha, _A, _B, _beta, _C),
...
+                    platform::Parallelisation::VECTORISE>;
+
+template class Blas<int32_t, Signature(_C <= _alpha, _A, _B, _beta, _C),
+                    Computes(_C <= _alpha * _A * T(_B) + _beta * _C),
+                    platform::Parallelisation::VECTORISE>;
+
+template class Blas<int64_t, Signature(_C <= _alpha, _A, _B, _beta, _C),
+                    Computes(_C <= _alpha * _A * T(_B) + _beta * _C),
+                    platform::Parallelisation::VECTORISE>;
+
```
iryna-melnyk, line 127: I don't get why one part of blas files contains explicit instantiation of Blas<fetch::fixed_point::FixedPoint<64, 64> and other doesn't. Please add it if needed

iryna-melnyk (dismissed):

libs/math/src/linalg/blas/gemm_nt_vector.cpp
baykaner: because there is no vectorisation support for fp128. So all files named *_vector.cpp do not have this

markos (dismissed):
Looks good

vhavrylov-ncube (approved):
Only lint-related changes since previous approval, reapproved.

PR #2294, opened 2020-01-13
VMTensor string constructor with charge estimate, benchmarks and tests

MissingNO57:


libs/vm-modules/src/math/tensor/tensor.cpp
```diff
@@ -83,9 +95,19 @@ void VMTensor::Bind(Module &module, bool const enable_experimental)
            COMPUTE_CHARGE_COST;
   };
 
+  auto tensor_string_constructor_charge_estimate = [](Ptr<String> const &str) -> ChargeAmount {
+    DataType size = static_cast<DataType>(str->string().size());
```
vhavrylov-ncube: auto const - becase you use static_cast, and it is a constant :)

baykaner: @MissingNO57 I probably missed something obvious, but why does this charge estimator use size and the previous constructor charge estimate use padded_size?

libs/vm-modules/src/math/tensor/tensor.cpp
```diff
@@ -538,6 +560,9 @@ TensorEstimator &VMTensor::Estimator()
 fixed_point::fp64_t const VMTensor::CONSTRUCTION_PADDED_SIZE_COEF = fixed_point::fp64_t("0.0028");
 fixed_point::fp64_t const VMTensor::CONSTRUCTION_CONST_COEF       = fixed_point::fp64_t("22");
 
+fixed_point::fp64_t const VMTensor::CONSTRUCTION_STRING_SIZE_COEF  = fixed_point::fp64_t("0.12");
```
baykaner, line 587: where are these coefficients coming from? if they're based on some benchmarking that you've done locally please make sure that work goes into the ledger-vm-misc repo and maybe we should add a comment here with a link to where the supporting benchmarking can be found justifying these constants?

libs/vm-modules/src/math/tensor/tensor.cpp
MissingNO57: This constructor uses Size of string, but previous one was using size of Tensor (Number of elements)

libs/vm-modules/src/math/tensor/tensor.cpp
MissingNO57: Yes, they are from running a benchmark on my machine. I will do that.

libs/vm-modules/src/math/tensor/tensor.cpp
```diff
@@ -83,9 +95,19 @@ void VMTensor::Bind(Module &module, bool const enable_experimental)
...
+
+    return static_cast<ChargeAmount>(CONSTRUCTION_STRING_SIZE_COEF * size +
+                                     CONSTRUCTION_STRING_CONST_COEF) *
+           COMPUTE_CHARGE_COST;
+  };
+
   auto interface =
       module.CreateClassType<VMTensor>("Tensor")
           .CreateConstructor(&VMTensor::Constructor, tensor_constructor_charge_estimate)
+          .CreateConstructor(&VMTensor::StringConstructor,
```
iryna-melnyk: AFAIK it should be experimental

libs/vm-modules/src/math/tensor/tensor.cpp
```diff
@@ -57,6 +57,13 @@ VMTensor::VMTensor(VM *vm, TypeId type_id)
   , estimator_(*this)
 {}
 
+VMTensor::VMTensor(VM *vm, TypeId type_id, std::string const &str)
+  : Object(vm, type_id)
+  , estimator_(*this)
+{
+  tensor_ = tensor_.FromString(str);
```
vhavrylov-ncube: FromString is a static method, so I guess it could be moved to initializer list as

VMTensor::VMTensor(VM *vm, TypeId type_id, std::string const &str)
  : Object(vm, type_id)
  , estimator_(*this)
  , tensor_(Tensor<DataType>::FromString(str))
{
}

vhavrylov-ncube (commented):
Looks good to me, let others review too.

iryna-melnyk (dismissed):
Looks good to me

iryna-melnyk (dismissed):
looks good

libs/vm-modules/tests/unit/math/tensor_tests.cpp
```diff
@@ -116,6 +116,35 @@ TEST_F(MathTensorTests, tensor_4_dim_fixed64_fill)
...
+              assert(d.at(0u64,1u64) == 2.0fp64);
+              assert(d.at(1u64,0u64) == 3.0fp64);
+              assert(d.at(1u64,1u64) == 4.0fp64);
+
+            endfunction
+          )";
+  ASSERT_TRUE(toolkit.Compile(STR_CONSTRUCT_SRC));
+  ASSERT_TRUE(toolkit.Run());
+}
+
```
iryna-melnyk, line 180: Probably it would be worth to add a test for malformed string

iryna-melnyk (changes requested):

baykaner (approved):

vhavrylov-ncube (commented):
Looks good to me, let's wait for other approvals.

baykaner (approved):

iryna-melnyk (approved):
looks good now

PR #2295, opened 2020-01-13
Test annotations for tooling.

troels-im:


libs/vm/src/analyser.cpp
```diff
@@ -494,9 +499,12 @@ void Analyser::ValidateFunctionAnnotations(NodePtr const &function_node)
...
     auto const &text = annotation->text;
-    if (text != INIT_ANNOTATION && text != ACTION_ANNOTATION && text != QUERY_ANNOTATION &&
-        text != PROBLEM_ANNOTATION && text != OBJECTIVE_ANNOTATION && text != WORK_ANNOTATION &&
-        text != CLEAR_ANNOTATION)
+    bool        test_annotation =
+        (!allow_test_annotations_) || (text != TEST_ANNOTATION && text != TEST_INIT_ANNOTATION);
+
+    if (test_annotation && text != INIT_ANNOTATION && text != ACTION_ANNOTATION &&
+        text != QUERY_ANNOTATION && text != PROBLEM_ANNOTATION && text != OBJECTIVE_ANNOTATION &&
+        text != WORK_ANNOTATION && text != CLEAR_ANNOTATION)
```
wkaluza: One thing we could improve this year is to have more meaningful error messages - I think we can do better here than "Invalid annotation".

```
if (text == TEST_ANNOTATION || text == TEST_INIT_ANNOTATION)
{
  if (!allow_test_annotations_)
  {
    AddError(annotation->line, "Test annotations are disabled '" + text + "'");
  }

  return;
}

if (text != INIT_ANNOTATION && text != ACTION_ANNOTATION && ...
```

libs/vm/include/vm/module.hpp
```diff
@@ -475,6 +475,16 @@ class Module
     return deserialization_constructors_;
   }
 
+  void EnableTestAnnotations()
+  {
+    test_annotations_ = true;
+  }
```
wkaluza, line 481: Does it make sense for the Module to know about test annotations? All it does is pass this Boolean to the Analyser. A simpler way would be to just inject the Boolean as an argument to the Compiler's constructor, where you could pass it to `Analyser::Initialise` directly.

wkaluza (changes requested):

wkaluza (approved):

PR #2298, opened 2020-01-13
Fix synergetic end to end tests

qati:
This PR fixes the following:

1. Naive synergetic miner from time to time was executing the same epoch multiple times. To fix this I've added executed epoch check.
2.  Synergetic fees end to end test


wkaluza (commented):
Looks good to me. It would be best if we first released the changes to the Python SDK that this relies on so that we can update `install-test-dependencies.sh` with a version number. Happy to approve once that is done.

wkaluza:
Closing: work to continue in #2418

PR #2300, opened 2020-01-13
Changed DataLoaders to have only one template

MissingNO57:


evsmithx (dismissed):
Looks good - I think this is a lot clearer than before.

baykaner (dismissed):

MissingNO57:
Already merged through #2307

PR #2301, opened 2020-01-13
Refactor/ml 467 unify ml ops serialisation tests

baykaner:
- moved layers & ops serialisation tests into ml/test/unit/serializers
- helper functions implemented to reduce repeated code
- refactors ops tests to anonymous namespaces
- build times should be dramatically faster now there is only 1 test that includes graph_builder


libs/ml/tests/unit/serializers/serialisation.cpp
```diff
@@ -0,0 +1,4407 @@
...
+  EXPECT_TRUE(prediction.at(1).AllClose(
+      new_prediction.at(1), fetch::math::function_tolerance<typename TypeParam::Type>(),
+      fetch::math::function_tolerance<typename TypeParam::Type>()));
+}
+
+//////////////////
+/// EMBEDDINGS ///
+//////////////////
+
+TYPED_TEST(SaveParamsTest, embeddings_saveparams_test)
```
vhavrylov-ncube: Do these two tests (for Embeddings I mean) use graph_builder?

libs/ml/include/ml/utilities/graph_builder.hpp
```diff
@@ -261,14 +273,14 @@ void BuildNodeAndInsertTrainables(NodeSaveableParams<T> const &nsp, std::string
     g->AddTrainable(node, name);
     break;
```
vhavrylov-ncube: I wonder if we could split this header file to both cpp and hpp file, and move all Ops includes to cpp file (and implementation of all functions that are declared here). This way we can reduce include amount in this particular header and reduce its impact on unit test .cpp files without merging many UT .cpp files to a large one.

vhavrylov-ncube: this header is also included in vm-modules Model (VMModel) but it actually does not need the whole scope of includes too and could be slightly refactored; the same for vm-modules Graph.

libs/ml/tests/unit/serializers/serialisation.cpp
baykaner: doesn't look like it; why are you asking? There are some tests in here that use BuildGraph and some that do not...

iryna-melnyk (commented):
looks much better now, let's see how it builds

iryna-melnyk (approved):

vhavrylov-ncube (approved):
LGTM

PR #2302, opened 2020-01-13
Move dataloader implementation to source files

MissingNO57:


libs/ml/include/ml/dataloaders/word2vec_loaders/sgns_w2v_dataloader.hpp
```diff
@@ -17,8 +17,10 @@
 //
 //------------------------------------------------------------------------------
 
+#include "math/tensor/tensor.hpp"
```
baykaner, line 20: you can just forward declare tensor instead of including it in all of these header files. Then include in the source file.

libs/ml/include/ml/dataloaders/word2vec_loaders/w2v_dataloader.hpp
```diff
@@ -17,17 +17,15 @@
 //
 //------------------------------------------------------------------------------
 
+#include "ml/dataloaders/word2vec_loaders/w2v_dataloader.hpp"
```
baykaner: this seems to be recursively including itself?

libs/ml/include/ml/dataloaders/word2vec_loaders/w2v_dataloader.hpp
```diff
@@ -17,17 +17,15 @@
...
+
 #include "core/random/lcg.hpp"
 #include "math/tensor/tensor.hpp"
 #include "meta/type_traits.hpp"
 #include "ml/dataloaders/dataloader.hpp"
```
baykaner, line 22: most of these includes can be moved into source and just forward declare the types you need for the header i think

libs/ml/tests/unit/dataloader/code2vec_context_loader.cpp
```diff
@@ -29,10 +29,8 @@ namespace test {
 
 TEST(C2vLoaderTest, loader_test)
 {
-  using TensorType  = fetch::math::Tensor<fetch::math::SizeType>;
-  using SizeType    = fetch::math::SizeType;
-  using LabelType   = TensorType;
-  using ContextType = TensorType;
+  using TensorType = fetch::math::Tensor<int64_t>;
```
baykaner: why hardcode int64?

libs/ml/src/dataloaders/word2vec_loaders/w2v_dataloader.cpp
```diff
@@ -0,0 +1,398 @@
...
+}
+
+template <typename T>
+T const W2VLoader<T>::WindowContextUnused = static_cast<T>(-1);
+
+///////////////////////////////
+/// EXPLICIT INSTANTIATIONS ///
+///////////////////////////////
+
+template class W2VLoader<float>;
```
baykaner, line 393: no integral types?

baykaner (changes requested):

libs/ml/src/dataloaders/word2vec_loaders/w2v_dataloader.cpp
MissingNO57: Yes, because there is static assert in class declaration
public:
  static_assert(fetch::meta::IsFloat<T> || math::meta::IsFixedPoint<T>,
                "The intended T is the typename for the data input to the neural network, which "
                "should be a float or double or fixed-point type.");

baykaner: i don't think that's true anymore though, right?
I mean, there may be many pieces that won't work well or need adjustment for full integral compatibility, but we can at least get everything integral compilable right?

libs/ml/src/dataloaders/tensor_dataloader.cpp
```diff
@@ -0,0 +1,256 @@
...
+    throw exceptions::InvalidMode("Unsupported dataloader mode.");
+  }
+  }
+}
+
+///////////////////////////////
+/// EXPLICIT INSTANTIATIONS ///
+///////////////////////////////
+
+// template class TensorDataLoader<math::Tensor<std::int8_t>>;
```
baykaner, line 244: please add a TODO(ML-464)

libs/ml/src/dataloaders/word2vec_loaders/sgns_w2v_dataloader.cpp
```diff
@@ -0,0 +1,692 @@
...
+byte_array::ConstByteArray GraphW2VLoader<TensorType>::GetVocabHash()
+{
+  return vocab_->GetVocabHash();
+}
+
+///////////////////////////////
+/// EXPLICIT INSTANTIATIONS ///
+///////////////////////////////
+
+//          template class GraphW2VLoader<math::Tensor<std::int8_t>>;
```
baykaner: missing TODO ML-464

baykaner (changes requested):

baykaner (dismissed):

baykaner (approved):

PR #2303, opened 2020-01-13
Refactor/ml 430 partial move graph builder implementations

baykaner:
- depends on #2291


libs/math/src/linalg/blas/gemm_nt_vector.cpp
```diff
@@ -117,6 +117,38 @@ void Blas<S, Signature(_C <= _alpha, _A, _B, _beta, _C),
...
+                    platform::Parallelisation::VECTORISE>;
+
+template class Blas<int32_t, Signature(_C <= _alpha, _A, _B, _beta, _C),
+                    Computes(_C <= _alpha * _A * T(_B) + _beta * _C),
+                    platform::Parallelisation::VECTORISE>;
+
+template class Blas<int64_t, Signature(_C <= _alpha, _A, _B, _beta, _C),
+                    Computes(_C <= _alpha * _A * T(_B) + _beta * _C),
+                    platform::Parallelisation::VECTORISE>;
+
```
iryna-melnyk: I don't see Blas<fetch::fixed_point::FixedPoint<64,64> here. is it intentional?

libs/ml/src/utilities/graph_builder.cpp
```diff
@@ -0,0 +1,563 @@
...
+    node->SetNodeSaveableParams(nsp, op_ptr);
+    g->AddTrainable(node, name);
+    break;
+  }
+  case ops::Convolution2D<T>::OpCode():
+  {
+    op_ptr = GetOp<ops::Convolution2D<T>>(nsp.op_save_params);
+    node->SetNodeSaveableParams(nsp, op_ptr);
+    g->AddTrainable(node, name);
+    break;
```
vhavrylov-ncube, line 125: pls pay attention, these

    node->SetNodeSaveableParams(nsp, op_ptr);
    g->AddTrainable(node, name);

are just copied to each case, so I guess we can move them outside of the switch at all. The only cases that do not contatin these two lines are cases that throw exceptions.

libs/ml/include/ml/model/model.hpp
```diff
@@ -21,17 +21,39 @@
...
+class CrossEntropyLoss;
+
+template <typename TensorType>
+class MeanSquareErrorLoss;
+
+template <typename TensorType>
+class SoftmaxCrossEntropyLoss;
+}  // namespace ops
+
+}  // namespace ml
```
AlexTeslenko: you included loss functions headers and at the same time provided forward declarations for them. I think one of the options is enough?

libs/ml/include/ml/utilities/graph_builder.hpp
```diff
@@ -18,65 +18,9 @@
...
-#include "ml/layers/convolution_1d.hpp"
-#include "ml/layers/convolution_2d.hpp"
-#include "ml/layers/fully_connected.hpp"
-#include "ml/layers/multihead_attention.hpp"
-#include "ml/layers/scaled_dot_product_attention.hpp"
-#include "ml/layers/self_attention_encoder.hpp"
-#include "ml/layers/skip_gram.hpp"
-
+#include "ml/core/subgraph.hpp"
 #include "ml/meta/ml_type_traits.hpp"
```
AlexTeslenko, line 21: I think if you have #include "ml/core/subgraph.hpp" you do not need to include "ml/core/graph.hpp" also

libs/ml/src/utilities/graph_builder.cpp
baykaner: done

libs/ml/src/utilities/graph_builder.cpp
```diff
@@ -0,0 +1,563 @@
...
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "ml/utilities/graph_builder.hpp"
+
+#include "ml/core/graph.hpp"
```
AlexTeslenko: I think you do not need this include here(you already included subgraph in "graph_builder.hpp").

libs/ml/include/ml/model/model.hpp
baykaner: good point. done

vhavrylov-ncube (dismissed):
LGTM

vhavrylov-ncube (approved):
LGTM

PR #2304, opened 2020-01-14
ML-427: moved model implementations to source files

iryna-melnyk:
Almost all methods (except variadic templates) were moved 
serializers were also moved


libs/ml/include/ml/model/model.hpp
```diff
@@ -70,6 +65,7 @@ class Model
   explicit Model(ModelConfig<DataType> model_config = ModelConfig<DataType>())
     : model_config_(std::move(model_config))
   {}
+
   Model(Model const &other)
   {
```
vhavrylov-ncube: I guess the constructor could be moved to cpp too.

libs/ml/src/model/model.cpp
```diff
@@ -0,0 +1,414 @@
...
+  for (auto const &met : metrics_)
+  {
+    ret.emplace_back(*(this->graph_ptr_->Evaluate(met).cbegin()));
+  }
+  return ret;
+}
+
+template <typename TensorType>
+template <typename... Params>
+void Model<TensorType>::SetData(Params... params)
```
baykaner: as before - you shouldn't move this type of implementation into source because you can't explicitly instantiate every possible set of types that Params could mean. move it back into header.

libs/ml/include/ml/model/model.hpp
```diff
@@ -17,19 +17,16 @@
 //
 //------------------------------------------------------------------------------
 
-#include "ml/core/graph.hpp"
 #include "ml/dataloaders/tensor_dataloader.hpp"
```
baykaner: you need to move most of these remaining includes into the source file

baykaner (changes requested):

libs/ml/include/ml/model/model.hpp
```diff
@@ -743,4 +366,5 @@ struct MapSerializer<ml::model::Model<TensorType>, D>
 };
```
vhavrylov-ncube: (optional) could static member methods of MapSerialiser be transferred to .cpp file too?

libs/ml/include/ml/model/model.hpp
iryna-melnyk: I removed as much as possible. types are used in function signatures, graph_saver is used in serialization

libs/ml/include/ml/model/model.hpp
iryna-melnyk: I'd better leave it as is

libs/ml/include/ml/model/model.hpp
iryna-melnyk: Nope, it seems they won't be accessible from the outer scope

libs/ml/src/model/model.cpp
iryna-melnyk: Done

libs/ml/include/ml/model/model.hpp
baykaner: what about forward declaration for types used in function signatures, and moving serialisation code to serialisers/ml_types to avoid including graph_saver?

libs/ml/include/ml/model/model.hpp
baykaner: why not move it?

libs/ml/src/model/model.cpp
```diff
@@ -0,0 +1,407 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "ml/model/model.hpp"
```
baykaner, line 19: you should definitely move most of the header includes here

baykaner (changes requested):

libs/ml/include/ml/model/model.hpp
iryna-melnyk: Done

libs/ml/include/ml/model/model.hpp
iryna-melnyk: I moved them to serialisers/ml_types.h instead

libs/ml/src/model/model.cpp
iryna-melnyk: Done

libs/ml/include/ml/model/model.hpp
iryna-melnyk: Done

libs/ml/include/ml/model/model.hpp
```diff
@@ -17,19 +17,12 @@
 //
 //------------------------------------------------------------------------------
 
-#include "ml/core/graph.hpp"
-#include "ml/dataloaders/tensor_dataloader.hpp"
-#include "ml/meta/ml_type_traits.hpp"
 #include "ml/model/model_config.hpp"
 #include "ml/ops/loss_functions/types.hpp"
```
iryna-melnyk: anticipating questions, 
1) model config is used during model construction:
 explicit Model(ModelConfig<DataType> model_config = ModelConfig<DataType>())
2) a forward declaration isn't applicable here because of such function declaration:
void Compile(OptimiserType optimiser_type, ops::LossType loss_type = ops::LossType::NONE...

baykaner: 👍

baykaner (dismissed):
nice work, looks good

libs/ml/include/ml/optimisation/learning_rate_params.hpp
```diff
@@ -36,8 +36,8 @@ struct LearningRateParam
   };
   LearningRateDecay mode                   = LearningRateDecay::NONE;
   DataType          starting_learning_rate = fetch::math::Type<DataType>("0.001");
-  DataType          ending_learning_rate   = starting_learning_rate / DataType{10000};
-  DataType          linear_decay_rate      = starting_learning_rate / DataType{10000};
+  DataType          ending_learning_rate   = starting_learning_rate / static_cast<DataType>(10000);
+  DataType          linear_decay_rate      = starting_learning_rate / static_cast<DataType>(10000);
```
iryna-melnyk, line 40: I would say these lines should be refactored in context of dedicated task

baykaner (approved):
LGTM

PR #2305, opened 2020-01-14
Feature/ml 428 optimisation to src

AlexTeslenko:
Feature/ml 428 optimisation to src


vhavrylov-ncube (commented):
Looks good, let's see if all test pass.

libs/ml/include/ml/optimisation/adagrad_optimiser.hpp
```diff
@@ -17,14 +17,9 @@
 //
 //------------------------------------------------------------------------------
 
-#include "math/standard_functions/sqrt.hpp"
 #include "ml/core/graph.hpp"
 #include "ml/optimisation/optimiser.hpp"
 
-#include <memory>
-#include <string>
-#include <vector>
```
iryna-melnyk, line 26: I would suggest to leave that. It seems to me more consistent along with usage of mentioned types in function declarations. But I'm not insist

libs/ml/include/ml/optimisation/adam_optimiser.hpp
```diff
@@ -17,16 +17,9 @@
 //
 //------------------------------------------------------------------------------
 
-#include "math/base_types.hpp"
-#include "math/standard_functions/pow.hpp"
-#include "math/standard_functions/sqrt.hpp"
 #include "ml/core/graph.hpp"
```
baykaner: I think you can move this include to the source code and just forward declare graph, right?

baykaner: i think this is true for all of these optimisers too right?

libs/ml/include/ml/optimisation/adagrad_optimiser.hpp
```diff
@@ -17,14 +17,9 @@
...
-
 namespace fetch {
 namespace ml {
 namespace optimisers {
```
baykaner, line 24: seems you didn't move over the OptimiserCode functions to source either?

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -26,9 +26,6 @@
 #include "ml/optimisation/learning_rate_params.hpp"
```
baykaner, line 22: you need to move as many of these includes to source as possible. For example, mean.hpp can definitely be moved.

libs/ml/include/ml/optimisation/optimiser.hpp
```diff
@@ -125,408 +122,6 @@ class Optimiser
...
-}
-
-template <typename T>
-std::shared_ptr<Graph<T>> Optimiser<T>::GetGraph()
-{
-  return graph_;
-}
-
 }  // namespace optimisers
 }  // namespace ml
```
baykaner: move the serialisation/deserialisation code out to serializers/ml_types.hpp
this will allow you to not have to include graph_builder.hpp in this file which is probably the most expensive include in the whole library

libs/ml/src/optimisation/rmsprop_ptimiser.cpp
```diff
@@ -0,0 +1,137 @@
+//------------------------------------------------------------------------------
```
baykaner: this filename seems to have a typo; should be rmsprop_optimiser

baykaner (changes requested):

libs/ml/include/ml/optimisation/adam_optimiser.hpp
AlexTeslenko: but I use Graph here. If I remove it from adam_optimiser.hpp  I will have:  error: use of undeclared identifier 'Graph'

baykaner (approved):

PR #2306, opened 2020-01-14
Make json parsing use fixed-point for floats

evsmithx:
JSON parsing uses float internally for numbers like "1.23". If you need to read a fixed-point from a file you then have to do a conversion from float to fixed-point which is unsafe. This PR modifies the JSON parsing to use fixed-points internally. A downside of this is that the fixed-point string parsing doesn't currently support scientific notation, so some test cases have had to be removed.


evsmithx:
Decided that actually this is overkill!

PR #2308, opened 2020-01-14
solution for storing entropy to file for quick crash recovery

n-hutton:
This PR adds the behaviour that after a crash, the beacon service should recover to an identical state (back slightly in time possibly). This includes jumping directly to the state it was in the state machine.

It will serialize itself to file every (in this PR) 100 entropy blocks generated. This limits the cost of this by keeping an object store for the signatures map, thus only updating new signatures to file (and cleaning old ones).

To maintain backwards compatability, it will first try to load the new file format (different file name). If this fails it will try to load the legacy file. Normal operation will write the loaded keys into the new file format (but leave the old file untouched).


libs/beacon/include/beacon/beacon_service.hpp
```diff
@@ -136,47 +166,52 @@ class BeaconService : public ledger::EntropyGeneratorInterface
   SignatureInformation GetSignatureShares(uint64_t round);
   /// @}
 
-  mutable Mutex                       mutex_;
-  CertificatePtr                      certificate_;
-  bool                                load_and_reload_on_crash_{false};
-  std::deque<SharedAeonExecutionUnit> aeon_exe_queue_;
+  // Related to recovering state after a crash
+  /// @{
+  uint16_t const save_periodicity_ = 100;
```
ejfitzgerald: Static and u

libs/muddle/internal/muddle_fake.hpp
```diff
@@ -20,6 +20,7 @@
 #include "fake_network.hpp"
 
 #include "muddle.hpp"
+#include "muddle/create_muddle_fake.hpp"
```
ejfitzgerald: ?

libs/constellation/src/constellation.cpp
```diff
@@ -384,6 +384,7 @@ Constellation::Constellation(CertificatePtr certificate, Config config)
   , http_network_manager_{"Http", HTTP_THREADS}
   , internal_identity_{std::make_shared<crypto::ECDSASigner>()}
   , external_identity_{std::move(certificate)}
+  , standalone_mode_{config.network_mode == NetworkMode::STANDALONE}
```
ejfitzgerald: !

ejfitzgerald (approved):

ejfitzgerald (changes requested):
.

ejfitzgerald (approved):

PR #2310, opened 2020-01-15
Fix vm-modules crypto example

evsmithx:
Fixed the crypto prediction example in vm-modules. However the data files have scientific notation in them, so running the example still requires switching on unsafe parsing mode in readCSV.


libs/vm-modules/examples/crypto_price_prediction/script/bitcoin.etch
```diff
@@ -44,6 +44,15 @@ function build_graph(graph : Graph)
 
 endfunction
 
+function add_leading_zero_dim(tensor : Tensor)
```
baykaner: maybe add a comment explaining why this is needed - it's not immediately obvious

baykaner (dismissed):

baykaner (approved):

PR #2312, opened 2020-01-15
Improvement/try-catch consensus

denis-trapeznikov:
One possible implementation of a nothrow consensus. Consensus interface is left intact (still working with blocks rather than pointers) but all the exception explicitly thrown are contained inside. In other words, program state monad is converted into that of zeroable value when control leaves the inwards of `Consensus` class.

Also discovered a very funny LGDR-699, it is also (probably) fixed here, since the whole fix took a replacement of one word in the code.


libs/ledger/include/ledger/consensus/consensus.hpp
```diff
@@ -82,6 +82,7 @@ class Consensus final : public ConsensusInterface
   Consensus &operator=(Consensus const &) = delete;
   Consensus &operator=(Consensus &&) = delete;
 
+  // TODO (nobody): the only reason this function is public is it's used in a single POS test.
```
ejfitzgerald: Can you create a JIRA issue to track this and update the code

libs/ledger/src/consensus/consensus.cpp
```diff
@@ -261,13 +261,20 @@ bool Consensus::VerifyNotarisation(Block const &block) const
...
+        auto ordered_notarisation_keys = aeon_block.block_entropy.aeon_notarisation_keys;
 
-      return notarisation_->Verify(previous_notarised_block->hash,
-                                   block.block_entropy.block_notarisation,
-                                   ordered_notarisation_keys, threshold);
+        return notarisation_->Verify(previous_notarised_block->hash,
+                                     block.block_entropy.block_notarisation,
+                                     ordered_notarisation_keys, threshold);
+      }
+      catch (...)
```
ejfitzgerald, line 288: Worth logging any error from here?

libs/ledger/src/consensus/consensus.cpp
```diff
@@ -280,24 +287,25 @@ bool Consensus::VerifyNotarisation(Block const &block) const
...
-                                                          current.block_entropy.EntropyAsU64());
-
-  if (std::find(qualified_cabinet_weighted.begin(), qualified_cabinet_weighted.end(), identity) ==
-      qualified_cabinet_weighted.end())
+  Block            beginning_of_aeon;
+  try
+  {
+    beginning_of_aeon = GetBeginningOfAeon(current, chain_);
+  }
+  catch (...)
```
ejfitzgerald: Again is it worth at least logging the error here?

libs/ledger/src/consensus/consensus.cpp
```diff
@@ -610,6 +627,17 @@ NextBlockPtr Consensus::GenerateNextBlock()
     notarisation_->NotariseBlock(*ret);
   }
 
+#ifndef NDEBUG
+  try
+  {
+    GetBeginningOfAeon(*ret, chain_);
```
ejfitzgerald, line 650: Is this supposed to be kept here?

ejfitzgerald (changes requested):
Looks like a nice change. Just a few comments. I think it is probably worth reviewing with @n-hutton which of the new catch blocks we want to log the exception message from

libs/ledger/src/consensus/consensus.cpp
denis-trapeznikov: See, earlier in the code, the weight is already calculated via `GetBlockGenerationWeight()`, so the beginning of aeon should not fail at this moment. Consequently, I've inserted this block as a debug-time assertion for that "should not".

libs/ledger/include/ledger/consensus/consensus.hpp
denis-trapeznikov: LGDR-698

libs/ledger/src/consensus/consensus.cpp
```diff
@@ -66,6 +66,7 @@ namespace {
 
 constexpr char const *LOGGING_NAME = "Consensus";
 const std::size_t     DIGEST_LENGTH_BYTES{32};
+constexpr char const *unknown_exception = "something wenth wrong";
```
ejfitzgerald:
```suggestion
constexpr char const *FATAL_ERROR = "Fatal error generated";
```

ejfitzgerald (changes requested):
Just a minor small suggestion looks okay. I can see that there is a little bit of feature creep on this PR

ejfitzgerald (approved):

PR #2315, opened 2020-01-16
Formatting

troels-im:


ejfitzgerald (changes requested):
Let’s chat about this, but this behaviour is useful to keep

PR #2318, opened 2020-01-17
Added functions for layer name finding and reporting

MissingNO57:
Prepared ground for deprecating StateDict


libs/ml/include/ml/core/graph.hpp
```diff
@@ -155,6 +155,10 @@ class Graph
 
   void ResetGradients();
 
+  std::vector<std::string> GetTrainableNames();
+  std::vector<std::string> GetNodeNames();
+  NodePtrType              GetNodeByName(std::string const &name) const;
```
baykaner: probably better just to call this GetNode

libs/ml/src/core/graph.cpp
```diff
@@ -1023,6 +1023,171 @@ void Graph<TensorType>::RecursiveApplyTwo(Val1Type &val_1, Val2Type &val_2,
...
+ * @param ret std::string name of node in format GRAPH1/...SUBGRAPHS../LEAF
+ * @param lookup_function
+ * @param level helper variable to remember on which level in graph function is
+ */
+template <typename TensorType>
+template <typename LookupFunction>
+void Graph<TensorType>::GetNamesRecursively(std::vector<std::string> &ret,
+                                            LookupFunction const lookup_function, std::string level)
+{
+  for (auto const &t : ((*this).*lookup_function)())
```
baykaner: probably cleaner to do:
this->(*lookup_function)

baykaner (changes requested):
very cool! a few minor changes, but otherwise seems great

baykaner (approved):

PR #2321, opened 2020-01-20
Added tests that ensures GetWeightReferences has always same order

MissingNO57:
Should be merged after #2319


baykaner (commented):
the tests you've added look good. let's wait until the other PR is merged in to make it easier to review properly

baykaner (approved):

PR #2322, opened 2020-01-21
Refactor/ml 430 utilities to src

AlexTeslenko:
Refactor/ml 430 utilities to src


libs/ml/include/ml/serializers/ml_types.hpp
```diff
@@ -26,6 +26,7 @@
 #include "ml/regularisers/reg_types.hpp"
 #include "ml/saveparams/saveable_params.hpp"
 #include "ml/utilities/graph_builder.hpp"
+//#include "ml/utilities/min_max_scaler.hpp"
```
baykaner: i guess you meant to remove this?

libs/ml/include/ml/utilities/min_max_scaler.hpp
```diff
@@ -35,147 +35,20 @@ class MinMaxScaler : public Scaler<TensorType>
 
   MinMaxScaler() = default;
 
-  void SetScale(TensorType const &reference_tensor) override;
-  void SetScale(DataType const &min_val, DataType const &max_val) override;
-  void Normalise(TensorType const &input_tensor, TensorType &output_tensor) override;
-  void DeNormalise(TensorType const &input_tensor, TensorType &output_tensor) override;
+  virtual void SetScale(TensorType const &reference_tensor) override;
```
baykaner: which object is inheriting from min_max_scaler and overriding SetScale?

libs/ml/include/ml/utilities/scaler.hpp
```diff
@@ -17,6 +17,8 @@
 //
 //------------------------------------------------------------------------------
 
+#include "math/tensor/tensor.hpp"
```
baykaner, line 20: can you explain why this is needed now but it wasn't before?

libs/ml/src/utilities/sparse_tensor_utilities.cpp
```diff
@@ -0,0 +1,289 @@
...
+    std::unordered_set<fetch::math::SizeType> const &update_rows,
+    fetch::math::SizeType                            sparsity_threshold);
+template void SparseAdd<math::Tensor<uint32_t>>(
+    math::Tensor<uint32_t> const &src, math::Tensor<uint32_t> &dst,
+    std::unordered_set<fetch::math::SizeType> const &update_rows,
+    fetch::math::SizeType                            sparsity_threshold);
+template void SparseAdd<math::Tensor<uint64_t>>(
+    math::Tensor<uint64_t> const &src, math::Tensor<uint64_t> &dst,
+    std::unordered_set<fetch::math::SizeType> const &update_rows,
+    fetch::math::SizeType                            sparsity_threshold);
```
baykaner: based on @markos 's comments i think maybe we shouldn't add the uint cases; please comment these out for now

baykaner (changes requested):

libs/ml/include/ml/utilities/scaler.hpp
AlexTeslenko: without this header I get error: scaler.hpp:29:29: error: implicit instantiation of undefined template 'fetch::math::Tensor<signed char, fetch::memory::SharedArray<signed char, 1> >'

libs/ml/src/utilities/sparse_tensor_utilities.cpp
AlexTeslenko: but uint instantiations needed for ml/ops/variable.cpp, because it also has uint instantiations. So if I comment them in utils I have to comment them in variable.cpp.

AlexTeslenko: even everywhere in ops where uint instantiated

libs/ml/include/ml/serializers/ml_types.hpp
```diff
@@ -3731,5 +3737,38 @@ struct MapSerializer<ml::optimisers::LearningRateParam<T>, D>
   }
 };
 
+/**
```
iryna-melnyk, line 3740: it might be worth to create a separate file for this routine (because ml_types is already 3740 lines)

libs/ml/src/utilities/mnist_utilities.cpp
```diff
@@ -0,0 +1,188 @@
...
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "math/tensor/tensor.hpp"
+#include <ml/utilities/mnist_utilities.hpp>
```
baykaner: use "" not <>

baykaner (changes requested):
trivial changes only

baykaner (approved):

PR #2323, opened 2020-01-21
Ops-based charge estimation framework skeleton

vhavrylov-ncube:
* extracted ChargeAmount and some constants to separate hedaers to improve include dependencies
* added dummy methods for Ops charge estimation
* added ChargeForward(...) cost getter to Graph and Model
* Implemented rough cost calculation for Dense layer (and its sub-ops)


libs/ml/src/layers/convolution_2d.cpp
```diff
@@ -54,16 +55,23 @@ Convolution2D<TensorType>::Convolution2D(SizeType const output_channels,
...
   this->GetNode(weights)->SetBatchOutputShape(
       {output_channels_, input_channels_, kernel_size_, kernel_size_, 1});
-  this->GetNode(input)->SetBatchOutputShape({output_channels_, 32, 32, 1});
+
+  // TODO(ML-470): Preliminary batch shape of a Conv2d layer (channels x 32(h) x 32(w) x
+  // 1(batch) ), is used here now, however, real width and height are to be set later on graph
+  // compilation (when expected Input shape of the Model/Graph is already known). Thus the
+  // convolutional weight init can be done on constructions, but this->input shape has to be
+  // inited only in this->CompleteInitialisation() override.
+  static constexpr SizeType DEFAULT_HEIGHT = 32;
```
baykaner, line 66: if these are just placeholders until the real values are set later, why not just set them both to 1?

baykaner: or better yet 0? that way they can't possibly be correct, and if, when debugging, we find that height and width are 0 we will know that they were never overwritten correctly.

libs/ml/include/ml/core/node.hpp
```diff
@@ -148,6 +149,9 @@ class Node
     return static_cast<bool>(cached_output_status_ == CachedOutputState::VALID_CACHE);
   }
 
+  fetch::vm::ChargeAmount ForwardCost(std::vector<fetch::math::SizeVector> const &input_shapes);
```
baykaner: naming convention might be better to give function names as ChargeX where X is the exact name of the function being charged for (i.e. ChargeEvaluate) etc.
I propose this because I expect that ultimately the right way to do this work is to figure out the charge of every function (even those like ComputeOutputShape).

libs/ml/include/ml/ops/estimation/charge_constants.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include "meta/vm_types.hpp"
+
+namespace fetch {
+namespace ml {
+namespace ops {
+namespace charge_cost {
+
+using fetch::vm::ChargeAmount;
+
+static constexpr ChargeAmount DROPOUT_PER_ELEMENT             = 3;
```
baykaner: where do these numbers come from by the way?

libs/ml/include/ml/core/node.hpp
vhavrylov-ncube: so for the forward pass it would be ChargeForward right?

libs/ml/include/ml/ops/estimation/charge_constants.hpp
vhavrylov-ncube: for now these constants are just dummies to be used in each Ops to return their charge, I will update the OR to show how are they used

libs/meta/include/meta/vm_types.hpp
```diff
@@ -0,0 +1,31 @@
+#pragma once
```
wkaluza: !

This file does not belong in `libs/meta` - that is a library meant for generic metaprogramming constructs. It knows nothing about the VM, charge estimation, or even the `vm` namespace and we should keep it this way.

wkaluza (changes requested):

libs/meta/include/meta/vm_types.hpp
vhavrylov-ncube: oh, sure, I have mistaken - it should be in ml/meta :)

wkaluza: Also no. I discussed this with Khan - architecturally it is dangerous to couple high-level business logic like VM charge estimation to generic ML code, even if it's just with a few types. My recommendation is that ML should have an internal representation (even if it's just `using MLChargeAmount = uint64_t;`) for how much effort an operation took, and expose a way to query that. Then estimators in the `vm_modules` lib which know about `MLChargeAmount` and `vm::ChargeAmount` can sum up the former and convert it into the latter for the virtual machine to use. This extra indirection avoids unnecessary coupling of ML code to VM charge estimation logic.

libs/ml/src/layers/convolution_2d.cpp
vhavrylov-ncube: it could be neither 0 nor 1 otherwise further calculations of output shape are being affected - a convolution of an image with kernel 3*3 and "valid" padding would produce a smaller image, and also it (input image) could not be less then a kernel size. So I have chosen some meaningful image size for this dummy. Anyway it would be corrected during implementation of a charge estimation for convolution (because real input inmage shape affects Conv performance greatly)

libs/meta/include/meta/vm_types.hpp
vhavrylov-ncube: agreed

libs/ml/include/ml/core/graph.hpp
```diff
@@ -151,6 +152,8 @@ class Graph
 
   std::vector<std::pair<std::string, std::vector<std::string>>> Connections();
 
+  fetch::ml::ops::MLChargeAmount ChargeForwardTo(std::string const &node_name);
```
baykaner: There is no "ForwardTo" method in Graph, I was imagining a naming convention with a ChargeX where X has a one to one correspondence with specific function calls. Is that not possible for some reason?

libs/ml/include/ml/ops/estimation/charge_constants.hpp
```diff
@@ -0,0 +1,44 @@
+#pragma once
```
baykaner: maybe this file should be moved to ml/charge_estimation/ops/constants.hpp?
Otherwise we're going to end up with many "estimation" subfolders all over the place.

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -149,6 +150,57 @@ class Ops
...
+      return 0;
+    }
+    fetch::math::SizeType total_elements = 1;
+    for (auto const &shape : shapes)
+    {
+      for (auto const &dimension : shape)
+      {
+        if (dimension == 0)
+        {
+          std::runtime_error("A dimension of size 0 found in tensor shape!");
```
baykaner: i guess this should be an ML exception of some kind

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -149,6 +150,57 @@ class Ops
...
+        }
+        total_elements *= dimension;
```
baykaner: is this correct? so if i have a custom op with three bias matrices that are just added to the incoming signal for example, then this will compute the total elements to be:
1 * n_elements in bias_1, and then that figure multiplied by dimension of bias_2, and then that figure multiplied by dimension of bias_3. But the number of elements is actually just dimension of bias_1 + d of bias_2 + d of bias_3 in that case, right?

I guess, i don't understand why ops.hpp's implementation isn't pure virtual anyway?

libs/meta/include/meta/vm_types.hpp
vhavrylov-ncube: I have separated MLChargeAmount in /ml/ scope and ChargeAmount in /vm/ scope, both are uint64 under the hood.

libs/ml/include/ml/core/graph.hpp
vhavrylov-ncube: the intention here was to achieve more smooth naming, because this method takes a node name as argument, so to emphasize the fact it is not an estimation for the _whole_ forward pass but only a forward pass "to" this particular node, I named it
... ForwardPassTo(first_convolutional_node)  <-- smth like that.

libs/ml/include/ml/ops/ops.hpp
vhavrylov-ncube: yes, that is a mistake! sure it should multiply dimensions only inside one of the shapes and then sum up across different shapes. The reason for a static method are: 1) it is used only in derivatives of Ops, 2) it is exactly the same always, because it just calculates how much element would be in a tensor of some given shape. 3) it does not use any internal state of a particular Op

libs/ml/include/ml/ops/ops.hpp
vhavrylov-ncube: I guess fetch::math::exceptions::WrongShape would fit here

libs/ml/include/ml/core/graph.hpp
vhavrylov-ncube: But you're right it is better to keep them consistent

libs/ml/include/ml/ops/estimation/charge_constants.hpp
vhavrylov-ncube: well, if we have a plan to estimate not only Ops but other ML stuff, it is reasonable, I'll move it

libs/ml/include/ml/ops/ops.hpp
vhavrylov-ncube: (this exception is available because we include a math::Tensor anyway)

libs/ml/include/ml/ops/estimation/charge_constants.hpp
vhavrylov-ncube: done, I've split it for clearer dependency tree

libs/ml/include/ml/charge_estimation/ops/constants.hpp
```diff
@@ -0,0 +1,40 @@
...
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include "ml/charge_estimation/types.hpp"
+
+namespace fetch {
+namespace ml {
+namespace ops {
```
baykaner, line 25: the namespace and the file directory are contrary here: i.e. charge_estimation/ops/ vs ops{ charge_cost{
this is a little unintuitive. Perhaps the namespaces for charge_cost and ops should be the other way around?

libs/ml/include/ml/charge_estimation/types.hpp
```diff
@@ -0,0 +1,28 @@
...
+//   limitations under the License.
+//
+//------------------------------------------------------------------------------
+
+#include <cstdint>
+
+namespace fetch {
+namespace ml {
+
+using MLChargeAmount = std::uint64_t;
```
baykaner: ideally, the naming would indicate that this is in fact not a charge amount but an estimate of the number of operations.

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -149,6 +151,60 @@ class Ops
...
   }
 
+  /**
+   * @brief ChargeForward
+   * @return estimated charge amount, necessary for performing a forward pass on data of given
+   * shapes.
+   */
+  virtual MLChargeAmount ChargeForward()
+  {
+    // TODO(VH): make a pure virtual call;
```
baykaner: so, why don't you just make this pure virtual like your todo suggests?

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -149,6 +151,60 @@ class Ops
...
+  }
+
+  /**
+   * @brief ChargeBackward
+   * @return estimated charge amount, necessary for performing a backward pass on data of given
+   * shapes.
+   */
+  virtual MLChargeAmount ChargeBackward() const
+  {
+    // TODO(VH): make a pure virtual call;
```
baykaner: again... why not just do this?

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -149,6 +151,60 @@ class Ops
...
+    {
+      SizeType elements_in_shape = 1;
+      for (SizeType dimension : shape)
+      {
+        if (dimension == 0)
+        {
+          throw fetch::math::exceptions::WrongShape("A dimension of size 0 found in tensor shape!");
+        }
+        elements_in_shape *= dimension;
+      }
```
baykaner: i believe there is a function in math/matrix_operations that might do this for you called Product. Might as well use that since it's more succinct and could potentially be better optimised in the future

libs/ml/src/core/graph.cpp
```diff
@@ -1111,6 +1111,21 @@ std::vector<std::pair<std::string, std::vector<std::string>>> Graph<TensorType>:
...
 }
 
+template <typename TensorType>
+fetch::ml::MLChargeAmount Graph<TensorType>::ChargeForward(const std::string &node_name)
+{
+  if (nodes_.find(node_name) == nodes_.end())
+  {
+    FETCH_LOG_ERROR(DESCRIPTOR, "The graph does not contain a node with name " + node_name);
+    return 0;
+  }
```
baykaner, line 1121: this won't catch cases where I want to call forward from the node in a subgraph. the node name wont be present in graph.nodes_; I think you want to use @MissingNO57's new functionality for finding a node in a graph - this makes a recursive call checking all subgraphs.

libs/ml/src/ops/matrix_multiply.cpp
```diff
@@ -285,6 +285,25 @@ std::vector<typename fetch::math::SizeType> MatrixMultiply<T>::ComputeOutputShap
   return output_shape;
 }
 
+template <typename T>
+MLChargeAmount MatrixMultiply<T>::ChargeForward()
+{
+  assert(!this->batch_input_shapes_.empty());
+
+  // TODO(VH): impl. for n-dimensional case, not only for 2D.
```
baykaner: if you are not going to do this implementation in this PR you need to raise a task for it and put the task number in the TODO

libs/ml/src/ops/matrix_multiply.cpp
```diff
@@ -285,6 +285,25 @@ std::vector<typename fetch::math::SizeType> MatrixMultiply<T>::ComputeOutputShap
...
+  assert(this->batch_input_shapes_.size() == 2);
+
+  // Assuming this is a matrix multiplication of weights * input_vector
```
baykaner, line 296: if you take a look at math/matrix_operations.hpp you can see that the Dot() method (which is what matmul operation invokes) uses blas/gemm_nn; so if we look at gemm_nn_novector.cpp we can count the number of operations in the non vectorised implementation. It looks to me like this should be something like:

resut.size() + n.width() * (1 + (result.height() * 2))
I haven't factored in assignment operations, but the compiler would probably remove some of these, anyway it's probably slightly off.

baykaner (changes requested):

libs/ml/include/ml/charge_estimation/ops/constants.hpp
vhavrylov-ncube: Agreed, done

libs/ml/include/ml/charge_estimation/types.hpp
vhavrylov-ncube: renamed to OperationsCount

libs/ml/include/ml/ops/ops.hpp
vhavrylov-ncube: this require immedtiate implementation across all Ops, which is huge PR; so I'd do it when all Ops would have their overriden implementation

libs/ml/include/ml/ops/ops.hpp
vhavrylov-ncube: the same as above, creating dummy empty override in each Ops will only create a huge change without any profit. It is better to do it in scope of implementation of estimation for each particular op

libs/ml/include/ml/ops/ops.hpp
vhavrylov-ncube: nice suggestion! I'll use it then.

libs/ml/src/ops/matrix_multiply.cpp
vhavrylov-ncube: Done, ML-482

libs/ml/src/ops/matrix_multiply.cpp
vhavrylov-ncube: Here I use a worst-case estimation which is for A[n;m] dot B[m;p] == O(n * m * p), or O(n^3) if the matrices are square and of size n. I know there are more effective algorithms, but at this stage I wouldn't dive into particular implementation, we will need to benchmark/measure it anyway.

libs/ml/src/core/graph.cpp
vhavrylov-ncube: For now it is intended to work with most parent Graph only; assuming we have FC and Conv layers as subgraphs, their froward pass charge estimation is calculated separately inside corresponding Layer code; and while in parent (model) graph, any Node is accessible.

libs/ml/include/ml/ops/ops.hpp
vhavrylov-ncube: ML-483 created for this, in order to prevent forgetting :)

libs/ml/src/ops/matrix_multiply.cpp
baykaner: okay we can leave it for next iteration then

baykaner (approved):

wkaluza (approved):

PR #2324, opened 2020-01-21
Persistent Bloom Filter

ejfitzgerald:


cmake/BuildTools.cmake
```diff
@@ -246,10 +246,18 @@ function (add_fetch_gbench
 
     # detect if the "DISABLED" flag has been passed to this test
     set(is_disabled FALSE)
+    set(include_gtest FALSE)
```
n-hutton, line 249: is this related/relevant?

libs/vm-modules/benchmark/CMakeLists.txt
```diff
@@ -10,5 +10,11 @@ include(${FETCH_ROOT_CMAKE_DIR}/BuildTools.cmake)
 # Compiler Configuration
 setup_compiler()
 
-add_fetch_gbench(benchmark_vm_modules_model fetch-vm-modules ../../vm-modules/benchmark/model)
-add_fetch_gbench(benchmark_vm_modules_tensor fetch-vm-modules ../../vm-modules/benchmark/tensor)
+add_fetch_gbench(benchmark_vm_modules_model
+                 fetch-vm-modules
+                 ../../vm-modules/benchmark/model
+                 GTEST)
```
n-hutton, line 16: what is going on here?

n-hutton (dismissed):

cmake/BuildTools.cmake
ejfitzgerald: Needed in order to build the benchmarks annoyingly

libs/bloom-filter/include/bloom_filter/metadata_store.hpp
```diff
@@ -0,0 +1,135 @@
...
+{
+  uint64_t version;
+  uint64_t length;
+};
+
+template <typename T>
+meta::IfIsPod<T> WriteToStream(std::ostream &stream, T const &element)
+{
+  auto const *raw = reinterpret_cast<char const *>(&element);
+  stream.write(raw, sizeof(T));
```
wkaluza, line 39: Shouldn't this check the stream's `badbit`?

libs/bloom-filter/include/bloom_filter/metadata_store.hpp
```diff
@@ -0,0 +1,135 @@
...
+}
+
+template <typename T>
+meta::IfIsPod<T> ReadFromStream(std::istream &stream, T &element)
+{
+  auto *raw = reinterpret_cast<char *>(&element);
+  stream.read(raw, sizeof(T));
```
wkaluza, line 46: As above, this operation can fail.

libs/bloom-filter/include/bloom_filter/metadata_store.hpp
ejfitzgerald: Good spot - This whole of this code is temporary at the moment, actually. In case of failure it would be detected by other means. And since we rebuild the whole of the bloom filter at the moment I am inclined not to fix this here at this time.

The next stage here will be for @n-hutton to integrate his changes on and remove this metdata file.

libs/bloom-filter/include/bloom_filter/metadata_store.hpp
ejfitzgerald: See above 😁

PR #2330, opened 2020-01-21
Refactor/borrow block generation weight

denis-trapeznikov:
This commit partly addresses, in a sense, the problem outlined in LGDR-698; in particular, it is an attempt to fix `fetch::ledger::Consensus`'s interface without disrupting the PoS tests.


wkaluza (changes requested):
From a design standpoint, I'm wholeheartedly opposed to this change. While having test-only public functionality is definitely a smell, this change replaces one public method with:

1. A forward declaration of a test class in production code (this alone is disqualifying, IMO)
2. A new test-only type
3. Public inheritance from said type in production code
4. Making the test more complex

...and all this to pretend that we don't have test-only functionality, when we still do - it's just messier to use now.

If you really want to make an improvement here, the way to do so is to think about the design of Consensus, the test, or both, where it pertains to block generation weight. Can the test infer the block generation weight by itself? Would mocking out the chain more cleverly help here? Does that test really need the block generation weight? Do we need this test in the first place?

We shouldn't write code which screams "yes, our design is imperfect but look, we feel really bad about that". We should design better code.

denis-trapeznikov:
> We should design better code.

0_o.

PR #2331, opened 2020-01-22
prevent out of order entropy gen

n-hutton:
It is possible in some extreme edge cases for nodes to do DKG on an aeon that they have already done it on. This could be possible if the whole network were to reset, trigger DKG 0 and complete it before syncing the main chain (from a relay perhaps).

This guard drops aeons that would generate signatures/entropy below already known/generated sigs/entropy.


jinmannwong (approved):
This looks fine but is it not possible instead to prevent them from running the DKG for an old aeon?

PR #2336, opened 2020-01-22
Add Simple TX Generator

ejfitzgerald:


wkaluza (approved):
For file-private functions I'd normally use an unnamed namespace rather than specify static linkage, but I can live with it.

PR #2337, opened 2020-01-22
Deploy Python tooling with pipenv

wkaluza:


Jenkinsfile
```diff
@@ -51,6 +52,21 @@ def run_full_build()
...
 
+def set_up_env_for_linux_pipenv(steps)
+{
+  withEnv([
+    "LC_ALL=C.UTF-8",
+    "LANG=C.UTF-8"])
+  {
+    steps()
+  }
+}
```
wkaluza: I'm not sure if we should have this here or in the image. Here is less invasive, but then again not having a UTF-8 locale feels very last-century on the part of the image. Thoughts?

ejfitzgerald: I have a mild preference for putting it in the image to be honest

ejfitzgerald (commented):
Basically looks all good to me. As you suggested this will need to be circulated through the team but certainly sorts out the dependency issue quite nicely.

Jenkinsfile
wkaluza: I'll give it a go.

wkaluza: Done.

ejfitzgerald (approved):

PR #2346, opened 2020-01-27
Etch binding for model.predict() estimation

vhavrylov-ncube:
- VMModel estimator method for .predict() call
- Conv layers construction splitted as for Dense: shape deduction is being done after locating input nodes in a Graph.
- fixed UTs according to new "input" layer behavior
- fixed UTs according to new .predict() estimation behavior
- correction of estimation computing for some Ops and Layers


libs/ml/tests/unit/layers/convolution_1d.cpp
```diff
@@ -44,12 +44,12 @@ TYPED_TEST(Convolution1DTest, set_input_and_evaluate_test)  // Use the class as
 
   SizeType const input_channels  = 3;
   SizeType const output_channels = 5;
-  SizeType const input_height    = 3;
+  SizeType const input_length    = 3;
```
iryna-melnyk, line 47: that renaming seems reasonable to me but it should comply with overall logic so i would rather wait for others opinions

baykaner (commented):
LGTM

kublai-baykaner (approved):

iryna-melnyk (approved):
lgtm

PR #2348, opened 2020-01-27
Feature/reactor alerts

n-hutton:


libs/core/include/core/reactor.hpp
```diff
@@ -50,6 +50,13 @@ class Reactor
   Reactor &operator=(Reactor const &) = delete;
   Reactor &operator=(Reactor &&) = delete;
 
+protected:
+  uint64_t execution_too_long_ms_{200};
```
n-hutton: getters and setters

libs/core/src/reactor.cpp
```diff
@@ -44,6 +44,25 @@ using WorkQueue = std::deque<fetch::core::WeakRunnable>;
 namespace fetch {
 namespace core {
 
+class BasicTimer
```
n-hutton: make core depend on libs moment

libs/core/src/reactor.cpp
```diff
@@ -65,11 +84,19 @@ Reactor::Reactor(std::string name)
         "The total number of times the reactor has failed to execute a runnable")}
   , expired_total_{CreateCounter("ledger_reactor_expired_total",
                                  "The total number of expired runnables")}
+  , too_long_total_{CreateCounter("ledger_reactor_too_long_total",
```
n-hutton, line 69: histogram

libs/core/src/reactor.cpp
```diff
@@ -147,6 +174,11 @@ Reactor::~Reactor()
     worker_->ApplyVoid([](auto &worker) { worker.join(); });
     worker_.reset();
   }
+
+  not_destructing_ = false;
```
n-hutton: rename and look at this

libs/core/tests/CMakeLists.txt
```diff
@@ -11,4 +11,5 @@ include(${FETCH_ROOT_CMAKE_DIR}/BuildTools.cmake)
 setup_compiler()
 
 fetch_add_test(core-unit-tests fetch-core unit/)
+fetch_add_slow_test(core-slow-tests fetch-core slow/)
```
n-hutton: put in integration

libs/core/include/core/reactor.hpp
```diff
@@ -50,6 +50,11 @@ class Reactor
   Reactor &operator=(Reactor const &) = delete;
   Reactor &operator=(Reactor &&) = delete;
 
+  uint64_t &ExecutionTooLongMs();
+  uint64_t &ThreadWatcherCheckMs();
```
denis-trapeznikov, line 55: Simply declare those two variables `public` maybe?

libs/core/include/core/runnable.hpp
```diff
@@ -41,7 +41,8 @@ class Runnable
   }
   virtual void Execute() = 0;
 
-  virtual char const *GetId() const = 0;
+  virtual std::string GetId() const    = 0;
+  virtual std::string GetDebug() const = 0;
```
denis-trapeznikov: How about a default implementation, for runnables that don't need to report any debug?

libs/core/include/core/state_machine.hpp
```diff
@@ -293,7 +307,26 @@ char const *StateMachine<S>::GetStateName() const
...
+char const *StateMachine<S>::GetPrevStateName() const
+{
+  char const *text = "Unknown";
+
+  if (mapper_)
+  {
+    text = mapper_(previous_state());
+  }
+
+  return text;
```
denis-trapeznikov, line 325: NP: this can be a single line.

libs/core/src/reactor.cpp
```diff
@@ -132,39 +137,76 @@ void Reactor::Stop()
...
 {
-  if (worker_)
-  {
-    worker_->ApplyVoid([](auto &worker) { worker.join(); });
-    worker_.reset();
-  }
+  StopWorker();
 }
 
 void Reactor::StopWorker()
```
denis-trapeznikov: A better name perhaps, since it stops both?

libs/core/include/core/reactor.hpp
```diff
@@ -59,16 +64,34 @@ class Reactor
...
   std::string const name_;
   Flag              running_{false};
+  Flag              not_destructing_{true};
 
   RunnableMap work_map_{};
   ThreadPtr   worker_{};
+  ThreadPtr   watcher_{};
+
+  // Keeping track of the last item executed
+  uint32_t     execution_counter_{0};
```
denis-trapeznikov: Are these plain uints always accessed under a locked mutex?

libs/core/tests/integration/state_machine_tests.cpp
```diff
@@ -0,0 +1,177 @@
...
+using namespace testing;
+
+class ReactorTests : public Test
+{
+public:
+  enum class State : uint8_t
+  {
+    A,
+    B,
+    C,
```
denis-trapeznikov: NP: IIRC trailing comma is non-standard in C++.

libs/core/tests/integration/state_machine_tests.cpp
```diff
@@ -0,0 +1,177 @@
...
+  state_machine_->RegisterHandler(State::C, static_cast<ReactorTests *>(this), &ReactorTests::OnSlowC);
+  // clang-format on
+
+  reactor_.Attach(state_machine_);
+  reactor_.Start();
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(reactor_.ExecutionTooLongMs() * 10));
+
+  EXPECT_NE(state_seen_, std::numeric_limits<uint8_t>::max());
+  EXPECT_NE(reactor_.ExecutionsTooLongCounter(), 0);
```
denis-trapeznikov, line 154: Can it be expected to equal 10 without a risk of being flaky?

n-hutton: Yeah, this should be half a second which should basically guarantee the reactor executes. If it doesn't that would indicate something has gone badly wrong

libs/core/tests/integration/state_machine_tests.cpp
n-hutton: fixed

libs/core/include/core/reactor.hpp
n-hutton: execution_counter was unsafe - executions_*_too_long_ are only used for testing but I made them atomic to be thorough

libs/core/src/reactor.cpp
n-hutton: renamed them to StopWorkerAndWatcher etc.

libs/core/include/core/state_machine.hpp
n-hutton: do you mean the whole function could be a ternary operator?

libs/core/include/core/reactor.hpp
n-hutton: This would have dot access of class variables which isn't our typical style (though of course we do for structs)

libs/core/include/core/runnable.hpp
n-hutton: makse sense. I have done this now

libs/core/include/core/reactor.hpp
denis-trapeznikov: Yes, exactly. Dot access to variables is indistinguishable from public methods returning mutable references to those variables, save for less typing and a lighter class interface.

libs/core/include/core/state_machine.hpp
denis-trapeznikov: Yes.

libs/core/include/core/reactor.hpp
n-hutton: I don't really care too much either way, @ejfitzgerald indicated he wanted it as getters/setters so that is what I would go with for this PR

libs/core/include/core/state_machine.hpp
n-hutton: I will probably leave it as is since it mirrors the function above it `GetStateName`

ejfitzgerald (approved):

PR #2350, opened 2020-01-27
Add validation for mined tx's from HTTP interface

denis-trapeznikov:
`ledger::BasicMiner` now uses `ledger::TransactionValidator` to check for apparently invalid transactions when possible.


libs/chain/include/chain/transaction.hpp
```diff
@@ -174,5 +176,9 @@ class Transaction
   friend class TransactionSerializer;
 };
 
+using TransactionPtr = std::shared_ptr<Transaction>;
+
+using TransactionIndex = std::unordered_map<Digest, TransactionPtr, DigestHashAdapter>;
```
ejfitzgerald:
```suggestion
using TransactionIndex = DigestMap<TransactionPtr>;
```

I  think the above it correct, in any case I think that it makes sense to use the helper type aliases

libs/ledger/include/ledger/block_packer_interface.hpp
```diff
@@ -17,18 +17,14 @@
...
 
+#include "chain/transaction.hpp"
+#include "chain/transaction_layout.hpp"
+
 #include <cstddef>
 #include <cstdint>
 
 namespace fetch {
 
-namespace chain {
```
ejfitzgerald, line 25: This was purposely done this way to break coupling, I don't think this is a good idea

libs/ledger/include/ledger/storage_unit/transaction_sinks.hpp
```diff
@@ -17,21 +17,18 @@
...
 
+#include "chain/transaction.hpp"
+
 #include <memory>
 #include <vector>
 
 namespace fetch {
-namespace chain {
-
-class Transaction;
```
ejfitzgerald, line 26: Again here, I think that we want to keep this to break coupling

libs/ledger/include/ledger/transaction_verifier.hpp
```diff
@@ -25,19 +26,14 @@
 #include <thread>
 
 namespace fetch {
-namespace chain {
-
-class Transaction;
```
ejfitzgerald, line 30: Again here

libs/ledger/src/miner/transaction_layout_queue.cpp
```diff
@@ -83,30 +86,29 @@ bool TransactionLayoutQueue::Remove(Digest const &digest)
  * @param digests The set of the transaction digests to be removed
  * @return The number of transaction layouts removed from the queue
  */
-std::size_t TransactionLayoutQueue::Remove(DigestSet const &digests)
+std::size_t TransactionLayoutQueue::Remove(DigestSet const &digests_to_remove)
 {
-  std::size_t count{0};
+  std::size_t current_population = digests_.size();
```
ejfitzgerald:
```suggestion
  std::size_t const current_population = digests_.size();
```

libs/ledger/src/miner/transaction_layout_queue.cpp
```diff
@@ -120,6 +122,7 @@ void TransactionLayoutQueue::Splice(TransactionLayoutQueue &other)
 {
   // extract the queue from other queue
   UnderlyingList input = std::move(other.list_);
+  other.list_.clear();
```
ejfitzgerald, line 125: Is this necessary since it has been moved?

ejfitzgerald (changes requested):
I made these comments a while ago and it looks like I forgot to submit them - sorry @denis-trapeznikov

libs/ledger/src/miner/transaction_layout_queue.cpp
denis-trapeznikov: Is it stated explicitly anywhere that a moved-from list is necessarily empty?
The best I'm aware of is that it is in a safe-to-destruct state.

libs/chain/include/chain/transaction.hpp
denis-trapeznikov: Thanks, I've made an almost a full circle and forgot about it.

libs/ledger/include/ledger/block_packer_interface.hpp
denis-trapeznikov: How about this?

ejfitzgerald:
@denis-trapeznikov looks like there might be missing commits from this branch. Can you make sure it is passing all the status checks and then assign back to me.

I think agree with the forward declaration header approach that you have taken here is probably the best bet

libs/ledger/src/miner/basic_miner.cpp
```diff
@@ -142,7 +174,7 @@ void BasicMiner::GenerateBlock(Block &block, std::size_t num_lanes, std::size_t
   DigestSet invalid{};
   for (auto const &layout : mining_pool_)
   {
-    if (chain::Transaction::Validity::VALID != chain::GetValidity(layout, block.block_number))
+    if (IsInvalid(layout, block.block_number))
```
ejfitzgerald: Okay so I think we need to make a subtle change here. I think we need to have a mechanism where we only remove transactions from the mining pool when there valid to / valid from check expires. i.e. the old check.

```
chain::Transaction::Validity::VALID != chain::GetValidity(layout, block.block_number)
```

We do however, still need to run the validator checks, however, because some are things like does this account have enough balance to proceed, they might go from being invalid initially to becoming valid again duration the valid from / valid until window.

I think that we should support the use case above. It would also be good to add and end 2 end test for this case also.

libs/core/include/core/digest.hpp
```diff
@@ -30,15 +30,13 @@ struct DigestHashAdapter
 {
   std::size_t operator()(Digest const &hash) const noexcept
   {
-    std::size_t value{0};
-
-    if (!hash.empty())
+    if (hash.empty())
```
n-hutton: rather than checking for empty here shouldn't it be `if(hash.size() < sizeof(std::size_t)` ?

libs/ledger/include/ledger/block_packer_interface.hpp
```diff
@@ -52,7 +47,7 @@ class BlockPackerInterface
    *
    * @param tx The reference to the transaction
    */
-  virtual void EnqueueTransaction(chain::Transaction const &tx) = 0;
+  virtual bool EnqueueTransaction(chain::TransactionPtr tx) = 0;
```
n-hutton, line 50: the comment needs to be updated here to include the return value

libs/ledger/include/ledger/dag/dag_sync_service.hpp
```diff
@@ -59,7 +59,7 @@ class DAGSyncService
   static constexpr char const *LOGGING_NAME = "DAGSyncService";
 
   using MuddleEndpoint = muddle::MuddleEndpoint;
-  using TransactionPtr = std::shared_ptr<chain::Transaction>;
+  using TransactionPtr = chain::TransactionPtr;
```
n-hutton, line 63: This file and any others that refer to the global TransactionPtr alias should also include tx_declaration.hpp

libs/ledger/src/miner/basic_miner.cpp
```diff
@@ -97,26 +105,50 @@ void BasicMiner::EnqueueTransaction(chain::Transaction const &tx)
...
+    return true;
+  }
+
+  duplicate_count_->increment();
+  FETCH_LOG_DEBUG(LOGGING_NAME, "Enqueued Transaction (duplicate) 0x", layout.digest().ToHex());
+  return false;
+}
+
+/**
+ * Check if transaction is inbvalid and should not be mined.
```
n-hutton: Minor spelling: 'invalid'

libs/ledger/src/miner/basic_miner.cpp
```diff
@@ -84,9 +86,15 @@ BasicMiner::BasicMiner(uint32_t log2_num_lanes)
  *
  * @param tx The reference to the transaction
  */
-void BasicMiner::EnqueueTransaction(chain::Transaction const &tx)
+bool BasicMiner::EnqueueTransaction(chain::TransactionPtr tx)
 {
-  EnqueueTransaction(chain::TransactionLayout{tx, log2_num_lanes_});
+  if (EnqueueTransaction(chain::TransactionLayout{*tx, log2_num_lanes_}))
+  {
+    txs_to_mine_.emplace(tx->digest(), std::move(tx));
```
n-hutton, line 96: Looks like txs_to_mine_ isn't protected by mining_pool_lock_ here

libs/ledger/src/miner/transaction_layout_queue.cpp
```diff
@@ -33,18 +33,15 @@ namespace ledger {
...
 
-  // ensure that this isn't already a duplicate transaction layout
-  if (digests_.find(digest) == digests_.end())
+  // try to add this digest and check if it's new
+  bool success = digests_.insert(digest).second;
+
+  if (success)
   {
-    // update the digest set and list
+    // this is a non-duplicate transaction
```
n-hutton, line 43: can you summarise the changes in this file? I don't fully understand what is going on

n-hutton (changes requested):
some minor comments, but does look like there is unprotected data access.

@ejfitzgerald @denis-trapeznikov I wonder if this is going to be too slow when calling GenerateBlock - there is now a whole bunch of serial transaction executions, possibly. Presumably these will involve RPC calls? I would expect this to bump the block packing time to a very long time before any execution actually happened.

It is also odd that this will only be a benefit to TXs submitted to miners through their HTTP interface? otherwise you can only do TX layout checks on TXs pulled from shards.

There is a minor race/edge case here also, where you populate mining_pool_ with valid TXs, then pack a small block, leaving some unmined TXs in mining_pool_ for the next block. It is assumed these are still valid, but they might not be. Probably this is not very critical.

n-hutton (approved):

ejfitzgerald (changes requested):
This current implementation looks like it will work only in the case where the transactions are submitted via the HTTP interface and not when they are synced from the poll to the lanes directly:

https://github.com/fetchai/ledger/blob/e655354c29460f31aa6019407991fac01562256f/libs/ledger/src/transaction_processor.cpp#L134

Supporting this case will require that the the transaction is pulled from the storage engine by the block packing call. Ideally this would be an async process. For the time being, I would request each of the transactions serially, but would only do so for the top of the mining queue.

Otherwise in the case where the mining pool is much larger than the block capacity, we waste time pulling all the transaction contents through the system.

Another approach might to be add the `from` field from the originating transaction to the transaction layout and then update the transaction validator to only need this information (so that all the checks can still be done, without the requirement to pull the entire transaction contents).

In a way, this work is related to @n-hutton 's work with pre-caching the transactions at the storage unit client level (this aids transaction execution performance). It would be good that any solution that we come up with for this plays nicely with this feature he is developing

ejfitzgerald (approved):
As discussed, we will leave the current implementation as is, and then work of implementing the same feature for transactions that are synced via the P2P interface between the lanes

PR #2351, opened 2020-01-28
Feature/ml 480 conv layer charge estimation

AlexTeslenko:
Feature/ml 480 conv layer charge estimation

### DIFF
--- a/libs/ml/src/ops/convolution_1d.cpp
+++ b/libs/ml/src/ops/convolution_1d.cpp
@@ -390,6 +390,29 @@ void Convolution1D<TensorType>::ReverseFillOutput(TensorType &gemm_output, Tenso
   }
 }
 
+template <typename TensorType>
+OperationsCount Convolution1D<TensorType>::ChargeForward()
+ {
+   assert(!this->batch_output_shape_.empty());
+   assert(this->batch_input_shapes_.size() == 2);
+
+   SizeType input_channels  = this->batch_input_shapes_.front().at(0);
+   SizeType batch_size      = this->batch_input_shapes_.front().at(2);
+   SizeType output_channels = this->batch_input_shapes_.back().at(0);
+   SizeType kernel_height   = this->batch_input_shapes_.back().at(2);
+
+   // output_height=number of stride_size steps over input size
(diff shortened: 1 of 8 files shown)


vhavrylov-ncube (commented):
LGTM but let other teammates approve too.

libs/ml/src/ops/convolution_1d.cpp
```diff
@@ -390,6 +390,33 @@ void Convolution1D<TensorType>::ReverseFillOutput(TensorType &gemm_output, Tenso
...
+
+  SizeType input_channels  = this->batch_input_shapes_.front().at(0);
+  SizeType batch_size      = this->batch_input_shapes_.front().at(2);
+  SizeType output_channels = this->batch_input_shapes_.back().at(0);
+  SizeType kernel_height   = this->batch_input_shapes_.back().at(2);
+
+  // output_height=number of stride_size steps over input size
+  SizeType output_height = (this->batch_input_shapes_.front().at(1) -
+                            this->batch_input_shapes_.back().at(2) + this->stride_size_) /
+                           this->stride_size_;
```
baykaner: These lines effectively duplicate a part of the ComputeOutputShape function. You should either abstract these two cases away into a separate private function (e.g. ComputeOutputHeight()), or alternatively you can invoke the ComputeOutputShape function here and use the result.

libs/ml/src/ops/convolution_2d.cpp
```diff
@@ -434,6 +434,39 @@ void Convolution2D<TensorType>::ReverseFillOutput(TensorType &gemm_output, Tenso
...
+
+  // output_height=number of stride_size steps over input height
+  SizeType output_height = (this->batch_input_shapes_.front().at(1) -
+                            this->batch_input_shapes_.back().at(2) + this->stride_size_) /
+                           this->stride_size_;
+
+  // output_width=number of stride_size steps over input width
+  SizeType output_width = (this->batch_input_shapes_.front().at(2) -
+                           this->batch_input_shapes_.back().at(3) + this->stride_size_) /
+                          this->stride_size_;
```
baykaner: same as for conv1d; this code duplication should be fixed

baykaner (changes requested):

baykaner (dismissed):

vhavrylov-ncube (approved):
LGTM

baykaner (approved):

PR #2352, opened 2020-01-28
added charge estimation for activations

AlexTeslenko:
added charge estimation for activations


baykaner (approved):
looks good for now! Naturally we'll need to do some benchmarking pretty soon to see what compute time actually looks like compared with estimates based on these constants

PR #2354, opened 2020-01-28
ML 432 bugfix to vmscaler

baykaner:
- updated normalise function to handle edge case with no variation in array


libs/ml/tests/unit/serializers_test/ops.cpp
```diff
@@ -0,0 +1,107 @@
...
+//      "1, -2, 3,-4, 5,-6, 7,-8;"
+//      "1,  2, 3, 4, 5, 6, 7, 8");
+//
+//  TensorType error = TensorType::FromString(
+//      "1, -1, 2, -2, 3, -3, 4, -4;"
+//      "5, -5, 6, -6, 7, -7, 8, -8");
+//
+//  fetch::ml::ops::Abs<TensorType> op;
+//
+//  // run op once to make sure caches etc. have been filled. Otherwise the test might be trivial!
```
iryna-melnyk: is it needed or forgotten?

iryna-melnyk (changes requested):

libs/ml/tests/unit/serializers_test/ops.cpp
baykaner: accidentally committed some stuff

iryna-melnyk (dismissed):
LGTM

vhavrylov-ncube (approved):
LGTM

PR #2356, opened 2020-01-28
Feature/ml 230 fixedpoint optimisations

markos:
FixedPoint class optimisations:
optimised CheckOverflow/CheckUnderflow() functions
optimised operators: +, -, *, /
average performance gain ~10%
optimised functions:
Exp(x): 9%, Pow(x, y): 15%.

### DIFF
--- a/libs/math/tests/unit/vectors/vector_tests.cpp
+++ b/libs/math/tests/unit/vectors/vector_tests.cpp
@@ -208,6 +208,7 @@ TYPED_TEST(VectorRegisterTest, minmax_tests)
     b[i] = fetch::math::Type<type>(
         std::to_string((static_cast<double>(std::rand()) / static_cast<double>(RAND_MAX)) *
                        static_cast<double>(fetch::math::numeric_max<type>()) / 2.0));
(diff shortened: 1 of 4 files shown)


libs/math/tests/unit/vectors/vector_tests.cpp
```diff
@@ -208,6 +208,7 @@ TYPED_TEST(VectorRegisterTest, minmax_tests)
     b[i] = fetch::math::Type<type>(
         std::to_string((static_cast<double>(std::rand()) / static_cast<double>(RAND_MAX)) *
                        static_cast<double>(fetch::math::numeric_max<type>()) / 2.0));
+    b[i]     = b[i] == 0 ? a[i] : b[i];
```
baykaner, line 211: is this just to make sure you don't divide by 0 later?

baykaner (approved):

libs/math/tests/unit/vectors/vector_tests.cpp
markos: yes

PR #2357, opened 2020-01-29
Bug/consensus stuck

denis-trapeznikov:


libs/ledger/src/consensus/consensus.cpp
```diff
@@ -596,10 +610,14 @@ NextBlockPtr Consensus::GenerateNextBlock()
...
   // into GetBlockGenerationWeight (important for first block of each aeon which specifies the
   // qual for this aeon)
-  if (EntropyGeneratorInterface::Status::OK !=
-      beacon_->GenerateEntropy(block_number, ret->block_entropy))
   {
-    return {};
+    auto entgen_status = beacon_->GenerateEntropy(block_number, ret->block_entropy);
+    if (entgen_status != EntropyGeneratorInterface::Status::OK)
+    {
+      TelemetryOnFail("failed_to_generate_entropy")->increment();
```
ejfitzgerald: Lets store these shared pointers as class variables in the same way as the other modules. So long as you only include "telemetry/telemetry.hpp" in the header file the coupling will be minor.

ejfitzgerald (changes requested):

ejfitzgerald (approved):

PR #2361, opened 2020-01-29
Fix serializer element count; typos

wkaluza:


libs/ledger/include/ledger/consensus/stake_manager.hpp
```diff
@@ -167,7 +167,7 @@ struct MapSerializer<ledger::StakeManager, D>
   template <typename Constructor>
   static void Serialize(Constructor &map_constructor, Type const &stake_manager)
   {
-    auto map = map_constructor(5);
+    auto map = map_constructor(4);
```
wkaluza, line 167: Could this break backwards compatibility? Looking at the code, it doesn't seem like it, but I've been wrong before.

ejfitzgerald: I think actually the way that our serialisers work this can be fine. It is quite easy to test, however, if you run up a small docker image with the version before the change and then with this one you can check

wkaluza: It works!

ejfitzgerald (approved):

PR #2364, opened 2020-01-30
Feature/ml 253 add keras style pooling layers

iryna-melnyk:
*pooling layers
*tests

### DIFF
--- a/libs/ml/src/ops/max_pool_1d.cpp
+++ b/libs/ml/src/ops/max_pool_1d.cpp
@@ -170,8 +170,13 @@ std::vector<fetch::math::SizeType> MaxPool1D<T>::ComputeOutputShape(
   // output_shape_[0]=number of output channels
   output_shape.emplace_back(inputs.at(0)->shape().at(0));
   // output_shape_[1]=number of stride_size steps over input size
-  output_shape.emplace_back((inputs.at(0)->shape().at(1) - (kernel_size_ - stride_size_)) /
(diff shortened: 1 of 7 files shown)


libs/ml/src/ops/max_pool_1d.cpp
```diff
@@ -170,8 +170,13 @@ std::vector<fetch::math::SizeType> MaxPool1D<T>::ComputeOutputShape(
   // output_shape_[0]=number of output channels
   output_shape.emplace_back(inputs.at(0)->shape().at(0));
   // output_shape_[1]=number of stride_size steps over input size
-  output_shape.emplace_back((inputs.at(0)->shape().at(1) - (kernel_size_ - stride_size_)) /
-                            stride_size_);
+  SizeType output_height = ((inputs.at(0)->shape().at(1) - kernel_size_) / stride_size_) + 1;
```
vhavrylov-ncube, line 173: "height" is a bit confusing for 1d convolution, m.b. let it be "length" ?

vhavrylov-ncube (commented):
looks good, but let others approve too.

libs/ml/src/ops/max_pool_1d.cpp
iryna-melnyk: agree but it needs to be discussed with @MissingNO57 i believe

baykaner (approved):

PR #2365, opened 2020-01-31
added backward charge estimation to losses

AlexTeslenko:
added backward charge estimation to losses


vhavrylov-ncube (commented):
Looks good, but it looks like we have no recursive computing of a backward pas from error layer until a particular node.

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -170,7 +170,7 @@ class Ops
    * @return estimated charge amount, necessary for performing a backward pass on data of given
    * shapes.
    */
-  virtual OperationsCount ChargeBackward() const
+  virtual OperationsCount ChargeBackward()
```
vhavrylov-ncube, line 173: Are you sure about removing the const specifier? This method is intended to be a getter-only, e.g. a const one.

vhavrylov-ncube (changes requested):
I would vote for keeping getters with "const" specifier

kublai-baykaner (approved):

vhavrylov-ncube (approved):
LGTM

PR #2366, opened 2020-01-31
Fix teardown segfault if no genesis file

wkaluza:


ejfitzgerald (approved):
Can you raise a backport PR to the release branch please also

PR #2367, opened 2020-01-31
LGDR-630 Refactor/remove heaviest chain endpoint

denis-trapeznikov:
GetHeaviestChain removed from both service and client.


ejfitzgerald (changes requested):
Can you remove the noise-c library update in this PR please

ejfitzgerald:
Otherwise it looks good

PR #2371, opened 2020-01-31
ML-252 added embeddings

iryna-melnyk:
*added bindings to embedding layer
*added smoke test

### DIFF
--- a/libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
+++ b/libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
@@ -444,7 +444,8 @@ TEST_F(VMModelTests, model_add_flatten)
 
 TEST_F(VMModelTests, model_add_embeddings)
 {
-  TestValidLayerAdding(R"(model.addExperimental("embeddings", 2u64, 3u64, true);)", IGNORE_CHARGE_ESTIMATION);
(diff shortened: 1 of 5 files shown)


baykaner (dismissed):

baykaner (dismissed):

libs/vm-modules/tests/unit/ml/ml_model_tests.cpp
```diff
@@ -1000,6 +1011,25 @@ TEST_F(VMModelTests, model_sequential_avgpool2d)
...
+TEST_F(VMModelTests, model_sequential_embeddings)
+{
+  static char const *sequential_model_src = R"(
+        function main()
+          var x = Tensor();
+          var str_vals = "1.5; 2.1; 3.1; 4.2; 5.1; 6.; 7.1; 8.1; 9.0;";
+          x.fromString(str_vals);
+          var model = Model("sequential");
+          model.addExperimental("input", x.shape());
+          model.addExperimental("embeddings", 10u64, 4u64, true);
```
baykaner: looks a bit confusing: aren't the inputs meant to be row indices? what actually happens internally when you run this?

iryna-melnyk: To be precise, static cast is happening
auto embedding_view = this->data_->View(static_cast<SizeType>(*input_it));
But you're absolutely right, I'll change the input to be less confusing

baykaner (approved):

PR #2380, opened 2020-02-03
Bug/quickfix for byte size

denis-trapeznikov:
A quick and dirty workaround for the recent ByteSize-related build failures, until we get all the versions sorted out.


ejfitzgerald:
@denis-trapeznikov I think that this should be sorted on master now. Can you raise a PR for the noise-c changes back to @jinmannwong repo and then submit a PR with the submodule reference made

ejfitzgerald (changes requested):
See previous comment

PR #2388, opened 2020-02-03
Remove unnecessary inline specifiers

wkaluza:
* Member functions are implicitly inline
* Functions marked as `constexpr` are implicitly inline
* Non-specialised function templates are implicitly inline


libs/vectorise/include/vectorise/fixed_point/fixed_point.hpp
```diff
@@ -568,7 +568,7 @@ class FixedPoint : public BaseFixedpointType
 };
 
 template <typename T>
-static inline bool IsMinusZero(T n);
+bool IsMinusZero(T n);
```
ejfitzgerald, line 571: Why the drop of `static`

wkaluza: What's the point of having it there? All it does is confer internal linkage, which is meaningless in a header file.

ejfitzgerald (approved):

PR #2394, opened 2020-02-04
Update SDK version

ejfitzgerald:


cyenyxe (approved):

cyenyxe (changes requested):
Actually the build is failing, I think due to some incompatibility in the interfaces.

cyenyxe (approved):

PR #2395, opened 2020-02-05
Improvement/recursive debug mutex

denis-trapeznikov:
This commit adds `std::recursive_mutex` as a possible base for `fetch::DebugMutex`, and brings a little bit more order into mutex debugging.


libs/core/tests/unit/mutex.cpp
```diff
@@ -116,4 +119,49 @@ TEST(DebugMutex, DISABLED_MultiThreadDeadlock2)
...
+  };
+
+  std::vector<std::thread> threads;
+  threads.emplace_back(f, 'a');
+  threads.emplace_back(f, 'b');
+
+  threads[0].join();
+  threads[1].join();
+
+  ASSERT_TRUE(rv == "aaaabbbb" || rv == "bbbbaaaa");
```
n-hutton: This test does not seem to test that recursive mutex detects a deadlock/infinite loop. Does recursive mutex throw if there is a deadlock, and on what condition will it throw?

denis-trapeznikov: Yes, that should be a separate test, as I thought this morning. This one only tests for the lock to be acquired and released recursively, when no deadlock occurs.

ejfitzgerald (changes requested):
I like the way that this is heading. Am I missing something that it looks like we can't register deadlocks across combinations of normal and recursive mutexs - is that correct?

Also noise-c! 😁

denis-trapeznikov:
> Am I missing something that it looks like we can't register deadlocks across combinations of normal and recursive mutexs - is that correct?

Well, chances are your mutex will be locked from line 0 of file ''.
This denial of logic can actually be addressed, but that would require a bit of filthy black macro magic ... unsure if we _seriously_ need it at this moment.

libs/core/include/core/mutex.hpp
```diff
@@ -33,106 +41,349 @@ namespace fetch {
...
+{
+protected:
+  using Mutex = SimpleDebugMutex;
+
+  struct OwnerId
+  {
+    std::thread::id id = std::this_thread::get_id();
+  };
+
+  static constexpr bool Populate(OwnerId & /*owner*/) noexcept
```
n-hutton: These methods feel a bit unclean - due to using templating as a kind of inheritence they basically do nothing for this class?

libs/core/include/core/mutex.hpp
```diff
@@ -33,106 +41,349 @@ namespace fetch {
...
+
+  static bool IsDeadlocked(OwnerId const &owner) noexcept;
+};
+
+class RecursiveDeadlock
+{
+protected:
+  using Mutex = RecursiveDebugMutex;
+
+  struct OwnerId
```
n-hutton: These classes seem confusingly named, insead of RecursiveDeadlock and OwnerId is LockAttempt and LockDetails not more accurate?

libs/core/include/core/mutex.hpp
```diff
@@ -33,106 +41,349 @@ namespace fetch {
...
+   * @param mutex Pointer to the mutex locked
+   * @param location Code point the lock is performed at
+   */
+  static void RegisterMutexAcquisition(MutexPtr mutex, LockLocation location = LockLocation())
   {
-    if (static_cast<bool>(throw_on_deadlock_))
+    auto &         instance = Instance();
+    GuardOfMutexes non_fetch_lock_to_avoid_recursion(instance.mutex_);
+
+    // Registering the matrix diagonal
```
n-hutton: I don't understand what diagonal means in this context?

libs/core/src/mutex.cpp
```diff
@@ -26,133 +28,112 @@ namespace fetch {
...
+ * Either throws a runtime error with a message,
+ * or prints that message on stderr and aborts,
+ * depending on throw_on_deadlock_ value.
+ *
+ * @param message Error message for the exception raised/to be printed before abort()
+ */
+void DeadlockHandler::DeadlockDetected(std::string message)
 {
-  FETCH_LOCK(mutex_);
+  if (static_cast<bool>(throw_on_deadlock_))
```
n-hutton, line 47: is this cast needed?

libs/core/tests/unit/mutex.cpp
```diff
@@ -27,51 +27,54 @@
...
   }
 }
 
 TEST(DebugMutex, MultiThreadDeadlock)
 {
-  fetch::MutexRegister::ThrowOnDeadlock();
-  fetch::DebugMutex m[5];
-  auto              f = [&m](int32_t n) {
-    std::lock_guard<fetch::DebugMutex> guard1(m[n]);
+  fetch::DeadlockHandler::ThrowOnDeadlock();
```
n-hutton: Seems odd to do it this way - the user will not know if both throw on deadlock and abort on deadlock are set what the priority would be. Wouldn't it make more sense to do `DeadlockHandler::BehaviourOnDeadlock(DeadlockBehaviour::Throw)`?

libs/core/tests/unit/mutex.cpp
```diff
@@ -116,4 +119,77 @@ TEST(DebugMutex, DISABLED_MultiThreadDeadlock2)
...
+{
+  fetch::DeadlockHandler::ThrowOnDeadlock();
+  RMutex m;
+
+  std::vector<std::thread> threads;
+
+  threads.emplace_back([&m] {
+    FETCH_LOCK(m);
+    FETCH_LOCK(m);
+    std::this_thread::sleep_for(std::chrono::seconds(4));
```
n-hutton: This test can fail based on race conditions and so is flaky. It is not obvious why there is a 4 second wait and will not scale to longer timeout durations. It also unneccessarily incresases our CI time to have sleeps when they could be avoided. I would suggest rewriting it to address these concerns

n-hutton (changes requested):
In general this is a good idea but I think there are the following issues:
-No high level comment in mutex.hpp that describes the behaviour of the deadlock detection, and how it achieves this
- @ejfitzgerald mentions that if we are to abort or even throw it should be on the timescales of 40m or so
- I do not think this would detect a deadlock where a thread locks recursive and non-recursive mutexes
-The unit tests do not run in our CI! Due to deadlock not being compiled in by default
-There is no test for a classic deadlock (priority inversion) which I find odd. Perhaps add the following:

```
TEST(DebugMutex, ClassicDeadlock)
{
  fetch::DeadlockHandler::ThrowOnDeadlock();

  Mutex a;
  Mutex b;

  std::atomic<bool> t1_acquired{false};
  std::atomic<bool> t2_acquired{false};
  std::atomic<int> throws{0};

  auto  thread_runme = [&throws](Mutex *one, Mutex *two, std::atomic<bool> *thread_acquired_mutex) {
    FETCH_LOCK(*one);

    *thread_acquired_mutex = true;

    while(*thread_acquired_mutex == true)
    {
      std::this_thread::sleep_for(std::chrono::milliseconds(5));
    }

    try
    {
      FETCH_LOCK(*two);
    }
    catch(...)
    {
      throws++;
    }
  };

  std::vector<std::thread> threads;

  std::thread t_one(thread_runme, &a, &b, &t1_acquired); // A, B
  std::thread t_two(thread_runme, &b, &a, &t2_acquired); // B, A

  while(!(t1_acquired && t2_acquired))
  {
    std::this_thread::sleep_for(std::chrono::milliseconds(5));
  }

  t1_acquired = false;
  t2_acquired = false;

  t_one.join();
  t_two.join();

  EXPECT_EQ(throws, 1);
}
```

libs/core/src/mutex.cpp
denis-trapeznikov: IIRC implicit casts break build.

libs/core/tests/unit/mutex.cpp
denis-trapeznikov: Well, it would be a refactor of current behaviour on itself, unsure if should go in the same PR.

libs/core/include/core/mutex.hpp
denis-trapeznikov: Whatever Troels meant when adding this comment? He had in mind, I assume, mutex-thread-location relations but I'm not sure the exact convolution of incidence tensor nailed down to an operator over a matrix diagonal can be instantly recovered here.

libs/core/include/core/mutex.hpp
denis-trapeznikov: Yes, they are no-ops for non-recursive. The mutex locking decomposition performed in this commit can be regarded as: same locking/unlocking/detection mechanisms are represented as A * B for recursive mutexes, but A * E for simple ones, where A is the most common set of actions, and E is identity matrix. These methods belong to the common interface of B and E; thus here they are present but do nothing.

libs/core/include/core/mutex.hpp
denis-trapeznikov: Unsure if LockAttempt is less confusing but changed it.

libs/core/tests/unit/mutex.cpp
denis-trapeznikov: 4 seconds are chosen as a time greater than assumed deadlock threshold of 2s. You've mentioned above that this should be 40m (0_o) which definitely exceeds _any_ CI time.

n-hutton (commented):
I don't know how much work you have done on this - is it ready for re-review?

denis-trapeznikov:
> I don't know how much work you have done on this - is it ready for re-review?

According to the label.

libs/core/tests/unit/mutex.cpp
```diff
@@ -22,56 +22,84 @@
...
   {
-    fetch::DebugMutex                  mutex;
-    std::lock_guard<fetch::DebugMutex> guard1(mutex);
-    EXPECT_THROW(std::lock_guard<fetch::DebugMutex> guard2(mutex), std::runtime_error);
+    Mutex                  mutex;
+    std::lock_guard<Mutex> guard1(mutex);
+    EXPECT_THROW(std::lock_guard<Mutex> guard2(mutex), std::runtime_error);
   }
 
+  static constexpr std::size_t TABLE_SIZE = 7;
```
n-hutton, line 60: What is the dining philosophers testing here? It seems racy, if you change the logic to this:

```
  for (std::size_t fork = 0; fork < TABLE_SIZE; ++fork)
  {
    Mutex &left_fork  = forks[fork];
    Mutex &right_fork = forks[(fork + 1) % TABLE_SIZE];
    dining_philosophers.emplace_back(runnable, std::ref(left_fork), std::ref(right_fork));
    std::this_thread::sleep_for(std::chrono::milliseconds(95));
  }
```

It will likely fail I believe

libs/core/src/mutex.cpp
```diff
@@ -17,144 +17,129 @@
...
+#include "moment/clocks.hpp"
 
+#include <cassert>
 #include <iostream>
 #include <sstream>
 
-namespace fetch {
+using namespace std::chrono_literals;
 
-#ifdef FETCH_DEBUG_MUTEX
```
n-hutton, line 26: @ejfitzgerald do we want to turn on the debug mutex by default?

n-hutton (commented):
Could you summarise the changes that have been made since the last review? Since there seems to be a fair amount of changes.

libs/core/src/mutex.cpp
ejfitzgerald: No I don't think so. Lets leave it as it currently is as an option to be enabled

PR #2402, opened 2020-02-06
Revert stake man seralizer fix

n-hutton:
This change would break backward compatability with previously generated genesis block hashes - I tested this manually to confirm. Caught when soaking on devnet - new nodes could not sync up since they started with a different genesis


wkaluza:
As discussed, I don't think we need this on master, just the current release branch.

PR #2406, opened 2020-02-06
Bootstrap v3 Support

ejfitzgerald:
Updated bootstrap configuration to support genesis file and parameter configuration of the requesting node

### DIFF
--- a/apps/constellation/settings.cpp
+++ b/apps/constellation/settings.cpp
@@ -70,9 +70,9 @@ public:
     return static_cast<int>(argc_);
   }
 
-  char **argv() const
+  char const *const *argv() const
   {
-    return const_cast<char **>(argv_.data());
+    return argv_.data();
   }
 
 private:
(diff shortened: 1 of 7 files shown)


n-hutton (dismissed):

apps/constellation/settings.cpp
```diff
@@ -43,6 +43,44 @@ const uint32_t DEFAULT_MAX_PEERS          = 3;
...
+  }
+
+  int argc() const
+  {
+    return static_cast<int>(argc_);
+  }
+
+  char **argv() const
+  {
+    return const_cast<char **>(argv_.data());
```
wkaluza: !

Just make `UpdateFromArgs` accept `char const **`. There's some knock-on effects from that, but it's better than casting away `const`. I know there is pressure for a new release, but this is a risk we can do without.

Also, you seem to be at least partly rewriting `SystemArgAdapter` from `setting_collection_tests.cpp`.

`UpdateBatch` does not seem to exist on `master`. Is this intentional?

apps/constellation/settings.cpp
```diff
@@ -43,6 +43,44 @@ const uint32_t DEFAULT_MAX_PEERS          = 3;
...
+    for (auto const &element : batch_)
+    {
+      assert(idx + 1 < argv_.size());
+
+      argv_[idx]     = element.first.c_str();
+      argv_[idx + 1] = element.second.c_str();
+
+      idx += 2;
+    }
+  }
```
wkaluza, line 66: Wouldn't an old-style for loop work better here?

wkaluza (changes requested):

wkaluza (approved):

PR #2411, opened 2020-02-10
Bugfix/fp32 vector multiplication fix

markos:
Fixes a bug in AVX2 multiplication, comparison to max should be done after shifting right.


MissingNO57 (approved):
LGTM

PR #2412, opened 2020-02-10
WIP: Feature/exp fp32 vectorisation

markos:
Exp() vectorisation, fp32 version
Implements Exp() to use ParallelDispatcher for all cases, but only fp32 is actually vectorised.
vectorised the Pade approximant but there is no integer division in AVX2, so it's not actually faster.
But it's currently bit-precise to the scalar version


libs/vectorise/include/vectorise/arch/avx2/math/exp.hpp
```diff
@@ -0,0 +1,310 @@
...
+static const fixed_point::fp64_t Exp_P04(0, 0x00410410);  //  1 / 1008
+static const fixed_point::fp64_t Exp_P05(0, 0x00022ACD);  //  1 / 30240
+
+static const fixed_point::fp64_t integer_mask(0xffffffff, 0);  //  1 / 30240
+
+static const fixed_point::fp64_t one_over_LN2{fixed_point::fp64_t::_1/fixed_point::fp64_t::CONST_LN2};  //  1 / CONST_LN2
+
+static const VectorRegister<fixed_point::fp32_t, 128> FP32_128_INTEGER_MASK(static_cast<fixed_point::fp32_t>(integer_mask));
+static const VectorRegister<fixed_point::fp32_t, 256> FP32_256_INTEGER_MASK(static_cast<fixed_point::fp32_t>(integer_mask));
+// static const VectorRegister<fixed_point::fp64_t, 128> FP64_128_INTEGER_MASK{integer_mask};
```
baykaner, line 44: we'll need to add a TODO and a task to Jira here regarding the fp64

baykaner (approved):

PR #2413, opened 2020-02-10
Persistent Transaction Status

ejfitzgerald:
This PR introduces the concept of allowing the ledger to store persistent transaction status information. This is off by default, but we will probably roll this change to the main network nodes


libs/constellation/src/constellation.cpp
```diff
@@ -359,6 +360,16 @@ Constellation::MessengerAPIPtr CreateMessengerAPI(Config const &cfg, muddle::Mud
...
 
+TransactionStatusPtr CreateTransactionStatusCache(Config const &config)
+{
+  if (config.persistent_tx_status)
+  {
+    return ledger::TransactionStatusInterface::CreatePersistentCache();
+  }
+
+  return ledger::TransactionStatusInterface::CreateTimeBasedCache();
+}
```
wkaluza, line 370: Wrap in unnamed namespace.

libs/ledger/include/ledger/transaction_status_cache.hpp
```diff
@@ -18,24 +18,18 @@
...
 namespace ledger {
 
 enum class TransactionStatus : uint8_t
 {
-  UNKNOWN,    ///< The status of the transaction is unknown
-  PENDING,    ///< The transaction is waiting to be mined
-  MINED,      ///< The transaction has been mined
-  EXECUTED,   ///< The transaction has been executed
-  SUBMITTED,  ///< Special case for the data based synergetic transactions
+  UNKNOWN = 0,  ///< The status of the transaction is unknown
```
wkaluza, line 28: Why the explicit 0?

libs/ledger/src/persistent_transaction_status_cache.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+  template <typename MapDeserializer>
+  static void DeserializeStatus(MapDeserializer &map, StatusEnum &status)
+  {
+    // read the raw value
+    uint8_t raw_enum_value{0};
+    map.ExpectKeyGetValue(STATUS, raw_enum_value);
+
+    // convert the raw value
+    auto converted_value = static_cast<StatusEnum>(raw_enum_value);
```
wkaluza, line 106: I don't think this trick works as well as one would reasonably expect. Can you double-check this cast does not blow up if `raw_enum_value` has a value not covered by `StatusEnum`? I recall trying to do something similar back on Windows, resulting in runtime errors. To my mind, the right way would be to exclude invalid values with `if (raw_enum_value < StatusEnum::INEXPLICABLE_FAILURE) { ... }` or similar.

libs/ledger/src/persistent_transaction_status_cache.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+  template <typename MapDeserializer>
+  static void DeserializeStatus(MapDeserializer &map, StatusEnum &status)
+  {
+    // read the raw value
+    uint8_t raw_enum_value{0};
+    map.ExpectKeyGetValue(EXECUTION_RESULT, raw_enum_value);
+
+    // convert the raw value
+    auto converted_value = static_cast<StatusEnum>(raw_enum_value);
```
wkaluza, line 168: Same as above.

libs/ledger/src/persistent_transaction_status_cache.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+#include "core/serializers/main_serializer.hpp"
+#include "logging/logging.hpp"
+
+namespace fetch {
+
+namespace serializers {
+
+template <typename D>
+struct MapSerializer<fetch::ledger::StakeUpdateEvent, D>
```
wkaluza, line 29: Don't these serializers need to be in a header file?

libs/ledger/src/time_based_transaction_status_cache.cpp
```diff
@@ -0,0 +1,160 @@
...
+{
+  {
+    FETCH_LOCK(mtx_);
+
+    auto const it = cache_.find(digest);
+    if (cache_.end() != it)
+    {
+      return it->second.status;
+    }
+  }
```
wkaluza: The outer scope here does nothing.

libs/ledger/src/time_based_transaction_status_cache.cpp
```diff
@@ -0,0 +1,160 @@
...
+    it->second.status.status               = EXECUTED_STATUS;
+    it->second.status.contract_exec_result = exec_result;
+  }
+
+  PruneCacheIfNecessary(now);
+}
+
+void TimeBasedTransactionStatusCache::PruneCache(Timestamp const &until)
+{
+  fetch::generics::MilliTimer timer{"TxStatusCache::Prune"};
```
wkaluza, line 127: Update timer name.

wkaluza (changes requested):

libs/ledger/src/persistent_transaction_status_cache.cpp
denis-trapeznikov:
```
$ cat omg.cpp 
#include <iostream>

enum class Omg {
	A,
	B
};

int main() {
	Omg x = static_cast<Omg>(3);
	std::cout << static_cast<int>(x) << '\n';
}
$ g++ -std=c++14 -Wpedantic omg.cpp 
$ ./a.out 
3
```
I don't see anything particular on CPPRef about casting values of underlying type to enum, even if they don't equal to any of the named constants.
(Though for C++17, the wording is definitely more verbose.)
Comparing against `INEXPLICABLE_FAILURE` would be probably the same as checking a pointer for non-nullness before deleting it.

wkaluza: I had a look and the only restriction appears to be this:

> Values of integer, floating-point, and enumeration types can be converted by static_cast or explicit cast, to any enumeration type. If the underlying type is not fixed and the source value is out of range, the result is unspecified (until C++17)undefined (since C++17). 

https://en.cppreference.com/w/cpp/language/enum

In this case we should be ok, because `raw_enum_value` is a `uint8_t`, and `StatusEnum` is an enum class of unspecified underlying type, which in practice means its an `int`. I withdraw my question above.

Just to check I'm not mad, I had a look at MSDN. It appears that Visual Studio's enum implementation is indeed non-conformant:

> The static_cast operator can explicitly convert an integral value to an enumeration type. If the value of the integral type does not fall within the range of enumeration values, the resulting enumeration value is undefined.

https://docs.microsoft.com/en-us/cpp/cpp/static-cast-operator?view=vs-2019

libs/ledger/src/time_based_transaction_status_cache.cpp
ejfitzgerald: I think that the name is good here to be honest.

libs/ledger/src/time_based_transaction_status_cache.cpp
ejfitzgerald: Strictly speaking the mutex is release slightly earlier - this was just a copy of the old code, but I take the larger point that this is not necessary in this case - Done

libs/ledger/src/persistent_transaction_status_cache.cpp
ejfitzgerald: Nope - I thought that you would be happy with this. They are only used here, in the future if they are needed elsewhere we can add them to a common header.

The serialisers as they are, add quite a bit to all the template code - so keep them in the `.cpp` file if we can I say.

libs/ledger/include/ledger/transaction_status_cache.hpp
ejfitzgerald: No real need, just think that it is clearer

libs/constellation/src/constellation.cpp
ejfitzgerald: Done

libs/ledger/src/time_based_transaction_status_cache.cpp
wkaluza: But we now have two different status caches - as it stands, the logs will not be explicit whether `TxStatusCache` refers to the time-based one or the persistent one.

libs/ledger/src/persistent_transaction_status_cache.cpp
wkaluza: That was not an objection - I'm merely surprised that it works, is all. I was under the impression that it is necessary for our serialisers to be in header files. Score!

libs/ledger/src/time_based_transaction_status_cache.cpp
ejfitzgerald: The persistent one doesn't have a prune (any will never have one I would guess). This is why this seemed like a fair name

wkaluza (dismissed):

wkaluza (approved):

PR #2414, opened 2020-02-10
Model estimation functions moved from Estimator to VMModel

MissingNO57:
Serialisation of Ops base class parameters fixed.
Added Evaluate estimation to Model.
Removed Evaluate and Predict estimations from ModelEstimator.
FIxed Predic and Evaluate estimation tests.

### DIFF
--- a/libs/ml/src/ops/activations/sigmoid.cpp
+++ b/libs/ml/src/ops/activations/sigmoid.cpp
@@ -33,7 +33,8 @@ Sigmoid<TensorType>::Sigmoid(SPType const &sp)
 template <typename TensorType>
 std::shared_ptr<OpsSaveableParams> Sigmoid<TensorType>::GetOpSaveableParams()
 {
-  return std::make_shared<SPType>();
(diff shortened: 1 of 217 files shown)


libs/ml/src/ops/activations/sigmoid.cpp
```diff
@@ -33,7 +33,8 @@ Sigmoid<TensorType>::Sigmoid(SPType const &sp)
 template <typename TensorType>
 std::shared_ptr<OpsSaveableParams> Sigmoid<TensorType>::GetOpSaveableParams()
 {
-  return std::make_shared<SPType>();
+  auto sp = std::make_shared<SPType>();
```
baykaner, line 36: what was the point in this change?

baykaner (dismissed):

libs/ml/src/ops/activations/sigmoid.cpp
MissingNO57: Point is I forgot to add base class savable params!

baykaner (approved):

PR #2419, opened 2020-02-11
Improvement/add not subtree txs to recent queue

denis-trapeznikov:
This commit implements LGDR-672.1.

Transactions now can be marked as obtained from a subtree sync and thus not recent.


libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
```diff
@@ -106,32 +98,30 @@ TransactionStoreSyncService::TransactionStoreSyncService(Config const &cfg, Mudd
...
-                                  &TransactionStoreSyncService::OnQuerySubtree);
-  state_machine_->RegisterHandler(State::RESOLVING_SUBTREE, this,
-                                  &TransactionStoreSyncService::OnResolvingSubtree);
-  state_machine_->RegisterHandler(State::QUERY_OBJECTS, this,
-                                  &TransactionStoreSyncService::OnQueryObjects);
-  state_machine_->RegisterHandler(State::RESOLVING_OBJECTS, this,
-                                  &TransactionStoreSyncService::OnResolvingObjects);
-  state_machine_->RegisterHandler(State::TRIM_CACHE, this,
-                                  &TransactionStoreSyncService::OnTrimCache);
+  // clang-format off
```
ejfitzgerald: Lets use the previous version (maybe introduce the clang format off and tidy a little. The main reason is that this way are are limited to using signature: `State(TransactionStoreSyncService::*)()` however, it is very useful when debugging to be able to change this to one of the other handlers like: `State(TransactionStoreSyncService::*)(State, State)`

libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
```diff
@@ -106,32 +98,30 @@ TransactionStoreSyncService::TransactionStoreSyncService(Config const &cfg, Mudd
...
     FETCH_UNUSED(new_state);
     FETCH_LOG_DEBUG(LOGGING_NAME, "*** Updating state to: ", ToString(new_state));
   });
 }
 
-TransactionStoreSyncService::~TransactionStoreSyncService()
-{
-  client_ = nullptr;
-}
+TransactionStoreSyncService::~TransactionStoreSyncService() = default;
```
ejfitzgerald: Move to the header?

libs/chain/include/chain/transaction.hpp
```diff
@@ -137,6 +137,8 @@ class Transaction
   bool Verify();
   bool IsVerified() const;
   bool IsSignedByFromAddress() const;
+  void SetFromSync();
```
ejfitzgerald: I think these names are a little vague. Maybe invert the logic and use `SetRecent(bool recent = true)` and `IsRecent()` - this will more closely match the other API too I guess

ejfitzgerald (changes requested):

libs/chain/include/chain/transaction.hpp
denis-trapeznikov: One thing I didn't like about this inversion was that it requires calling `SetRecent` everywhere a recent tx may be received, while `SetFromSync` is only needed after a sync, and all other tx handling sites, either real or imaginary, can be left intact, unaware of this bit.

ejfitzgerald: Sure okay that is fair - Can we be more explicit then and go with. `SetFromSubtreeSync` etc? Thoughts?

denis-trapeznikov: Changed to .*FromSubtreeSync, don't have a more descriptive idea either.

ejfitzgerald (approved):

PR #2421, opened 2020-02-12
Bloom filter performance improvements

n-hutton:
The bloom filter performance can be improved by only filling it with blocks that go to disk (during healthy operation this will be a single heavy chain). This avoids having to manage the bloom filter for multiple forks but does mean that checks for duplicates will usually have to parse around 10 blocks (in memory).

Minor unrelated debug also added.

### DIFF
--- a/libs/ledger/src/chain/main_chain.cpp
+++ b/libs/ledger/src/chain/main_chain.cpp
@@ -2154,7 +2154,7 @@ DigestSet MainChain::DetectDuplicateTransactions(BlockHash const &           sta
 
   bloom_filter_false_positive_count_->add(potential_duplicates.size() - duplicates_disk.size());
   bloom_filter_walk_count_->add(blocks_walked);
-  bloom_filter_positive_count_->add(duplicates_on_disk);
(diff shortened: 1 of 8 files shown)


libs/ledger/src/chain/main_chain.cpp
```diff
@@ -2004,35 +2025,78 @@ DigestSet MainChain::DetectDuplicateTransactions(BlockHash const &           sta
...
+        }
+      }
+    }
+
+    // termination condition
+    if (!LookupBlockFromCache(block->previous_hash, block))
+    {
+      // Sanity check - there should be the continuation of this on disk
+      IntBlockPtr block_dummy;
+      assert(LookupBlock(block->previous_hash, block_dummy));
```
n-hutton, line 2067: ifndef here

ejfitzgerald (dismissed):

ejfitzgerald (dismissed):

ejfitzgerald (dismissed):

ejfitzgerald (approved):

PR #2422, opened 2020-02-12
Model estimations improvement

MissingNO57:
Deprecated ModelEstimator


libs/ml/include/ml/ops/weights.hpp
```diff
@@ -81,8 +81,14 @@ class Weights : public fetch::ml::ops::Variable<T>
                          WeightsInitialisation mode = WeightsInitialisation::XAVIER_GLOROT,
                          SizeType              seed = 123456789);
 
+  static OperationsCount ChargeInitialise(uint64_t in_size, uint64_t out_size,
+                                          WeightsInitialisation mode);
+  // static OperationsCount        ChargeInitialise(uint64_t data_size, WeightsInitialisation mode);
```
evsmithx: commented code

libs/ml/src/layers/fully_connected.cpp
```diff
@@ -161,6 +161,129 @@ OperationsCount FullyConnected<TensorType>::ChargeBackward() const
...
+  // AddInputNode, SetOutputNode
+  op_cnt += 2;
+
+  // If inputs count is known, the initialisation can be completed immediately.
+  if (in != AUTODETECT_INPUTS_COUNT)
+  {
+    // Set batch_input_shapes_
+    op_cnt += 1;
+
+    // ComputeBatchOutputShape
```
evsmithx: Should op_cnt be incremented here?

libs/ml/src/layers/fully_connected.cpp
```diff
@@ -161,6 +161,129 @@ OperationsCount FullyConnected<TensorType>::ChargeBackward() const
...
+
+    // SetBatchInputShapes, SetBatchOutputShape
+    op_cnt += 2;
+
+    op_cnt += ChargeCompleteConstruction(false, init_mode, time_distributed, in, out);
+  }
+
+  FETCH_UNUSED(regulariser);
+  FETCH_UNUSED(regularisation_rate);
```
evsmithx, line 258: No operations for the regularisation?

libs/ml/src/ops/weights.cpp
```diff
@@ -130,6 +130,18 @@ void Weights<TensorType>::Initialise(TensorType &array, uint64_t in_size, uint64
   }
 }
 
+/**
+ * interface to call standard weights initialisation routines. defaults to xavier
+ * @param mode  An enum indicating which type of initialisation to perform
```
evsmithx: doc needs updating

libs/ml/src/layers/fully_connected.cpp
MissingNO57: I am not sure how expensive is that, maybe I need to implement function EstimateComputeBatchOutputShape.

libs/ml/src/layers/fully_connected.cpp
MissingNO57: It's just assigning 1 ENUM value - regulation type, so it's always constant.

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -205,6 +205,15 @@ class Ops
     return total_elements;
   }
 
+  /**
+   * Default Op construction charge
+   * @return
+   */
+  static OperationsCount ChargeConstruct()
+  {
+    return 1;
```
baykaner, line 228: add a constant to ml/charge_estimation for op construction and invoke it here (even if the value remains as 1)

libs/ml/src/layers/fully_connected.cpp
```diff
@@ -161,6 +161,130 @@ OperationsCount FullyConnected<TensorType>::ChargeBackward() const
...
+  SizeType weights_shape_size = total_outputs * total_inputs;
+  SizeType biases_shape_size  = total_outputs;
+
+  // 2 x SetBatchOutputShape
+  op_cnt += 2;
+
+  // Create weights Tensor
+  op_cnt += weights_shape_size;
+
+  op_cnt += fetch::ml::ops::Weights<TensorType>::ChargeInitialise({total_outputs, total_inputs});
```
baykaner: just a note - this will need to change a little after you merge with my PR because the tensor initialisation is being moved out of the op constructors

libs/ml/src/ops/dataholder.cpp
```diff
@@ -112,6 +112,12 @@ OperationsCount DataHolder<TensorType>::ChargeBackward() const
   return cost;
 }
 
+template <typename TensorType>
+OperationsCount DataHolder<TensorType>::ChargeConstruct()
+{
+  return 1;
```
baykaner: I doubt this is correct. Constructing a DataHolder means constructing a tensor, which is more complex than 1 primitive operation. Add a constant to ml/charge_estimation (maybe called EMPTY_TENSOR_CONSTRUCTION or something like that) and use it here; i'm sure it will get reused elsewhere too

libs/ml/src/optimisation/adam_optimiser.cpp
```diff
@@ -157,6 +157,55 @@ void AdamOptimiser<T>::ResetCache()
...
+      break;
+    }
+  }
+
+  OperationsCount op_cnt{1};
+  for (auto &train : graph->GetTrainables())
+  {
+    auto     weight    = train->GetWeights();
+    SizeType data_size = TensorType::PaddedSizeFromShape(weight.shape());
+    op_cnt += data_size * 5;
```
baykaner: it's not super great to have an unexplained value (5) here. At least this should be a named constant which hints at what the 5 is doing and why the value is 5.

baykaner (changes requested):

libs/ml/src/ops/dataholder.cpp
MissingNO57: Dataholder has pointer to Tensor and that tensor is constructed on SetData call, so it is really constant.

baykaner: ah.... this also is changing with the my other PR I think.
We can leave it for now and come back to it when we do finetuning

baykaner (approved):

evsmithx (approved):

PR #2424, opened 2020-02-13
added becnhmarks for losses charage estimations

AlexTeslenko:
added becnhmarks for losses charage estimations


libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -772,6 +774,41 @@ void VMModel::LayerAddReshape(const fetch::vm::Ptr<String> &
   }
 }
 
+void VMModel::LayerAddLoss(const fetch::vm::Ptr<String> &layer,
```
kublai-baykaner, line 777: you don't seem to bind this anywhere?
Shouldn't this be bound under the experimental bindings?

Also you should add a test using etch code (see vm_modules/test/unit/ml/jml_model_tests.cpp)

kublai-baykaner (changes requested):

libs/vm-modules/src/ml/model/model.cpp
AlexTeslenko: I use it in vm_modules/benchmark/model/model.ccp in BM_AddLoss function

PR #2427, opened 2020-02-13
Feature/performance improvement tx verify

n-hutton:
Place a small cache in front of the TX verifier and drop TXs which have already been put into it or the storage - this probably alleviates transaction thrashing (although does not fix the root cause).

### DIFF
--- a/libs/ledger/include/ledger/storage_unit/transaction_store_sync_service.hpp
+++ b/libs/ledger/include/ledger/storage_unit/transaction_store_sync_service.hpp
@@ -92,6 +92,8 @@ public:
   static constexpr uint64_t TX_FINDER_PROTO_LIMIT = 1000;
   // Limit the amount a single rpc call will provide
   static constexpr uint64_t PULL_LIMIT = 10000;
+  // Maximum size the recently seen cache can become
+  static constexpr uint64_t RECENTLY_SEEN_CACHE_SIZE = 10000;
 
   struct Config
   {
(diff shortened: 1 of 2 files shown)


libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
```diff
@@ -494,5 +505,35 @@ void TransactionStoreSyncService::OnTransaction(TransactionPtr const &tx)
   }
 }
 
+bool TransactionStoreSyncService::AlreadySeen(chain::Transaction const &tx)
+{
+  auto const &digest = tx.digest();
+  bool        result{false};
+
+  if (store_.Has(digest))
```
n-hutton, line 513: Questionable whether this should be here, for performance reasons

libs/ledger/include/ledger/storage_unit/transaction_store_sync_service.hpp
```diff
@@ -145,6 +147,12 @@ class TransactionStoreSyncService : public TransactionSink
   State OnResolvingObjects();
   State OnTrimCache();
 
+  // Avoid processing transactions that have been recently
+  // seen or are already in storage
+  bool                                   AlreadySeen(chain::Transaction const &tx);
+  std::set<byte_array::ConstByteArray>   recently_seen_txs_;
```
ejfitzgerald, line 153: Unordered set + type alises?

libs/ledger/include/ledger/storage_unit/transaction_store_sync_service.hpp
```diff
@@ -145,6 +147,12 @@ class TransactionStoreSyncService : public TransactionSink
...
+  std::deque<byte_array::ConstByteArray> recently_seen_txs_ordered_;
```
ejfitzgerald, line 154: I think actually that I would wrap this into a object by itself i.e. a SlidingWindowSet or similar? Thoughts?

PR #2430, opened 2020-02-17
ML-519 Added charge estimation for dataloaders

evsmithx:
Added charge estimation for dataloaders


libs/ml/src/dataloaders/tensor_dataloader.cpp
```diff
@@ -49,6 +49,27 @@ typename TensorDataLoader<TensorType>::ReturnType TensorDataLoader<TensorType>::
...
 }
 
+template <typename TensorType>
+OperationsCount TensorDataLoader<TensorType>::ChargeGetNext()
+{
+  OperationsCount cost = 1;
+  for (auto it : one_sample_label_shape_)
+  {
+    cost *= it;
+  }
```
baykaner: it would be good to add comments explaining exactly which operations in GetNext incur this cost.
It's difficult for me to tell why this might be correct, so some explanation is absolutely necessary.
Presumably in this case this loop should have the comment for this line?:
"  TensorType ret_labels = labels_.View(*this->current_cursor_).Copy(one_sample_label_shape_); "

Also, you don't need to write for loops like this for vectors; the math library has a Product function for doing this which means less repeated code and probably a slightly better optimisation.

libs/ml/src/dataloaders/tensor_dataloader.cpp
```diff
@@ -49,6 +49,27 @@ typename TensorDataLoader<TensorType>::ReturnType TensorDataLoader<TensorType>::
...
+  OperationsCount data_cost = 1;
+  for (auto it : one_sample_data_shapes_)
+  {
+    for (auto it2 : it)
+    {
+      data_cost *= it2;
+    }
+  }
+  cost += data_cost;
+  cost += 5;
```
baykaner, line 69: add a named constant which better describes where this "5" magic number comes from

libs/ml/src/dataloaders/tensor_dataloader.cpp
```diff
@@ -79,6 +100,21 @@ bool TensorDataLoader<TensorType>::AddData(std::vector<TensorType> const &data,
...
+                                                            const TensorType &             labels)
+{
+  OperationsCount cost = data.size();
+  cost += TensorType::PaddedSizeFromShape(labels.shape());
+  for (SizeType i{0}; i < data.size(); i++)
+  {
+    cost += TensorType::PaddedSizeFromShape(data.at(i).shape());
+    cost += data.at(i).shape().size();
+  }
+  cost += 50;
```
baykaner: magic number needs a comment and/or name constant

libs/vm-modules/tests/unit/ml/ml_estimator_tests.cpp
```diff
@@ -60,7 +60,7 @@ TEST_F(VMMLEstimatorTests, vmdataloader_constructor_have_infinite_charge)
   )";
 
   EXPECT_TRUE(toolkit.Compile(TEXT)) << stdout.str();
-  EXPECT_FALSE(toolkit.Run());
+  EXPECT_TRUE(toolkit.Run());
 }
```
baykaner: add some tests for consisitency of charge estimations

baykaner (changes requested):

libs/vm-modules/tests/unit/ml/dataloader_tests.cpp
```diff
@@ -0,0 +1,77 @@
...
+//
+//------------------------------------------------------------------------------
+
+#include "vm_modules/ml/dataloaders/dataloader.hpp"
+#include "vm_modules/ml/utilities/conversion_utilities.hpp"
+#include "vm_modules/vm_factory.hpp"
+
+#include "gmock/gmock.h"
+
+namespace {
```
baykaner, line 25: why not put all of it inside an anonymous namespace?

evsmithx: Fair point!

libs/ml/src/dataloaders/tensor_dataloader.cpp
```diff
@@ -79,6 +100,27 @@ bool TensorDataLoader<TensorType>::AddData(std::vector<TensorType> const &data,
...
 }
 
+template <typename TensorType>
+OperationsCount TensorDataLoader<TensorType>::ChargeAddData(const std::vector<TensorType> &data,
+                                                            const TensorType &             labels)
+{
+  OperationsCount cost = 2;  // setting one_sample_label_shape
+  cost += 4;                 // resizing data vector
+  OperationsCount label_copy_cost = TensorType::PaddedSizeFromShape(labels.shape());
+  cost += label_copy_cost;
```
baykaner, line 110: seems like these should be one line + add a comment to be consistent with the prior two lines:
cost += PaddedSizeFromShape() // label copy cost

baykaner (approved):

PR #2431, opened 2020-02-17
Increased debug for http connections

n-hutton:
Increase the debug for http connections, plus lock the is_open variable to prevent possible races (and actually move it to where the socket is not open)


apps/muddle-node/main.cpp
```diff
@@ -23,8 +23,8 @@
 #include "core/macros.hpp"
 #include "core/synchronisation/protected.hpp"
 #include "crypto/ecdsa.hpp"
+#include "http/http_server.hpp"
```
n-hutton, line 26: renamed to http_server since I prefer to have file names match class names. Also moved functions into .cpp so not header only file.

libs/http/include/http/http_server.hpp
```diff
@@ -0,0 +1,118 @@
...
+               ViewType const &view, Authenticator authenticator);
+
+  void AddModule(HTTPModule const &module);
+
+  std::vector<MountedView> views();
+
+  std::vector<MountedView> views_unsafe();
+
+private:
+  void SendToManager(HandleType client, HTTPResponse const &res);
```
n-hutton, line 103: this is pretty much identical  except making SendToManager private (since it should be).

libs/http/include/http/response.hpp
```diff
@@ -31,7 +31,7 @@
 namespace fetch {
 namespace http {
 
-class HTTPResponse : public std::enable_shared_from_this<HTTPResponse>
+class HTTPResponse
```
n-hutton, line 34: no need for this.

libs/http/src/http_server.cpp
```diff
@@ -0,0 +1,290 @@
+//------------------------------------------------------------------------------
```
n-hutton, line 1: Nothing changed here from the header

libs/http/include/http/connection.hpp
```diff
@@ -235,12 +241,15 @@ class HTTPConnection : public AbstractHTTPConnection,
   }
 
 private:
+  using RMutex = std::recursive_mutex;
+
   asio::ip::tcp::tcp::socket socket_;
   HTTPConnectionManager &    manager_;
   ResponseQueueType          write_queue_;
-  Mutex                      write_mutex_;
+  RMutex                     write_mutex_;
```
n-hutton, line 249: This is the only logical change to the code (protect the is_open_, plus indicate it is not open on CloseConnection rather than Close.

ejfitzgerald (approved):

PR #2433, opened 2020-02-17
performance improvement tidy

n-hutton:
Tidy up the performance improvements with dedicated caching class. Follows the comments on PR #2427 

Will port the whole change to master after this PR goes in.


libs/storage/include/storage/have_seen_recently_cache.hpp
```diff
@@ -0,0 +1,74 @@
...
+    while (seen_deque_.size() > size_)
+    {
+      seen_set_.erase(seen_deque_.back());
+      seen_deque_.pop_back();
+    }
+  }
+
+private:
+  // To achieve the functionality the class maintains both a set and a deque
+  // of the elements added
```
wkaluza: This is clear from the code. Could you use the comment to explain why having both a set and a deque is beneficial?

libs/storage/include/storage/have_seen_recently_cache.hpp
```diff
@@ -0,0 +1,74 @@
...
+  {
+    seen_set_.insert(item);
+    seen_deque_.push_front(item);
+
+    while (seen_deque_.size() > size_)
+    {
+      seen_set_.erase(seen_deque_.back());
+      seen_deque_.pop_back();
+    }
+  }
```
wkaluza: Might be able to use `fetch::core::TrimToSize` here.

libs/storage/include/storage/have_seen_recently_cache.hpp
```diff
@@ -0,0 +1,74 @@
...
+ * This class provides the functionality that you can add elements to it,
+ * they will be held in the order they were added up to a fixed size,
+ * and you can query whether they are still there (have been recently seen).
+ *
+ * Due to the implementation behaviour, adding the same item when it is
+ * already Seen may result in it returning it as not been seen sooner than expected.
+ *
+ */
+template <typename T>
+class HaveSeenRecentlyCache
```
denis-trapeznikov: Is `Have` really needed in the name? Smth like `RecentlySeenCache` would be easier to pronounce imo.

libs/storage/include/storage/have_seen_recently_cache.hpp
denis-trapeznikov: And add a test for a trimmed cache perhaps.

libs/storage/include/storage/have_seen_recently_cache.hpp
```diff
@@ -0,0 +1,74 @@
...
+{
+public:
+  explicit HaveSeenRecentlyCache(uint64_t size)
+    : size_{size}
+  {}
+
+  bool Seen(T const &item)
+  {
+    return seen_set_.find(item) != seen_set_.end();
```
denis-trapeznikov: `core::IsIn()` from core/containers/is_in.hpp

libs/storage/include/storage/have_seen_recently_cache.hpp
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  void Add(T const &item)
+  {
+    seen_set_.insert(item);
+    seen_deque_.push_front(item);
+
+    while (seen_deque_.size() > size_)
```
denis-trapeznikov: As a side note, should it even be a `while`? In a single thread, can this condition hold for longer than a single iteration?

libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
```diff
@@ -515,21 +516,13 @@ bool TransactionStoreSyncService::AlreadySeen(chain::Transaction const &tx)
...
-      recently_seen_txs_.erase(recently_seen_txs_ordered_.back());
-      recently_seen_txs_ordered_.pop_back();
-    }
+    result = true;
   }
   else
   {
-    result = true;
+    recently_seen_txs_.Add(digest);
   }
```
denis-trapeznikov, line 526: A microimprovement: this whole block logic can be hidden in `HaveSeenRecentlyCache::Add()` which can in turn return whether an insertion did actually happen, same as `std::(unordered_)?set::(insert|emplace)` do it.

libs/storage/include/storage/have_seen_recently_cache.hpp
n-hutton: I have rewritten it:

```
  // To achieve the functionality of both time ordering and fast lookup the class maintains both a
  // set and a deque of the elements added
```

libs/storage/include/storage/have_seen_recently_cache.hpp
n-hutton: @denis-trapeznikov  There is already a test for trimmed cache `TEST(HaveSeenRecentlyCache, cache_size_limit)`

@wkaluza In addition to trim to size, it also needs to be from the back, plus it needs to remove from the map so I would be wary about using a free function for that. Also, I can't see a TrimToSize in core?

libs/storage/include/storage/have_seen_recently_cache.hpp
n-hutton: Makes sense. Dropped the 'have' in the class name.

libs/storage/include/storage/have_seen_recently_cache.hpp
n-hutton: sure

libs/storage/include/storage/have_seen_recently_cache.hpp
n-hutton: No, but it is good defensive programming practice to do this. For example if you added a setter for the cache size (probably someone will want to do this), it could hold for more than a single iteration immediately after changing the cache size.

libs/ledger/src/storage_unit/transaction_store_sync_service.cpp
n-hutton: I did consider this, but, this makes it a bit ambigous to the reader. The logical thing to return from Add is 'did it add to the cache', which you would expect to be always true.

ejfitzgerald (approved):
Looks good to me

PR #2436, opened 2020-02-18
Feature/ml 472 vm modules boston house prices

baykaner:
- moved vm-modules test toolkit into vm-modules/test-utilities
- implemented vm-modules ml integration test for boston housing data


libs/vm-modules/tests/integration/ml/boston_house_prices.cpp
```diff
@@ -0,0 +1,85 @@
...
+      model.fit(data, label, batch_size);
+
+      // evaluate performance
+      var loss = model.evaluate();
+
+      // make predictions on data
+      var predictions = model.predict(data);
+
+    endfunction
+  )";
```
evsmithx: This etch code presumably overlaps a lot with the etch script in vm-modues/examples - could the two be unified?

baykaner: seems like a good idea :-)

evsmithx (dismissed):
Nice solution

evsmithx (approved):

PR #2441, opened 2020-02-20
Remove empty statements

wkaluza:


libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -436,7 +436,8 @@ State MainChainRpcService::OnWaitForBlocks()
 
     state_machine_->Delay(std::chrono::milliseconds{100 * consecutive_failures_});
     return State::REQUEST_NEXT_BLOCKS;
-  case PromiseState::SUCCESS:;
+  case PromiseState::SUCCESS:
+    break;
```
denis-trapeznikov, line 440: What exactly does this `break` break?

denis-trapeznikov (dismissed):

libs/ledger/src/protocols/main_chain_rpc_service.cpp
wkaluza: Prevents accidental fallthrough if someone extends this `switch` in the future. It's less readable to just have a semicolon.

denis-trapeznikov: > It's less readable to just have a semicolon.

Questionnable.

ejfitzgerald: I fully agree with @wkaluza - this change is much more readable I think that we should add this as an entry in our style guide.

ejfitzgerald (approved):

PR #2442, opened 2020-02-20
Refactor/spelling correction

denis-trapeznikov:
s/ize/ise/g in filenames.


n-hutton (changes requested):
'Serialise' isn't really in common usage - I would not recoomend changing it at this point, although you could see what @ejfitzgerald thinks?

PR #2443, opened 2020-02-20
Decrease build concurrency on MacOS

wkaluza:


ejfitzgerald (approved):

wkaluza:
The Mac build for this and other PRs is failing, but this change is meant to fix the problem with the Mac builds in the first place. I'm going to merge this without completing CI.

PR #2449, opened 2020-02-21
Adding input size to ChargeForwards

evsmithx:
Adding input size as a parameter to Ops::ChargeForwards() so that charge calculations can use the full size of the input tensors and not just the per-datapoint shape.


libs/ml/include/ml/ops/ops.hpp
```diff
@@ -167,15 +160,31 @@ class Ops
...
   /**
    * @brief ChargeForward
+   * This returns the cost *per-datapoint*, not the whole batch cost
    * @return estimated charge amount, necessary for performing a forward pass on data of given
    * shapes.
    */
   virtual OperationsCount ChargeForward() const
   {
-    // TODO(ML-483): make a pure virtual method after all Ops have their overrides;
+    // todo: should be removed in favour of ChargeForward below
```
baykaner: this either needs to be implemented or have a jira task added

libs/ml/include/ml/ops/ops.hpp
```diff
@@ -167,15 +160,31 @@ class Ops
...
+  }
+
+  /**
+   * This returns the cost for the *whole* batch (not per-datapoint cost) and the shape of the Op
+   * output
+   */
+  virtual std::pair<OperationsCount, math::SizeVector> ChargeForward(
+      std::vector<math::SizeVector> const &input_shapes)
+  {
+    // Todo: all Ops should implement their own ChargeForward(input_shapes)
```
baykaner: as above

libs/ml/include/ml/utilities/utils.hpp
```diff
@@ -17,12 +17,30 @@
...
+#include <string>
+#include <vector>
+
 namespace fetch {
 namespace ml {
 namespace utilities {
 
 std::string GetStrTimestamp();
 
+template <class TensorType>
```
baykaner, line 37: a function comment stating what this function does would be lovely

libs/ml/src/core/node.cpp
```diff
@@ -114,10 +108,25 @@ OperationsCount Node<TensorType>::ChargeForward(
       throw std::runtime_error("Unable to lock weak pointer.");
     }
 
-    cost += input_node_ptr->ChargeForward(visited_nodes);
+    auto cost_and_outputshape = input_node_ptr->ChargeForward(visited_nodes);
```
baykaner, line 112: would be nice to have 1 or 2 extra comments to spell out the logic in this function

libs/ml/src/ops/dataholder.cpp
```diff
@@ -83,15 +88,16 @@ bool DataHolder<TensorType>::SetData(TensorType const &data)
...
-  if (!data_)
+  if (!future_data_shape_.empty())
   {
-    throw std::runtime_error("Data is not set for " + std ::string(Descriptor()) +
-                             ", shape computing is not possible.");
+    return future_data_shape_;
   }
-  return data_->shape();
+
+  throw std::runtime_error("future_data_shape_ is not set for " + std ::string(Descriptor()) +
```
baykaner: maybe worth flipping the return and throw positions here, so instead check for:
if (future_data_shape_.empty())

libs/vm-modules/include/vm_modules/ml/model/model.hpp
```diff
@@ -175,8 +175,13 @@ class VMModel : public fetch::vm::Object
                           math::SizeType const &dimensions, math::SizeType const &data_points,
                           bool stub);
 
-  fetch::ml::OperationsCount ChargeForward() const
+  fetch::ml::OperationsCount ChargeForward(math::SizeVector input_shape) const
   {
+    // todo: this logic should be in model_->ChargeForward()
```
baykaner: again, either do this or raise a task for it

libs/vm-modules/src/ml/model/model.cpp
```diff
@@ -963,13 +961,18 @@ ChargeAmount VMModel::EstimateEvaluate()
...
-  FETCH_LOG_INFO(LOGGING_NAME, " forward pass estimated batch cost is " +
-                                   std::to_string(batch_size) + " * " + std::to_string(cost) +
-                                   " = " + std::to_string(batch_cost));
+  //  SizeType const     batch_size = model_->dataloader_ptr_->Size();
+
+  // todo: this logic should be in model_->ChargeForward()
+  //  auto input_node = model_->graph_ptr_->GetNode(model_->input_);
+  //  auto dataholder =
+  //    std::dynamic_pointer_cast<fetch::ml::ops::DataHolder<TensorType>>(input_node->GetOp());
+  //  dataholder->SetFutureDataShape(data->GetConstTensor().shape());
```
baykaner: leftover code?

baykaner (changes requested):

baykaner (approved):

PR #2452, opened 2020-02-24
rewrite the bloom search logic due to possible bug

n-hutton:
During deployment, it seemed as if the main chain incorrectly calculated where to search the bloom filters (link time issue?). The logic is rewritten to flag such a bug were it to happen.

### DIFF
--- a/libs/ledger/src/chain/main_chain.cpp
+++ b/libs/ledger/src/chain/main_chain.cpp
@@ -2089,8 +2089,6 @@ DigestSet MainChain::DetectDuplicateTransactions(BlockHash const &           sta
             ? (valid_until - chain::Transaction::MAXIMUM_TX_VALIDITY_PERIOD)
             : 0;
 
-    static_assert(chain::Transaction::MAXIMUM_TX_VALIDITY_PERIOD == 40000, "");
-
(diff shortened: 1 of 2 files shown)


libs/ledger/src/chain/main_chain.cpp
```diff
@@ -2082,11 +2082,18 @@ DigestSet MainChain::DetectDuplicateTransactions(BlockHash const &           sta
...
   {
+    auto const valid_until = tx_layout.valid_until();
+
     auto const default_valid_from =
-        tx_layout.valid_until() - std::min(MAXIMUM_TX_VALIDITY_PERIOD, tx_layout.valid_until());
+        valid_until >= chain::Transaction::MAXIMUM_TX_VALIDITY_PERIOD
+            ? (valid_until - chain::Transaction::MAXIMUM_TX_VALIDITY_PERIOD)
+            : 0;
+
+    static_assert(chain::Transaction::MAXIMUM_TX_VALIDITY_PERIOD == 40000, "");
```
ejfitzgerald: I am not sure that I would include this in the review. This was only for our local testing I thought?

n-hutton: OK, I will remove it

ejfitzgerald (approved):

PR #2454, opened 2020-02-24
fix possible segv in main chain when modifing iterator

n-hutton:
The mainchain can make an unsafe memory access in the following way:

`auto const &child{ref_it->second};`
calls
CacheReference(hash, child, true);

Which has the following:

```
    // forget all other references
    forward_references_.erase(siblings.first, siblings.second);
    // keep unique this one
    forward_references_.emplace(hash, next_hash);
```

The next_hash here will cause a segv since it has been erased from the multimap the line above. The fix copies the next_hash in order to make the call.

In general the forward reference stuff should probably be refactored as part of planned cleanup - on looking into this bug there were a number of code smells.

### DIFF
--- a/libs/ledger/src/chain/main_chain.cpp
+++ b/libs/ledger/src/chain/main_chain.cpp
@@ -388,7 +388,7 @@ void MainChain::KeepBlock(IntBlockPtr const &block) const
     auto forward_refs{forward_references_.equal_range(hash)};
     for (auto ref_it{forward_refs.first}; ref_it != forward_refs.second; ++ref_it)
     {
-      auto const &child{ref_it->second};
+      auto const child{ref_it->second};
       if (block_store_->Has(storage::ResourceID(child)))
       {
         record.next_hash = child;


ejfitzgerald (approved):

n-hutton:
Side notes: this is probably the issue seen before that prompted the inclusion of 
```
    MilliTimer tmr("ChildrenOnStore", 2000);
    children_on_storage_checks_total_->increment();
```

to be added so it should be removed in future.

libs/ledger/src/chain/main_chain.cpp
```diff
@@ -388,7 +388,7 @@ void MainChain::KeepBlock(IntBlockPtr const &block) const
     auto forward_refs{forward_references_.equal_range(hash)};
     for (auto ref_it{forward_refs.first}; ref_it != forward_refs.second; ++ref_it)
     {
-      auto const &child{ref_it->second};
+      auto const child{ref_it->second};
```
denis-trapeznikov, line 391: This actually performs a hash copy on each iteration. An alternative would be to only perfrom copy before the call to `CacheReference` below, or even smth like change the call to `CacheReference` to `CacheReference(hash, record.next_hash, true);`.

n-hutton: This seems like the cleanest (minimal) solution to me - as I say I think there should be wider refactors (the forward reference stuff should probably be managed in its own class). I don't think the copy should matter here since byte array is shared and so there isn't really an underlying memory copy.

PR #2455, opened 2020-02-25
Bug/undeprecated file io

denis-trapeznikov:
This commit fixes (at least partly) https://fetchai.atlassian.net/browse/LDGR-595

Old binary-based file I/O is now replaced with a more strict model, requiring a boilerplate for every binary-stored type, to avoid deprecated or undefined behaviour.


n-hutton (changes requested):
This does not build (?). This PR has a lot of templates in it, especially buffer_io.hpp. I don't really see that it adds value, while it significantly reduces readability

PR #2457, opened 2020-02-25
Added benchmarks for AdamOptimiser

evsmithx:
Adding benchmarks for optimisation and starting on fine-tuning


libs/ml/benchmark/optimisation/optimisation.cpp
```diff
@@ -0,0 +1,285 @@
...
+  }
+}
+
+static void OptimiserRunArguments(benchmark::internal::Benchmark *b)
+{
+  std::vector<std::int64_t> batch_size{1, 10, 100};
+
+  for (auto const &bs : batch_size)
+  {
+    b->Args({5, bs, bs, bs, bs, 100, 0});
```
baykaner: these seems confusing. batch size is set at position 1, 2, 3 & 4.
Looking at Optimiser run batch_size takes the config position 0, and uses this batch size for input, hidden, & output ?

baykaner (changes requested):

baykaner (approved):

PR #2459, opened 2020-02-26
Refactor/ml 525 fine tune coefs relu

baykaner:
- implemented benchmarks & charge estimation relu


MissingNO57:
Continues in #2473

PR #2460, opened 2020-02-26
Improvement/peer blacklist

denis-trapeznikov:
This commit adds a temporary blacklist for alien peers.


libs/core/tests/unit/temporary_blacklist_tests.cpp
```diff
@@ -0,0 +1,57 @@
...
+using namespace std::chrono_literals;
+
+using TestedClass = fetch::core::TemporaryBlacklist<int>;
+
+TEST(TemporaryBlacklistTests, Main)
+{
+  TestedClass blacklist(1s);
+
+  blacklist.Blacklist(10);
+  std::this_thread::sleep_for(500ms);
```
wkaluza: We have an adjustable clock to simulate passage of time in tests. Calls `sleep` in test code bloat execution time unnecessarily.

wkaluza (changes requested):

libs/ledger/src/protocols/main_chain_rpc_service.cpp
```diff
@@ -240,11 +240,18 @@ MainChainRpcService::Address MainChainRpcService::GetRandomTrustedPeer() const
...
+    do
+    {
+      // generate a random peer index
+      std::size_t const index = rng() % direct_peers.size();
 
-    // select the address
-    address = direct_peers[index];
+      // select the address
+      address = direct_peers[index];
+    } while (alien_peers_.IsBlacklisted(address));
```
wkaluza, line 254: use `std::find_if_not`

wkaluza (changes requested):

libs/ledger/src/protocols/main_chain_rpc_service.cpp
denis-trapeznikov: > use std::find_if_not

Hmm, @wkaluza, do you suggest employing a custom randomized `ExecutionPolicy`? One problem is, std algorithms accepting policies were only standardised in 17 (and even then most STL implementations did not support them). Even then, that would be an overcomplication of random choice, IMO.

libs/core/tests/unit/temporary_blacklist_tests.cpp
denis-trapeznikov: > We have an adjustable clock to simulate passage of time in tests.

Where exactly?

wkaluza: `adjustable_clock.hpp`

libs/ledger/src/protocols/main_chain_rpc_service.cpp
wkaluza: You're right, the randomisation would be tricky. Let's leave it as it is, then.

n-hutton (changes requested):
The tests for this segfault

PR #2463, opened 2020-02-28
LGDR-733 Bug/telemetry uniqueness

denis-trapeznikov:
A proposal to fix aforementioned bug, this is currently a blocker for #2460.


libs/telemetry/include/telemetry/registry.hpp
```diff
@@ -81,14 +86,33 @@ class Registry
 
 private:
   using MeasurementPtr = std::shared_ptr<Measurement>;
-  using Measurements   = std::vector<MeasurementPtr>;
+  using Measurements =
+      std::unordered_map<std::string, std::unordered_map<std::type_index, MeasurementPtr>>;
```
ejfitzgerald: As discussed previously, we need to actually match on name + labels. I think the better solution for this here would be to make an unordered_set of MeasurementPtr with a custom hasher that combines measurement name and labels.

The case that you below allowing more than one type per name is not quite right

libs/core/tests/integration/containers/set_intersection.cpp
```diff
@@ -47,10 +47,10 @@ TEST(SetIntersection, BasicSetMapCheck1)
   auto const c = a & b;
 
   EXPECT_EQ(c.size(), 4);
-  EXPECT_TRUE(core::IsIn(c, 3));
-  EXPECT_TRUE(core::IsIn(c, 4));
-  EXPECT_TRUE(core::IsIn(c, 5));
-  EXPECT_TRUE(core::IsIn(c, 6));
+  EXPECT_TRUE(meta::IsIn(c, 3));
```
ejfitzgerald: Ideally the meta library was supposed to just be a compile time meta programming helpers. How come this was necessary to move it?

ejfitzgerald (changes requested):

libs/chain/src/transaction_serializer.cpp
```diff
@@ -47,17 +47,18 @@ using serializers::MsgPackSerializer;
...
+constexpr int8_t  UNIT_MEGA          = -2;
+constexpr int8_t  UNIT_KILO          = -1;
+constexpr int8_t  UNIT_DEFAULT       = 0;
+constexpr int8_t  UNIT_MILLI         = 1;
+constexpr int8_t  UNIT_MICRO         = 2;
+constexpr int8_t  UNIT_NANO          = 3;
+constexpr int8_t  CONTRACT_PRESENT   = 1;
+constexpr int8_t  CHAIN_CODE_PRESENT = 2;
+constexpr int8_t  SYNERGETIC_PRESENT = 3;
+constexpr uint8_t IDENTITY_KEY       = 0x4;
```
ejfitzgerald: This should actually be called something like: Uncompressed secp256k1 public key or something similar - we can do this in the next PR

libs/telemetry/src/registry.cpp
```diff
@@ -183,11 +124,36 @@ void Registry::Collect(std::ostream &stream)
...
 
+std::size_t Registry::FastMeasurementHash::operator()(MeasurementPtr const &measurement) const
+{
+  assert(measurement);
+
+  auto const &labels = measurement->labels();
+
+  std::string flattened = std::accumulate(
+      begin(labels), end(labels), std::string(), [](std::string accum, auto &&element) {
+        return std::move(accum) + element.first + KEY + element.second + VALUE;
```
ejfitzgerald: Do we not need to sort the keys here first to ensure a consistent hash. Also it might be worth (maybe in the future) to switching to a XOR based hash aggregation as the performance of this with multiple labels might be expensive

ejfitzgerald (approved):

PR #2467, opened 2020-02-28
refactored all ops to remove old style ChargeForward methods

baykaner:


libs/ml/src/ops/abs.cpp
```diff
@@ -111,12 +111,14 @@ std::vector<math::SizeType> Abs<TensorType>::ComputeOutputShape(
 }
 
 template <typename TensorType>
-OperationsCount Abs<TensorType>::ChargeForward() const
+std::pair<OperationsCount, math::SizeVector> Abs<TensorType>::ChargeForward(
+    std::vector<math::SizeVector> const &input_shapes)
 {
   assert(!this->batch_input_shapes_.empty());
```
evsmithx: Now that we are no longer using batch_input_shapes_ in this method I think we can get rid of the asserts

libs/ml/src/ops/avg_pool_1d.cpp
```diff
@@ -180,17 +180,20 @@ std::vector<math::SizeType> AvgPool1D<TensorType>::ComputeOutputShape(
...
                                          this->batch_output_shape_.at(2);
-  OperationsCount cost =
+  OperationsCount op_cnt =
       fetch::ml::charge_estimation::ops::DIVISION_PER_ELEMENT * num_output_shape_ops +
       fetch::ml::charge_estimation::ops::ADDITION_PER_ELEMENT * num_output_shape_ops *
           this->kernel_size_;
-  return cost;
+
+  auto output_shape = ComputeOutputShape(input_shapes);
+  return std::make_pair(op_cnt, output_shape);
```
evsmithx, line 194: ChargeForward should now return the cost for the full batch, not the per-datapoint cost. Assuming the old method returned the correct per-datapoint cost then this method should return op_cnt * batch_size

libs/ml/src/ops/avg_pool_2d.cpp
```diff
@@ -199,17 +199,20 @@ std::vector<math::SizeType> AvgPool2D<TensorType>::ComputeOutputShape(
...
       this->batch_output_shape_.at(2) * this->batch_output_shape_.at(3);
-  auto cost = static_cast<OperationsCount>(
+  auto op_cnt = static_cast<OperationsCount>(
       fetch::ml::charge_estimation::ops::DIVISION_PER_ELEMENT * num_output_shape_ops +
       fetch::ml::charge_estimation::ops::ADDITION_PER_ELEMENT * num_output_shape_ops *
           static_cast<OperationsCount>(this->kernel_size_ * this->kernel_size_));
-  return cost;
+
+  auto output_shape = ComputeOutputShape(input_shapes);
+  return std::make_pair(op_cnt, output_shape);
```
evsmithx, line 214: op_cnt *= batch size again

libs/ml/src/ops/convolution_1d.cpp
```diff
@@ -411,28 +411,30 @@ void Convolution1D<TensorType>::ReverseFillOutput(TensorType &gemm_output, Tenso
...
-  OperationsCount cost = horizontal_stride_width * horizontal_stride_height *
-                         vertical_stride_width *
-                         fetch::ml::charge_estimation::ops::MULTIPLICATION_PER_ELEMENT;
+  OperationsCount op_cnt = horizontal_stride_width * horizontal_stride_height *
+                           vertical_stride_width *
+                           fetch::ml::charge_estimation::ops::MULTIPLICATION_PER_ELEMENT;
 
-  return cost;
+  auto output_shape = ComputeOutputShape(input_shapes);
+  return std::make_pair(op_cnt, output_shape);
```
evsmithx: *= batch size again (I think)

libs/ml/src/ops/convolution_2d.cpp
```diff
@@ -449,32 +449,32 @@ void Convolution2D<TensorType>::ReverseFillOutput(TensorType &gemm_output, Tenso
...
-  OperationsCount cost = horizontal_stride_width * horizontal_stride_height *
-                         vertical_stride_width *
-                         fetch::ml::charge_estimation::ops::MULTIPLICATION_PER_ELEMENT;
+  OperationsCount op_cnt = horizontal_stride_width * horizontal_stride_height *
+                           vertical_stride_width *
+                           fetch::ml::charge_estimation::ops::MULTIPLICATION_PER_ELEMENT;
 
-  return cost;
+  auto output_shape = ComputeOutputShape(input_shapes);
+  return std::make_pair(op_cnt, output_shape);
```
evsmithx: batch size

libs/ml/src/ops/reshape.cpp
```diff
@@ -124,14 +124,17 @@ std::vector<math::SizeType> Reshape<TensorType>::ComputeOutputShape(
...
-  assert(!this->batch_output_shape_.empty());
-  OperationsCount cost = fetch::ml::charge_estimation::ops::RESHAPE_PER_ELEMENT *
-                             this->TotalElementsIn({this->batch_output_shape_}) +
-                         fetch::ml::charge_estimation::ops::ASSIGN_PER_ELEMENT *
-                             this->TotalElementsIn({this->batch_output_shape_});
-  return cost;
+  assert(!this->batch_input_shapes_.empty());
+
+  OperationsCount op_cnt = (fetch::ml::charge_estimation::ops::RESHAPE_PER_ELEMENT +
+                            fetch::ml::charge_estimation::ops::RESHAPE_PER_ELEMENT) *
```
evsmithx: This should be ASSIGN_PER_ELEMENT?

evsmithx (changes requested):

libs/ml/src/ops/abs.cpp
baykaner: good point

libs/ml/src/ops/avg_pool_1d.cpp
baykaner: input shapes contains that info though right? so if i just swap out the 'batch shapes' for input shapes it'll include the batch size for free i think.

libs/ml/src/ops/avg_pool_2d.cpp
baykaner: as above

libs/ml/src/ops/convolution_1d.cpp
baykaner: here I've already done the swap

libs/ml/src/ops/convolution_2d.cpp
baykaner: also here

libs/ml/src/ops/reshape.cpp
baykaner: I haven't fine-tuned it. So either is fine at this stage.

libs/ml/src/ops/avg_pool_1d.cpp
evsmithx: Yep, but you're using batch_output_shape_ above!

libs/ml/src/ops/avg_pool_2d.cpp
evsmithx: Again, you're using batch_output_shape!

libs/ml/src/ops/convolution_1d.cpp
evsmithx: Yep, agreed

libs/ml/src/ops/reshape.cpp
evsmithx: Okay, I just noticed it had changed from the previous implementation!

libs/ml/src/ops/avg_pool_2d.cpp
baykaner: not anymore - i pushed changes

libs/ml/src/ops/avg_pool_1d.cpp
evsmithx: Github fail, it's fixed

evsmithx (dismissed):

evsmithx (dismissed):

MissingNO57 (approved):
LGTM

PR #2468, opened 2020-02-28
Disable pipenv check build step

wkaluza:


ejfitzgerald (approved):

wkaluza:
Bypassing CI after discussion with Ed. Too many builds are failing on `pipenv check` due to an unrelated network issue. We'll wait for `pyup.io` to stop misbehaving or reintroduce `pipenv check` somewhere where it doesn't cause this much disruption.

PR #2471, opened 2020-02-29
VM Return instruction bugfix

robertdickson:
Fixes a bug in the handling of the Return instruction when invoked from inside an etch struct's constructor

### DIFF
--- a/libs/vm/src/parser.cpp
+++ b/libs/vm/src/parser.cpp
@@ -1440,8 +1440,8 @@ ExpressionNodePtr Parser::ParseExpressionStatement()
 bool Parser::IsStatementKeyword(Token::Kind kind) const
 {
   return (kind == Token::Kind::Persistent) || (kind == Token::Kind::Contract) ||
-         (kind == Token::Kind::EndContract) || /* (kind == Token::Kind::Struct) || */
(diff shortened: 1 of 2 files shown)


libs/vm/src/parser.cpp
```diff
@@ -1440,8 +1440,8 @@ ExpressionNodePtr Parser::ParseExpressionStatement()
 bool Parser::IsStatementKeyword(Token::Kind kind) const
 {
   return (kind == Token::Kind::Persistent) || (kind == Token::Kind::Contract) ||
-         (kind == Token::Kind::EndContract) || /* (kind == Token::Kind::Struct) || */
-         /* (kind == Token::Kind::EndStruct) || */ (kind == Token::Kind::Function) ||
+         (kind == Token::Kind::EndContract) || (kind == Token::Kind::Struct) ||
```
ejfitzgerald, line 1443: Not sure we want to enable this at this point do we?

robertdickson: It's not enabling them though -- it's just removing one of the ways of disabling structs. Structs are actually still disabled in the parser.cpp with a commented out block.

ejfitzgerald: 👍

ejfitzgerald (approved):

PR #2472, opened 2020-03-02
ML-531 Converted to new Ops::ChargeBackwards

evsmithx:
Converted all Ops to use the new ChargeBackwards method which takes the input sizes as an argument.


MissingNO57 (approved):
LGTM

PR #2475, opened 2020-03-04
Benchmark/transaction serializer

denis-trapeznikov:


libs/chain/benchmark/tx_serialisation.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include <vector>
+
+using fetch::chain::TransactionSerializer;
+using fetch::crypto::ECDSASigner;
+using Storage      = std::vector<fetch::byte_array::ConstByteArray>;
+using Transactions = std::vector<fetch::chain::Transaction>;
+
+void TxSerialisation(benchmark::State &state)
+{
+  state.PauseTiming();
```
n-hutton, line 36: I didn't think the timer needed to be paused here - isn't the benchmarking only done in the `for (auto _ : state)` loop?

libs/chain/benchmark/tx_serialisation.cpp
```diff
@@ -0,0 +1,99 @@
...
+    TransactionList input = GenerateTransactions(std::size_t(state.range(1)), signer, small_tx);
+
+    const auto            num_txs = input.size();
+    Transactions          output(num_txs);
+    Storage               cells(num_txs);
+    TransactionSerializer sr;
+    std::size_t           in_errors{}, out_errors{};
+
+    state.ResumeTiming();
+    for (std::size_t i{}; i < num_txs; ++i)
```
n-hutton, line 73: I might have considered populating a vector and using the implicit serializer, this is equivalent though of course

n-hutton (approved):
Looks good - might be worth also making three tests out of this: serialization only, deserialization only, ser + deser.

